PR #12, opened 2019-11-14
Regroup Travis CI setup

SaschaWillems:
## Description

This PR regroups the Travis CI yaml file and gets rid of the empty shell build steps.

## Checklist:

- n/a My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [x] I have performed a self-review of my own code
- n/a I have commented my code, particularly in hard-to-understand areas
- n/a I have made corresponding changes to the documentation
- n/a I have tested my sample on at least one compliant Vulkan implementation
- n/a I have stated on what implementation the sample has been tested so that others can test on different implementations and platforms
- n/a My changes do not add any new compiler warnings
- n/a Vulkan validation layer output is clean on at least one compliant implementation
- n/a Any dependent changes (e.g. assets) have been merged and published in downstream modules
- n/a I have used existing framework/helper functions where possible
- n/a I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- n/a My changes build and run on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,9 +30,6 @@ jobs:
   allow_failures:
     - name: "Snake Case (Optional)"
   include:
-
-jobs:
-  include:
     - stage: Check
       name: "Doxygen"
       script: 
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,53 +30,38 @@ jobs:
   allow_failures:
     - name: "Snake Case (Optional)"
   include:
+
+jobs:
+  include:
     - stage: Check
-
-    - name: "Doxygen"
-      script:
-        - docker exec VKB sh -c "cd /home/vulkan-samples && mkdir -p doxygen && doxygen ./docs/doxygen/doxyfile"
-        - docker exec VKB sh -c "cd /home/vulkan-samples && if [ $(stat -c%s ./doxygen/warnings.txt) -gt 0 ]; then cat ./doxygen/warnings.txt; exit 1; fi"
-
-    - name: "Copyright Headers"
-      script:
-        - docker exec VKB sh -c "cd /home/vulkan-samples && python3 /usr/local/bin/check_copyright_headers.py master"
-
-    - name: "Clang Format"
-      script:
-        - docker exec VKB sh -c "cd /home/vulkan-samples && python3 /usr/local/bin/clang_format.py -v --diff master > ./clang-format-report.txt"
-        - docker exec VKB sh -c "cd /home/vulkan-samples && if [ $(grep -c 'clang-format did not modify any files' ./clang-format-report.txt) -eq 0  ] && [ $(grep -c 'no modified files to format' ./clang-format-report.txt) -eq 0  ]; then cat ./clang-format-report.txt; exit 1; fi"
-
-    - name: "Clang Tidy"
-      script:
-        - docker exec VKB sh -c "cd /home/vulkan-samples && cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -H. -Bbuild/clang"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -30,53 +30,46 @@ jobs:
...
-
-    - name: "Copyright Headers"
-      script:
-        - docker exec VKB sh -c "cd /home/vulkan-samples && python3 /usr/local/bin/check_copyright_headers.py master"
-
-    - name: "Clang Format"
-      script:
+    - script: 
+      -  docker exec VKB sh -c "cd /home/vulkan-samples && python3 /usr/local/bin/check_copyright_headers.py master"
+      name: "Copyright Headers"
```
TomAtkinsonArm, line 40: I see that the names are now under the script tag. Although the pipeline seems to be working and there are no wasted jobs this seems counter intuitive. Also if we are putting the names under the script tag shouldn't the Doxygen job follow the same format?

.travis.yml
```diff
@@ -30,53 +30,46 @@ jobs:
   allow_failures:
     - name: "Snake Case (Optional)"
   include:
```
TomAtkinsonArm, line 32: Can you confirm that the snake case check is still optional in the pipeline. I can't seem to find the view i was on that i posted on slack were it explicitly said this job is optional. This job is required to be optional as the check outputs a lot of false positives when working with third party libraries. The flex i wrote for snake case check is a does not take into account the source of a token merely says if a token is snake case or not (hence the false positives).

TomAtkinsonArm (changes requested):
Unsure about the job structure. Although it works it seems counter intuitive. There must be a way we achieve this result but keeping the script looking intuitive..

.travis.yml
SaschaWillems: Yeah, that actually is counter intuitive, but that's how it's described in the Travis documentation: https://docs.travis-ci.com/user/build-stages/#naming-your-jobs-within-build-stages

.travis.yml
SaschaWillems: Judging by the documentation, ```allow_failures``` always refers to a job. And as the Snake Case test is a stage, this won't work. So unless I'm mistaken, you need to make Snake Case a separate job if you want it to be optional.

TomAtkinsonArm: When i had the allow failures in section in the main job includes i had that table output (slack) from Travis stating it was optional. If you can find this page showing the job is optional then this syntax is okay. Else it may need more testing (withing the main job list, where it is... etc)

.travis.yml
TomAtkinsonArm: I see now that when i was reading the docs i was reading it name then script by default. A little broken from Travis but it works i'm happy :)

.travis.yml
TomAtkinsonArm: I found the page i'm talking about! Seems like the only way to access it is through Github in the checks section below (click details on the Travis tests). Comparing it to my screen shot the snake case currently is not optional. I'm assuming this can be fixed by moving allowed_failures back to the main job block?

SaschaWillems: You're right, snake case is no longer optional for the changes in my PR. I'm off to a business trip, but I'll try to fix that once I get home.

SaschaWillems: Snake case is now optional (again). See: https://github.com/KhronosGroup/Vulkan-Samples/pull/12/checks?check_run_id=308540944

TomAtkinsonArm (approved):

wasimabbas-arm (approved):

ghost (approved):

PR #27, opened 2020-01-03
Fix ShaderResource::size value for 8 bit types

jherico:
## Description

The use of 7 looks like a simple typo that would easily go undetected until someone actually tried to use a double or other 64 bit type in a shader.

## Checklist:

I'm not spending my whole day on this.

### DIFF
--- a/framework/spirv_reflection.cpp
+++ b/framework/spirv_reflection.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2019, Arm Limited and Contributors
+/* Copyright (c) 2019-2020, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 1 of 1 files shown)


wasimabbas-arm (commented):
Bug fix LGTM, please revert the copyright changes then we can merge it.

ZandroFargnoli (changes requested):
@jherico Thank you for the fixes, I see you have a few PRs changing many of the files. You should only need to change the copyright for files you've added/modified.

If you can revert your commits and just add `-2020` to the `.cpp` file we can then merge it.

jherico:
@wasimabbas-arm @ZandroFargnoli fixed copyright

ZandroFargnoli (approved):
LGTM 👍

PR #31, opened 2020-01-07
Index buffer size is using vertex_buffer_size

jherico:
## Description

Apparent simple copy paste error.  When creating the index buffer for the GUI, it uses `vertex_buffer_size` instead of `index_buffer_size`

## Checklist:

Ugh, seriously, guys, come on.

### DIFF
--- a/framework/gui.cpp
+++ b/framework/gui.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2019, Arm Limited and Contributors
+/* Copyright (c) 2018-2020, Arm Limited and Contributors
  * Copyright (c) 2019, Sascha Willems
  *
  * SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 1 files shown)


wasimabbas-arm:
LGTM, please revert the copyright headers commit.

ZandroFargnoli (changes requested):
@jherico Thank you for the fixes, I see you have a few PRs changing many of the files. You should only need to change the copyright for files you've added/modified.

If you can revert your commits and just add -2020 to the `.cpp` file we can then merge it.

jherico:
@ZandroFargnoli  @wasimabbas-arm copyright headers reverted.  However, the gui file I changed contains two copyright lines, one for "arm and contributors" which I updated, and one specifically for Sascha, which I didn't touch.  I don't know if the CI check is smart enough to differentiate between them.

wasimabbas-arm:
> @ZandroFargnoli @wasimabbas-arm copyright headers reverted. However, the gui file I changed contains two copyright lines, one for "arm and contributors" which I updated, and one specifically for Sascha, which I didn't touch. I don't know if the CI check is smart enough to differentiate between them.

This is a good question. We should probably change the copyright script to look for all copyrights. I have created https://github.com/KhronosGroup/Vulkan-Samples/issues/42

PR #24, opened 2019-12-16
Arm Release v1.3.0

ZandroFargnoli:
## Description

Read the [full release article](https://github.com/ARM-software/vulkan_best_practice_for_mobile_developers/releases/tag/v1.3.0) for a more complete description of the PR.

Please read the [contribution guidelines](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md)

## Checklist:

Do not submit your PR without all of the below being checked:

- [x] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have tested my sample on at least one compliant Vulkan implementation
- [x] I have stated on what implementation the sample has been tested so that others can test on different implementations and platforms
- [x] If my sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [x] My changes do not add any new compiler warnings
- [x] Vulkan validation layer output is clean on at least one compliant implementation
- [x] Any dependent changes (e.g. assets) have been merged and published in downstream modules
- [x] I have used existing framework/helper functions where possible
- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [x] My changes build and run on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)

PR #34, opened 2020-01-07
Validation

### LINKED ISSUE #29: Default image usage bits for vkb::Swapchain wrapper appear incorrect
The predominate usage for a swapchain image is to attach it to a framebuffer and render directly to it, or the render to something else and blit to the swapchain image.  This corresponds to a usage of `VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT` and `VK_IMAGE_USAGE_TRANSFER_DST_BIT`. 

However, the default flags in the swapchain constructors [here](https://github.com/KhronosGroup/Vulkan-Samples/blob/af0dbc2ec3a872688265e355d997070fd50813ac/framework/core/swapchain.h#L82) and [here](https://github.com/KhronosGroup/Vulkan-Samples/blob/af0dbc2ec3a872688265e355d997070fd50813ac/framework/core/swapchain.h#L95) are `VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT` (correct) and `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` (_incorrect_).

### LINKED ISSUE #30: Instance creation should have validation layers applied to it.
The code in `vkb::Instance::Instance` constructs a `VkInstance` and then (when either `VKB_DEBUG` or `VKB_VALIDATION_LAYERS` are defined) populates a `VkDebugReportCallbackCreateInfoEXT` and  creates a debug report callback.

However, the `VK_EXT_debug_report` extension was specifically designed to ensure that validation could be applied even to instance creation.

if `VKB_DEBUG` or `VKB_VALIDATION_LAYERS` are defined, then the `VkDebugReportCallbackCreateInfoEXT` declaration should be before instance creation code, and the `VkDebugReportCallbackCreateInfoEXT` structure should be set as the `pNext` of `VkInstanceCreateInfo`

### LINKED ISSUE #33: Use VK_LAYER_KHRONOS_validation layer
Then enabling validation on the hello_triangle example it tries to use the following layers

    VK_LAYER_GOOGLE_threading
    VK_LAYER_LUNARG_parameter_validation
    VK_LAYER_LUNARG_object_tracker
    VK_LAYER_LUNARG_core_validation
    VK_LAYER_GOOGLE_unique_objects

But these layers do not exist on my up to date system.  See [this report](http://vulkan.gpuinfo.org/displayreport.php?id=7266#tabinstancelayers) on vulkan.gpuinfo.org.

The code should attempt to first use `VK_LAYER_KHRONOS_validation`, then fall back to `VK_LAYER_LUNARG_standard_validation`, and then fall back to the above list, and if that still doesn't validate against the existing layers, it should use no validation layers (possibly emitting a warning).

### LINKED ISSUE #37: The parallel compilation option is gated by `VKB_WARNINGS_AS_ERRORS`
The `/MP` option enables Visual Studio to build in parallel.  This drastically reduces build time on systems with sufficient cores.  However, this option is only [turned on](https://github.com/KhronosGroup/Vulkan-Samples/blob/af0dbc2ec3a872688265e355d997070fd50813ac/framework/CMakeLists.txt#L350) inside a `if(${VKB_WARNINGS_AS_ERRORS})` conditional block.  If wanrings as errors are turned off, then parallel build is disabled.

### LINKED ISSUE #38: Incorrect barrier in GLTF upload
[source access mask](https://github.com/KhronosGroup/Vulkan-Samples/blob/af0dbc2ec3a872688265e355d997070fd50813ac/framework/gltf_loader.cpp#L320) should be `VK_ACCESS_TRANSFER_WRITE_BIT`

jherico:
## Description

Fixes #29

Switches the default image usage flag bits for the swapchain images to use `VK_IMAGE_USAGE_TRANSFER_DST_BIT` instead of `VK_IMAGE_USAGE_TRANSFER_SRC_BIT`.  

Fixes #30 

In both the hello_triangle example and the common instance.cpp code, the `VkDebugReportCallbackCreateInfoEXT` structure is used to populate the `pNext` member of the `VkInstanceCreateInfo` when validation is enabled.

Fixes #33 

Both the hello_triangle and common instance.cpp now attempt to use the recommended validation layer `VK_LAYER_KHRONOS_validation`, before falling back to `VK_LAYER_LUNARG_standard_validation`, which in turn will fall back to the list of explicit validation layers.  This should ensure the examples work with validation across a broad range of versions of Vulkan drivers. Relying on the list included in hello_triangle will fail on modern drivers as reported in the linked issue.  On the other hand, older Vulkan drivers may not recognize the newer `VK_LAYER_KHRONOS_validation`, so a set of fallbacks is preferable to the fixed approach. 

Fixes #37 

The `/MP` compile option should be added regardless of whether warnings are being treated as errors.

Fixes #38 

Fixed invalid source access mask in post-image-upload barrier in the GLTF loader.

Tested on http://vulkan.gpuinfo.org/displayreport.php?id=7266

## Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have tested my sample on at least one compliant Vulkan implementation
- [X] If my sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [X] Vulkan validation layer output is clean on at least one compliant implementation
- [X] Any dependent changes (e.g. assets) have been merged and published in downstream modules

### DIFF
--- a/framework/core/instance.cpp
+++ b/framework/core/instance.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2019, Arm Limited and Contributors
+/* Copyright (c) 2018-2020, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -158,7 +158,33 @@ Instance::Instance(const std::string &              application_name,
 	std::vector<const char *> active_instance_layers(required_validation_layers);
 
 #ifdef VKB_VALIDATION_LAYERS
-	active_instance_layers.push_back("VK_LAYER_KHRONOS_validation");
+	// Optimal validation layers
+	std::vector<const char *> default_debug_layers{{
+	    "VK_LAYER_KHRONOS_validation",
+	}};
+	// Alternative layers
+	if (!validate_layers(default_debug_layers, instance_layers))
+	{
+		default_debug_layers = {
+		    "VK_LAYER_LUNARG_standard_validation",
+		};
+	}
+	// Fallback layers
+	if (!validate_layers(default_debug_layers, instance_layers))
+	{
+		default_debug_layers = {
+		    "VK_LAYER_GOOGLE_threading",
+		    "VK_LAYER_LUNARG_parameter_validation",
+		    "VK_LAYER_LUNARG_object_tracker",
+		    "VK_LAYER_LUNARG_core_validation",
+		    "VK_LAYER_GOOGLE_unique_objects",
+		};
+	}
+	if (!validate_layers(default_debug_layers, instance_layers))
+	{
+		default_debug_layers = {};
+	}
+	active_instance_layers.insert(active_instance_layers.end(), default_debug_layers.begin(), default_debug_layers.end());
 #endif
 
 	if (validate_layers(active_instance_layers, instance_layers))
@@ -183,6 +209,15 @@ Instance::Instance(const std::string &              application_name,
 
 	VkInstanceCreateInfo instance_info = {VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO};
 
+#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
+	VkDebugReportCallbackCreateInfoEXT debug_report_info = {VK_STRUCTURE_TYPE_DEBUG_REPORT_CREATE_INFO_EXT};
+
+	debug_report_info.flags       = VK_DEBUG_REPORT_ERROR_BIT_EXT | VK_DEBUG_REPORT_WARNING_BIT_EXT;
+	debug_report_info.pfnCallback = debug_callback;
+
+	instance_info.pNext = &debug_report_info;
+#endif
+
 	instance_info.pApplicationInfo = &app_info;
 
 	instance_info.enabledExtensionCount   = to_u32(extensions.size());
@@ -201,12 +236,7 @@ Instance::Instance(const std::string &              application_name,
 	volkLoadInstance(handle);
 
 #if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
-	VkDebugReportCallbackCreateInfoEXT info = {VK_STRUCTURE_TYPE_DEBUG_REPORT_CREATE_INFO_EXT};
-
-	info.flags       = VK_DEBUG_REPORT_ERROR_BIT_EXT | VK_DEBUG_REPORT_WARNING_BIT_EXT;
-	info.pfnCallback = debug_callback;
-
-	result = vkCreateDebugReportCallbackEXT(handle, &info, nullptr, &debug_report_callback);
+	result = vkCreateDebugReportCallbackEXT(handle, &debug_report_info, nullptr, &debug_report_callback);
 	if (result != VK_SUCCESS)
 	{
 		throw std::runtime_error("Could not create debug callback.");
--- a/samples/api/hello_triangle/hello_triangle.cpp
+++ b/samples/api/hello_triangle/hello_triangle.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2019, Arm Limited and Contributors
+/* Copyright (c) 2018-2020, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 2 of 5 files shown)


wasimabbas-arm:
@jherico I don't think we should update copyright notice in source/header files if we haven't actually changed anything in these source files in that year. IANAL but this is how understand this works.

jherico:
@wasimabbas-arm then someone needs to fix the CI server to recognize that, because currently it fails a PR if any file doesn't have 2020 in the copyright block, whether it's been touched by the PR or not.

jherico:
@wasimabbas-arm I've updated the PR to remove the global copyright update in favor of updated copyrights only where I've touched the code.  This will presumably fail CI again.

jherico:
I don't know whether it's possible, but its seems like the stage 1 checks should not trigger cancellation of the stage 2 checks.  A stage 1 failure in CI should still mark the PR as failed, but it's important to know whether the build succeeds independently of the formatting / tidy / copyright checks, etc.

wasimabbas-arm:
@jherico We have fixed this in one of our PR https://github.com/KhronosGroup/Vulkan-Samples/pull/24/commits/c284969decb3c5bfe3b8bf5cfb85d241f1469994 
@ZandroFargnoli has submitted a fix it's in review. I would suggest we stick to that one and reject your commit. The other changes look sensible though.

wasimabbas-arm:
Sorry my comment was not fully correct. @ZandroFargnoli changes shouldn't affect your commit.

Everything looks good to me now.

samples/api/hello_triangle/hello_triangle.cpp
```diff
@@ -207,11 +207,33 @@ void HelloTriangle::init_instance(Context &                        context,
...
-	active_instance_layers.push_back("VK_LAYER_LUNARG_core_validation");
-	active_instance_layers.push_back("VK_LAYER_GOOGLE_unique_objects");
+	// Optimal validation layers
+	std::vector<const char *> default_debug_layers{{
+	    "VK_LAYER_KHRONOS_validation",
+	}};
+	// Alternative layers
+	if (!validate_layers(default_debug_layers, instance_layers))
+	{
+		default_debug_layers = {
```
ZandroFargnoli, line 217: Can we add a warning here to let the developer know a fallback is being attempted:

`LOGW("Attempted to enable VK_LAYER_KHRONOS_validation but it wasn't found, falling back to VK_LAYER_LUNARG_standard_validation")`

We can do the same for the further fallbacks.

samples/api/hello_triangle/hello_triangle.cpp
```diff
@@ -207,11 +207,33 @@ void HelloTriangle::init_instance(Context &                        context,
 	std::vector<const char *> active_instance_layers(required_instance_layers);
 
 #ifdef VKB_VALIDATION_LAYERS
-	active_instance_layers.push_back("VK_LAYER_GOOGLE_threading");
-	active_instance_layers.push_back("VK_LAYER_LUNARG_parameter_validation");
-	active_instance_layers.push_back("VK_LAYER_LUNARG_object_tracker");
-	active_instance_layers.push_back("VK_LAYER_LUNARG_core_validation");
-	active_instance_layers.push_back("VK_LAYER_GOOGLE_unique_objects");
+	// Optimal validation layers
```
ZandroFargnoli, line 210: I can foresee edits being made to this in the future and for debugging purposes. Since we've written this twice, could we potentially move it to a `vk_common.*` function that handles this entire process for us?

ZandroFargnoli (changes requested):

samples/api/hello_triangle/hello_triangle.cpp
jherico: I can, but the point of hello_triangle.cpp appears to be avoiding delegating functionality like that to external files.  Most of what's in that file is duplicated logic.  If you still want me to put the logic in a common place, let me know and I'll do that.

samples/api/hello_triangle/hello_triangle.cpp
jherico: I'll try to add that today.

samples/api/hello_triangle/hello_triangle.cpp
ZandroFargnoli: Yes it is, it's meant to be a one file solution, and not have much code outside, however in practice this is much harder to achieve. 

I would argue selection logic for validation layers is something that wouldn't be necessary to keep in the file. 

If we had a function written that was like: `std::vector<const char*> validation_layers = select_validation_layers(available_validation_layers, requested_validation_layers)` then I think this is sufficient, and if the developer wanted to step into the code they are very much welcome too.

ZandroFargnoli: This, or we keep it in the file but simplify the selection logic for `hello_triangle` (separate to `instance.*`) to a list of known layers and we just select any that are available

framework/core/instance.cpp
```diff
@@ -158,7 +158,33 @@ Instance::Instance(const std::string &              application_name,
 	std::vector<const char *> active_instance_layers(required_validation_layers);
 
 #ifdef VKB_VALIDATION_LAYERS
-	active_instance_layers.push_back("VK_LAYER_KHRONOS_validation");
+	// Optimal validation layers
+	std::vector<const char *> default_debug_layers{{
```
ZandroFargnoli, line 162: This doesn't build on Android for me

jherico: Will investigate

ghost:
@jherico - Do you have any updates on this? :)

jherico:
I'm going to break the PR into several smaller PR's individually addressing the bugs.

framework/core/instance.cpp
jherico: I wasn't able to get an Android build compiling for reasons that appear unrelated to this PR.

Issue #36, opened 2020-01-08
SemaphorePool has invalid rvalue reference ctor


jherico:
The `SemaphorePool` class declares an rvalue reference constructor like this: 

    SemaphorePool(SemaphorePool &&other) = default;

However, the class uses RAII sematics, and will destroy all the contained semaphores when it is destroyed.  If anyone ever actually uses the `SemaphorePool(SemaphorePool &&other)` while the pool is non-empty, then both the new instance and the `other` instance will contain the same set of semaphores.  This would lead to semaphores being used after their destroyed or semaphores being double destroyed.

`SemaphorePool(SemaphorePool &&other)` should either be marked `delete` or it should be explicitly written to wipe the `other.semaphores` vector.

PR #40, opened 2020-01-10
Fix double destroy bug in semaphore pool

jherico:
## Description

Fix for #36 

Ensure that the `SemaphorePool` rvalue reference constructor clears the vector of `VkSemaphore` objects in the source instance.

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have tested my sample on at least one compliant Vulkan implementation
- [x] If my sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [x] Vulkan validation layer output is clean on at least one compliant implementation
- [x] Any dependent changes (e.g. assets) have been merged and published in downstream modules

### DIFF
--- a/framework/semaphore_pool.h
+++ b/framework/semaphore_pool.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2019, Arm Limited and Contributors
+/* Copyright (c) 2019-2020, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -31,7 +31,7 @@ class SemaphorePool
 
 	SemaphorePool(const SemaphorePool &) = delete;
 
-	SemaphorePool(SemaphorePool &&other) = default;
+	SemaphorePool(SemaphorePool &&other);
 
 	~SemaphorePool();
 
--- a/framework/fence_pool.cpp
+++ b/framework/fence_pool.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2019, Arm Limited and Contributors
+/* Copyright (c) 2019-2020, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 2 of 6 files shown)


framework/semaphore_pool.h
```diff
@@ -31,7 +31,7 @@ class SemaphorePool
 
 	SemaphorePool(const SemaphorePool &) = delete;
 
-	SemaphorePool(SemaphorePool &&other) = default;
```
wasimabbas-arm, line 34: I think this is a typo. This should follow the design of the rest of the class and we should delete the move constructor.

The samples not build with `= delete`?
As a reference http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines C.21 backs this theory up.
`C.21: If you define or =delete any default operation, define or =delete them all`

wasimabbas-arm: @jherico what do you think about this? Can you do the fix so we can merge this one too?

jherico:
Yeah, sorry, I've been distracted looking for a job.   Will update today.

wasimabbas-arm (dismissed):

jherico:
Looks like removing the move ctor doesn't work.  `RenderFrame` has a `SemaphorePool` member, and `RenderFrame` is stored in a `std::vector` inside `RenderContext`.

Investigating further, it appears that `FencePool` has the same bug, which is also used inside `RenderFrame`.

In order to resolve this bug by changing the `FencePool` and `SemaphorePool` move constructors to `= delete` I would also need to delete the `RenderFrame` move constructor and figure out how to fix the compilation failure of the `std::vector<RenderFrame>` member inside RenderContext.

jherico:
I've also discovered that `RenderTarget` has the same bug.  It must have a non-default move operator.

An alternative solution to all this would be to delete all these move operators and then make `RenderContext` contain a vector of `std::unique_ptr<RenderFrame>`.  I think I'll make a second PR for that solution and you guys can pick.

ZandroFargnoli:
@jherico Thank you very much for giving us two solutions to chose from. I think PR 50 have the preferable changes so I will close this to allow for us to review that instead :)

PR #82, opened 2020-02-23
Fix UI scaling and mouse interaction on retina displays

### LINKED ISSUE #76: Mouse interaction does not work properly on Mac Retina displays
The logic in gui.cpp takes into account the physical size of the monitor in order to scale content (via `dpi_factor`) however, it does not handle a disparity between the window coordinate system used for input events and the framebuffer surface pixel coordinate system.  

On retina mac displays input events come in at half the resolution of the actual screen.  This can be be seen in a disparity between the values returned by `glfwGetWindowSize` and `glfwGetFramebufferSize`.  In GLFW 3.3 and up this is available from the convenience function `glfwGetWindowContentScale`.  

See the behavior here...
ezgif-7-c56ed2a1f02a

jherico:
## Description

Closes #76 

## General Checklist:

### DIFF
--- a/framework/gui.cpp
+++ b/framework/gui.cpp
@@ -1,5 +1,5 @@
 /* Copyright (c) 2018-2020, Arm Limited and Contributors
- * Copyright (c) 2019, Sascha Willems
+ * Copyright (c) 2019-2020, Sascha Willems
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -40,6 +40,7 @@ VKBP_ENABLE_WARNINGS()
 #include "core/shader_module.h"
 #include "imgui_internal.h"
 #include "platform/filesystem.h"
+#include "platform/window.h"
 #include "rendering/render_context.h"
 #include "timer.h"
 #include "utils/graphs.h"
@@ -93,9 +94,10 @@ const ImGuiWindowFlags Gui::options_flags = Gui::common_flags;
 
 const ImGuiWindowFlags Gui::info_flags = Gui::common_flags | ImGuiWindowFlags_NoInputs;
 
-Gui::Gui(VulkanSample &sample_, const float dpi_factor, const float font_size, bool explicit_update) :
+Gui::Gui(VulkanSample &sample_, const Window &window, const float font_size, bool explicit_update) :
     sample{sample_},
-    dpi_factor{dpi_factor},
+    content_scale_factor{window.get_content_scale_factor()},
+    dpi_factor{window.get_dpi_factor() * content_scale_factor},
     explicit_update{explicit_update}
 {
 	ImGui::CreateContext();
@@ -1014,8 +1016,8 @@ bool Gui::input_event(const InputEvent &input_event)
 	{
 		const auto &mouse_button = static_cast<const MouseButtonInputEvent &>(input_event);
 
-		io.MousePos = ImVec2{mouse_button.get_pos_x(),
-		                     mouse_button.get_pos_y()};
+		io.MousePos = ImVec2{mouse_button.get_pos_x() * content_scale_factor,
+		                     mouse_button.get_pos_y() * content_scale_factor};
 
 		auto button_id = static_cast<int>(mouse_button.get_button());
 
--- a/framework/gui.h
+++ b/framework/gui.h
@@ -36,6 +36,8 @@
 
 namespace vkb
 {
+class Window;
+
 /**
  * @brief Helper structure for fonts loaded from TTF
  */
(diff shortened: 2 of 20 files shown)


framework/gui.cpp
```diff
@@ -93,11 +94,16 @@ const ImGuiWindowFlags Gui::options_flags = Gui::common_flags;
 
 const ImGuiWindowFlags Gui::info_flags = Gui::common_flags | ImGuiWindowFlags_NoInputs;
 
-Gui::Gui(VulkanSample &sample_, const float dpi_factor, const float font_size, bool explicit_update) :
+Gui::Gui(VulkanSample &sample_, const std::unique_ptr<Window> &window, const float font_size, bool explicit_update) :
```
ZandroFargnoli: Why are we passing a pointer? When would we want the window to be null? This should be a const reference IMO

framework/gui.cpp
```diff
@@ -93,11 +94,16 @@ const ImGuiWindowFlags Gui::options_flags = Gui::common_flags;
...
     sample{sample_},
-    dpi_factor{dpi_factor},
     explicit_update{explicit_update}
 {
+	if (window)
+	{
+		retina_factor = window->get_retina_factor();
```
ZandroFargnoli: If we pass the reference to the window in the constructor, it means it has to be initialised. We can then initialise these values up inside the member initialisation braces

framework/platform/glfw_window.h
```diff
@@ -44,6 +44,8 @@ class GlfwWindow : public Window
 
 	float get_dpi_factor() const override;
 
+	float get_retina_factor() const override;
```
ZandroFargnoli: `Retina` is an Apple concept/brand. This should really just be `get_x_content_scale` as this doesn't necessarily pertain to only macOS, but all platforms.

framework/gui.h
```diff
@@ -470,6 +472,9 @@ class Gui
 	/// Scale factor to apply to the size of gui elements (expressed in dp)
 	float dpi_factor{1.0f};
 
+	///  Scale factor to apply due to a difference between the window and GL pixel sizes
+	float retina_factor{1.0f};
```
ZandroFargnoli: Rename to `x_content_scale_factor`

framework/platform/platform.h
```diff
@@ -100,6 +100,11 @@ class Platform
 
 	Window &get_window() const;
 
+	const std::unique_ptr<Window> &get_window_ptr() const
```
ZandroFargnoli: Why does this function need to be added? We already have `Window &get_window() const;`, which invalidates the checklist item `I have used existing framework/helper functions where possible`

ZandroFargnoli: If you have an argument for adding this, can we place implementation inside the `.cpp`

framework/platform/window.h
```diff
@@ -67,6 +67,14 @@ class Window
 	 */
 	virtual float get_dpi_factor() const = 0;
 
+	/**
+     * @return The scale factor for systems with heterogenous window and pixel coordinates
+     */
+	virtual float get_retina_factor() const
```
ZandroFargnoli: Can implementation be kept inside the `.cpp` file?

ZandroFargnoli (changes requested):
Make sure you also modify `bldsys/cmake/template/sample/sample.cpp.in` so when people wish to generate a new sample it doesn't come with compilation errors out the box

framework/platform/window.h
```diff
@@ -67,6 +67,14 @@ class Window
 	 */
 	virtual float get_dpi_factor() const = 0;
 
+	/**
+     * @return The scale factor for systems with heterogenous window and pixel coordinates
```
ZandroFargnoli: `heterogenous` -> `heterogeneous`

framework/platform/window.h
ZandroFargnoli: And typo `heterogeneous` (not `heterogenous`)

framework/gui.cpp
jherico: I was trying to reproduce the behavior of the old signature, which allowed everything after `sample_` to be an optional parameter, but given the current usage pattern, I agree that passing a const ref is better.  will fix.

framework/gui.cpp
jherico: will fix

framework/platform/glfw_window.h
jherico: I've changed it to `get_content_scale_factor` because the fact that it's the x factor is part of an implementation detail inside glfw that supports non-uniform scaling (and thus non-square pixels).  I'd rather either support non-uniform scale all the way through (which would require further changes to `Gui` or just assume uniform scaling for now.

framework/gui.h
jherico: changed to `content_scale_facter` per above.

framework/platform/platform.h
jherico: will remove

framework/platform/window.h
jherico: will fix

framework/platform/window.h
jherico: will fix

framework/platform/glfw_window.h
ZandroFargnoli: Okay, I agree with this

ZandroFargnoli (approved):
LGTM

PR #97, opened 2020-03-28
OpenGL inter-operability example using VK_KHR_EXTERNAL_*

jherico:
## Description

A new sample that uses `VK_KHR_external_semaphore`, `VK_KHR_external_memory` along with platform specific extensions to show inter-operability between an OpenGL context and a Vulkan renderer.  

image

While many systems are migrating to Vulkan support, projects often still dependencies on third party libraries that might only support OpenGL, or where the Vulkan support is unstable or experimental.  For example, Qt supports rendering of it's QML user interface to an OpenGL framebuffer, but doesn't yet support rendering to Vulkan.  It is therefore valuable to be able to share resources between an OpenGL context and a Vulkan one to allow piecemeal migration of large systems.

The example creates an offscreen (hidden) OpenGL context and shares with it two semaphores as well as a block of memory used to back a texture in both Vulkan and OpenGL.  The OpenGL context uses the texture as a framebuffer color attachment and renders procedural content to it.  The Vulkan instance renders treats the texture as a sampled image texture and renders it to a 3D scene.  The bulk of the Vulkan side is copied from the `texture_loading` sample.  The OpenGL code is entirely original and based on my earlier OpenGL interoperability example in my fork of Sascha Willems Vulkan samples.  The procedural shader code is from [shadertoy](https://www.shadertoy.com/view/Xd23Dh), is by Inigo Quilez and is covered under the MIT license.

The memory allocation in the Vulkan side does _not_ use the VMA allocation routines used elsewhere in the code as I have no idea if it's possible to use the required `VkExportMemoryAllocateInfo` functionality with VMA.  

I have tested this on 

* nVidia RTX 2080 
  * Windows 10 64 bit with driver version 445.75
  * Ubuntu 19.10 with driver version 440.64

This PR depends directly on GLFW, so it won't build on Android.  Additionally, Mac OpenGL doesn't (and likely will never) support the required extension.

## General Checklist:

## Sample Checklist

- [X] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)


samples/extensions/open_gl_interop/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+add_sample_with_tags(
+    ID ${FOLDER_NAME}
+    CATEGORY ${CATEGORY_NAME}
+    AUTHOR "Bradley Austin Davis"
+    NAME "OpenGL Interoperability"
+    DESCRIPTION "Example showing sharing resources between OpenGL and Vulkan"
+    TAGS
+        "opengl")
+
+endif()
```
rex-remind101, line 57: possibly missing a newline

TomAtkinsonArm:
Great work!

I've had a quick look over this sample and currently have one concern. You are tightly coupled with GLFW; I'm assuming this is for creating the OpenGL context. Is there anyway we can support more platforms by changing this? Specifically Android?

jherico:
@TomAtkinsonArm I've added a codepath for Android using EGL offscreen contexts.  However, I'm almost certain the current code will fail to build on `DIRECT_TO_DISPLAY` setups and I have no way to testing the Android code path on any of my devices.

ghost:
Hi @jherico, thanks for submitting this sample, it is a popular one internally and we think it is a nice addition. 

Quick question, why the GLAD fork instead of the main version?

jherico:
The primary glad repository contains not an instance of glad, but code for a glad generator.  In theory I could have integrated the main repository and used CMake to execute the generator to produce the appropriate GL loader, but that would have made the build process significantly more fragile and added a dependency on Python to execute the generator.  

Instead I opted to create a fork of glad, and then create branches with the output of the generator needed for the two specific API's, OpenGL and OpenGL ES.  

There are a few possible paths...

* Leave it as it is
* Leave the glad dependencies as submodules, but switch to a fork of Glad owned by KhronosGroup
* Switch to embedding the glad dependencies as files under third-party instead of submodules
  * Unfortunately this would require fixing the CI servers to ignore these files in the linting and copyright headers checks
* Switch to the main glad repository and modify the CMake process to call the generator appropriately and create these files at configure or build time.

aurolou (changes requested):
When I run the sample, I get following errors:

[error] [samples\extensions\open_gl_interop\open_gl_interop.cpp:145] OpenGL: glTextureStorageMem2DEXT has generated an error (GL_INVALID_OPERATION)
[error] [samples\extensions\open_gl_interop\open_gl_interop.cpp:145] OpenGL: glFramebufferTexture failed because (depending on the operation) a referenced binding point is empty; a referenced name is not the name of an object; or the given name is otherwise not valid to this operation (GL_INVALID_VALUE)
[error] [samples\extensions\open_gl_interop\open_gl_interop.cpp:145] OpenGL: glDrawArrays has generated an error (GL_INVALID_FRAMEBUFFER_OPERATION)
[error] [samples\extensions\open_gl_interop\open_gl_interop.cpp:145] OpenGL: glDrawArrays failed because the currently set GL_DRAW_FRAMEBUFFER binding has incomplete status 'GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT' (GL_INVALID_FRAMEBUFFER_OPERATION)

Tested on a system with
RX 5700 XT
Driver: 20.4.1

TomAtkinsonArm:
Is there a specific reason why we depend on GL 4.6 and not a more widely supported version. Also do you know if there is a fallback method of achieving interoperability? Or if we should gracefully inform the user that there device is not supported? I say this as currently my laptop only supports upto GL4.5 and does not support the GL_EXT_semaphores

jherico:
@TomAtkinsonArm no, there's no good reason to specify anything over 4.3.

As for fallback mechanisms.... that's a more complicated topic.  If the Vulkan side doesn't support the required extensions then I'm assuming it will fail initialization due to the `add_instance_extension` calls in the sample ctor, long before it ever gets to OpenGL initialization. 

If a Vulkan driver **does** support the appropriate external objects extensions, then presumably an OpenGL context backed by the same hardware and driver authors would support the corresponding required OpenGL extensions.  It wouldn't make much sense for an ISV to release one half and not the other, so I'm not sure there's much point in writing a bunch of code to handle the case of "ISV driver authors were being jerks".

While there might be fallback approaches to conveying images between Vulkan and OpenGL in the absence of the extension, the sample is specifically about the extension itself.

jherico:
@aurolou I was able to reproduce what you're seeing.

When switching from the DSA code to the non-DSA code I made a small error (failed to force the `color` texture instance to be a 2D texture through some other means than `glCreateTextures(GL_TEXTURE_2D, 1, &color)`

Adding a `glBindTexture(GL_TEXTURE_2D, color)` after the `glGenTextures(1, &color)` seems to have fixed the problem

jherico:
@KhronosKris I've modified the use of glad to rely on the primary repository.

jherico:
@KhronosKris I rolled back my change to the glad dependency because of the travis build failures, which I'm not in a position to fix:  https://travis-ci.com/github/KhronosGroup/Vulkan-Samples/jobs/320793427

To elaborate, the GLAD generation process wants to fetch content over the network and is using a Python2 script to do so.  The Python script is failing due to SSL errors.  In theory, you can set a value in the environment to tell it to disregard SSL validity, but my attempt to do so from the Vulkan-Examples CMake file doesn't seem to work.  That CMake invokes another CMake execution to configure GLAD, then yet another to execute the build.  That build process is indirectly executing the Python script, so it's not super surprising that an environment variable set at the top doesn't make it down to the actual script.  I'll look into the possibility of having a patch that gets applied at configure time in order to resolve the problem.

ghost:
> The primary glad repository contains not an instance of glad, but code for a glad generator. In theory I could have integrated the main repository and used CMake to execute the generator to produce the appropriate GL loader, but that would have made the build process significantly more fragile and added a dependency on Python to execute the generator.
> 
> Instead I opted to create a fork of glad, and then create branches with the output of the generator needed for the two specific API's, OpenGL and OpenGL ES.
> 
> There are a few possible paths...
> 
> * Leave it as it is
> * Leave the glad dependencies as submodules, but switch to a fork of Glad owned by KhronosGroup
> * Switch to embedding the glad dependencies as files under third-party instead of submodules
>   
>   * Unfortunately this would require fixing the CI servers to ignore these files in the linting and copyright headers checks
> * Switch to the main glad repository and modify the CMake process to call the generator appropriately and create these files at configure or build time.

This makes total sense, thanks @jherico. 

I think that options 2 or 4 are probably the most logical, I'll create and issue and discuss with some of the others.

jherico:
@KhronosKris I'm still pursuing option 2.  My initial attempt worked locally but failed on Travis because of the Python 2 SSL issue.  I've created a small patch to the GLAD code to prefer Python 3, and I'm trying to see if it resolves the SSL issue.  I've also created a PR against the upstream GLAD repo, that if accepted means I won't need my fork (or presumably a Khronos hosted repository).

Dav1dde:
Personally I'd recommend to embedd the generated files and update them when needed (e.g. new extensions).

Embedding via cmake is also a viable option, a while ago I added the specs to the glad repository and you can use the `GLAD_REPRODUCIBLE` to use the embedded files, which also means if the specification gets updated in the future and breaks glad your build is still going to succeed.

Sorry I just losely followed, if you need a patch to glad or anything else let me know.

jherico:
@Dav1dde I initially tried embedding the generated GLAD files directly, but any files that aren't in submodules get subjected to the code hygiene checks that validate against clang format, and look for specific copyright headers.  The GLAD files don't pass such checks.

My second attempt was to use generated GLAD files but hosted in my fork of GLAD on two branches...  but it's not appealing to the owners of this repository to have a submodule dependency on someone's random fork.  If they make a Khronos-owned repository with the same files I think that will satisfy everyone.

TomAtkinsonArm:
You should be able to add these files directly to .copyrightignore and they will be ignored from the check

aurolou (dismissed):

TomAtkinsonArm:
Hi @jherico,

I've been looking at your sample for the past few days. I cant seem to get the android version to work on my device that supports the extension. Sorry for adding this requirement at the beginning i didn't realize how many issues it would uncover.

Currently there are some big issues in the framework that need to be resolved before android support can be adopted in this sample - One of these being custom windows so we can properly support the desktop and android windows that you have created.

Unfortunately, I will not be able to revisit this sample properly for the next month as I am currently working on these framework changes and other responsibilities. However I still believe this could be merged in a partial state.

Option 1:
Revert android changes. Normal PR stuff. Merge as a windows only sample. Revisit at a later date with wider platform support.

Option 2:
Wait for platform changes - could take some time. Make required changes for the new platform. Normal PR stuff.

Option 3:
Any other ideas?

Stuff that can be changed now:
- Add extension description to samples/extensions/README.md
- I'm going to code review today :)

framework/api_vulkan_sample.h
```diff
@@ -207,6 +207,13 @@ class ApiVulkanSample : public vkb::VulkanSample
 	 */
 	void draw_model(std::unique_ptr<vkb::sg::SubMesh> &model, VkCommandBuffer command_buffer);
 
+	/**
+	 * @brief Synchronously execute a block code within a command buffer, then submit the command buffer and wait for completion.
+	 * @param f a block of code which is passed a command buffer which is already in the begin state.
+	 * @param signalSemaphore An optional semaphore to signal when the commands have completed execution.
+	 */
+	void with_command_buffer(const std::function<void(VkCommandBuffer command_buffer)> &f, VkSemaphore signalSemaphore = nullptr);
```
TomAtkinsonArm: Should use VK_NULL_HANDLE instead of nullptr

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+				LOGW("OpenGL: {}", message);
+				break;
+			case GL_DEBUG_SEVERITY_LOW:
+				LOGI("OpenGL: {}", message);
+				break;
+			default:
+			case GL_DEBUG_SEVERITY_NOTIFICATION:
+				LOGD("OpenGL: {}", message);
+		}
+	}
```
TomAtkinsonArm: Can be moved to the sample class

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+			std::string strError;
+			strError.insert(strError.end(), errorLog.begin(), errorLog.end());
+
+			// Provide the infolog in whatever manor you deem best.
+			// Exit with failure.
+			glDeleteShader(shader);        // Don't leak the shader.
+			LOGE("OpenGL: Shader compilation failed", strError.c_str());
+		}
+		return shader;
+	}
```
TomAtkinsonArm: Can be moved to the sample class

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+		GLuint program = glCreateProgram();
+		GLuint vs      = loadShader(vertexShaderSource, GL_VERTEX_SHADER);
+		GLuint fs      = loadShader(fragmentShaderSource, GL_FRAGMENT_SHADER);
+		glAttachShader(program, vs);
+		glAttachShader(program, fs);
+		glLinkProgram(program);
+		glDeleteShader(vs);
+		glDeleteShader(fs);
+		return program;
+	}
```
TomAtkinsonArm: Can be used to the sample class

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+	{
+#if defined(VK_USE_PLATFORM_ANDROID_KHR)
+		timespec now;
+		clock_gettime(CLOCK_MONOTONIC, &now);
+		float result = now.tv_sec + float(now.tv_nsec / 1000000LL) / 1000.0f;
+		return result;
+#else
+		return glfwGetTime();
+#endif
+	}
```
TomAtkinsonArm: Not needed see - framework/timer.h

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+		glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 4);
+		glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 6);
+		glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
+		glfwWindowHint(GLFW_OPENGL_DEBUG_CONTEXT, 1);
+		glfwWindowHint(GLFW_VISIBLE, GLFW_FALSE);
+		window = glfwCreateWindow(SHARED_TEXTURE_DIMENSION, SHARED_TEXTURE_DIMENSION, "OpenGL Window", nullptr, nullptr);
+		glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
+		glfwMakeContextCurrent(window);
+		gladLoadGL();
+#endif
```
TomAtkinsonArm: It would be clearer to have an Android window and a Desktop Window (in this case Windows only Window). However framework changes are needed to implement these custom windows

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+
+		glDebugMessageCallback(debugMessageCallback, NULL);
+		glEnable(GL_DEBUG_OUTPUT_SYNCHRONOUS);
+
+		program   = buildProgram(OPENG_VERTEX_SHADER, OPENGL_FRAGMENT_SHADER);
+		startTime = getTime();
```
TomAtkinsonArm: Can use Timer from framework/timer.h

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+
+		glUseProgram(program);
+		glProgramUniform3f(program, 0, (float) SHARED_TEXTURE_DIMENSION, (float) SHARED_TEXTURE_DIMENSION, 0.0f);
+
+		glViewport(0, 0, SHARED_TEXTURE_DIMENSION, SHARED_TEXTURE_DIMENSION);
+	}
+
+	void render()
+	{
+		float time = (float) (getTime() - startTime);
```
TomAtkinsonArm: Again this is where timer would be useful :)

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+		glBindFramebuffer(GL_FRAMEBUFFER, fbo);
+		glFramebufferTexture(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, color, 0);
+
+		glUseProgram(program);
+		glProgramUniform3f(program, 0, (float) SHARED_TEXTURE_DIMENSION, (float) SHARED_TEXTURE_DIMENSION, 0.0f);
+
+		glViewport(0, 0, SHARED_TEXTURE_DIMENSION, SHARED_TEXTURE_DIMENSION);
+	}
+
+	void render()
```
TomAtkinsonArm: could this be moved to the update method in the sample class

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+#endif
+	}
+
+  private:
+#if defined(VK_USE_PLATFORM_ANDROID_KHR)
+	EGLConfig  eglConf{nullptr};
+	EGLSurface eglSurface{nullptr};
+	EGLContext eglCtx{nullptr};
+	EGLDisplay eglDisp{nullptr};
+#else
```
TomAtkinsonArm: Would EGL_NO_CONTEXT, EGL_NO_SURFACE, EGL_NO_DISPLAY be better defaults?

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+
+		glGenFramebuffers(1, &fbo);
+		glBindFramebuffer(GL_FRAMEBUFFER, fbo);
+		glFramebufferTexture(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, color, 0);
+
+		glUseProgram(program);
+		glProgramUniform3f(program, 0, (float) SHARED_TEXTURE_DIMENSION, (float) SHARED_TEXTURE_DIMENSION, 0.0f);
+
+		glViewport(0, 0, SHARED_TEXTURE_DIMENSION, SHARED_TEXTURE_DIMENSION);
+	}
```
TomAtkinsonArm: GL stuff that is not part of windowing could be moved to prepare in OpenGlInterop, The window stuff could be moved to a `create_window` method?

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+		// Failure to flush the operation can cause the GL driver to sit and wait for
+		// sufficient additional commands in the buffer before it flushes automatically
+		// but depending on how the waits and signals are structured, this may never
+		// occur.
+		glFlush();
+	}
+
+	void destroy()
+	{
+		glFinish();
```
TomAtkinsonArm: destroy could be moved to the applications `virtual void finish();` see framework/platform/application.h

samples/extensions/open_gl_interop/open_gl_interop.cpp
```diff
@@ -0,0 +1,931 @@
...
+		},
+		                    sharedSemaphores.glReady);
+	}
+}
+
+void OpenGLInterop::prepare_opengl_context()
+{
+	glWindow = std::make_unique<OpenGLWindow>();
+	glWindow->create(shareHandles, sharedTexture.allocationSize);
+}
```
TomAtkinsonArm: Referring to my other comment, the windowing could be initiated here

samples/extensions/open_gl_interop/open_gl_interop.h
```diff
@@ -0,0 +1,114 @@
...
+struct VertexStructure
+{
+	float pos[3];
+	float uv[2];
+	float normal[3];
+};
+
+class OpenGLInterop : public ApiVulkanSample
+{
+	friend class OpenGLWindow;
```
TomAtkinsonArm: A lot of the functionality of this friend class can be moved into `OpenGLInterop` making it less useful. I think at some point custom windows will be needed for different platforms to initialize opengl but currently the friend class confuses things. My previous comments mention moving functionality back into `OpenGLInterop`

samples/extensions/open_gl_interop/open_gl_interop.h
```diff
@@ -0,0 +1,114 @@
...
+	{
+		Handle memory{INVALID_HANDLE_VALUE};
+		Handle glReady{INVALID_HANDLE_VALUE};
+		Handle glComplete{INVALID_HANDLE_VALUE};
+	} shareHandles;
+
+	struct SharedTexture
+	{
+		VkImage        image{nullptr};
+		VkDeviceMemory memory{nullptr};
```
TomAtkinsonArm: VK_NULL_HANDLE

samples/extensions/open_gl_interop/open_gl_interop.h
```diff
@@ -0,0 +1,114 @@
...
+		VkDeviceSize   size{0};
+		VkDeviceSize   allocationSize{0};
+		VkSampler      sampler{nullptr};
+		VkImageView    view{nullptr};
```
TomAtkinsonArm: VK_NULL_HANDLE

samples/extensions/open_gl_interop/open_gl_interop.h
```diff
@@ -0,0 +1,114 @@
...
+		//VkImageLayout  image_layout{VK_IMAGE_LAYOUT_UNDEFINED};
+
+	} sharedTexture;
+
+	struct Semaphores
+	{
+		VkSemaphore glReady{nullptr};
+		VkSemaphore glComplete{nullptr};
```
TomAtkinsonArm: VK_NULL_HANDLE

samples/extensions/open_gl_interop/open_gl_interop.h
```diff
@@ -0,0 +1,114 @@
...
+	{
+		glm::mat4 projection;
+		glm::mat4 model;
+		glm::vec4 view_pos;
+	} ubo_vs;
+
+	VkPipeline            pipeline{nullptr};
+	VkPipelineLayout      pipeline_layout{nullptr};
+	VkDescriptorSet       descriptor_set{nullptr};
+	VkDescriptorSetLayout descriptor_set_layout{nullptr};
```
TomAtkinsonArm: VK_NULL_HANDLE

TomAtkinsonArm (dismissed):
- I've tested on windows with success
- Android didn't end up working and i couldn't figure out the fix.
- Not using `VK_NULL_HANDLE` broke my android build hence the comments
- A lot of the android stuff can be removed if we choose to delay supporting this platform

A nit pick would be using snake case consistently for variables and function names to match the rest of the project. This however can be rectified at a later date with find and replace so isn't a huge concern

Don't forget to add a paragraph about the sample in the samples/extensions/README.md.

Good Job! :+1:

samples/extensions/open_gl_interop/open_gl_interop.cpp
TomAtkinsonArm: I was confused between `OpenGLWindow::render()`, `OpenGLInterop::render(float)` and `draw()`. Could all these be concatenated together in `OpenGLInterop::render(float)`?

It may result in a bigger function but would be less confusing

samples/extensions/open_gl_interop/open_gl_interop.cpp
TomAtkinsonArm: Or moved into `prepare_opengl_context`

samples/extensions/open_gl_interop/open_gl_interop.cpp
jherico: If you mean a visible OpenGL backed window where the source texture is displayed, I did consider that and used it while debugging, but you run into all sorts of issues related to v-sync.  

Without any special effort, the GL window will be locked to the refresh rate of the device (usually 60 Hz) and because the example is single threaded, that results in the Vulkan output also being locked to that rate or slower.

samples/extensions/open_gl_interop/open_gl_interop.h
jherico: Moving the helper class functionality into the main sample class pretty much requires that I move all the GLAD / EGL headers from `open_gl_interop.cpp` into `open_gl_interop.h`.  Alternatively I could move the functions as you suggest and have `open_gl_interop.h` forward declare a `struct OpenGLState` type and declare a pointer to it.   But even then, moving a function like `static void debugMessageCallback(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length, const GLchar *message, const void *userParam)` into the header implies that `GLenum`, `GLsizei`, etc are all exposed.  

I don't really anticipate other examples needing to use OpenGL so I wanted to encapsulate it's use to this one PR as much as possible.

jherico:
Thanks for all the feedback Tom

TomAtkinsonArm:
> Hi @jherico,
> 
> I've been looking at your sample for the past few days. I cant seem to get the android version to work on my device that supports the extension. Sorry for adding this requirement at the beginning i didn't realize how many issues it would uncover.
> 
> Currently there are some big issues in the framework that need to be resolved before android support can be adopted in this sample - One of these being custom windows so we can properly support the desktop and android windows that you have created.
> 
> Unfortunately, I will not be able to revisit this sample properly for the next month as I am currently working on these framework changes and other responsibilities. However I still believe this could be merged in a partial state.
> 
> Option 1:
> Revert android changes. Normal PR stuff. Merge as a windows only sample. Revisit at a later date with wider platform support.
> 
> Option 2:
> Wait for platform changes - could take some time. Make required changes for the new platform. Normal PR stuff.
> 
> Option 3:
> Any other ideas?
> 
> Stuff that can be changed now:
> 
> * Add extension description to samples/extensions/README.md
> * I'm going to code review today :)

gary-sweet (dismissed):
I can only do a build sanity check for this, and it doesn't currently build for platforms that don't have glfw support.

```
Vulkan-Samples/samples/extensions/open_gl_interop/open_gl_interop.cpp:40:25: fatal error: GLFW/glfw3.h: No such file or directory
 # include <GLFW/glfw3.h>
                         ^
compilation terminated.
```

jherico:
I'll take a look at that.  I'm trying to avoid making the check super-complicated, but it looks like I'm going to have to do some preprocessor hackery.

TomAtkinsonArm:
Hi @jherico,
I'm back from my absence, what do we need to do to get this merged? Can i help in any way? Have you made any more progress?

jherico:
As far as I know the main blocking issue is getting the code compiling on non-GLFW platforms.  I'll try to find time for that this week.

@TomAtkinsonArm you suggested that a lot of the `OpenGLWindow` class functionality could be moved into the main example class, but I pointed out that would almost certainly require moving the OpenGL headers into the main example header, which was undesirable.

jherico:
@gary-sweet should now build properly on non-GLFW systems (Direct2Display and Android) but doesn't do anything on those systems.

TomAtkinsonArm:
Hi @jherico,

### Android

I spent a week trying to add Android support. I don't think that it is currently achievable, at least not on the devices i have to hand. When using a pixel 3, which supports the required extensions, vkGetSemaphoreFdKHR results in VK_ERROR_OUT_OF_HOST_MEMORY and i have not made any substantial progress that would fix this since.

Because of this it may be best to leave the android support here and pick it up at a later date if we have time or if device support improves. In [OffscreenContext Proposal](https://github.com/jherico/Vulkan-Samples/pull/3) the android implementation has moved to a single `ifdef` in `offscreen_context.cpp`. This segment can be commented with a depreciation not or a to do note.

In [GLAD submodule proposal](https://github.com/jherico/Vulkan-Samples/pull/2), I have demonstrated (using your last commit changes) how we can omit the sample from specific platforms without the need for a big ifdef around the sample.

### Proposals

I don't see the extra time debugging this as a loss as i think there are still a few aspects that might be tweaked with the sample. I felt it best to create a proposal hopefully you will see a nice diff in the PR's. This way i am not wasting your time with review comments and having cryptic conversations.

[OffscreenContext Proposal](https://github.com/jherico/Vulkan-Samples/pull/3) - Possibly a replacement for the friend class
[GLAD submodule proposal](https://github.com/jherico/Vulkan-Samples/pull/2) - Build system tweaks (GLAD)

I don't want it to feel like these are being forced on to your sample it was merely just an easier way to communicate my ideas :)
If you have specific comments about these proposals then their PR's are the best place to take the discussion.

### Other Errors

On windows in gl_interop the desired image is not outputted and i have validation layer errors. This is not a priority as the sample runs and i would much rather put in the effort to resolve this after with an issue after the release of this sample.

Thank you for all the effort so far. Apologies that my replies have not been too consistent.

jherico:
@TomAtkinsonArm thanks for your feedback and the obvious time and effort put into these proposals.  I've integrated both of them into my code.  However, my main concern about migrating from a submodule-based version of glad to one embedded directly in the primary repository is that the CI checks will now "see" these files, so additional work has to be done to make sure they either pass, or are ignored by, checks like the copyright header and formatting check.

I've done my best to that end in [this commit](https://github.com/KhronosGroup/Vulkan-Samples/pull/97/commits/a282988955363c13377436e68e74f69989eee5cb) by using clang to reformat the GLAD files and adding the file names to `.copyrightignore`.  However, I don't see any available mechanism to make the clang-tidy checks ignore files.  Further, since the python scripts that actually execute these checks are in the docker image, not in this repository, there's no easy means of adding such a mechanism, at least not from here. 

I _could_ copy the existing scripts from the docker `/usr/local/bin` folder into `bldsys/ci` (or something similar) and then modify the Travis yml to reference the local location, and _then_ I'd be able to modify the tidy check to ignore files based on a pattern, similar to the copyright check.  Alternatively, I could create a new docker image that makes the required changes.

TomAtkinsonArm:
It looks like the only job that failed so far, other then snake case check, was the Linux build when linking the framework to tests. This will be because of my change of OBJECT library to STATIC library. It should be a quick fix.

TomAtkinsonArm:
Added another PR to fix the CI Linux build [Link test framework with test targets](https://github.com/jherico/Vulkan-Samples/pull/4)

TomAtkinsonArm:
Fixes #106

gary-sweet:
I can confirm that this no longer breaks the direct-to-display build. Thanks.
Since I can't test the rest of it, I don't feel it's appropriate to add an official 'review' however.

AlejandroCosin (approved):

gary-sweet (approved):
Builds ok for Broadcom, but can't test the actual functionality.

r-potter (approved):
Builds fine on Android. Unable to test.

aurolou (approved):

Issue #28, opened 2020-01-05
Texture and scene files are missing for the texture_mipmap_generation sample


jherico:
The [cpp file](https://github.com/KhronosGroup/Vulkan-Samples/blob/af0dbc2ec3a872688265e355d997070fd50813ac/samples/api/texture_mipmap_generation/texture_mipmap_generation.cpp#L318) for this example references `textures/checkerboard_rgba.ktx` and `scenes/tunnel_cylinder.gltf` but these files are not present under the assets folder.  

Running the example throws an exception when the KTX loader fails to find the file.

SaschaWillems:
The PR containing those files hasn't been merged into master yet: https://github.com/KhronosGroup/Vulkan-Samples-Assets/pull/5

@wasimabbas-arm: Can you take a look at this and merge the asset PR if possible?

jherico:
@SaschaWillems thanks... I chekced the PRs on this repository and looked at the assets repo, but didn't think to check the PRs there.  This tends to be the problem with git submodules in my experience... keeping them in sync becomes a pain in the ass.

ghost:
Asset now merged in the asset repo.

jherico:
The upstream repository has had the file added, but _this_ repository hasn't had it's submodule commit ID updated.  Also, the current submodule commit 1c6c656 isn't in the master branch, so either 1c6c656 needs to get merged into master or master needs to be merged into 1c6c656, and the resulting commit added to the assets repo, and referenced in this repo.

wasimabbas-arm:
https://github.com/KhronosGroup/Vulkan-Samples/pull/45 fixes that please close the issue if you agree.

Issue #26, opened 2019-12-19
The `colorAttachmentCount` member of default subpass in `VkSubpassDescription` is not set


unknownue:
If `subpasses` is empty, the code should create single subpass with a ColorAttachment by default.
It seems that the the `colorAttachmentCount` member of this subpass in `VkSubpassDescription` is not set.

https://github.com/KhronosGroup/Vulkan-Samples/blob/6c17e141670b5ab515d4f7b0a380c0224f5e775d/framework/core/render_pass.cpp#L122-L147

PR #99, opened 2020-04-08
Constant data sample

ZandroFargnoli:
## Description

This PR adds a new sample called Constant Data.

Constant Data is a sample designed to show a few of the different methods in Vulkan to use constant data in shaders, along with the performance implications they may bring. (Read the [tutorial ](https://github.com/KhronosGroup/Vulkan-Samples/blob/46628d42918ff09cf6d6687707ba5613eec8d74f/samples/performance/constant_data/constant_data_tutorial.md) for full rundown)

(Khronos Group members: see Gitlab issue #1423)

This PR adds the following framework changes:
* We now defer calls to `vkCmdPushConstants` to draw time. This is since pipeline layouts are flushed at the draw call, there was a chance that we could be invalidating our push constants if the two pipeline layouts ranges don't match.
* Geometry subpass has been reworked to give derived subpasses a means to override certain aspects of the subpass.
* Added load/store cycles stat.
* Descriptor set layouts now know their set binding.
* Added `VK_BUFFER_USAGE_STORAGE_BUFFER_BIT` to render frame buffer pools, added a multiplier for the default block size.

## General Checklist:

## Sample Checklist

Tested on:
* Mali S10 (Android)
* Adreno S10 (Android)
* Desktop


gary-sweet:
I have a couple of problems when running this new sample.

1) When I select 'Descriptor Sets' or 'Dynamic Descriptor Sets' I get a stream of errors on the console which just keep repeating:

```
[debug] Building #26 cache object (N3vkb12ShaderModuleE)
[debug] Building #27 cache object (N3vkb12ShaderModuleE)
[debug] Building #28 cache object (N3vkb12ShaderModuleE)
[debug] Building #29 cache object (N3vkb12ShaderModuleE)
[debug] Building #30 cache object (N3vkb12ShaderModuleE)
[debug] Building #31 cache object (N3vkb12ShaderModuleE)
[error] [framework/buffer_pool.cpp:140] Ignore buffer allocation update
[debug] Building #3 cache object (N3vkb14PipelineLayoutE)
[debug] Building #3 cache object (N3vkb19DescriptorSetLayoutE)
[debug] Building #5 cache object (N3vkb16GraphicsPipelineE)
[debug] Building #3 cache object (N3vkb14DescriptorPoolE)
[debug] Building #0 cache object (N3vkb13DescriptorSetE)
[error] [framework/buffer_pool.cpp:140] Ignore buffer allocation update
[debug] Building #1 cache object (N3vkb13DescriptorSetE)
[error] [framework/buffer_pool.cpp:140] Ignore buffer allocation update
[debug] Building #2 cache object (N3vkb13DescriptorSetE)
...
```
2) When I select 'Single Pre-allocated Buffer Array' half the geometry in the scene goes black.

gary-sweet:
3) You appear to have deleted shaders/triangle.frag which is needed by a couple of samples.

gary-sweet:
Re 1) : Your handling of 'push_constant_limit' seems wrong. It looks like you are either expecting 128 or 256 push constants. That might be the case on Arm GPUs, but our limit is 134,217,727‬. Simply creating a UBO that size doesn't really seem like a sensible plan, nor does checking for exactly 256 in several places.

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+- [Update-after-bind Descriptor Sets](#update-after-bind-descriptor-sets)
+- [Buffer Object Arrays](#buffer-object-arrays)
+- [Further reading](#further-reading)
+- [Best practice summary](#best-practice-summary)
+
+## **What is constant data?**
+
+### **Introduction**
+
+Constant data is a form of information that is supplied to our graphics pipeline to help with shader computations.
```
gary-sweet: "graphics pipelines" rather than "our graphics pipeline"

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+It differs from other data (e.g. input vertex data) in the sense that it remains *constant* across every shader invocation of a draw call.
+
+This is important as because of this assumption, the data can be **shared** between shader stages, as we know it isn't going to be changed throughout the runtime of a single draw call in a render pipeline.
+
+The next section aims to cover the constant data theory, starting at the shader level before moving to the basics of how to plug in your data using Vulkan.
+
+### **Constant data in Vulkan shaders**
+
+Constant data is implemented in our shader code by using global variables.
```
gary-sweet: Remove "our"

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+*You can read more about shader stage inputs and outputs [here](https://www.khronos.org/opengl/wiki/Type_Qualifier_(GLSL)#Shader_stage_inputs_and_outputs).*
+
+#### *Constant Types*
+
+Constant types are global variables that have either the `uniform` or `buffer` storage type, these are *uniform buffer objects* and *shader storage buffer objects* respectively. They describe data which remains constant across an entire draw call, meaning that the values stay the same across the different shader stages and shader invocations.
+
+**Uniform buffer objects (UBOs)** are the more commonly used of the two. They can hold up to 16KB of data. It is important to note that they are not *compile-time* constant, so they are not given the `const` type. However, trying to edit the data in shader code will result in a compile-time error. These values use a `layout binding` and, when working with multiple `VkDescriptorSet`s, we will even give it a `layout set`. *You can read more about UBOs [here](https://www.khronos.org/opengl/wiki/Uniform_Buffer_Object).*
```
gary-sweet: > They can hold up to 16KB of data.

On Arm maybe. Our limit is much higher, as are other vendors.

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+**Shader storage buffer objects (SSBOs)** are like special types of uniform buffer objects, denoted by the storage type `buffer`. Unlike UBOs they can be written to, meaning the values *can* be changed in the shaders so therefore they don't always represent data that is constant. Having said this, they can hold up to 128MB of data as opposed to the 16KB with UBOs. *You can read more about SSBOs [here](https://www.khronos.org/opengl/wiki/Shader_Storage_Buffer_Object).*
```
gary-sweet: > they can hold up to 128MB of data as opposed to the 16KB with UBOs

Arm specific again. We have much higher limits.

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+*You can read more about interface blocks [here](https://www.khronos.org/opengl/wiki/Interface_Block_(GLSL)).*
+
+### **Vulkan API**
+
+We've covered how constant data is implemented in the shader, however to push the data from the application to the shader we need to use Vulkan.
+
+We do this with the use of `VkBuffer`s, which is Vulkan's implementation of buffer memory.
+
+Buffers in Vulkan are just chunks of memory used for storing data, which can be read by the graphics card.
```
gary-sweet: > graphics card

GPU

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+* [Inline uniform buffer objects](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_inline_uniform_block.html) (click to read more)
+* [Push descriptors](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_push_descriptor.html) (click to read more)
+
+**Inline uniform buffer objects** and **push descriptors** are not covered by this tutorial, please use the links above to learn more about them.
+
+## **Sample Overview**
+
+### **Introduction**
+
+The sample uses a mesh heavy scene which has 1856 meshes with 475 KB of mesh data. This is to demonstrate a use case where many different calls to pushing uniform data will occur during a single frame. The constant data that is being sent is the per-mesh model matrix, the camera view projection matrix, and if 256 bytes of push constants are available it will push a scale matrix and some extra padding.
```
gary-sweet: > and if 256 bytes of push constants are available

Needs to be generalised

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+A performance graph is displayed at the top with two charts, one showing frame time, and one showing the load/store cycles.
```
gary-sweet, line 168: Only on Arm systems right now

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+When an option is changed, the descriptor sets are flushed and recreated with their new setup, and the respective render pipeline/subpass.
+
+## **Push Constants**
+
+### **Introduction**
+
+Push constants are usually the first method newer Vulkan programmers will stumble upon when beginning to work with constant data. They are straightforward to use and integrate nicely into any codebase, making them a great option to send simple data to your shaders.
+
+The downside to push constants is that they have very strict limitations on how much data can be sent.
```
gary-sweet: on some platforms.

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+The Vulkan spec requires drivers to support at least 128 bytes of push constants, and most modern implementations of Vulkan will support 256 bytes.
```
gary-sweet: or much more. It is important to query the limits for the device to determine the sensible choice at runtime.

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+### **Introduction**
+
+Push constants are usually the first method newer Vulkan programmers will stumble upon when beginning to work with constant data. They are straightforward to use and integrate nicely into any codebase, making them a great option to send simple data to your shaders.
+
+The downside to push constants is that they have very strict limitations on how much data can be sent.
+The Vulkan spec requires drivers to support at least 128 bytes of push constants, and most modern implementations of Vulkan will support 256 bytes. 
+
+***Note:** To determine how many bytes your system supports, you can query the physical device for its `VkPhysicalDeviceLimits` and check the value `maxPushConstantsSize`.*
+
+Having said this, 128/256 bytes is still a useful amount of data, even if it isn't exactly scalable to a full game scenario. In the case of 128 bytes, we can at least send two float 4x4 matrices (2 * 4 * 16 = 128). This, for example, can hold our world matrix and our view-projection matrix.
```
gary-sweet, line 195: **even** 128/256 bytes is still useful

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+    uint32_t                                    offset,
+    uint32_t                                    size,
+    const void*                                 pValues);
+```
+
+### **Performance**
+
+![push constants](images/push_constants_performance.jpg)
+
+In early implementations of Vulkan, this was usually the fastest way of pushing data to your shaders. In more recent times, we have observed that *overall* they can be slower. If performance is something you are trying to maximise, descriptor sets may be the way to go.
```
gary-sweet: This seems like a platform specific statement.

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+For all the descriptor set sections below, we will use one such resource known as a **buffer object**, as this will be what we use to store our MVP data.
+
+A **buffer object** in Vulkan is a type of `VkBuffer`, created with the respective buffer usage flag. For uniform buffer objects we use the `VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT` flag, and for shader storage buffer objects we use the `VK_BUFFER_USAGE_STORAGE_BUFFER_BIT` flag. These map to the shader `<storage>` type `uniform`s and `buffer`s respectively. 
+
+### **Performance**
+
+![single ubo](images/descriptor_set_performance.jpg)
+
+While it is not straightforward to perform a 1:1 comparison between push constants and descriptor sets, the sample does show static descriptor sets outperforming push constants.
```
gary-sweet, line 267: outperforming push constants **on an Arm Mali.**

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+When we compare the frametime with [push constants](#push-constants) we can see it is the same (16.7ms), however it is the load/store cycles we want to look at. They drop from 266 k/s to 123 k/s, showing that the GPU is worked more in the case of push constants to achieve the same visual results.
```
gary-sweet: Arm specific numbers here

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+## **Dynamic Descriptor Sets**
+
+### **Introduction**
+
+Dynamic descriptor sets differ to the regular descriptor sets because they allow an offset to be specified when we aere *binding* (`vkCmdBindDescriptorSets`) the descriptor set. This dynamic offset can be used in addition to the base offset used at the time of updating the descriptor set.
```
gary-sweet: Typo - 'aere'

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+1. Allocating one giant **uniform buffer object** containing all the world matrices of the meshes in your scene. 
+2. Allocating a **descriptor set** with a binding containing the `VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC` flag that then points to the **UBO** you just created.
+3. In our draw call, for each mesh, *dynamically* offset into the giant **uniform buffer object**.
+
+### **Performance**
+
+![dynamic ubo](images/dynamic_descriptor_set_performance.jpg)
+
+In the screenshot above we can see the load/store cycles stay roughly around the same compared to [static uniform buffer objects](#uniform-buffer-objects). However, the frame time goes up from 16.7 ms to 20.9 ms. This is due to the extra time you need to spend every frame determining the dynamic offsets, that you need to send in the bind call (`vkCmdBindDescriptorSets`).
```
gary-sweet: Arm specific numbers

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+This should come with zero performance costs, and as a result this method is designed purely for offering flexibility to your codebase. 
+
+## **Buffer Object Arrays**
+
+### **Introduction**
+
+Another approach, which can be likened to a dynamic descriptor set, is a buffer object array. This is the concept of allocating all of your constant data *upfront* in a large buffer, and writing the entire buffer to a descriptor set. This means in any one shader invocation we have access to all of the model data for the entire scene, at the benefit of only needing to bind one descriptor set per entire draw call.
+
+You can use either a `uniform` or a `buffer` storage type in your shader code to achieve this. However, since `uniform`s can only carry 16 KB of data, this tutorial will use a `buffer`.
```
gary-sweet: Arm specific limit. In general wouldn't it be better to say that if the shader is only reading data and the buffer is smaller than the UBO size limit, a UBO would be better. When the shader writes data, use a storage buffer?

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+
+This on Mali is not a recommend practice as it disables a compiler optimisation technique known as **pilot shaders**. 
+
+Pilot shaders are a technique that allows us to determine what calculations can be "piloted" into your GPU's register so that when the data needs to be read it doesn't take a full read cycle from the GPU RAM.
+
+To show this we can use Streamline to profile the sample, here are the read cycles for using a single descriptor set per mesh against the pre allocated buffer array:
+
+![loadcycles](images/loadcycles.png)
+
+A few different stats are affected in the GPU by using this, but the main thing is the **full read** in the **Mali Core Load/Store Cycles**.
```
gary-sweet: This section needs to be clearly marked as Arm specific

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,399 @@
...
+* Do make constant data directly available in the shader if it is pre-determinable, such as with the use of specialization constants.
+
+**Avoid**
+
+* Avoid indexing in the shader if possible, such as dynamically indexing into `buffer` or `uniform` arrays.
+
+**Impact**
+
+* Failing to use the correct method of constant data will negatively impact performance, causing either reduced FPS and/or increased BW and load/store activity.
+* Register mapped uniforms are effectively free. Any spilling to buffers in memory will increase load/store cache accesses to the per thread uniform fetches.
```
gary-sweet: Arm specific

gary-sweet (changes requested):

framework/rendering/render_frame.h
```diff
@@ -60,6 +60,13 @@ class RenderFrame
 	 */
 	static constexpr uint32_t BUFFER_POOL_BLOCK_SIZE = 256;
 
+	// A map of the supported usages to a multiplier for the BUFFER_POOL_BLOCK_SIZE
+	const std::unordered_map<VkBufferUsageFlags, uint32_t> supported_usage_map = {
+	    {VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT, 1},
+	    {VK_BUFFER_USAGE_STORAGE_BUFFER_BIT, 2},        // x2 the size of BUFFER_POOL_BLOCK_SIZE since SSAOs are normally much larger than other types of buffers
```
SaschaWillems: Shouldn't that be SSBO instead of SSAO?

ZandroFargnoli:
> I have a couple of problems when running this new sample.
> 
> 1. When I select 'Descriptor Sets' or 'Dynamic Descriptor Sets' I get a stream of errors on the console which just keep repeating:
> 
> ```
> [debug] Building #26 cache object (N3vkb12ShaderModuleE)
> [debug] Building #27 cache object (N3vkb12ShaderModuleE)
> [debug] Building #28 cache object (N3vkb12ShaderModuleE)
> [debug] Building #29 cache object (N3vkb12ShaderModuleE)
> [debug] Building #30 cache object (N3vkb12ShaderModuleE)
> [debug] Building #31 cache object (N3vkb12ShaderModuleE)
> [error] [framework/buffer_pool.cpp:140] Ignore buffer allocation update
> [debug] Building #3 cache object (N3vkb14PipelineLayoutE)
> [debug] Building #3 cache object (N3vkb19DescriptorSetLayoutE)
> [debug] Building #5 cache object (N3vkb16GraphicsPipelineE)
> [debug] Building #3 cache object (N3vkb14DescriptorPoolE)
> [debug] Building #0 cache object (N3vkb13DescriptorSetE)
> [error] [framework/buffer_pool.cpp:140] Ignore buffer allocation update
> [debug] Building #1 cache object (N3vkb13DescriptorSetE)
> [error] [framework/buffer_pool.cpp:140] Ignore buffer allocation update
> [debug] Building #2 cache object (N3vkb13DescriptorSetE)
> ...
> ```
> 
> 2. When I select 'Single Pre-allocated Buffer Array' half the geometry in the scene goes black.

1. What device/GPU is this? My console is clean

2. Again, what GPU is this? Perhaps storage  buffers aren't supported? I should implement a check for this in that case

gary-sweet:
> What device/GPU is this? My console is clean
> 
> Again, what GPU is this? Perhaps storage buffers aren't supported? I should implement a check for this in that case

This is one of our own Broadcom GPUs.
Storage buffers are supported.
I suspect these are related to my earlier comment about your mishandling of 'push_constant_limit'.

ZandroFargnoli:
> Re 1) : Your handling of 'push_constant_limit' seems wrong. It looks like you are either expecting 128 or 256 push constants. That might be the case on Arm GPUs, but our limit is 134,217,727‬. Simply creating a UBO that size doesn't really seem like a sensible plan, nor does checking for exactly 256 in several places.

Do you mean in the sample or in the framework? The command buffer should be flexible to whatever the push constants limit is. The sample prefers to use a 256 byte uniform, however because the Vulkan spec only requires 128 bytes we need a path to catch the implementations that don't use 256+

gary-sweet:
> we need a path to catch the implementations that don't use 256+

Yes, but there are several places in the code where you have:
"if (push_constant_limit == 256)" not ">="

samples/performance/constant_data/constant_data_tutorial.md
ZandroFargnoli: This is written against the GLSL spec, these are the guaranteed limits. I believe most implementations of Vulkan let you use more. Do I change the wording? It probably isn't a great idea to maintain a list of these limits

samples/performance/constant_data/constant_data_tutorial.md
ZandroFargnoli: Again, https://www.khronos.org/opengl/wiki/Shader_Storage_Buffer_Object

gary-sweet:
The allocation errors are due to this code in ConstantData::DescriptorSetSubpass::update_uniform():

```
        // Here you allocate a buffer sizeof(MVPUniform)
	auto allocation = render_frame.allocate_buffer(VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT, sizeof(MVPUniform), thread_index);
        ...
        // This makes a vector the same size
	auto data = vkb::to_bytes(mvp);

        // This resizes it (very large on our platform - I think you're only expecting it to make it smaller)
	data.resize(push_constant_limit);

        // This barfs because the underlying buffer is too small for the data
	allocation.update(data);

```

samples/performance/constant_data/constant_data_tutorial.md
gary-sweet: That's an OpenGL spec. What does Vulkan say?

gary-sweet: I also think you've misread it. It says they are a minimum of those sizes (not maximum).

gary-sweet: So I'm saying rather than "They can hold up to 16KB of data" you should be saying "They can hold at least 16KB of data, often a lot more" etc.

samples/performance/constant_data/constant_data_tutorial.md
ZandroFargnoli: How would you like to approach the quoting of numbers? Shall I take readings from a few different vendors?

gary-sweet: I think we just need to make it clear that these are example figures on Arm GPUs. I think that's what happened in the other samples.

ZandroFargnoli:
> > What device/GPU is this? My console is clean
> > Again, what GPU is this? Perhaps storage buffers aren't supported? I should implement a check for this in that case
> 
> This is one of our own Broadcom GPUs.
> Storage buffers are supported.
> I suspect these are related to my earlier comment about your mishandling of 'push_constant_limit'.

The scene going black is most likely a `maxStorageBufferRange` issue. I expected the descriptor set to complain about this (Do you see the error message that is at `descriptor_set.h` line 94)

gary-sweet:
> Do you see the error message that is at `descriptor_set.h` line 94

Nothing odd in the console, just a lot of black geometry.

r-potter:
I haven't managed to debug the issue yet, but the single pre-allocated buffer array variant appears to show some issues on Adreno GPUs. 

On older GPUs (540, 630) it causes the app to crash. On newer devices (640, 650), it runs but with artefacts (i.e. floor surface gets completely dropped).

gary-sweet:
The latest code changes fix both problems for me. Thanks for that.
There are still things in the doc I need to review.

Broadcom's recommendations don't align with Arm's, at the moment push constants are much faster for us (that's why they exist I guess). We'll need to decide how best to deal with that.

Also, on our platforms all of the modes under test are heavily CPU bound. Is that really the intention of these tests? Are there lots of memory allocations happening, or cached memory buffers being flushed all the time perhaps? If so, is that a realistic scenario?

gary-sweet:
> Also, on our platforms all of the modes under test are heavily CPU bound. 

Some of this cost is in our driver, and we have some changes ongoing that will help mitigate some of that cost. This still seems fairly CPU heavy though. Is that the same for others?

samples/performance/constant_data/constant_data_tutorial.md
gary-sweet: ... very strict limitations on how much data can be sent **on some platforms.**

samples/performance/constant_data/constant_data_tutorial.md
gary-sweet: support **at least** 256 bytes, **sometimes much more.**

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,405 @@
...
+* [Inline uniform buffer objects](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_inline_uniform_block.html) (click to read more)
+* [Push descriptors](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_push_descriptor.html) (click to read more)
+
+**Inline uniform buffer objects** and **push descriptors** are not covered by this tutorial, please use the links above to learn more about them.
+
+## **Sample Overview**
+
+### **Introduction**
+
+The sample uses a mesh heavy scene which has 1856 meshes (475 KB of mesh data). This is to demonstrate a use case where many different calls to pushing constant data will occur during a single frame.
```
gary-sweet: I think we need to point out that having lots of constant data for each mesh like this is not recommended practice but is being used to artificially exaggerate the performance differences in this example.

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,405 @@
...
+    uint32_t                                    offset,
+    uint32_t                                    size,
+    const void*                                 pValues);
+```
+
+### **Performance**
+
+![push constants](images/push_constants_performance.jpg)
+
+In early implementations of Vulkan on Arm Mali, this was usually the fastest way of pushing data to your shaders. In more recent times, we have observed on Mali devices that *overall* they can be slower. If performance is something you are trying to maximise, descriptor sets may be the way to go.
```
gary-sweet: may be the way to go **on Mali, but other devices may favour push constants.**

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,405 @@
...
+Here is an example of using a `buffer` in shader code:
+
+```
+layout(set = 0, binding = 1) buffer MeshArray
+{
+	mat4 model_matrices[];
+} mesh_array;
+```
+
+***Note:** If you plan to write to `buffer`s inside shaders, you have to query the physical device features and check that `vertexPipelineStoresAndAtomics` is enabled.*
```
gary-sweet: If you plan to write to buffers inside fragment shaders, you have to query the physical device features and check that fragmentStoresAndAtomics is enabled (vertexPipelineStoresAndAtomics when storing in vertex, tessellation, and geometry shaders).

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,405 @@
...
+    vkCmdDrawIndexed(command_buffer, mesh.index_buffer.size(), 1, 0, 0, instance_index++);
+}
+```
+
+In the code snippet above we can see that we bind our descriptor set once, and for each mesh bind its vertex and index buffers and then execute a draw call with an incrementing value for `uint32_t firstInstance`. This `uint32_t` will be substituted in wherever `gl_InstanceIndex` exists in the shader code, which will pull out the required model matrix to position the mesh inside our world.
+
+
+### **Performance**
+
+This on Mali is not a recommend practice as it disables a compiler optimisation technique known as **pilot shaders**.
```
gary-sweet: This is one of the fastest methods for us, similar to push-constant performance.

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,405 @@
...
+**Do**
+
+* Do keep constant data small, where 128 bytes is a good rule of thumb.
+* Do use push constants if you do not want to set up a descriptor set/UBO system.
+* Do make constant data directly available in the shader if it is pre-determinable, such as with the use of specialization constants.
+
+**Avoid**
+
+* Avoid indexing in the shader if possible, such as dynamically indexing into `buffer` or `uniform` arrays.
+
```
gary-sweet, line 400: Avoid having lots of separate constant data per frame.

gary-sweet (changes requested):

samples/performance/constant_data/constant_data_tutorial.md
ZandroFargnoli: I've removed it altogether as writing is really relevant to constant data

shaders/constant_data/buffer_array.frag
```diff
@@ -0,0 +1,117 @@
...
+	mat4 model;
+	mat4 view_proj;
+
+#ifdef PUSH_CONSTANT_LIMIT_256
+	mat4 scale;
+	mat4 padding;
+#endif
+};
+
+layout(set = 0, binding = 1) uniform MVPUniformArray
```
r-potter: This seems to be the root cause of the issue on older Adreno devices. The compiler fails trying to create an 1800(ish) * 128 byte uniform.

framework/core/command_buffer.h
```diff
@@ -124,33 +128,25 @@ class CommandBuffer
...
 
 	template <typename T>
-	void push_constants_accumulated(const T &value, uint32_t offset = 0)
+	void push_constants(const T &value)
 	{
-		push_constants_accumulated(std::vector<uint8_t>{reinterpret_cast<const uint8_t *>(&value),
-		                                                reinterpret_cast<const uint8_t *>(&value) + sizeof(T)},
-		                           offset);
-	}
+		auto data = to_bytes(value);
```
gary-sweet, line 139: This seems like a very expensive way to do sizeof(T). to_bytes() creates a new vector and copies the data doesn't it?

framework/core/command_buffer.h
```diff
@@ -124,33 +128,25 @@ class CommandBuffer
...
-		push_constants(offset,
-		               std::vector<uint8_t>{reinterpret_cast<const uint8_t *>(&value),
-		                                    reinterpret_cast<const uint8_t *>(&value) + sizeof(T)});
+		if (size > max_push_constants_size)
+		{
+			LOGE("Push constant limit exceeded ({} / {} bytes)", size, max_push_constants_size);
+			throw std::runtime_error("Cannot overflow push constant limit");
+		}
+
+		push_constants(to_bytes(value));
```
gary-sweet: Is there not a more efficient way to get the data to push_constants() without making an intermediate vector?

shaders/constant_data/buffer_array.frag
ZandroFargnoli: Good spot!

framework/core/command_buffer.h
ZandroFargnoli: It is marked as `inline`, so it should do exactly what it did before.

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,403 @@
...
+- [Update-after-bind Descriptor Sets](#update-after-bind-descriptor-sets)
+- [Buffer Object Arrays](#buffer-object-arrays)
+- [Further reading](#further-reading)
+- [Best practice summary](#best-practice-summary)
+
+## **What is constant data?**
+
+### **Introduction**
+
+Constant data is a form of information that is supplied to the graphics pipeline to help with shader computations.
```
sfricke-samsung:
```suggestion
Constant data is a form of information that is supplied to the pipeline to help with shader computations.
```

Compute/RayTracing pipelines all use constant data

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,403 @@
...
+
+In theory, this data can be anything we want it to be, for instance it can be used for things such as calculating where an object should be placed inside our world, or computing the overall brightness of an object based on the lights in the scene.
+
+It differs from other data (e.g. input vertex data) in the sense that it remains *constant* across every shader invocation of a draw call.
```
sfricke-samsung, line 53: input vertex data is constant though, but it is not uniform. I feel there is a blending of `constant` data  and `uniform` data here

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,403 @@
...
+
+Interface blocks are still global variables, and technically still follow the global variable format that was mentioned at the start of this chapter. However, the difference is that they have to be given a user-defined type. They work exactly the same way as a `struct` in GLSL/C++. For example, to access the model matrix in this interface block, you'd use `per_mesh_data.model_matrix`. 
+
+*You can read more about interface blocks [here](https://www.khronos.org/opengl/wiki/Interface_Block_(GLSL)).*
+
+### **Vulkan API**
+
+We've covered how constant data is implemented in the shader, however to push the data from the application to the shader we need to use Vulkan.
+
+We do this with the use of `VkBuffer`s, which is Vulkan's implementation of buffer memory.
```
sfricke-samsung:
```suggestion
We do this mainly with the use of `VkBuffer`s, which is Vulkan's implementation of buffer memory.
```

Push constants don't have a VkBuffer backing it

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,403 @@
...
+
+The following links are useful for learning how to create a Vulkan buffer in your application:
+
+* https://vulkan-tutorial.com/Vertex_buffers/Vertex_buffer_creation
+* https://vulkan-tutorial.com/Vertex_buffers/Staging_buffer#page_Abstracting-buffer-creation
+* https://vulkan-tutorial.com/Uniform_buffers/Descriptor_layout_and_buffer#page_Uniform-buffer
+
+#### *The Methods*
+
+Generally, there are two ways to push your constant data. It is either with **push constants** or **descriptor sets**.
```
sfricke-samsung: Specialization Constants, Inline uniform buffer objects and push descriptors?

I would remove "generally" and just say "there are various ways, but only covering a subset"

sfricke-samsung (changes requested):

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,403 @@
...
+
+A few different stats are affected in the Mali GPU by using this, but the main thing is the **full read** in the **Mali Core Load/Store Cycles**.
+
+## **Further reading**
+
+* The [Vulkan 1.2 spec](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html)
+* "Writing an efficient Vulkan renderer" by Arseny Kapoulkine https://zeux.io/2020/02/27/writing-an-efficient-vulkan-renderer/
+* Alexander Overvoorde's [Vulkan Tutorial on Descriptors](https://vulkan-tutorial.com/Uniform_buffers/Descriptor_layout_and_buffer) guide
+* Vulkan Fast Paths http://32ipi028l5q82yhj72224m8j-wpengine.netdna-ssl.com/wp-content/uploads/2016/03/VulkanFastPaths.pdf
+* Performance root signature of descriptor sets https://gpuopen.com/performance-root-signature-descriptor-sets/
```
SaschaWillems: It seems that they have reworked the whole gpuopen.com page, and the link no longer works. Haven't been able to find that page via the search, so maybe remove that link.

samples/performance/constant_data/constant_data_tutorial.md
```diff
@@ -0,0 +1,403 @@
...
+
+**Do**
+
+* Do keep constant data small, where 128 bytes is a good rule of thumb.
+* Do use push constants if you do not want to set up a descriptor set/UBO system.
+* Do make constant data directly available in the shader if it is pre-determinable, such as with the use of specialization constants.
+
+**Avoid**
+
+* Avoid indexing in the shader if possible, such as dynamically indexing into `buffer` or `uniform` arrays.
```
SaschaWillems: Not sure on this. At least this should have some explanation as to why people should avoid this. This is something done so commonly that without explanation this may cause some discussions.

samples/performance/constant_data/constant_data_tutorial.md
ZandroFargnoli: Will remove

PR #114, opened 2020-04-29
Android Launcher Refactor

TomAtkinsonArm:
## Description

Reorganize / Simplify the android launcher.

* Refactored most components
* Added a sample store - Moves sample processing logic to one class instead of scattered among classes
* Added a Utils library for common misc methods
* Added a Notification library for notification classes
* Chained permissions to follow androids expected permission flow `W/Activity: Can request only one set of permissions at a time`
* Moved view logic into view classes
* Moved most inline class definitions or function definitions to separate classes in the sample file
Updated the filter dialog to discard a changed configuration of filters if they are not applied when the filter closes. This means it will only ever show the current filter configuration when opening
* CategoryComparator -> PredefinedOrderComparator - Now can be used with any set and any predefined order

## General Checklist:

## Sample Checklist

Tested all samples on an S9 using the launchers batch mode feature


gary-sweet (commented):
Can't test on Android - but doesn't appear to break anything for me.

r-potter (approved):
LGTM

Crashing samples (such as running the conservative rasterization sample on every phone I tested) leave the performance and extensions lists unpopulated. The extensions one will repopulate later, but the performance one never does.

However, the same crashing samples on the previous app bring down the entire app, so this doesn't make the situation worse than it currently is.

TomAtkinsonArm:
Hi @r-potter, Thanks for taking the time to review

How can i reproduce this? Where there any particular steps you took?

r-potter:
Hi @TomAtkinsonArm 

On a Mali S20 (and also various Adreno devices that I had to hand):

1. Launch the Vulkan Samples app
2. Select the Extensions tab
3. Launch the conservative rasterization sample

On the old app, this will crash the entire app back to the desktop. On this branch it instead returns to the launch screen, but with unpopulated tabs. Clicking between the tabs repopulates some, but not all tabs.

wasimabbas-arm:
> On the old app, this will crash the entire app back to the desktop. On this branch it instead returns to the launch screen, but with unpopulated tabs. Clicking between the tabs repopulates some, but not all tabs.

Do I understand correctly the crash is not introduced by this PR but I guess either way crash is an issue anyways. So we/someone should fix that?

r-potter:
Correct. The actual crash is not related to the PR. It's an unrelated sample that crashed anyway (and should be fixed). I'm happy to take Tom's MR as is. 

Fixing the unpopulated tabs following a crash would be nice, but to me it's not critical given that samples aren't supposed to crash in the first place.

PR #143, opened 2020-06-10
Note on OnContentRectChanged callback in surface_rotation sample

### LINKED ISSUE #113: Handling Android orientation in surface rotation tutorial
A code snippet in the [surface rotation tutorial](https://github.com/KhronosGroup/Vulkan-Samples/blob/master/samples/performance/surface_rotation/surface_rotation_tutorial.md#rotation-in-android) suggests using `APP_CMD_CONTENT_RECT_CHANGED` to handle orientation changes. 
However, this event is never generated by the current version of NDK without some extra work which is done in the framework (using `onContentRectChanged`).
This is not mentioned in the tutorial, so the advice is a little bit misleading (see this [issue](https://github.com/ARM-software/vulkan_best_practice_for_mobile_developers/issues/36)).
Probably it is good to add a little **Note** section with an explanation.
This NDK [issue](https://github.com/android/ndk/issues/1139) is already reported and it's very likely that it will be fixed in the version 22, so it would be good to keep in mind (and probably update the code and the tutorial).

TomAtkinsonArm:
## Description

Added a note for the use of a callback to patch NDK expected behaviour

Fixes #113

### DIFF
--- a/samples/performance/surface_rotation/surface_rotation_tutorial.md
+++ b/samples/performance/surface_rotation/surface_rotation_tutorial.md
@@ -113,6 +113,29 @@ void on_app_cmd(android_app *app, int32_t cmd)
 }
 ```
 
+Note that android NDK 21 and below does not support `APP_CMD_CONTENT_RECT_CHANGED`. This is fixed in version 22+ which at the date of this addition has not been released. This can be patched in the short term by registering a callback that mimics the expected behavior of `OnContentRectChanged`. The callback we use in this project is:
+
+```
+void on_content_rect_changed(ANativeActivity *activity, const ARect *rect)
+{
+	... log content rect dimensions
+	
+	struct android_app *app = reinterpret_cast<struct android_app *>(activity->instance);
+	auto                cmd = APP_CMD_CONTENT_RECT_CHANGED;
+
+	app->contentRect = *rect;
+
+	if (write(app->msgwrite, &cmd, sizeof(cmd)) != sizeof(cmd))
+	{
+		... failed to write message to the app glue command stream
+	}
+}
+```
+
+The above triggers the `APP_CMD_CONTENT_RECT_CHANGED` event and the new content rect dimensions should be accessible by the `app->contentRect` attribute.
+
+There are other alternatives to `CONTENT_RECT_CHANGED`, an example can be found here [Handling Device Orientation Efficiently in Vulkan With Pre-Rotation](https://android-developers.googleblog.com/2020/02/handling-device-orientation-efficiently.html)
+
 ## Swapchain re-creation
 
 We need to sample the current transform from the surface:


samples/performance/surface_rotation/surface_rotation_tutorial.md
```diff
@@ -112,6 +112,27 @@ void on_app_cmd(android_app *app, int32_t cmd)
 	}
 }
 ```
+##### Addition June 10th 2020
+Note that android NDK 21 and below does not support `APP_CMD_CONTENT_RECT_CHANGED`. This is fixed in version 22+ which at the date of this addition has not been released. This can be patched in the short term by registering a callback that mimics the expected behavior of `OnContentRectChanged`. The callback we use in this project is:
```
JoseEmilio-ARM, line 116: maybe add a link here to the place where this happens in the framework

samples/performance/surface_rotation/surface_rotation_tutorial.md
```diff
@@ -112,6 +112,27 @@ void on_app_cmd(android_app *app, int32_t cmd)
 	}
 }
 ```
+##### Addition June 10th 2020
```
JoseEmilio-ARM: not sure if we need the date here

samples/performance/surface_rotation/surface_rotation_tutorial.md
TomAtkinsonArm: The android_platform file is definitely going to be changed more in the future than the tutorial so the link has a chance to be outdated. This method is an exact copy of our implementation without the logs

samples/performance/surface_rotation/surface_rotation_tutorial.md
```diff
@@ -112,6 +112,27 @@ void on_app_cmd(android_app *app, int32_t cmd)
...
+	app->contentRect = *rect;
+
+	if (write(app->msgwrite, &cmd, sizeof(cmd)) != sizeof(cmd))
+	{
+		... failed to write message to the app glue command stream
+	}
+}
+```
+
+The above triggers the `APP_CMD_CONTENT_RECT_CHANGED` event and the new content rect dimensions should be accessible by the `app->contentRect` attribute.
```
JoseEmilio-ARM, line 135: we could somehow link to the recommendations in https://android-developers.googleblog.com/2020/02/handling-device-orientation-efficiently.html

TomAtkinsonArm: This article doesn't seem to cover onContentRectChanged? Are you meaning that we should add a link to the tutorial with extra resources?

JoseEmilio-ARM: Yes it covers other ways of detecting a rotation depending on the Android version so we could say something like 'see this blog for other options'.

TomAtkinsonArm: I like the idea. "See this blog for more options on Handling Device Orientation Efficiently in Vulkan With Pre-Rotation"

where <> is also the link to the article?

TomAtkinsonArm: I think I wonder what <> is in markdown cos it hid my comment

PR #178, opened 2020-08-27
Add a PostProcessingPipeline

PaoloJovon-arm:
## Description
This branch adds a new `PostProcessingPipeline` (& supporting utilities). It is similar to a `RenderPipeline`, but it's specialized for multiple fullscreen & compute passes - potentially on different render targets - as seen in complex post-processing effects.  
It uses SPIR-V reflection to bind descriptors by name, and it submits image memory barriers automatically based on what is sampled / used as input or output attachment / used as storage image (unless the image is already in the right layout).

Please see the modified MSAA sample for what the API looks like: [preparation](https://github.com/PaoloJovon-arm/Vulkan-Samples/blob/4ec40eff2a31222eef9864c11822e5a23b4edf9f/samples/performance/msaa/msaa.cpp#L120) and [drawing](https://github.com/PaoloJovon-arm/Vulkan-Samples/blob/4ec40eff2a31222eef9864c11822e5a23b4edf9f/samples/performance/msaa/msaa.cpp#L568).

## Framework additions
- `PostProcessingPipeline` and all related types - see files.
  - `PostProcessingRenderPass` + `PostProcessingSubpass`: RPs and SPs in the graphics pipeline (that auto-transition input/sampled/output attachments).
  - `PostProcessingComputePass`: wraps one (or more) compute dispatch[es] (with automatic storage image transitions).
  - Other pass types can be implemented by subclassing.
- `SampledImage`, i.e. a small handle type that can reference either an `ImageView` or a (`RenderTarget*`, attachment number).
  - If a render target is not specified, a default one (for `PostProcessingRenderPass`: the one passed to `draw()`) is used.
  - This makes the API to bind images sensibly more convenient to use.

## Framework changes
These are mostly to allow for multiple `vkb::RenderTarget`s in the same frame (potentially spanning different mip-maps or array layers of images).

- `ImageView`s can now be constructed for a specific array layer and/or mipmap level.
- `RenderTarget`s can now be constructed from a list of `ImageView`s (not just from `Image`s directly).
  - Useful when rendering to particular array layers of 2D_ARRAY textures, or mipmaps of a texture.
- Information about what accessors are marked as NonReadable/NonWritable in SPIR-V is kept when parsing.
  - For inference of readonly/writeonly attributes for storage image binding points in shaders.
- `CommandBuffer` can now bind image views without an associated sampler.
   - For storage images and separate image/sampler descriptors.
- Descriptors are hashed when being put in the descriptor cache. If the hash of a binding mismatches with the previous one, the descriptor will always be updated.
  - This prevents a condition where they would not be updated due to a pending write-update for the same set/binding number (but with different content) - it happens frequently with particular effects, e.g. multipass blurs.
  - If not applied, the framework sometimes does not update texture handles when it should - causing corruption of the entire pipeline.
- MSAA sample's `PostProcessingSubpass` has been removed - a `PostProcessingPipeline` is now in its place.

## General Checklist:


gary-sweet (approved):
All appears to work for me without issue.
Don't have time to look over the changes in any depth though I'm afraid.

aurolou (approved):

mattruschnv (approved):

huawei-ahcox:
Hi @PaoloJovon-arm,
It seems like the MSAA sample only exercises a subset of the functionality.
PostProcessingComputePass is one example of a class that isn't exercised by any sample
and doesn't seem to be referenced outside its own header and implementation file.
To what extent were features relating to compute pipelines like this tested during
development by internal samples (perhaps ones which were not considered interesting enough
to submit to this repo)?
Are there other samples utilising more of this post processing stack which are planned to
be submitted soon?
Thanks for any insights!

PaoloJovon-arm:
Hi @huawei-ahcox,
The compute passes were briefly tested internally (to make sure that the abstractions worked); that said, to my knowledge, there are no public / planned samples that make use of the functionality as of yet.  
Since the implementation was only a few lines of code (and some post-processing pipelines in the wild seem to make heavy use of compute shaders), I left them in the final PR in the hope that they would be useful.

PR #194, opened 2020-10-26
Added slide up/down to freecamera.

huawei-ahcox:
Put slide down and up on Q and E, fitting in to the WSAD cluster to give the less common but still useful axis of translational control on the same hand.
Lets the models be navigated more freely.

Does not relate to any issue. Just a simple quality of life improvement.

### DIFF
--- a/framework/scene_graph/scripts/free_camera.cpp
+++ b/framework/scene_graph/scripts/free_camera.cpp
@@ -1,4 +1,5 @@
-/* Copyright (c) 2019, Arm Limited and Contributors
+/* Copyright (c) 2019-2020, Arm Limited and Contributors
+ * Copyright (c) 2020, Andrew Cox, Huawei Technologies Research & Development (UK) Limited
  *
  * SPDX-License-Identifier: Apache-2.0
  *
--- a/framework/scene_graph/scripts/free_camera.cpp
+++ b/framework/scene_graph/scripts/free_camera.cpp
@@ -72,6 +72,14 @@ void FreeCamera::update(float delta_time)
 	{
 		delta_translation.x += TRANSLATION_MOVE_STEP;
 	}
+	if (key_pressed[KeyCode::Q])
+	{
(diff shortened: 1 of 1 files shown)


abbaswasim (dismissed):

huawei-ahcox:
The build fail is in `649.2 Copyright Headers | failed`

Maybe just a glitch: I didn't change the existing copyright notice. 

Where can I read the python script this step runs?

abbaswasim:
ah yes it will fail because the copyright notice needs updating. The policy is if you have changed the file in 2020 the copyright notice should reflect that. its currently saying 2019. The scripts are in private docker images that not everyone has access to.

You should add a line for your affiliation to the copyright header to the file you have edited.

So

```
/* Copyright (c) 2019, Arm Limited and Contributors
   Copyright (c) 2020, Your affiliation
```

Having said that not sure what our scripts would actually expect. It might be expecting the Arm copyright to be updated as well and that might not be the intended result. @TomAtkinsonArm what do you think?

huawei-ahcox:
> You should add a line for your affiliation to the copyright header to the file you have edited.
> 
> So
> 
> ```
> /* Copyright (c) 2019, Arm Limited and Contributors
>    Copyright (c) 2020, Your affiliation
> ```
> 
> Having said that not sure what our scripts would actually expect. It might be expecting the Arm copyright to be updated as well and that might not be the intended result. @TomAtkinsonArm what do you think?

How about just: 

    Copyright (c) 2019-2020, Arm Limited and Contributors

The "and Contributors" bit would seem to be there to cover randoms like me passing by.

Alternatively, with the years spread for the Arm line:
,
    Copyright (c) 2019-2020, Arm Limited and Contributors
    Copyright (c) 2020, Andrew Cox, Huawei Technologies Research & Development (UK) Limited

abbaswasim (approved):

gary-sweet (approved):

PR #193, opened 2020-10-22
Noted Python 3 build dependency

huawei-ahcox:
Cmake configure on a Windows PC without Python 3 installed results in::

CMake Error at C:/Program Files/CMake/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
  Could NOT find PythonInterp (missing: PYTHON_EXECUTABLE) (Required is at
  least version "3")
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:458 (_FPHSA_FAILURE_MESSAGE)
  C:/Program Files/CMake/share/cmake-3.18/Modules/FindPythonInterp.cmake:169 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  third_party/glslang/CMakeLists.txt:175 (find_package)

After installing Python 3 (Python 3.9 in this case), that error goes away.

## Description

Please include a summary of the change, new sample or fixed issue. Please also include relevant motivation and context.
Please read the [contribution guidelines](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md)

Fixes #<issue number>

## General Checklist:

Please ensure the following points are checked:

- [ ] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [ ] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [ ] I have commented any added functions (in line with Doxygen)
- [ ] I have commented any code that could be hard to understand
- [ ] My changes do not add any new compiler warnings
- [ ] My changes do not add any new validation layer errors or warnings
- [ ] I have used existing framework/helper functions where possible
- [ ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [ ] My changes do not add any regressions
- [ ] I have tested every sample to ensure everything runs correctly
- [ ] This PR describes the scope and expected impact of the changes I am making

## Sample Checklist

If your PR contains a new or modified sample, these further checks must be carried out *in addition* to the General Checklist:
- [ ] I have tested the sample on at least one compliant Vulkan implementation
[...]

PR #196, opened 2020-10-31
Update glslang submodule

SaschaWillems:
## Description

This PR updates the glslang third party submodule. This is necessary for using newer GLSL shader functionality like recent extension. The submodule in this repo hasn't been updated in a while, and the glslang code had at least one breaking change.

## General Checklist:

- [n/a] I have commented any added functions (in line with Doxygen)
- [n/a] I have commented any code that could be hard to understand


SaschaWillems:
CI currently fails, as the new build_info.h doesn't seem to get generated for clang tidy. Any ideas on this?

SaschaWillems:
The Windows CI build failes due to an unsuitable python version:

> CMake Error at C:/Program Files/CMake/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:164 (message):
  Could NOT find PythonInterp: Found unsuitable version "2.7.9", but required
  is at least "3" (found C:/ProgramData/chocolatey/bin/python.exe)

Not sure how to fix this.

wasimabbas-arm:
I think this issue was reported, or rather someone added an update to the docs to remind Python is a requirement and needs to be up to date.  [here](https://github.com/KhronosGroup/Vulkan-Samples/pull/193)

This has recently changed. Not an issue with the samples themselves but glslang's requirement has changed since we have used it.

I think we need to install Python 3 on our windows runners. The hostname is a weird one from Travis logs so this is not one of our runners. So perhaps requires a dependency to the CI scripts. @TomAtkinsonArm what do you think?

wasimabbas-arm:
By the way do you know if this dependency if for something like tests or extra stuff we can disable in `glslang` as a workaround for now?

SaschaWillems:
Looks like Python 3 is only required for building glslang's external dependencies: https://github.com/KhronosGroup/glslang/commit/8ba49b862cfb3449ffdfecd2e6d9c2fd5e855f1e

Not sure if we would need those or not. If we don't, we can' disable that cmake flag by default.

wasimabbas-arm:
> Not sure if we would need those or not. If we don't, we can' disable that cmake flag by default.

This would be my preference. If there is an easy CMake option to do this.

SaschaWillems:
Just checked, and it doesn't look like this can be easily controlled by a CMake option. The current version of glslang will always use python to create a build information file. Even though we don't need that file with this PR anymore, the creation of that file is a fixed part of the CMake script for glslang. Judging by the [documentation for the windows travis environment](https://docs.travis-ci.com/user/reference/windows/), it's stuck on Python 2.7.9.

But with travis using chocolaty, it may be possible to add it like this:

      before_install:
        - choco install python3
        - export PATH="/c/Python37:/c/Python37/Scripts:$PATH"

I'll try if this actually works.

SaschaWillems:
With the gitlab actions CI, this now fails at clang format, because it doesn't seem to be able to find the clang-format executable: https://github.com/KhronosGroup/Vulkan-Samples/runs/1359299764?check_suite_focus=true

TomAtkinsonArm:
> With the gitlab actions CI, this now fails at clang format, because it doesn't seem to be able to find the clang-format executable: https://github.com/KhronosGroup/Vulkan-Samples/runs/1359299764?check_suite_focus=true

Looking into it!

TomAtkinsonArm:
#204 fixes the missing binary

SaschaWillems:
Thanks. CI is now working as intended 👍

gary-sweet (changes requested):
This update breaks the CMake build for me.  Unfortunately, I don't have time to look into it right now. Does this happen for anyone else?

Note: I do have python3 installed. I suspect if the cmake did find_package(Python3) it would work.

CMake Error at /usr/share/cmake-3.15/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find PythonInterp: Found unsuitable version "2.7.17", but
  required is at least "3" (found /usr/bin/python)
Call Stack (most recent call first):
  /usr/share/cmake-3.15/Modules/FindPackageHandleStandardArgs.cmake:376 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.15/Modules/FindPythonInterp.cmake:160 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  third_party/glslang/CMakeLists.txt:233 (find_package)
  third_party/glslang/CMakeLists.txt:239 (find_host_package)

r-potter (dismissed):
Tested on Note 9 (Adreno) and Note10 (Mali). No issues related to the PR (although it did reveal https://github.com/KhronosGroup/Vulkan-Samples/issues/217 during testing)

SaschaWillems:
@gary-sweet: Can you check if python3 is in your path? If it is, can you do a clean build. It's possible that CMake has cached the 2.x python binary.

SaschaWillems:
I have updated the sub module again, adding support for the finalized ray tracing extensions. This is a requirement for updating the ray tracing samples, so reviews would be very much appreciated :)

gary-sweet (approved):
Removing the entire build folder did indeed resolve the python3 errors, thanks.

I had previously searched the cmake cache for python variables, but forgot to look in the advanced vars, which is why I didn't do this initially.

SaschaWillems:
Thanks for the confirmation. As this problem may affect people that already cloned the repo, we may need to put a note somewhere to inform them about this possibly breaking change.

JoseEmilio-ARM (approved):

SaschaWillems:
I'd very much appreciate one additional approval for this to be merged :) It's a requirement for using the final ray tracing functions, and as soon as this is approved and merged, I can submit a PR to update the ray tracing sample.

marty-johnson59:
Since this is needed for RT samples, there's some time pressure here as we hope to release these on the 15th.  Can we get one more review?  Perhaps NVIDIA or AMD (@mattruschnv, @aurolou)?  Thanks!

r-potter (approved):
Retested on Mali (Note 10) and Adreno (Note 9). No new issues

aurolou (approved):

PR #197, opened 2020-11-02
Add GLSL shader files to project structure via CMAKE

SaschaWillems:
## Description

This PR adds a function to the CMAKE build system that adds GLSL shader files to the sample projects where applicable. For every sample where the shaders are located in a folder with the same name as the samples (e.g. shaders\hdr). The CMAKE setup will now gather all GLSL files in that folder and adds them to the project structure. For samples without explicit shader folders, no shaders will be added to the project file. 

image

This allows for easy access of the shaders directly from within the IDE.

## General Checklist:

- [n/a] I have commented any added functions (in line with Doxygen)
- [n/a] My changes do not add any new compiler warnings
- [n/a] My changes do not add any new validation layer errors or warnings
- [n/a ] I have used existing framework/helper functions where possible

### DIFF
--- a/bldsys/cmake/sample_helper.cmake
+++ b/bldsys/cmake/sample_helper.cmake
@@ -150,6 +150,22 @@ function(add_sample)
         list(APPEND SRC_FILES ${TARGET_FILES})
     endif()
 
+    # Gather GLSL shader files
+    set(SHADER_DIR_GLSL "${PROJECT_SOURCE_DIR}/shaders/${TARGET_ID}")
+    file(GLOB SHADERS_GLSL 
+        "${SHADER_DIR_GLSL}/*.vert"
+        "${SHADER_DIR_GLSL}/*.frag"
+        "${SHADER_DIR_GLSL}/*.comp"
+        "${SHADER_DIR_GLSL}/*.geom"
+        "${SHADER_DIR_GLSL}/*.tesc"
+        "${SHADER_DIR_GLSL}/*.tese"
+        "${SHADER_DIR_GLSL}/*.mesh"
+        "${SHADER_DIR_GLSL}/*.task"
+        "${SHADER_DIR_GLSL}/*.rgen"
+        "${SHADER_DIR_GLSL}/*.rchit"
+        "${SHADER_DIR_GLSL}/*.rmiss"
+    )    
+
     add_project(
         TYPE "Sample"
         ID ${TARGET_ID}
@@ -162,7 +178,10 @@ function(add_sample)
         FILES
             ${SRC_FILES}
         LIBS
-            ${TARGET_LIBS})
+            ${TARGET_LIBS}
+        SHADERS_GLSL
+            ${SHADERS_GLSL})
+
 endfunction()
 
 function(add_sample_with_tags)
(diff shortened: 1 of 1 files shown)


bldsys/cmake/sample_helper.cmake
```diff
@@ -184,6 +203,21 @@ function(add_sample_with_tags)
...
+        "${SHADER_DIR_GLSL}/*.comp"
+        "${SHADER_DIR_GLSL}/*.geom"
+        "${SHADER_DIR_GLSL}/*.tesc"
+        "${SHADER_DIR_GLSL}/*.tese"
+        "${SHADER_DIR_GLSL}/*.mesh"
+        "${SHADER_DIR_GLSL}/*.task"
+        "${SHADER_DIR_GLSL}/*.rgen"
+        "${SHADER_DIR_GLSL}/*.rchit"
+        "${SHADER_DIR_GLSL}/*.rmiss"
+    )
```
wasimabbas-arm, line 220: Curious if this actually works as expected? I am not sure how this `${PROJECT_SOURCE_DIR}` works when used before the `add_project` call (this makes a `project` call later)?

Also nothing else seems to be dependent on these `add_sample*` functions. So perhaps a better option would be to move this outside to its own function? Will save on the repetition of code and offer more freedom if we want to make this logic more complex, looking for shaders in other places for example.

gpx1000:
Instead of doing another glob.  I have a more cmake target specific way of doing this:

```cmake
function(target_add_spirv_shader TARGET INPUT_FILE)
    find_program(GLSLC glslc)
    set(optionalArgs ${ARGN})
    set(OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/shader-spv")
    list(LENGTH optionalArgs numArgs)
    if(${numArgs} GREATER 1)
        message(FATAL_ERROR "target_add_spirv_shader called incorrectly, add_spirv_shader(target INPUT_FILE [OUTPUT_FILE])")
    endif()
    if(${numArgs} EQUAL 1)
        list(GET optionalArgs 0 OUTPUT_DIR)
    endif()
    file(MAKE_DIRECTORY ${OUTPUT_DIR})
    get_filename_component(bare_name ${INPUT_FILE} NAME_WE)
    get_filename_component(extension ${INPUT_FILE} LAST_EXT)
    string(REGEX REPLACE "[.]+" "" extension ${extension})
    set(OUTPUT_FILE "${OUTPUT_DIR}/${bare_name}-${extension}.spv")
    add_custom_command( PRE_BUILD
            OUTPUT ${OUTPUT_FILE}
            COMMAND ${GLSLC} ${INPUT_FILE} -o ${OUTPUT_FILE}
            MAIN_DEPENDENCY ${INPUT_FILE}
            WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
            )
    add_custom_target(${bare_name}-${extension} DEPENDS ${OUTPUT_FILE})
    add_dependencies(${TARGET} ${bare_name}-${extension})
endfunction()
```

Now with the above, you can use it as a target against any library and the dependency will work its way up the chain: i.e.
```cmake
add_library( foo STATIC foo.cpp)
target_add_spirv_shader(foo bar.vert ${CMAKE_CURRENT_BINARY_DIR})

add_exeutable( fooBin main.cpp )
target_link_libraries(fooBin public foo)
```
That should be enough to do the right thing at the sample level and keep the shaders near the sample without a glob.

SaschaWillems:
I went with the glob way because it doesn't require one to manually specify the shader file names. If there is a better suited solution then I'm all for it :)

gpx1000:
As annoying as writing out the files are, using glob is frowned upon because it makes life difficult if you wish to create new files; you have to remember to touch the CMakeLists.txt file unless you don't use CMake as your primary build source.

Personally, I like the main feature of being able to make the folder structure make sense:
example/foo/shaders/
example/foo/*cpp
example/foo/CMakeLists.txt

The only place that needs to worry about shaders is in the example folder structure.  CMakeLists.txt just needs to know about the local example shaders.  By the same token, you can easily specify where the compiled shader winds up and locally use that from the example/foo/main.cpp etc

Another advantage is when you use cpack to create an installer, it will identify each shader as a target thus it'll auto get packaged.

SaschaWillems:
Closing this as discussed on today's call. Steven's solution is way better :)

SaschaWillems:
We talked about this on the recent call, so I'd like to pick this up again. 

@gpx1000 : If you have some time to spare, can you take a look at this? The snippet you posted seems to be geared towards using CMake to also compile the shaders, which is not what we do with these samples.

gpx1000:
There's nothing wrong with this at all.  My main idea was / is to supplement this path so that CMake gets an idea of assets (files that exist that the binary would depend upon existing).  This way if you were to pack with CPack for an installer then the shaders go to the right place.  As an example, if the sample labels in the CMake that it depends upon XXX.gltf XXX.ktx XXX.vert XXX.frag etc then individually simply linking against the shader as a cmake target all of those assets would come along with it when being packaged it also would show up in various IDEs (including VS) as an asset dependent upon each target.
It also affords the chance to make changes to assets (i.e. compile the shader, run minification, optimizations, static analysis etc.) or compile to gltf with compression etc.
I *still* think it's worth accepting this PR as is, but I'm only trying to say that there's more here we can do.

SaschaWillems:
Thanks a lot for your comments. So I'll go ahead and do a new PR that just adds the shaders to the solution, and then we can maybe built a more advanced solution like the one you mentioned on top of that.

PR #198, opened 2020-11-02
WIP GitHub Actions

TomAtkinsonArm:
## Description

Due to recent changes of the Travis CI platform it makes sense to prioritize the move from Travis to GitHub Actions

[Here is an example PR using GitHub Actions](https://github.com/TomAtkinsonArm/Vulkan-Samples/pull/1), see the checks section.

image

## Workflows

I have added the same capabilities as Travis using three separate workflows

### Quality Checks

This covers Doxygen Syntax, Copyright Headers, Snake Case and Clang Format Checks

### Builds

Build jobs for MacOS, Windows and Ubuntu in Release.

We can cover Debug modes by adding Debug to the 

> `build_types` property in the matrix strategy. This was not added as we do not cover this with Travis currently

### External

External Projects tests for all platforms to test adding the project as a submodule

## Execution

Each workflow will be executed in parallel and each job will also be executed in parallel up to GitHubs parallel job limit (20 jobs Windows and Linux, 5 jobs MacOS). We should have improved completion speeds compared to Travis due to this. Before we executed the jobs in stages to reduce upfront job count but as this is less of an issue with GitHub actions I decided to remove this bottleneck for the time being

## Checklists

This PR does not affect any code standard of the project


SaschaWillems:
This looks good to me 👍

PR #199, opened 2020-11-02
Fix Subpasses sample on some platforms

### LINKED ISSUE #94: Render subpasses sample issues on some platforms
Render Subpasses sample shows flickering and fails to render using renderpasses on some platforms (Adreno 540).

Seems to get fixed after forcing a swapchain re-creation.
Video: [render_passes_bug.zip](https://github.com/KhronosGroup/Vulkan-Samples/files/4348900/render_passes_bug.zip)

carlosadcq:
## Description

When toggling values in the 'Subpasses' sample, it recreates the render targets with the new parameters. The resource caching system was not taking into account that an ImageView that may have been destroyed and recreated with a different image could have the same ImageView handle in some platforms.

I believe the framework has a deeper issue since when a RenderContext gets recreated, it destroys the respective Image and ImageViews but does not remove them from the resource_cache. That may need to be addressed as well although I don't know where's the best place to do that and will leave it out of the scope of this fix. A possibility for this is to add a:
`device.get_resource_cache().clear_framebuffers();`
at the end of the void RenderContext::recreate() function but I'm not 100% sure that's the best place for it.

Fixes #94 

## General Checklist:

- [x] I have tested every sample to ensure everything runs correctly. Tested on Adreno and Nvidia GPUs.

### DIFF
--- a/framework/common/resource_caching.h
+++ b/framework/common/resource_caching.h
@@ -445,6 +445,7 @@ struct hash<vkb::RenderTarget>
 		for (auto &view : render_target.get_views())
 		{
 			vkb::hash_combine(result, view.get_handle());
+			vkb::hash_combine(result, view.get_image().get_handle());
 		}
 
 		return result;
--- a/framework/common/resource_caching.h
+++ b/framework/common/resource_caching.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2020, Arm Limited and Contributors
+/* Copyright (c) 2018-2021, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
--- a/framework/rendering/render_context.cpp
+++ b/framework/rendering/render_context.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2019-2020, Arm Limited and Contributors
+/* Copyright (c) 2019-2021, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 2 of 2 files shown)


carlosadcq:
I'd be happy to add a fix for clearing the FBO cache when render context is restored, I just need some guidance on where would we think is the best place to do this within the framework.

gary-sweet (dismissed):

carlosadcq:
With the last commit, it ensures that FBO's are destroyed in cache as this 'Recreate' function was already creating new RenderTargets.

r-potter:
I still see some issues with this, at least on Adreno 630. To reproduce:

1. Open the subpasses sample
2. Repeatedly toggle between Enabled and Disabled for Transient attachments. Eventually both buttons start flickering, although the rest of the display is static
3. Slightly shift the camera. The entire rendered scene flickers

carlosadcq:
Hi @r-potter, I apologize for the delay.. I finally had a chance to test on multiple devices to try to reproduce this again. I can now reproduce what you are reporting. From my testing this seems to happen on newer devices though, and seem to be multiple issues that result in flickering type scenarios. In terms of the original bug report on Snapdragon 835 I can confirm the issue is not reproducible there anymore. On a Snapdragon 855 and 865 I see the original issue solved as well by this changelist (just toggling the 'transient attachments' used to flicker and now it doesn't), but these newer devices have a separate issue where flickering happens when toggling Subpasses/Renderpasses which result in flickering once the viewport gets 'corrupted'. I'll continue to look into this and submit a fix that hopefully tackles them all.

carlosadcq:
We've confirmed this PR fixes the original flickering reported in https://github.com/KhronosGroup/Vulkan-Samples/issues/94. The additional flickering was fixed in a recent driver (v514). Once this driver is adopted by OEMs and pushed in OTA updates, this sample should work as expected.

PR #223, opened 2020-12-06
#1 Plugins Base

TomAtkinsonArm:
## The Idea

There are lots of features in Vulkan Samples, and reason to add more in the future. In order to make the addition of new features and maintenance of old features easier we are putting forward the idea of plugins. Plugins should encapsulate everything that a feature needs and should interact with the application through the observer pattern. These features can then configure the platform and react to hooks. Hooks are predefined methods that an plugin can implement and subscribe too when they want to interact with the platform or be notified of changes in the platform. Plugins should be configurable at both run and compile time (if an plugin is created for testing purposes then it should only be available if VKB_BUILD_TESTS=ON).

## Why are plugins needed?

We have lots of features such as batch mode, benchmark mode, fps logging, app swapping... Most of these features are hard coded into the VulkanSamples class. This, over time, has created a coupling between Platform, Application and VulkanSamples - which in itself is an Application. In practice we always load the same application (VulkanSamples) which then loads another application (The actual sample we want to load). This app can then be swapped at run time with batch mode but again Vulkan Samples is performing this swap.

To summarize, one class, lots of responsibilities and feature specific code being spread over the framework.

## How are plugins implemented?

With plugins, all of the features that VulkanSamples contains can be split up and maintained separately. Some of the ones that you will see in this series of PRs are Batch Mode, Benchmark Mode, StopAfterX, FpsLogger and HeadlessMode.

An plugin can define a `Flag` or `FlagGroup` which are used by the Parser to generate and parse the CLI. Flags replace using strings when interacting with the parser - statements such as `get<int>("headless")` were used to retrieve values, now the flag itself can be used to retrieve values `Flag headless_flag{...}` `get<int>(headless_flag)`. This removes human error and allows for different types of flags without knowing the required CLI output for docopt. I admit that it seems counter intuitive as adding flags replaces the simplicity of docopt. However, I wanted an easier way for Plugins to manage their own activation and configuration from the same file. I did not want to replace the work surrounding parsing and using the docopt CLI.

Plugins have different intents, of which some combinations do not work. An exclusion/inclusion list could have been used to specify the interoperability of plugins, however, I think this list would be large and contain too many duplicates. Such a list would be hard to maintain. To fix this I introduced tags with some help from @samwalls-arm. An plugin generally tags itself with its name and behavioral intent.

As an example, Batch mode has the tags BatchMode and FullControl whilst stop after frame X has tags StopAfter and Stopping. As their names imply, these two plugins will not mix, we can't have batch mode stopping halfway at the frame specified by StopAfter as this would create undefined behaviour.

An inclusion/exclusion list would look at their names BatchMode does not work with StopAfter. We would have to define this for each combination of plugins. Instead we look at their intent, FullControl plugins do not mix with Stopping plugins. These simple rules end up creating a pretty accurate CLI which can also be added too in the future without effecting other plugins or maintaining a list of interoperable plugins.

More detail on this can be found in the Plugins PR <link>

## PR layout

- PR 1 (Current PR) - Remove unused features and clean up some parts of the build system to make PR diffs easier to read
- PR 2 Apps - Remove the current samples.h and tests.h and replace with a combined apps.h with a unified app loading interface
- PR 3 Plugins - Adds the base plugin interface and the minimum plugins required to preserve launching a sample
- PR 4 Batch Mode - Adds batch mode (to be released)
- PR 5 Benchmark Mode - Adds benchmark mode (to be released)
- PR 6 More Plugins - Adds more plugins which provide more functionality (to be released)

**Goals of this PR:**
- Remove `VKB_SYMLINKS` - Unused feature
- Remove `VKB_ENTRYPOINTS` - Functionality replaced by CLI
- Bump CMake version to 3.12
- Remove sample.h, test.h and main.cpp generation. Replaced in PR 2
- Remove Vulkan Samples

## General Checklist:

## Sample Checklist


gary-sweet:
I have to admit I find the name 'Extensions' somewhat confusing since it seems to have nothing to do with Vulkan extensions. Surely that is going to lead to misunderstandings down the line. Perhaps 'Features' would be a better name?

tomadamatkinson:
> I have to admit I find the name 'Extensions' somewhat confusing since it seems to have nothing to do with Vulkan extensions. Surely that is going to lead to misunderstandings down the line. Perhaps 'Features' would be a better name?

Possibly Platform Extensions? Or Platform Features?

SaschaWillems:
Maybe modules? It's a more commonly used software architecture term.

TomAtkinsonArm:
I've been stuck between both plugins and modules. Although the wording here doesn't need to be 100% correct; The plugin definition looks more correct in this case.

Plugin
>Plug-in, also called add-on or extension, computer software that adds new functions to a host program without altering the host program itself.

Module
>A module is a software component or part of a program that contains one or more routines. One or more independently developed modules make up a program

I'm going to update everything with Plugins now and also rebase to fix an issue in one of the other PR

marty-johnson59:
Hi Team, per discussion this week, Tom's 3 PRs (#223, #224 and #225) are quite significant and will need help from the team to review.  Please check them out and comment (or approve) as needed. Thanks!

@r-potter, @aurolou, @ZandroFargnoli , @mattruschnv, @AlejandroCosin, @carlosadcq, @gary-sweet, @wasimabbas-arm

TomAtkinsonArm:
Rebased

PR #226, opened 2020-12-08
Update ray tracing sample to the final ray tracing extensions

SaschaWillems:
## Description

This PR updates the ray tracing sample to use the final cross-vendor ray tracing extensions. It also slightly reworks the code to make it easier to read, adds additional comments to make it easier to follow. 

**Note:** This PR also updates the Vulkan headers and Volk sub modules to Vulkan 1.2.162, which is required for the new extensions.

Due to the changes and split into multiple extensions, the updated sample requires the following extensions:
- VK_KHR_ray_tracing_pipeline
- VK_KHR_acceleration_structure
- VK_KHR_buffer_device_address
- VK_KHR_deferred_host_operations
- VK_EXT_descriptor_indexing
- VK_KHR_spirv_1_4
- VK_KHR_shader_float_controls

It also requires Vulkan 1.1. due to VK_KHR_spirv_1_4.

To make sure it doesn't throw any validation errors or warnings, I have tested this with a recent build of the Validation layers from the source, as the SDK is not yet available. 

**Note**: Trying to run this sample with old validation layers not supporting the new extensions will probably cause errors, so if you want to test with validation, you need to build the layers from source.

## General Checklist:


JoseEmilio-ARM (approved):

JoseEmilio-ARM (dismissed):
Tested on RTX 2070

gary-sweet:
I can confirm that this doesn't break the build or other samples for Broadcom.
I cannot test the new functionality.

marty-johnson59:
Hate to keep leaning on my favorite DT vendors, but as this is DT only and we need quick turn to release the samples next week (to align with the SDK launch), can we get reviews from @mattruschnv and @aurolou?  Thanks!

samples/extensions/raytracing_basic/raytracing_basic.cpp
```diff
@@ -238,266 +136,377 @@ uint64_t RaytracingBasic::get_buffer_device_address(VkBuffer buffer)
...
+		| hit       |
+		\-----------/
+*/
+
+void RaytracingBasic::create_shader_binding_tables()
+{
+	const uint32_t           handle_size             = ray_tracing_pipeline_properties.shaderGroupHandleSize;
+	const uint32_t           handle_alignment        = ray_tracing_pipeline_properties.shaderGroupHandleAlignment;
+	const uint32_t           group_count             = static_cast<uint32_t>(shader_groups.size());
+	const uint32_t           sbt_size                = handle_alignment * group_count;
```
AaronHaganAMD: I think this should be handle_size * group_count;

samples/extensions/raytracing_basic/raytracing_basic.cpp
```diff
@@ -238,266 +136,377 @@ uint64_t RaytracingBasic::get_buffer_device_address(VkBuffer buffer)
...
+
+	// Copy the pipeline's shader handles into a host buffer
+	std::vector<uint8_t> shader_handle_storage(sbt_size);
+	VK_CHECK(vkGetRayTracingShaderGroupHandlesKHR(get_device().get_handle(), pipeline, 0, group_count, sbt_size, shader_handle_storage.data()));
+
+	// Copy the shader handles from the host buffer to the binding tables
+	uint8_t *data = static_cast<uint8_t *>(raygen_shader_binding_table->map());
+	memcpy(data, shader_handle_storage.data(), handle_size);
+	data = static_cast<uint8_t *>(miss_shader_binding_table->map());
+	memcpy(data, shader_handle_storage.data() + handle_alignment, handle_size);
```
AaronHaganAMD: I believe this should be handle_size instead of handle_alignment..

samples/extensions/raytracing_basic/raytracing_basic.cpp
SaschaWillems: I'm not sure on this. Wouldn't this result in the buffer being too small on implementations where `shaderGroupHandleSize` is smaller than `shaderGroupHandleAlignment`? I guess we need a function that properly aligns the size instead.

SaschaWillems:
@AaronHaganAMD: Thanks for reviewing my code. I have pushed a small update that aligns the shader group sizes where required. I can't really test this, as on my implementation the group size is the same as the group size alignment, but I think it should be fine now. It would be great of you could take a look at the changes.

samples/extensions/raytracing_basic/raytracing_basic.cpp
```diff
@@ -238,266 +136,383 @@ uint64_t RaytracingBasic::get_buffer_device_address(VkBuffer buffer)
...
+	acceleration_structure_build_range_info.primitiveCount                                           = 1;
+	acceleration_structure_build_range_info.primitiveOffset                                          = 0;
+	acceleration_structure_build_range_info.firstVertex                                              = 0;
+	acceleration_structure_build_range_info.transformOffset                                          = 0;
+	std::vector<VkAccelerationStructureBuildRangeInfoKHR *> acceleration_build_structure_range_infos = {&acceleration_structure_build_range_info};
+
+	// Accelerations structures can be build either on the host or the device, with the former only possible if the implementation supports it
+	if (acceleration_structure_features.accelerationStructureHostCommands)
+	{
+		// Implementation supports building acceleration structure building on host
```
dgkoch: Is there any good reason to prefer building on the host (which is optional), instead of just using device build all the time?

samples/extensions/raytracing_basic/raytracing_basic.cpp
```diff
@@ -238,266 +136,383 @@ uint64_t RaytracingBasic::get_buffer_device_address(VkBuffer buffer)
...
+
+	VkAccelerationStructureBuildRangeInfoKHR acceleration_structure_build_range_info;
+	acceleration_structure_build_range_info.primitiveCount                                           = 1;
+	acceleration_structure_build_range_info.primitiveOffset                                          = 0;
+	acceleration_structure_build_range_info.firstVertex                                              = 0;
+	acceleration_structure_build_range_info.transformOffset                                          = 0;
+	std::vector<VkAccelerationStructureBuildRangeInfoKHR *> acceleration_build_structure_range_infos = {&acceleration_structure_build_range_info};
+
+	// Accelerations structures can be build either on the host or the device, with the former only possible if the implementation supports it
+	if (acceleration_structure_features.accelerationStructureHostCommands)
```
dgkoch: same comment about preferring host builds

samples/extensions/raytracing_basic/raytracing_basic.cpp
SaschaWillems: Not really. Having that path in there may be misleading. Will rework that.

huawei-ahcox: Might be nice to have it controlled from the outside (e.g. commandline) so people can compare build time and quality.

dgkoch: I don't think there is a point for such a basic sample. I'd just keep it simple and only use the guaranteed device path in this sample.

SaschaWillems:
@dgkoch: I have removed the if-statement and AS are now alway built on the device. I added a comment on host builds, so people know that this is optionally available.

dgkoch (approved):
LGTM

samples/extensions/raytracing_basic/raytracing_basic.cpp
AaronHaganAMD: Thank you for the update, this is passing for us now.

JoseEmilio-ARM (approved):

aurolou (approved):

PR #234, opened 2021-01-07
Make physical device selection a bit smarter

baryluk:
When selecting the physical device, try to select CPU `deviceType` last. Do so, by
ordering devices in somehow sensible way, at least for graphics work.

It could be even more smart, by returning ordered list, which can be then 
against the queue and swapchain / present support. But lets not complicate 
too much for a moment.

The motivation, is some users of Mesa, when compile Intel (`anv`) and Gallium
swrast (`lavapipe`) Vulkan drivers, Vulkan driver will load them in random order,
and because `anv` devices usually are integrated, the current code will usually
select `lavapipe`, which is a (multi-threaded) CPU implementation, which does
work, but is not what user wants in 99.9% of cases, until it is selected
explicitly.

I believe Vulkan Samples, should do the right thing, and not cut corners,
otherwise people will copy wrong naive approaches, and make things
harder in the future.

## General Checklist:

## Sample Checklist

## Testing checks

I tested this on Mesa 20.3.2 and Mesa master-git on Linux, amd64. With `radv` (AMD Radeon Fury X) on and `lavapipe` (on 16 core machine) drivers, as well all 4 combinations of them installed or not, as well with different settings of `VK_LAYER_MESA_device_selection`, as well changes in `deviceType` preference in the source code, to make sure the preference order is respected as intended.

I did check multiple samples, including hello_triangle and some other ones too.

### DIFF
--- a/framework/core/instance.cpp
+++ b/framework/core/instance.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2020, Arm Limited and Contributors
+/* Copyright (c) 2018-2021, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -381,20 +381,54 @@ void Instance::query_gpus()
 
 PhysicalDevice &Instance::get_suitable_gpu()
 {
-	assert(!gpus.empty() && "No physical devices were found on the system.");
-
-	// Find a discrete GPU
-	for (auto &gpu : gpus)
+	if (gpus.size() == 0)
 	{
-		if (gpu->get_properties().deviceType == VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU)
+		throw std::runtime_error("No physical devices found.");
+	}
+
+	uint32_t count_device_type[VK_PHYSICAL_DEVICE_TYPE_CPU + 1];
+	memset(count_device_type, 0, sizeof(count_device_type));
+
+	for (uint32_t i = 0; i < gpus.size(); i++)
+	{
+		const auto deviceType = gpus[i]->get_properties().deviceType;
+		assert(deviceType <= VK_PHYSICAL_DEVICE_TYPE_CPU);
+		count_device_type[deviceType]++;
+	}
+
+	const VkPhysicalDeviceType device_type_preference[] = {
+	    VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU,
+	    VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU,
+	    VK_PHYSICAL_DEVICE_TYPE_VIRTUAL_GPU,
+	    VK_PHYSICAL_DEVICE_TYPE_CPU,
+	    VK_PHYSICAL_DEVICE_TYPE_OTHER};
+
+	VkPhysicalDeviceType search_for_device_type = VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU;
+	for (uint32_t i = 0; i < sizeof(device_type_preference) / sizeof(VkPhysicalDeviceType); i++)
+	{
+		if (count_device_type[device_type_preference[i]])
 		{
-			return *gpu;
+			search_for_device_type = device_type_preference[i];
+			break;
 		}
 	}
 
-	// Otherwise just pick the first one
-	LOGW("Couldn't find a discrete physical device, picking default GPU");
-	return *gpus.at(0);
+	int gpu_number = -1;
+	for (uint32_t i = 0; i < gpus.size(); i++)
+	{
+		const auto deviceType = gpus[i]->get_properties().deviceType;
+		assert(deviceType <= VK_PHYSICAL_DEVICE_TYPE_CPU);
+		if (deviceType == search_for_device_type)
+		{
+			LOGI("Using GPU {}: {}, type: {}", i, gpus[i]->get_properties().deviceName, deviceType);
+			gpu_number = i;
+			break;
+		}
+	}
+
+	assert(gpu_number >= 0);
+
+	return *gpus.at(gpu_number);
 }
 
 bool Instance::is_enabled(const char *extension) const
--- a/samples/api/hello_triangle/hello_triangle.cpp
+++ b/samples/api/hello_triangle/hello_triangle.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2020, Arm Limited and Contributors
+/* Copyright (c) 2018-2021, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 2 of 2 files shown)


framework/core/instance.cpp
```diff
@@ -381,20 +381,54 @@ void Instance::query_gpus()
...
 PhysicalDevice &Instance::get_suitable_gpu()
 {
-	assert(!gpus.empty() && "No physical devices were found on the system.");
+	if (gpus.size() == 0)
+	{
+		throw std::runtime_error("No physical devices found.");
+	}
+
+	uint32_t count_device_type[VK_PHYSICAL_DEVICE_TYPE_CPU + 1];
+	memset(count_device_type, 0, sizeof(count_device_type));
```
huawei-ahcox, line 390: ```[VK_PHYSICAL_DEVICE_TYPE_END_RANGE + 1]``` ?

framework/core/instance.cpp
```diff
@@ -381,20 +381,54 @@ void Instance::query_gpus()
...
+
+	for (uint32_t i = 0; i < gpus.size(); i++)
+	{
+		const auto deviceType = gpus[i]->get_properties().deviceType;
+		assert(deviceType <= VK_PHYSICAL_DEVICE_TYPE_CPU);
```
huawei-ahcox, line 395: You removed an assert at the top of the function. Seems a shame to add one back.

framework/core/instance.cpp
```diff
@@ -381,20 +381,54 @@ void Instance::query_gpus()
...
+	for (uint32_t i = 0; i < sizeof(device_type_preference) / sizeof(VkPhysicalDeviceType); i++)
 	{
-		if (gpu->get_properties().deviceType == VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU)
+		if (count_device_type[device_type_preference[i]])
 		{
-			return *gpu;
+			search_for_device_type = device_type_preference[i];
+			break;
 		}
 	}
```
huawei-ahcox, line 415: An integrated or a virtual GPU would still be better than a software implementation.

baryluk: And? This is exactly the point of my patch.

framework/core/instance.cpp
baryluk: They have different purpose.

The one at the top is still there, and turned into exception. Usage of the assert is wrong, as it will be removed in Release mode, but can still happen in runtime, and it is expected to happen, and must be handled, not ignored.

The one I added here, is impossible to happen, and tests internal API invariant. It should never happen at runtime, unless Vulkan API changes in weird and incompatible way.

framework/core/instance.cpp
baryluk: There is no such thing as `VK_PHYSICAL_DEVICE_TYPE_END_RANGE`.

ZandroFargnoli:
@baryluk Thanks for the contribution! Could the description of your PR please include the [checklist](https://github.com/KhronosGroup/Vulkan-Samples/blob/master/docs/pull_request_template.md) of items required to have a PR merged.

baryluk:
> @baryluk Thanks for the contribution! Could the description of your PR please include the [checklist](https://github.com/KhronosGroup/Vulkan-Samples/blob/master/docs/pull_request_template.md) of items required to have a PR merged.

Done.

garyjsweet (commented):
Logic looks ok to me, and builds and runs fine for Broadcom.
I do see an odd mix of CamelCase and snake_case variable names however.
Not sure why the snake_case CI test hasn't complained.

SaschaWillems:
This is a good idea, although I do see two potential things we should address before merging:
- The term "best GPU" is arguable, and with the new logic of automatically selecting a GPU based on it's type we should add a way of explicitly selecting a device by adding a command line argument to override the automatic selection
- I'm not sure if we should add the GPU selection to the hello_triangle sample. That sample is intentionally simple, and just selecting a GPU based on it's type and calling that "best" may cause a false impression for people starting out with that sample.

baryluk:
> This is a good idea, although I do see two potential things we should address before merging:
> 
>     * The term "best GPU" is arguable, and with the new logic of automatically selecting a GPU based on it's type we should add a way of explicitly selecting a device by adding a command line argument to override the automatic selection
> 
>     * I'm not sure if we should add the GPU selection to the hello_triangle sample. That sample is intentionally simple, and just selecting a GPU based on it's type and calling that "best" may cause a false impression for people starting out with that sample.

Fair enough on both points. But considering the current code does a blind selection and doesn't allow configuring it either, it can probably be addressed in a separate Pull Request.

marty-johnson59:
@baryluk can you submit the PR for this fix?  Then I'll push for reviews so we can merge this one.  Thanks!

baryluk:
@marty-johnson59 

> @baryluk can you submit the PR for this fix? Then I'll push for reviews so we can merge this one. Thanks!

I am not sure I understand. What "this fix"?

marty-johnson59:
I was referring to your comment above about creating a separate PR for Sascha's suggestion: "add a way of explicitly selecting a device by adding a command line argument to override the automatic selection."  Is that something you can take on?

baryluk:
> I was referring to your comment above about creating a separate PR for Sascha's suggestion: "add a way of explicitly selecting a device by adding a command line argument to override the automatic selection." Is that something you can take on?

Sure, I can do that. Can take few days, I have some other projects going on too.

framework/core/instance.cpp
abbaswasim: This will, as you know, fail if a new device type is introduced in the future. I would instead suggest you use the size of the list you have created below? `device_type_preference` wdyt?

framework/core/instance.cpp
```diff
@@ -381,20 +381,54 @@ void Instance::query_gpus()
...
+	memset(count_device_type, 0, sizeof(count_device_type));
+
+	for (uint32_t i = 0; i < gpus.size(); i++)
+	{
+		const auto deviceType = gpus[i]->get_properties().deviceType;
+		assert(deviceType <= VK_PHYSICAL_DEVICE_TYPE_CPU);
+		count_device_type[deviceType]++;
+	}
+
+	const VkPhysicalDeviceType device_type_preference[] = {
```
abbaswasim, line 399: The project uses `std::array` and `std::vector` throughout the code. Not sure why you went with raw arrays and C style `sizeof(device_type_preference) / sizeof(VkPhysicalDeviceType)`?

framework/core/instance.cpp
```diff
@@ -381,20 +381,54 @@ void Instance::query_gpus()
...
+	    VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU,
+	    VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU,
+	    VK_PHYSICAL_DEVICE_TYPE_VIRTUAL_GPU,
+	    VK_PHYSICAL_DEVICE_TYPE_CPU,
+	    VK_PHYSICAL_DEVICE_TYPE_OTHER};
 
-	// Find a discrete GPU
-	for (auto &gpu : gpus)
+	VkPhysicalDeviceType search_for_device_type = VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU;
```
abbaswasim, line 406: slight nit pick, `search_for_device_type` sounds like a `verb/action`, it's usually better to have `nouns/name` for variable names. (not really a rule though)
For example `search_for_device_type` should be something like `outcome_of_search_for_device`(not great name though but gets the point across) on the other hand `search_for_device_type` could be a good name for a function.

Similar concern with other variable names.

framework/core/instance.cpp
huawei-ahcox: Shame that VK_PHYSICAL_DEVICE_TYPE_END_RANGE and the like stopped being generated by default. It seems useful to have them for cases like this.
We had them at Vulkan Docs 1.2.139: <https://github.com/KhronosGroup/Vulkan-Headers/blob/4c19ae6b95e44c821d5306adf94842defa57ba21/include/vulkan/vulkan_core.h>
They were gone in 1.2.140: <https://github.com/KhronosGroup/Vulkan-Headers/blob/0c5351f5e9114d3e9033aeae51d036a3d201c082/include/vulkan/vulkan_core.h>
The reasoning is mostly in the private issue but the change is documented here:
<https://github.com/KhronosGroup/Vulkan-Docs/commit/cc5a1ec67405bab282fb037b065a88e0afcf3483>
<https://github.com/KhronosGroup/Vulkan-Docs/issues/1230>
Reasoning is roughly that there are fiddly details in generating the names of some of these and they are becoming less useful anyway given that enum ranges may be sparse going forward, and will not be compacted to be continuous in future spec revisions.

framework/core/instance.cpp
marty-johnson59: @baryluk, any updates on this one?  I still see some questions about how to handle the ambiguity around "best" device - and an over-ride mechanism for those who want to choose what's best for them (separate PR?).  Also questions re: other comments above.  Thanks.

framework/core/instance.cpp
gpx1000: Not to nitpick, but technically it can happen in runtime without need for API changes in a weird and incompatible way.  All that's needed is for an additional device type enum to be defined.  While I'm unaware of a current need for a new device type, I wouldn't peg it as weird behavior if the enums are reordered or if a new device type enum were to be created.  I would agree if the device type enum listed a sentinal end that always comes with the guarantee that no enum would come after it, then it makes sense.

marty-johnson59:
Hi @baryluk, we're closing this for now as it seems to be a stale PR.  Please re-open if this still needs further work.  Thanks!

PR #242, opened 2021-01-20
Update Windows build instructions for newer CMake

berniegp:
## Description

The undocumented CMake `-H` option was removed in CMake 3.13 and replaced with `-S`
See https://cgold.readthedocs.io/en/latest/glossary/-S.html

I have only tested this on Windows, but I suspect the same changes need to apply to the other platforms as well.

## General Checklist:

## Sample Checklist

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -134,7 +134,7 @@ adb shell setprop security.perf_harden 0
 
 ## Dependencies
 
-- CMake v3.10+ (known to work with 3.10.2)
+- CMake v3.10+ (known to work with 3.10.2 and 3.19.3)
 - Python 3
 - Visual Studio 2017 or above
 - [clang-format-8](#clang-format-and-visual-studio)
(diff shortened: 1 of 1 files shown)


gpx1000:
Just to point out the slightly more idiomatic way to do this are these commands:
mkdir build && cd build
cmake ..
cmake --build .

Then all one needs to do to work with windows/linux/mac etc is something like this:
mkdir build-win && cd build-win
cmake -G "Visual Studio 15 2017 Win64" ..
cmake --build . --target SomeTargetName --config Release

One could exclude that last line as all it does is run the build from the command line without need for opening up the IDE.  The -S directory and the -H directory flags are really provided for out of source builds which is not the way CMake was originally designed.

All that said, I think what is here in the current PR is the right solution given the "small change" to get the current correct information in the documentation and the above suggestion would probably fit for a different PR with a larger scope.

gpx1000 (approved):

SaschaWillems (approved):

PR #239, opened 2021-01-18
Add buffer device address sample

HansKristian-Work:
## Description

Adds a sample showing how one can use buffer device addresses. The sample implements a simple scheme where "bindless VBO" is implemented through BDA. The extension is quite abstract to visualize since it just allows for raw pointers, and this was deemed the smallest reasonable sample that can demonstrate why this extension can be useful.

## General Checklist:

Please ensure the following points are checked:

- [x] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [x] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
- [x] I have commented any added functions (in line with Doxygen)
- [x] I have commented any code that could be hard to understand
- [x] My changes do not add any new compiler warnings
- [x] My changes do not add any new validation layer errors or warnings
- [x] I have used existing framework/helper functions where possible
- [x] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [x] My changes do not add any regressions
- [x] I have tested every sample to ensure everything runs correctly
- [x] This PR describes the scope and expected impact of the changes I am making

## Sample Checklist

If your PR contains a new or modified sample, these further checks must be carried out *in addition* to the General Checklist:
- [x] I have tested the sample on at least one compliant Vulkan implementation
- [x] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [x] I have stated on what implementation the sample has been tested so that others can test on different implementations and platforms
- [x] Any dependent assets have been merged and published in downstream modules
- [x] For new samples, I have added a paragraph with a summary to the appropriate chapter in the [samples readme](./../samples/README.md)

Tested on RADV, ANV and Arm Mali.

PR #249, opened 2021-02-18
Small copyright fix

JoseEmilio-ARM:
Copyright comment was missing the opening tag


marty-johnson59:
If this is just a copyright tweak, I don't think we need to go through all the review/approvals here.  LMK if anyone feels differently - else I'll go ahead and merge this one

JoseEmilio-ARM:
Already merged as part of #239

Issue #240, opened 2021-01-18
What is the expected behaviour of samples that cannot be executed due to unsupported extensions?


r-potter:
In reviewing #213 and #216, I noted that they have different failure behaviour on devices that don't support the required extensions. 

#213 still ran, but noted that the extension was unsupported and that the checkbox was a no-op (which seemed nice).
#216 failed to start, and logged the missing extension as the cause in the error log (which is what most of the other samples do in this case)

Neither option seems unreasonable, but I wonder if consistency would be a positive here in future?

JoseEmilio-ARM:
The calls to enable a device/instance extension have a second parameter that indicates whether the extension is 'optional'. If it is, the sample will still run and it is responsible for handling the necessary changes to render the scene without using the extension. We found this useful in samples where we wanted to show several features and did not want to crash if just one of them was not supported. In cases where the sample is all about a particular extension, not marking it as optional causes a crash.

In 213 all extensions are marked as optional, whereas in 216 a comment points out that they are not https://github.com/KhronosGroup/Vulkan-Samples/pull/216/files#diff-923f75ffa3930c2a37b4d8b0e84c2b5ac3b14de8bb77bb0027415bf0c9b74fe1R33

I agree that ideally we would never crash (particularly due to a bug on Android where the sample menu does not re-populate), it might be better to introduce some mechanism for samples to have a fallback, for example rendering a message on the screen and letting the user exit.

TomAtkinsonArm:
I have noticed the error handling issues a lot more after it was mentioned in the last call!

The main area of the project that i have run into similar issues is using batch mode and it prematurely closing as an error is thrown when choosing extensions.

Currently we catch runtime errors in the `main` function which means that if an error occurs at any point in a samples execution the entire app will close immediately. This is the behaviour seen in samples where the extension is not present and the application closes. Other than this there is no set in stone fallback.

Adding a fallback to each sample would be a great option but would result in time to review and implement this on older samples. We could implement a failure screen which shows error information and allows for the app to gracefully exit. This would probably take the form of a `VulkanSample` application that only renders the UI with a helpful message such as "extensions XXX not available".

What are your thoughts on future improvements? Are there other options?

TomAtkinsonArm:
Graceful closure implemented in plugins

PR #254, opened 2021-03-13
Plugins

### LINKED ISSUE #263: Add option to close console window after running sample using --benchmark on Windows
WindowsPlatform::terminate currently pauses waiting for input after running the samples in benchmark mode on Windows i.e running:  "vulkan_samples --sample msaa --benchmark 100" will run the sample for 100 frames, exit the sample but keep the console window open waiting for user input - the enter key to be pressed.

Is there any possibility of adding a command line option or similar to avoid this behaviour?

### LINKED ISSUE #266: VSync differences between samples based on different classes, no way to toggle vsync
It looks like API-Samples are always running with uncapped framerates, where as the samples based on the ARM-framework always run with capped framerate.

There is currently no way to toggle this, and from a quick look at the framework, it looks like this requires a more complex change to make this work.

This should be unified in such a way that vsync can be enabled/disabled via a command line switch and the swap chain then picks the proper present mode, no matter what framework the sample is based on.

TomAtkinsonArm:
## Description

To all reviewers, thank you for your patience with this PR, there are a lot of changes and it has been a while for it to be in a reviewable state. I have provided a change log to try and make the review process easier. I attempted to add all changes but I may have missed some.

The goal is to replace the current functionality of the application with a plugin based system. This removes a coupling between `Platform` <-> `Application` <-> `VulkanSamples` and redistributes features across individual plugins. Future feature requests or CLI toggles can now be accomplished via the plugin structure without the need to change the framework in major ways.

For instance, #295 can be solved by adding a hook when an instance is created `on_create_instance(VkInstanceCreateInfo* info)`. A plugin can be created which provides CLI flags to toggle the different validation layer states. The correct struct can then be appended to `info.pNext` to configure the validation layers. Effect on the framework is a single `platform->on_create_instance()` call.

#### Common

- CMake bump to 3.12
    - Possibly a concern as the android build system ships with 10. However, most developers will have a newer version of CMake if they are active C++ and graphics developers. 3.12 brings a few features that clean up parts of the build system
- C++17 Optional class
    - A mock of the C++17 optional class which allows the use of optional in the project, can later be replaced when we upgrade to C++17 down the line
- Tags
    - Struct based tagging, used with plugins to identify plugins and group them by behaviour
- Removed Entrypoints
    - A decision made a while back where they where less common
    - May still be desired to reduce compilation time, although entrypoints don't fix the root cause they did reduce the amount to be compiled
- Removed Symlinks
    - Generally not used
- Removed unused parameters in some classes and functions which are not used

#### Applications, Samples and Tests (apps library)

- Applications create functions and ids are stored in AppInfo
- SampleInfo expands on AppInfo by providing sample specific meta which is configured by the build system
- TestInfo expands on AppInfo, no meta is currently available
- The app static library now provides some helper functions to retrieve sub sets of samples and tests

#### Plugin Framework

- Plugin
    - Hook based interface used to implement custom behaviour at different points of the platforms lifecycle
    - Current Hooks: OnUpdate, OnAppStart, OnAppClose, OnAppError, OnPlatformClose, PostDraw
    - Tag based identification so that both plugin class names and behaviour types can be used to define interoperability and see if plugins are activated.  See `associate_plugins` and `Platform::using_plugin<T>()`
- Plugin Base
    - Enforces the use of Tags
    - Partially implements some of the common components of the plugin interface

Hooks must be enabled via enums in the platforms constructor for their hook method to be called in the lifecycle. This is a pre-emptive optimisation which should allow for lots more plugins without a larger impact on performance. A map is created which relates hooks to plugin pointers. This means that only a plugin which has previously stated that it wants to use a specific hook will be called for that hook. See `Platform::hooks` and the `HOOK` definition at `platform.cpp:406`

#### Plugins

- Batch Mode
    - Moved from `platform` and `VulkanSamples` to `BatchMode`
    - Provides the same CLI and behaviour
    - An extra try/catch is added to the platform lifecycle to attempt to continue to the next app if there is a failure
        - Partly addresses #240
        - Still some work to be done with error handling and propagation in the future
- Benchmark Mode
    - Moved to `BenchmarkMode`
    - Changed to outputting stats after an application exits to ensure interoperability with as many plugins as possible
- FPS Logger
    - Gives the user the option to see FPS statistics in the log
    - Disables by default, enabled by `--log-fps`
- Screenshot
    - Allows for a screenshot to be taken and saved at a specific frame `--screenshot <frame_num>`
    - Optional output name with `--screenshot-output <name>`
    - Path is currently locked due to how the filesystem and screenshot function works (out of scope)
- Start Sample
    - Running a Sample is now a sub command instead of a flag `sample <sample name>`
- Start Test
    - Running a Test is now a sub command instead of a flag `test <test name>`
    - Test plugin is only available when `VKB_BUILD_TESTS` is enabled
    - May be possible to keep the plugin when `VKB_BUILD_TESTS` is not defined. This would be a nice place to display a message to build the project with `VKB_BUILD_TESTS`
- Stop After
    - Stop after a set amount of frames
    - Currently not interoperable with Batch Mode, could be adjusted to work along side batch mode so that a user can stop after a frame count or stop after a duration
- Window Options
    - Alter the window configuration (VSYNC, extent, headless/fullscreen/fullscreen borderless with glfw)
    - Each window can now support each of these features if possible
    - Addresses #266
- Force Close
   - Allows the application to close without halting
   - Addresses #263

#### CLI

- CLI is rebased on the recent CLI transition away from docopt
- `HelpFormatter` now provides custom logic for displaying the CLI
- The original CLI system in plugins has been replaced with the new CLI/CLI11 system

##### Control Flow

- Plugins are passed to `associate_plugins` 
- Plugins associate by exclusion or inclusion lists which each plugin stores
- Plugins are then passed to the CLI Parser
    - In this case CLI11 creates sub apps for each plugin (Their mechanism of grouping re-occuring elements, you cant redefine flags)
    - The groups are then composed together by the inclusion and exclusion lists
    - The original command visitor is used to generate flags in the subgroups
- Parser is now used to parse command line arguments and retrieve values for each argument

#### Android Platform

- Move getSamples from `VulkanSamples` application to android platform
- Android events are now passed through the platform first
   - The platform should be the only known truth when it comes to handling events. This allows a more unified response to events across all platforms
- Poll events replaced by `process_android_events`
    - `process_android_events` processes android lifecycle events, returns false if the app should close (`destroyRequested`)
- `main_loop` replaced by processing android events within `AndroidWindow::process_events`
    - Now there is only one version of the main application loop that needs to be maintained

##### Android App

- SampleListView now launches a sample by using the `launchSample` helper
- Use `--es "cmd" "normal command argument line here"` or `--esa "cmd" "normal","command","argument","line","here"` when launching the android app from a command line
    - Allows for any valid CLI command to be passed to the android app
- Added a few helpers for launching Samples and Tests to remove duplication

#### Application & VulkanSamples Application

An issue was the coupling of Platform <-> Application <-> Vulkan Samples. I was able to remove this coupling by treating the platform as the only source of truth. All logic will be passed back to the platform and propagate outwards. An example of this is the responses to android events. No more `platform->get_app()->do_extra_platform_stuff()`.

- Step functionality has been spread across plugins / platform
    - Removes the coupling between platform lifecycle and application lifecycle
- Platform input event processing has been moved to the platform
- Removed setters and getters that are related to plugin code
    - Before the codebase had a lot of `platform->get_app()->is_headless()` sort of calls.
- `parse_options` removed
    - Parsing configuration state is now left to plugins/parsers/platform

- VulkanSamples application deleted
    - This application combined all of the major logic used for each plugin and the platform lifecycle
    - Before: Platform -> loads VulkanSamples -> loads applications / many applications. Now: The platform manages everything
    - CLI responsibility passed to parsers
    - Batch mode code moved to batch mode
    - Application loading given to platform
    - Finding create functions given to AppInfo
    - Reacting to CLI flags handled by plugins

## General Checklist:

## Sample Checklist

No changes to samples or the vulkan framework have been implemented


SaschaWillems:
I can't seem to run any sample with this PR. I did a clean checkout, clean cmake and clean build, and I always get an "unknown command" error in the argument parser, even I'm not passing any arguments and just run the binary from the command line:

image

For the code on master, running the binary without arguments shows a list of available examples, which seems to be missing in this PR.

Any idea if this is a fault on my side and how to fix this?

I'd like to the review and this give some feedback ;)

framework/platform/plugins/parser.cpp
```diff
@@ -0,0 +1,396 @@
...
+		for (auto ext : passives)
+		{
+			compatable.push_back(ext);
+		}
+
+		usage.insert({entrypoint, compatable});
+	}
+
+	const char *spacer   = "  ";
+	const char *app_name = "VulkanSamples";
```
gary-sweet: The app name is actually "vulkan_samples". This causes all the usage to display the wrong command.

gary-sweet:
Is there a way to see the list of samples now? "vulkan_samples --help" doesn't do it, and I can't see an option to list them either.

gary-sweet:
Batch mode only runs samples until it hits one that is unsupported, then it aborts. To be fair I think it's always done this, but it should really just skip over any samples that fail because of unsupported features.

app/plugins/window_options/window_options.cpp
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+#include "window_options.h"
+
+namespace plugins
+{
+vkb::Flag width_flag = {"width",
+                        vkb::Flag::Type::FlagWithOneArg,
+                        "Initial window width"};
+
```
gary-sweet, line 21: In direct-to-display mode, most of these arguments do nothing at all. I think there were previously not displayed in the  help text in that build mode. Is that still possible?

app/plugins/benchmark_mode/benchmark_mode.cpp
```diff
@@ -0,0 +1,59 @@
...
+	// Benchmark is active
+}
+
+void BenchmarkMode::on_update(float delta_time)
+{
+	elapsed_time += delta_time;
+	total_frames++;
+}
+
+void BenchmarkMode::on_app_start(const std::string &app_id)
```
gary-sweet, line 50: I'm finding benchmark mode not very useful the way it's set up right now.
The timing includes all of the loading and setup time, so even though the app is running at 60fps for a minute, the benchmark is reporting 31.6fps.  We should only start timing when rendering begins, not on app start I would have thought.

gary-sweet:
The usage allows filtering by tags and categories, but does not say what the valid tags or categories are:

[info]   --T=<arg>...                  Filter samples by tags
[info]   --C=<arg>...                  Filter samples by categories

gary-sweet:
I'm seeing issues with --screenshot as well, but they seem to pre-exist this change, so I'll look into those separately.

SaschaWillems:
This breaks all samples for me on Android. The launcher works fine, though samples are now sorted alphabetical, but I can't run any of the samples. It just crashes back to the home screen without an error message. 

In the Android Device Monitor I can see the cause of the crash: 03-26 18:27:58.712: A/DEBUG(24725): Abort message: 'JNI DETECTED ERROR IN APPLICATION: GetStringUTFChars received NULL jstring

gary-sweet:
> I'm seeing issues with --screenshot as well, but they seem to pre-exist this change, so I'll look into those separately.

I've fixed the screenshot issues here : https://github.com/KhronosGroup/Vulkan-Samples/pull/261

TomAtkinsonArm:
> I can't seem to run any sample with this PR. I did a clean checkout, clean cmake and clean build, and I always get an "unknown command" error in the argument parser, even I'm not passing any arguments and just run the binary from the command line:
> 
> image
> 
> For the code on master, running the binary without arguments shows a list of available examples, which seems to be missing in this PR.
> 
> Any idea if this is a fault on my side and how to fix this?
> 
> I'd like to the review and this give some feedback ;)

@SaschaWillems, can you share your attempted usage?

SaschaWillems:
> @SaschaWillems, can you share your attempted usage?
I have tried with all available command line options that work in master, e.g:
--sample hello_triangle
hello_triangle

Both arguments run the hello_triangle in master, but in this branch I'm always getting above error, no matter what command line arguments I used.

It looks like docopt has problems parsing the arguments, as I always end up in the final catch block:

image

I did multiple cleanups, deleted the CMake cache, etc. to no avail.

TomAtkinsonArm:
> It looks like docopt has problems parsing the arguments

`There was insufficient memory to determine whether the regular expression could match the specified character sequence.`

The error is coming from regex. Unsure which change created this

TomAtkinsonArm:
Similar issues are seen here https://github.com/docopt/docopt.cpp/issues/49. Seems like the only suggested fix is to use boost regex...

framework/platform/plugins/parser.cpp
TomAtkinsonArm: Fixed

TomAtkinsonArm:
> Batch mode only runs samples until it hits one that is unsupported, then it aborts. To be fair I think it's always done this, but it should really just skip over any samples that fail because of unsupported features.

@gary-sweet this is an issue in the pre-existing framework. It would require a rework of the error handling logic and i think would be better placed in a separate PR

app/plugins/window_options/window_options.cpp
TomAtkinsonArm: Fixed by passing the width and height values chosen by the platform

app/plugins/benchmark_mode/benchmark_mode.cpp
TomAtkinsonArm: image

So here we start the app and render the first frame without calling update on any plugin. After this we calculate delta time and call on_update with the correct time of the frame. then pass 60fps to the active app. This means that all graphs shown by the app will state 60fps but the plugin itself will go off the correct time value.

I believe the justification is that when running benchmark the rendered frames must be consistent between benchmarks hence the static fps. I'm a little bit shaky on the perfect solution but this may explain the discrepancies between the actual and visual reading of fps.

TomAtkinsonArm:
@SaschaWillems @gary-sweet, Thank you for starting the reviews. I think the commands themselves can be reworked to provide a more consistent interface for showing what values can be used for each parameter.

The regex issue seems to be a breaking bug. I rebuilt the app several times and now plugins work again without changes. Seems to be intermittent and compiler dependant... It is also a known issue that doesn't have an official alternative other than boost for now

TomAtkinsonArm:
@SaschaWillems Docopt should not fail to parse arguments now. Please have another go and update me if this fix works locally for you :)

SaschaWillems:
Unfortunately it still fails for me, both in release and debug. The only difference is that I now get the reason for the failure printed out:

image

This is with MSVC 2019 16.9.4.

TomAtkinsonArm:
This is also the error i encountered. I fixed this by temporarily removing the regex stack limitations in my fork of docopt. Does this error still occur for you after syncing submodules and pulling my "fix"?

SaschaWillems:
Do I have to manually pull in your docopt fork, or has this been fixed in the docopt repo?

SaschaWillems:
With your docopt fix I'm getting a crash deep inside MSVCS's memory allocation. I'll try to investigate this later to rule out any errors on my side.

TomAtkinsonArm:
Hey @gary-sweet and @SaschaWillems, I think this is finally in a reviewable state. I have hidden all comments that are outdated. The ones remaining I need to look over :)

Less major changes now as this piggy backs of the recent CLI and vsync PRs

gary-sweet:
> Is there a way to see the list of samples now? "vulkan_samples --help" doesn't do it, and I can't see an option to list them either.

This still appears to be the case with the latest version.

gary-sweet:
This is mostly looking really good for me under testing :-)

It's still missing some stuff in the 'help' as I mentioned before and I do have a couple of issues with headless mode.
1) It just segfaults when I try to use it with anything but the performance samples.
2) The fps seems significantly slower when running headless (45fps vs 100fps for example).

gary-sweet:
> 1. It just segfaults when I try to use it with anything but the performance samples.

The crash appears to be at the end of `ApiVulkanSample::submit_frame()`. It calls `device->get_queue_by_present(0)` which fails because there are no queues that support 'present' since we're in headless mode.

Note: It may have always been like this. I've only debugged with this PR in place.

TomAtkinsonArm:
Thanks for looking at this Gary

> It may have always been like this. I've only debugged with this PR in place.

Would you be able to test the same scenario on master to verify that it is indeed introduced by Plugins? Also what command did you use so that I can test on my side as well

In terms of the sample list, I've left a few comments that further up which I intend to investigate and fix when I get chance to give this PR another pass :)

gary-sweet:
> Would you be able to test the same scenario on master to verify that it is indeed introduced by Plugins?

Sure. I'll give it a try.

> Also what command did you use so that I can test on my side as well

`./vulkan_samples sample instancing --headless`

gary-sweet:
> > Would you be able to test the same scenario on master to verify that it is indeed introduced by Plugins?
> 
> Sure. I'll give it a try.

As suspected, it also crashes in master, so it's not related to this change. I've raised [this issue](https://github.com/KhronosGroup/Vulkan-Samples/issues/314) for it.

gpx1000:
moving it into the if conditional check results in a fix to the crash.

gary-sweet:
> moving it into the if conditional check results in a fix to the crash.

Sure. This will also skip the waitIdle, and I wasn't sure if it was needed. It could use vkDeviceWaitIdle() instead I guess.

gpx1000:
I think it's fair to remove the wait idle completely.  In the comment above it where synchronization is a typo (missing the h), it's correctly pointing out that proper synchronization in the samples is the real way to not have to do the expensive waitidle.

Given how often Synchronization is a pain point, maybe having more rigor in the samples at the cost of being more verbose isn't a bad idea?  In other words, the samples themselves should be rewritten with an eye towards proper synchronization.

gary-sweet:
> I think it's fair to remove the wait idle completely. 

I'll defer to those who know the framework better to decide on that.

TomAtkinsonArm:
The majority of the API samples where submitted by @SaschaWillems, so the best guidance on this would come from Sascha. This does require a fix but I think this PR has already made enough changes to the project. So to avoid bloating the PR more and as this is a pre-existing issue, should this be resolved in a separate PR?

TomAtkinsonArm:
Added a samples list, slightly less compacted version of showing samples compared to master. I prefer this as we can add more lines stating author or tags if we want. The table version is restricted by console width.

See start sample plugin

image

TomAtkinsonArm:
The global catch is now removed, replaced by returning error codes. The newer try/catch within the main loop will catch any `vkb::Application` specific errors meaning the error handling out side of this can be more graceful. Added `ErrorCode::Help` for showing the help menu and `ErrorCode::Close` to close the application without help (useful for `vulkan_samples samples` and other CLI output tasks in the future)

SaschaWillems:
As for the waitIdle, my samples still require it. They don't have proper per-frame resources yet, so they wouldn't work anymore. But as Garry noted, we should defer that to another PR? I'm working on updating my own samples to proper sync and want to update our API samples too, but that takes some time.

CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.12)
```
gpx1000, line 18: Just curious about why the version bump?

SaschaWillems:
I'm having trouble testing this. I'm on VS2019 and when running the app starts the requested sample, the sample loads all of it's resource but then fails with the following error and closes:

image

I did several clean builds, both in release and debug. Idk why this isn't working for me. Any ideas?

SaschaWillems:
A clean checkout from the repo that the branch is based on helped. It's now working as expected :)

SaschaWillems:
To rule out any problems with me switching between PRs, I checked out from your repo and I'm encountering the same problem.

It also seems that you can no longer run a sample by specifying it's name as the sole argument, but rather by "sample name_of_the_sample". Is that correct? Took me some time to figure out, but maybe I'm specifying the wrong arguments?

gary-sweet:
> The majority of the API samples where submitted by @SaschaWillems, so the best guidance on this would come from Sascha. This does require a fix but I think this PR has already made enough changes to the project. So to avoid bloating the PR more and as this is a pre-existing issue, should this be resolved in a separate PR?

Yes. That's why I raised a separate issue.

CMakeLists.txt
TomAtkinsonArm: Introduced some useful macros which removed the need for our own versions. I think it may have also introduced the `glob(CONFIGURE_DEPENDS)` which reruns glob if cmake detects file changes (glob but without the bad bits)

TomAtkinsonArm:
> I did several clean builds, both in release and debug. Idk why this isn't working for me. Any ideas?

@SaschaWillems what command line arguments did you use?

From a glance, it looks like the app has finished and the platform is trying to load the next requested app before it closes...

SaschaWillems:
> @SaschaWillems what command line arguments did you use?

```sample hello_triangle```

Just running with

```hello_triangle```

doesn't seem to work anymore.

**Comment by TomAtkinsonArm** [comment]
I removed the `vulkan_samples <sample_id>` positional as it introduces ambiguity. It is much more stable and extendable to follow a `vulkan_samples command <positional> --options` pattern where the command has a meaning.

For instance now we can use `vulkan_sample sample <sample_id>` to load a given sample, `vulkan_sample samples` to list all samples, `vulkan_sample batch --C=<> --T=<>` to run a sub set of samples... So removing the positional removes ambiguity and makes the cli more consistent.

**Comment by TomAtkinsonArm** [comment]
> > @SaschaWillems what command line arguments did you use?
> 
> `sample hello_triangle`
> 
> Just running with
> 
> `hello_triangle`
> 
> doesn't seem to work anymore.

Ill give this a go now and try to track down the error :)

**Comment by TomAtkinsonArm** [comment]
Yep I introduced this bug in my recent commits, its to do with the `app_requested` check

**Comment by TomAtkinsonArm** [comment]
Should be resolved now @SaschaWillems

- `main_loop`
- No app requested -> Close
- while platform alive
    - load app if requested
    - update
    - process inputs

- On error
    - Propagate error to plugins
        - This is where batch mode will queue the next app to load
    - if there is an app requested continue with the loop
    - no app requested -> fatal error (high chance this is a single app run that has errored)

Not the perfect system, but this is how I am attempting to reduce crashes in batch mode

File: CMakeLists.txt (hunk 4 above)
Comment (TomAtkinsonArm) [reply CMakeLists.txt:18]: image

**Comment by SaschaWillems** [comment]
I can confirm that it works as expected now :)

The new argument pattern of "sample SAMPLE_NAME" is fine be me.

**Review by SaschaWillems — COMMENTED** [review]
LGTM now with all the recent changes.

File: framework/platform/parsers/CLI11.cpp [hunk 5]
```diff
@@ -172,4 +153,83 @@ std::vector<std::string> CLI11CommandParser::get_command_value(Command *command)
 
 	return it->second->results();
 }
+
+// TODO: Explain this
```
JoseEmilio-ARM: Also need to update the new command-line pattern in the documentation (see https://github.com/KhronosGroup/Vulkan-Samples#usage)

gpx1000 (dismissed):
Seems to work for me.  Excellent job on this massive PR!

framework/platform/window.h
```diff
@@ -17,27 +17,69 @@
...
+		OptionalExtent        extent;
+	};
+
+	struct Properties
+	{
+		std::string title     = "";
+		Mode        mode      = Mode::Default;
+		bool        resizable = true;
+		Vsync       vsync     = Vsync::Default;
+		Extent      extent    = {1024, 768};
```
SaschaWillems: With this default extent, all samples start with a 4:3 ratio, which differs from what's currently used in master (1280x760, 16:6).

SaschaWillems (changes requested):
This looks very good now, with the only thing that needs fixing being the default window size.

TomAtkinsonArm:
Changed the default window size to @SaschaWillems suggested 1280x760

SaschaWillems:
> Shouldn't this be 1280 x 720??

Correct. Looks like I used the wrong number im my comment. Sorry for that.

TomAtkinsonArm:
Should be correct now 🤣

Thanks all

gary-sweet (approved):
Seems fine, other than the known pre-existing issues which will be dealt with elsewhere.

SaschaWillems (approved):

gpx1000 (approved):
LGTM

PR #261, opened 2021-03-29
Fix vkb::screenshot()

### LINKED ISSUE #259: screenshot option creates a sampleable, linear destination image which isn't guaranteed to be supported
The screenshot utility code requests that the destination image of copy be both linear (VK_IMAGE_TILING_LINEAR) and sampleable (VK_IMAGE_USAGE_SAMPLED_BIT).

Linear is obviously correct, but some hardware (including ours) cannot sample from linear images.
The image doesn't actually get sampled from, so just removing the VK_IMAGE_USAGE_SAMPLED_BIT will resolve the issue.

gary-sweet:
## Description

Addresses the incorrect use of VK_IMAGE_USAGE_SAMPLED_BIT as described in issue #259.

Changes vkb::screenshot() to use VkCopyImageToBuffer  rather than VkCopyImage.

This simplifies the code, and I believe is a better practice that trying to read image data directly.
The previous code also didn't work for me. It either produced empty images, or images that started with a horizontal offset for me.

Tested only on Linux devices.

Fixes #259

## General Checklist:

### DIFF
--- a/framework/common/utils.cpp
+++ b/framework/common/utils.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2020, Arm Limited and Contributors
+/* Copyright (c) 2018-2021, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -58,19 +58,15 @@ void screenshot(RenderContext &render_context, const std::string &filename)
 	auto &frame          = render_context.get_last_rendered_frame();
 	auto &src_image_view = frame.get_render_target().get_views().at(0);
 
-	auto width  = render_context.get_surface_extent().width;
-	auto height = render_context.get_surface_extent().height;
+	auto width    = render_context.get_surface_extent().width;
+	auto height   = render_context.get_surface_extent().height;
+	auto dst_size = width * height * 4;
 
-	core::Image dst_image{render_context.get_device(),
-	                      VkExtent3D{width, height, 1},
-	                      format,
-	                      VK_IMAGE_USAGE_SAMPLED_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT,
-	                      VMA_MEMORY_USAGE_GPU_TO_CPU,
-	                      VK_SAMPLE_COUNT_1_BIT,
-	                      1, 1,
-	                      VK_IMAGE_TILING_LINEAR};
-
-	core::ImageView dst_image_view{dst_image, VK_IMAGE_VIEW_TYPE_2D};
+	core::Buffer dst_buffer{render_context.get_device(),
+	                        dst_size,
+	                        VK_BUFFER_USAGE_TRANSFER_DST_BIT,
+	                        VMA_MEMORY_USAGE_GPU_TO_CPU,
+	                        VMA_ALLOCATION_CREATE_MAPPED_BIT};
 
 	const auto &queue = render_context.get_device().get_queue_by_flags(VK_QUEUE_GRAPHICS_BIT, 0);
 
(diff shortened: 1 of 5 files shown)


framework/common/utils.cpp
```diff
@@ -58,37 +58,31 @@ void screenshot(RenderContext &render_context, const std::string &filename)
 	auto &frame          = render_context.get_last_rendered_frame();
```
gpx1000, line 53: the local variable format define at line 55 is no longer used

We also no longer use the following includes:
#include "core/pipeline_layout.h"
#include "core/shader_module.h"
#include "scene_graph/components/image.h"
#include "scene_graph/components/mesh.h"
#include "scene_graph/components/pbr_material.h"
#include "scene_graph/components/sampler.h"
#include "scene_graph/components/texture.h"
#include "scene_graph/components/transform.h"

framework/common/utils.cpp
```diff
@@ -109,26 +103,26 @@ void screenshot(RenderContext &render_context, const std::string &filename)
 	swizzle          = std::find(bgr_formats.begin(), bgr_formats.end(), src_image_view.get_format()) != bgr_formats.end();
```
gpx1000: it makes sense to just define swizzle here as line 99 definition with setting to false is never read and proves unused until here.

framework/core/command_buffer.h
```diff
@@ -1,4 +1,4 @@
-/* Copyright (c) 2019-2020, Arm Limited and Contributors
+/* Copyright (c) 2019-2021, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
```
gpx1000, line 4: line 81: move constructors should be marked noexcept

line 161: End of the line has an extra ';'.  The macro LOGE handles the ';'

framework/core/command_buffer.cpp
```diff
@@ -455,6 +455,12 @@ void CommandBuffer::copy_buffer_to_image(const core::Buffer &buffer, const core:
 	                       to_u32(regions.size()), regions.data());
 }
 
+void CommandBuffer::copy_image_to_buffer(const core::Image &image, VkImageLayout image_layout, const core::Buffer &buffer, const std::vector<VkBufferImageCopy> &regions)
```
gpx1000, line 458: this method can be made const.

gpx1000 (dismissed):
Both the old version of using a screenshot and this method work in Linux when tested locally.  I think we should change the key binding from printscr to somehting else as that key is bound by the system in linux to capture the whole screen.

The comments I found are minor and shouldn't hold up the acceptance of this PR.  If there's time it'd be great to address them otherwise, this LGTM;

framework/core/command_buffer.cpp
gary-sweet: It could, but doing so would make it inconsistent with all the other similar methods around it. In general, I prefer to fit in with the existing code (unless it's blatantly wrong of course). If we were to make this const, then we'd also want to make a whole lot more of these const, and that's well beyond the scope of this fix.

framework/core/command_buffer.h
gary-sweet: Neither of those lines have been modified by this PR, and similar issues likely exist in many other places, so I consider them beyond the scope of this change.

framework/common/utils.cpp
gary-sweet: Agreed. This was the original code, but does look a bit odd. I'll fix it up.

framework/common/utils.cpp
gary-sweet: Done

framework/common/utils.cpp
gary-sweet: Done

SaschaWillems (approved):

SaschaWillems:
LGTM and works as expected on an NVIDIA RTX 2060.

One thing I noticed, which is not related to this PR at all, is that the screenshot feature only seems to work in samples using the ARM framework, but not in the samples using the API sample base class as the input handler is never called in these. That may be worth a separate issue.

ZandroFargnoli (approved):
LGTM! Tested on Windows 👍

marty-johnson59:
Assuming Steven is still OK with this (from comments, it looks like he is), this looks like it's ready to merge.

PR #264, opened 2021-04-09
Add async compute performance sample

HansKristian-Work:
## Description

This sample demonstrates how multiple queues should be used on TBDR (Arm) architectures to keep all queues busy with work. The sample uses a compute shader post-processing pipeline, which normally is a problem on mobile, but it can be made to work well if we use multiple queues to remove starvation of fragment hardware queue.

Also demonstrates more direct use of semaphores, queue ownership transfers.

Tested on Arm, RADV and NV.

## General Checklist:

## Sample Checklist


framework/core/device.cpp
```diff
@@ -40,7 +40,27 @@ Device::Device(PhysicalDevice &gpu, VkSurfaceKHR surface, std::unordered_map<con
 	{
 		const VkQueueFamilyProperties &queue_family_property = gpu.get_queue_family_properties()[queue_family_index];
 
-		queue_priorities[queue_family_index].resize(queue_family_property.queueCount, 1.0f);
+		if (gpu.has_high_priority_graphics_queue())
+		{
+			uint32_t graphics_queue_family = get_queue_family_index(VK_QUEUE_GRAPHICS_BIT);
+			if (graphics_queue_family == queue_family_index)
+			{
+				queue_priorities[queue_family_index].reserve(queue_family_property.queueCount);
```
SaschaWillems, line 48: I think this should be ```resize``` instead of ```reserve```. With the latter, this crashes for my on Windows with MSVC2019.

SaschaWillems (changes requested):
This is a great sample, as usual 👍 

I noticed a few problems tough, see bewlo.

framework/rendering/render_context.cpp
```diff
@@ -409,6 +415,19 @@ void RenderContext::end_frame(VkSemaphore semaphore)
...
+	if (acquired_semaphore)
+	{
+		release_owned_semaphore(acquired_semaphore);
+		acquired_semaphore = VK_NULL_HANDLE;
+	}
+}
+
+VkSemaphore RenderContext::consume_acquired_semaphore()
+{
+	assert(frame_active && "Frame is not active, please call begin_frame");
```
SaschaWillems, line 427: This error is triggered on some of the existing samples for me:

image

Tested with the MSAA and afbc samples, that shows above error in debug mode after the first few frames and then exists.

SaschaWillems (changes requested):
A change in the framework causes some of the existing samples to crash for me.

framework/core/device.cpp
HansKristian-Work: Fixed.

framework/rendering/render_context.cpp
HansKristian-Work: Fixed.

samples/performance/async_compute/async_compute.cpp
```diff
@@ -0,0 +1,809 @@
...
+{
+	return {
+	    std::max(1u, extent.width >> level),
+	    std::max(1u, extent.height >> level),
+	    std::max(1u, extent.depth >> level)};
+}
+
+void AsyncComputeSample::prepare_render_targets()
+{
+	// To make this sample demanding enough to saturate mobile devices, use 4K.
```
gary-sweet: Using a size here that is larger than the swapchain causes significant performance issues on some of our devices (which are designed to be very small and low power). You noted that this was to allow comparing of cross-device performance, but I don't believe that's the aim of these performance samples; they allow you to compare various options on a single device.

samples/performance/async_compute/async_compute.cpp
```diff
@@ -0,0 +1,809 @@
...
+	    {*device, size, VK_FORMAT_D32_SFLOAT,
+	     VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT | VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT,
+	     VMA_MEMORY_USAGE_GPU_ONLY},
+	    {*device, size, VK_FORMAT_D32_SFLOAT,
+	     VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT | VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT,
+	     VMA_MEMORY_USAGE_GPU_ONLY},
+	};
+
+	// 8K shadow-map overkill to stress devices.
+	vkb::core::Image shadow_target{*device, VkExtent3D{8 * 1024, 8 * 1024, 1}, VK_FORMAT_D16_UNORM,
```
gary-sweet: 8K images aren't universally supported by Vulkan (or some of our devices).  You should be using 4K maximum for guaranteed support, or checking the limits and selecting the best-fit power-of-two size perhaps.

gary-sweet (commented):
Just a couple of comments after a brief test.  Unfortunately on some of our devices, this breaks some of the limits and runs very slowly.

samples/performance/async_compute/async_compute.cpp
HansKristian-Work: I hit vsync on the devices I test on unless I really push it. The intention is to be able to see a difference when toggling the options, not run fast on all devices. (The sample is tagged Arm as well, FWIW).

samples/performance/async_compute/async_compute.cpp
HansKristian-Work: Fixed.

samples/performance/async_compute/async_compute.cpp
SaschaWillems: Samples based on the ARM framework are always locked to vsync. We discussed this on the last call and hopefully have fix for that soon ;)

samples/performance/async_compute/async_compute.cpp
```diff
@@ -0,0 +1,820 @@
...
+	                                     {VK_ATTACHMENT_LOAD_OP_CLEAR, VK_ATTACHMENT_STORE_OP_DONT_CARE}});
+
+	set_render_pipeline(std::move(blit_render_pipeline));
+
+	stats->request_stats({
+	    vkb::StatIndex::frame_times,
+	    vkb::StatIndex::gpu_cycles,
+	    vkb::StatIndex::gpu_vertex_cycles,
+	    vkb::StatIndex::gpu_fragment_cycles,
+	});
```
gary-sweet: I think you need to add "vkb::CounterSamplingConfig{vkb::CounterSamplingMode::Continuous}" as a second argument to request_stats here. That will prevent the Vulkan extension for performance counters being used.

The problem with the Vulkan counters extension is that it is designed to operate on a single command buffer (it's a query pool thing). This sample uses at least four different command buffers, each of which is starting and stopping stats sampling, and the Vulkan counters then just don't work.

You can then remove all the begin_sampling() and end_sampling() calls.

samples/performance/async_compute/async_compute.cpp
```diff
@@ -0,0 +1,820 @@
...
+
+	return true;
+}
+
+void AsyncComputeSample::render_shadow_pass()
+{
+	auto &queue          = *early_graphics_queue;
+	auto &command_buffer = render_context->get_active_frame().request_command_buffer(queue);
+	command_buffer.begin(VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT);
+	stats->begin_sampling(command_buffer);
```
gary-sweet: Here for example

samples/performance/async_compute/async_compute.cpp
gary-sweet: Looks good, thanks.

samples/performance/async_compute/async_compute.cpp
gary-sweet: > The sample is tagged Arm as well, FWIW

I'm not sure why it's tagged Arm. What here is Arm specific?

samples/performance/async_compute/async_compute.cpp
HansKristian-Work: Ye, I guess I can fix that.

samples/performance/async_compute/async_compute.cpp
HansKristian-Work: It was tagged like that since it's designed to demonstrate performance differences on Mali. It's somewhat implementation specific how this works (how many queues do you have, can queues pre-empt, is it TBDR, etc).

gary-sweet: Sure, but that's true of many of the performance samples, which aren't tagged as Arm specific.
My understanding was that the tag was for things that really are vendor specific.

samples/performance/async_compute/async_compute.cpp
HansKristian-Work: Continous seems to be broken. The output is very jumpy and weird, and I don't know why. I removed begin/end_sampling which should work. I cannot reproduce any issues here. I just pushed a commit which removes the begin/end_sampling calls.

gary-sweet: The pipeline_barriers sample uses continuous sampling. Does that work correctly?
We don't support continuous sampling here, someone at Arm would need to look into that.

gary-sweet: > I removed begin/end_sampling which should work.

It does, but you'll also need to remove the stats that are listed. Right now, they appear in the UI but are always 0, since nothing is updating them any more.  This is why continuous sampling mode would be better.  The counters won't display when continuous mode isn't supported then, but would still work when they can.

samples/performance/async_compute/async_compute.cpp
HansKristian-Work: Fixed.

samples/performance/async_compute/async_compute.cpp
HansKristian-Work: Found some bugs in the Continuous implementation. Fixed those, and flipped the sample over to continuous.

gary-sweet: Thanks a lot. That looks much better.

samples/performance/async_compute/async_compute.cpp
gary-sweet: Thanks.

samples/performance/async_compute/async_compute.cpp
```diff
@@ -0,0 +1,815 @@
...
+void AsyncComputeSample::update(float delta_time)
+{
+	if (last_async_enabled != async_enabled)
+	{
+		setup_queues();
+	}
+
+	// We can potentially get more overlap if we double buffer the HDR render target.
+	// In this scenario, the next frame can run ahead a little further before it needs to block.
+	if (double_buffer_hdr_frames)
```
gary-sweet, line 649: I think you need an 'else' here to reset forward_render_target_index to 0 when switching out of double buffer mode.
Double-buffer mode (for me) gives a 'lumpy' framerate (faster then slower every second or so). I noticed that when switching double-buffer mode off, the lumpiness continued.  Adding the 'else' case fixes the issue for me.

HansKristian-Work: Not exactly sure why, but it's somewhat cleaner.

gary-sweet (approved):
Thanks for the fixes.

gpx1000 (approved):

SaschaWillems (approved):
LGTM. Works fine for me on Win 10 + NV RTX 2060 and my Google Pixel 4a with an Adreno 618.

HansKristian-Work:
@marty-johnson59 Good to go.

PR #271, opened 2021-04-23
Add support for Xlib and Wayland WSI

caramelli:
This change adds support for Xlib and Wayland WSI.
Note that the glfw submodule has been updated for the GLFW_X11_XCB_VULKAN_SURFACE hint.

Nicolas Caramelli


third_party/CMakeLists.txt
```diff
@@ -57,11 +57,27 @@ elseif(WIN32)
...
+    find_package(PkgConfig QUIET)
+    if (VKB_WSI_SELECTION STREQUAL XCB)
+        pkg_check_modules(XCB xcb QUIET)
+        if (XCB_FOUND)
+            target_compile_definitions(vulkan INTERFACE VK_USE_PLATFORM_XCB_KHR)
+        endif()
+    endif()
+    if (VKB_WSI_SELECTION STREQUAL XLIB)
+        pkg_check_modules(X11 x11 QUIET)
+        if (X11_FOUND)
```
gary-sweet, line 71: Could we make direct-to-display part of VKB_WSI_SELECTION as well, rather than a fallback? Something like VKB_WSI_SELECTION=DIRECT would do it.

The code, as it stands, makes it difficult to switch from XCB, X11 or Wayland to direct-to-display, since the XCB_FOUND, X11_FOUND and WAYLAND_FOUND cache entries are still present.

This version worked for me, but you might want to add an error in the case that nothing is set in a final else case.
```
elseif(UNIX)
    # Choose WSI based on VKB_WSI_SELECTION
    find_package(PkgConfig QUIET)
    if (VKB_WSI_SELECTION STREQUAL XCB)
        pkg_check_modules(XCB xcb QUIET)
        if (XCB_FOUND)
            target_compile_definitions(vulkan INTERFACE VK_USE_PLATFORM_XCB_KHR)
        endif()
    elseif (VKB_WSI_SELECTION STREQUAL XLIB)
        pkg_check_modules(X11 x11 QUIET)
        if (X11_FOUND)
            target_compile_definitions(vulkan INTERFACE VK_USE_PLATFORM_XLIB_KHR)
        endif()
    elseif (VKB_WSI_SELECTION STREQUAL WAYLAND)
        pkg_check_modules(WAYLAND wayland-client QUIET)
        if (WAYLAND_FOUND)
            target_compile_definitions(vulkan INTERFACE VK_USE_PLATFORM_WAYLAND_KHR)
        endif()
    elseif (VKB_WSI_SELECTION STREQUAL DIRECT)
        set(DIRECT_TO_DISPLAY TRUE)
        set(DIRECT_TO_DISPLAY TRUE PARENT_SCOPE)
        target_compile_definitions(vulkan INTERFACE VK_USE_PLATFORM_DISPLAY_KHR)
    endif()
endif()
```

You'd need to add DIRECT to `set(VKB_WSI_SELECTION "XCB" CACHE STRING "Select WSI target (XCB, XLIB, WAYLAND)")` as well of course.

We also have some issues building for Wayland with this at the moment. Those are under investigation.

caramelli: Direct-to-Display was selected as a fallback if X11 was not available but that's true, this VKB_WSI_SELECTION option can now be used to select Direct-to-Display.
I updated the third_party/CMakeLists.txt as suggested (but using D2D instead of DIRECT)
If I'm not wrong, the VKB_DIRECT_2_DISPLAY option can be removed (it's strange to me that the definition of VK_USE_PLATFORM_DISPLAY_KHR was not set for the build when using VKB_DIRECT_2_DISPLAY ...)

gary-sweet: Thanks for this.

I'm now wondering if all the QUIET flags for pkg_check_modules should actually be REQUIRED instead. There is no longer a fallback, so if you request a specific WSI it needs to be supported.

gary-sweet (changes requested):
One of my colleagues is still investigating Wayland issues. We'll reply further when we have more information.

third_party/CMakeLists.txt
caramelli: Nice find! REQUIRED is now used instead of QUIET

tomek-brcm:
Hi Nicolas,

I've fixed 2 issues with cross-compilation environment where target system headers and binaries are different to the build system counterparts. I've sent you a merge request with those changes (https://github.com/caramelli/Vulkan-Samples/pull/1) so it's easy to see what I've done but it makes more sense to squash those into your commit.

caramelli:
Thanks for this merge request.

I understand the GLFW_INCLUDE_NONE (GL/gl.h missing on most embedded systems)

Regarding the target_include_directories(volk PUBLIC ${WAYLAND_INCLUDE_DIRS}), I guess this is due to the #include "vulkan_wayland.h" done in vulkan.h
So I think we should also add:
if (VKB_WSI_SELECTION STREQUAL XLIB)
    target_include_directories(volk PUBLIC ${X11_INCLUDE_DIRS})
endif()
if (VKB_WSI_SELECTION STREQUAL XCB)
    target_include_directories(volk PUBLIC ${XCB_INCLUDE_DIRS})
endif()

However, do we really need the pkg_check_modules(WAYLAND wayland-client QUIET) as we already called pkg_check_modules(WAYLAND wayland-client REQUIRED)?

tomek-brcm:
> Regarding the target_include_directories(volk PUBLIC ${WAYLAND_INCLUDE_DIRS}), I guess this is due to the #include "vulkan_wayland.h" done in vulkan.h

Exactly. The Wayland WSI pulls in wayland-client.h 

> So I think we should also add:
> if (VKB_WSI_SELECTION STREQUAL XLIB)
> target_include_directories(volk PUBLIC ${X11_INCLUDE_DIRS})
> endif()
> if (VKB_WSI_SELECTION STREQUAL XCB)
> target_include_directories(volk PUBLIC ${XCB_INCLUDE_DIRS})
> endif()

I didn't check other WSI choices but that looks about right. I'd go with elseif() though as this is a one off selection and that's how WSI is tested earlier in this file. 

> However, do we really need the pkg_check_modules(WAYLAND wayland-client QUIET) as we already called pkg_check_modules(WAYLAND wayland-client REQUIRED)?

No, we don't really have to repeat that.

caramelli:
Commit done. Hope you are okay with it.

tomek-brcm:
Yes, thanks! 
I closed my pull request now as both fixes are now merged into yours.

gary-sweet (dismissed):

gpx1000 (dismissed):

SaschaWillems:
At least one sample is using the ```VKB_DIRECT_2_DISPLAY``` option: https://github.com/KhronosGroup/Vulkan-Samples/blob/30e0ef953f9492726945d2042400a3808c8408f5/samples/extensions/open_gl_interop/CMakeLists.txt#L44
I think that needs to be changed to properly work with this PR.

caramelli:
I updated the commit to fix Vulkan-Samples/samples/extensions/open_gl_interop/CMakeLists.txt. Thanks!

gary-sweet (dismissed):

gpx1000 (dismissed):

TomAtkinsonArm (dismissed):
The changes here LGTM. This approval is for code review only, I have no machine available at this time to test these changes

marty-johnson59:
@caramelli, this looks ready to merge, but we need to resolve the conflicting file issue with options.h before Github lets me do that.  Can you help resolve?  Thanks

caramelli:
#undef Bool in options.h was required for Xlib to avoid a conflict with the Bool type definition in docopt_value.h
Since docopt has been removed, this change in options.h is therefore no longer necessary.

marty-johnson59:
OK, still not letting me merge w/o resolving the conflict.  Is this something I can do (sorry, I'm actually not very fluent in Github yet..)

PR #274, opened 2021-05-17
mlefrancois/rt_reflection

mklefrancois:
## Description

Render a simple scene showing the basics of ray tracing, including reflection and shadow rays. The sample creates some geometries and create a bottom acceleration structure for each, then make instances of those, using different materials and placing them at different locations.

## Test Environment

* Windows 10
* GeForce RTX 3060 Ti
* Driver: 466.27
* Vulkan: 1.2.176.1

## General Checklist:

## Sample Checklist


SaschaWillems (dismissed):
Awesome sample. Thank you very much for contributing this :)

Runs fine on my RTX 2060.

CI is currently failing due to the copyright check. Judging by the CI log you need to update the copyright to 2021 in the `CMakeLists.txt` of your sample to get all CI stages working.

samples/extensions/ray_tracing_reflection/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2019-2020, Sascha Willems
```
gpx1000: FWIW; is the attribution correct?  The date likely at least needs to be updated.

gpx1000 (dismissed):
Thanks for the high quality pull request!  This looks like a great sample of high quality.

samples/extensions/ray_tracing_reflection/CMakeLists.txt
mklefrancois: Thanks, I fixed it.

SaschaWillems (approved):

gpx1000 (dismissed):

SaschaWillems (dismissed):

shaders/ray_tracing_reflection/closesthit.rchit
```diff
@@ -0,0 +1,145 @@
...
+	float specular                  = kEnergyConservation * pow(max(dot(V, R), 0.0), kShininess);
+
+	return vec3(mat.specular * specular);
+}
+
+void main()
+{
+	int objId = gl_InstanceCustomIndexEXT;
+
+	int mat_idx = mat_index[objId].i[gl_PrimitiveID];
```
cforfang: Might seem like `objId` here and below (`materials[objId]`) should be surrounded by `nonuniformEXT`?

shaders/ray_tracing_reflection/closesthit.rchit
```diff
@@ -0,0 +1,145 @@
...
+	vec3 pos;
+	vec3 nrm;
+};
+
+// clang-format off
+layout(set = 0, binding = 0) uniform accelerationStructureEXT topLevelAS;
+layout(set = 0, binding = 3, scalar) buffer _materials { WaveFrontMaterial m[]; } materials[];
+layout(set = 0, binding = 4, scalar) buffer _vertices { Vertex v[]; } vertices[];
+layout(set = 0, binding = 5)         buffer _indices { uint i[]; } indices[];
+layout(set = 0, binding = 6)         buffer _mat_index { int i[]; } mat_index[];
```
cforfang: I believe the above bindings would be OpTypeRuntimeArray, as they are unsized. Ref. VUID-StandaloneSpirv-OpTypeRuntimeArray-04680, I don't belive that is allowed except for the _last_ binding in a set.

In practice, as the bindings are not unsized at API level (the layouts use `obj_models.size()`), that's probably why this works fine in practice. But as far as I can tell this is not valid usage.

cforfang: Hm, I'm misreading this case somewhat. What I'm basically looking for confirmation for is that it's OK/not OK to have the descriptor counts be unsized in the SPV, _as long as_ they are sized at API level (which they are in this case) -- or similar.

What I have in mind is that _if_ they are unsized at _API level_ (at layout creation time), the binding must be _last_ in the set -- not the case here -- but this case is _not_ that; this is just dropping the array size in the SPV (but still providing them at API level). So far I can't see anything saying this is not valid usage -- so presumably it's OK? :)

cforfang: Right, so reading up a bit, the key might appear to be that this was invalid prior to SPV_EXT_descriptor_indexing and co, which adds (quoting SPV_EXT_descriptor_indexing):

> The RuntimeDescriptorArrayEXT capability is used to indicate that a module uses arrays of resources which are declared with OpTypeRuntimeArray.

... and (from VK_EXT_descriptor_indexing -- emphasis mine):

> The final binding in a descriptor set layout can have a variable size (**and unsized arrays of resources are allowed** in the GL_EXT_nonuniform_qualifier and SPV_EXT_descriptor_indexing extensions).

This seems to allow what is needed. Note that VK_DESCRIPTOR_BINDING_VARIABLE_DESCRIPTOR_COUNT_BIT is _not_ needed, as the size is specified at layout creation time, not descriptor set allocation time.

I will say, it's not totally clear to me what language invalidates VUID-StandaloneSpirv-OpTypeRuntimeArray-04680 as seen in the 1.2 spec:
> OpTypeRuntimeArray **must only** be used for the last member of an OpTypeStruct that is in the StorageBuffer or PhysicalStorageBuffer storage class decorated as Block, or that is in the Uniform storage class decorated as BufferBlock

So '_must only be used for \<blah\>_' -- where the use here is _not_ \<blah\>, making it sounds like the code here goes against this particular VU -- but then the 1.2 spec _also_ says:
> OpTypeRuntimeArray **must only** be used for an array of variables with storage class Uniform, StorageBuffer, or UniformConstant, or for the outermost dimension of an array of arrays of such variables if the runtimeDescriptorArray feature is enabled,

Maybe I'm missing something, but these two statements seems to not agree? Probably I'm either misreading this ,or the VU is out of date. Either way, by the latter the use case here seems OK (assuming runtimeDescriptorArray  is enabled of course -- which it explicitly is in the code here)

shaders/ray_tracing_reflection/closesthit.rchit
mklefrancois: Yes, sorry I missed this while translating.

SaschaWillems (dismissed):

shaders/ray_tracing_reflection/closesthit.rchit
mklefrancois: Thanks for looking more deeply into this. I have double check with a colleague (Jeff Bolz), co-author of the extension, and he find no issue with the current usage. Maybe the specification could be made more clear. 

Another way to achieve the same thing, would be to use [buffer_reference](https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GLSL_EXT_buffer_reference.txt), but it will introduce an extension that wasn't use in the existing samples.

gpx1000 (dismissed):
Runs well here lgtm;

SaschaWillems:
Works fine for me on an RTX 2060 on Windows 10. The readme is a great addition too 👍

SaschaWillems (approved):

gpx1000 (approved):

gpx1000:
tested and works here.  NVidia AMD Linux and Windows.

marty-johnson59:
Hi folks, I'd like to get our final approval on this one so we can release it (been stuck for a few weeks).  Please help review if able.  Thanks!

mcj

PR #290, opened 2021-06-18
Adds surface compatibility detection to gpu selection logic

### LINKED ISSUE #289: ubuntu 20.04 running hello_triangle fails with "Failed to find a create function for"
I built the samples project per the linux build guide. I have vulkan-sdk from the lunarg repo. Build did not show any errors.

```
$ ./build/linux/app/bin/Release/x86_64/vulkan_samples hello_triangle
[info] Logger initialized
[info] Window created
[error] [build/linux/app/main.cpp:73] Failed to find a create function for
```
(Yes, the last line ends with "for")

* [My vulkaninfo](https://pastebin.com/PNkhynqD)
* I have both clang and gcc installed. I'm not sure which one cmake picked.
* Xubuntu 20.04
* nvidia 1080 using proprietary driver 460.80
* radeon x570 using mesa (I think)

memory-thrasher:
## Description

My multi-gpu system would crash consistently because the default monitor and "first" gpu were not connected. This adds a check for surface compatibility to the gpu selection logic. The surface, which was already created, had to be passed in.

Fixes #289 

## Testing/checklist Notes:

I do not have a mac or android compiler handy. Builds on windows and ubuntu.

Two samples, synchronization_2 and fragment_shading_rate, did not work before for me or after this change, complaining of a missing extension.

I have no idea what the checklist item about adding regressions means. To me, regression is testing things that previously worked.

## General Checklist:

- [X ] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [X ] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [X ] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
- [ X] I have commented any added functions (in line with Doxygen)
- [X ] I have commented any code that could be hard to understand
- [X ] My changes do not add any new compiler warnings
- [X ] My changes do not add any new validation layer errors or warnings
- [X ] I have used existing framework/helper functions where possible
- [ see above ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [ see above ] My changes do not add any regressions
- [ see above ] I have tested every sample to ensure everything runs correctly
- [X ] This PR describes the scope and expected impact of the changes I am making

## Sample Checklist

N/A


framework/core/instance.cpp
```diff
@@ -415,7 +415,7 @@ void Instance::query_gpus()
 	}
 }
 
-PhysicalDevice &Instance::get_suitable_gpu()
+PhysicalDevice &Instance::get_any_gpu()
```
gpx1000: maybe better to name this get_first_gpu()

framework/platform/parser.h
```diff
@@ -217,7 +217,8 @@ class CommandParser
 		auto values = get_command_value(command);
 
 		Type type{};
-		assert(convert_type(values, &type) && "Failed to retrieve value. Type unsupported");
+		auto status = convert_type(values, &type);
+		assert(status && "Failed to retrieve value. Type unsupported");
```
gpx1000: We should probably wait to get the merge from #288 so we don't take it twice and cause a merge conflict.

samples/api/hello_triangle/hello_triangle.cpp
```diff
@@ -277,40 +277,42 @@ void HelloTriangle::init_device(Context &                        context,
...
+		std::vector<VkExtensionProperties> device_extensions(device_extension_count);
+		VK_CHECK(vkEnumerateDeviceExtensionProperties(context.gpu, nullptr, &device_extension_count, device_extensions.data()));
 
-	for (uint32_t i = 0; i < queue_family_count; i++)
-	{
-		VkBool32 supports_present;
-		vkGetPhysicalDeviceSurfaceSupportKHR(context.gpu, i, context.surface, &supports_present);
+		if (!validate_extensions(required_device_extensions, device_extensions))
+		{
+			throw std::runtime_error("Required device extensions are missing, will try without.");
```
gpx1000: I think this might need to be rethought a little bit.  As this code now no longer looks at only the first device, it probably makes sense to only throw if no device has all requested extensions in support.

samples/api/hello_triangle/hello_triangle.cpp
```diff
@@ -277,40 +277,42 @@ void HelloTriangle::init_device(Context &                        context,
...
-		if ((queue_family_properties[i].queueFlags & VK_QUEUE_GRAPHICS_BIT) && supports_present)
+		for (uint32_t i = 0; i < queue_family_count; i++)
 		{
-			context.graphics_queue_index = i;
-			break;
+			VkBool32 supports_present;
+			vkGetPhysicalDeviceSurfaceSupportKHR(context.gpu, i, context.surface, &supports_present);
+
+			// Find a queue family which supports graphics and presentation.
+			if ((queue_family_properties[i].queueFlags & VK_QUEUE_GRAPHICS_BIT) && supports_present)
```
gpx1000, line 301: This is certainly a good thing to check.  It might make sense to include checking for which device is the "best" out of candidate devices or show how to create more selection criteria.  One big area that likely would be ideal to check is ensure the device is a discrete vs in built if there's more than one as it's popular for laptops to have both.

gpx1000 (commented):
Thanks for the submission.  After we have passed the CI checks, I'll run the code locally and ensure it works.  This was just a top of head pass.

framework/core/instance.cpp
memory-thrasher: I agree.

framework/platform/parser.h
memory-thrasher: Yeah. I'll revert that file in my next commit.

samples/api/hello_triangle/hello_triangle.cpp
memory-thrasher: Fair. I do not want to make the hello app too complex, but I can definitely reorder this in a way that makes more sense.

samples/api/hello_triangle/hello_triangle.cpp
memory-thrasher: Yeah. I have heard that some laptops even require both gpus be put in play, the discrete does the renders and then the app has to copy the completed framebuffer to the cpu-embedded gpu for presentation. So that affects not only which gpu, but how many and the creation options for the framebuffers, and the existence of a staging swapchain as a copy_dst.

I also agree about needing better logic to find the "best" gpu. But the definition of best needs some discussion. I have been playing with the idea of a scoring system, one element of which would be based on the types of available queues (bonus points to any gpu with a queue family that supports render, compute, and present). As a user though, my definition of best would more likely depend on either the more powerful gpu, or the one that happens to be directly connected to the display that contains the target frame; and I am not sure how to detect either case. Thoughts?

samples/api/hello_triangle/hello_triangle.cpp
memory-thrasher: > Required device extensions are missing, will try without.

Assuming that is accurate, it does not make any sense to me to use this to influence the gpu selection for this specific sample, but still worth a check. Extension support might be something to consider for the framework gpu selection logic, though.

framework/platform/parser.h
memory-thrasher: On second thought, I hate leaving my user branch in a broken state, so yeah, just complete the other pr first.

SaschaWillems:
Just a general idea, as this is related to samples crashing because the first gpu in the system doesn't support required extension: Wouldn't it be a better option to select the suitable gpu by finding the first gpu that supports the extensions required by the sample?

memory-thrasher:
> Just a general idea, as this is related to samples crashing because the first gpu in the system doesn't support required extension: Wouldn't it be a better option to select the suitable gpu by finding the first gpu that supports the extensions required by the sample?

Good thought. I do think the surface extension should probably be one of the things that are checked. In my case, however, both gpus support the requisite extensions (or at least, claim to), but the first gpu did not have a queue that supported the surface because the surface was on a monitor not connected to that gpu.

TomAtkinsonArm:
The current code LGTM - I need to look at how we score GPUs / could score in the future, so I will do an official review pass later

CI enabled, you should now be able to see which compilers or quality checks fail :)

memory-thrasher:
Wow the windows build is picky on int type sizes. I am not seeing those up-cast errors at all, even as warnings. So now it has static_cast for the for loop.

gpx1000 (approved):
LGTM; works on the platforms I've tested.

Overall I think we should make a sample that demonstrates a couple of options for selecting the best card.  However, for now I think this is the right thing to do and well implemented here.

gpx1000 (approved):

TomAtkinsonArm (approved):

gary-sweet (approved):

SaschaWillems (approved):

PR #294, opened 2021-06-30
Introduce vulkan.hpp-based version of hello_triangle.

### LINKED ISSUE #304: Vulkan HPP Framework Discussion
Hey @asuessenbach, It would be great to discuss the Vulkan HPP Framework changes before the next call :)

### Overview

> Demonstrate more use cases with the Vulkan HPP API. This may require the recreation of another framework using Vulkan HPP instead of the C style API. A selection of API and performance samples would be ported to show its benefits ~ paraphrased from previous conversations

My biggest concern with duplicating the framework is that when larger parts are refactored in one of the frameworks then these will need to be reflected in the other; the same applies for samples. Replicating existing samples would mean that contributors will need to take more factors into account - there is already an ever increasing check list on PRs.

### Is this a good direction?

I'm worried that direct ports of samples will generate more work in the future without providing much value for the developers looking at these samples. For instance, the inner workings of hello triangle have not been changed. We are simply seeing the same code written in a different style. This is similar to to-do list examples across programming languages in the sense that the added value for a developer is learning dramatically different syntax. In this case, the C style API covers all the Vulkan syntax that a developers needs and the HPP is "syntactic sugar".

It may be a better use of time to find how the HPP API can be used in comparison to the C style API when tackling specific problems in Vulkan. Are there any scenarios where we can show one API greatly simplifying concepts for developers or reducing the size of code footprints for specific features?

One example of this could be showing the benefits of: not using a shared context, current framework; using a shared context, shared pointer to a struct of common handles and using the HPP style API to not require tracking lower level handles - device, instance etc... Or possibly rethinking resource management with tracking structs of handles vs tracking hpp vulkan objects... Can hello triangle be written in a more simple way using HPP...

### Summary

Overall, I think it is great to add more HPP. It just may be more beneficial to demonstrate its pros/cons through new samples rather than maintain two copies of the current codebase

asuessenbach:
## Description

Add a new sample `hpp_hello_triangle`, which is a clone of the `hello_triangle` sample illustrating the usage of the C++ bindings provided by vulkan.hpp.
In order to make that work, a change is needed to the pure virtual function `Window::create_surface`, now having a `VkInstance` instead of a `vkb::Instance &` as its only argument. `AndroidWindow`, `GlfwWindow` and `HeadlessWindow` are adjusted accordingly.
As a consequence, usage of `vkb::Instance` in the `hello_triangle` sample could be removed, which reduces its use of the framework even more.

I've built and tested the new sample on Windows.

## General Checklist:

## Sample Checklist

### DIFF
--- a/samples/api/hello_triangle/hello_triangle.cpp
+++ b/samples/api/hello_triangle/hello_triangle.cpp
@@ -1068,8 +1068,6 @@ void HelloTriangle::teardown(Context &context)
 		vkDestroyDebugReportCallbackEXT(context.instance, context.debug_callback, nullptr);
 		context.debug_callback = VK_NULL_HANDLE;
 	}
-
-	vk_instance.reset();
 }
 
 HelloTriangle::HelloTriangle()
@@ -1085,7 +1083,7 @@ bool HelloTriangle::prepare(vkb::Platform &platform)
 {
 	init_instance(context, {VK_KHR_SURFACE_EXTENSION_NAME}, {});
 
-	vk_instance = std::make_unique<vkb::Instance>(context.instance);
+	context.surface = platform.get_window().create_surface(context.instance);
 
 	context.surface = platform.get_window().create_surface(*vk_instance);
 	context.swapchain_dimensions.width  = platform.get_window().get_width(),
--- a/samples/api/hello_triangle/hello_triangle.cpp
+++ b/samples/api/hello_triangle/hello_triangle.cpp
@@ -1068,6 +1068,8 @@ void HelloTriangle::teardown(Context &context)
 		vkDestroyDebugReportCallbackEXT(context.instance, context.debug_callback, nullptr);
 		context.debug_callback = VK_NULL_HANDLE;
 	}
+
+	vk_instance.reset();
 }
 
 HelloTriangle::HelloTriangle()
@@ -1083,7 +1085,9 @@ bool HelloTriangle::prepare(vkb::Platform &platform)
 {
 	init_instance(context, {VK_KHR_SURFACE_EXTENSION_NAME}, {});
 
-	context.surface = platform.get_window().create_surface(context.instance);
+	vk_instance = std::make_unique<vkb::Instance>(context.instance);
+
+	context.surface = platform.get_window().create_surface(*vk_instance);
 
 	context.surface = platform.get_window().create_surface(*vk_instance);
 	context.swapchain_dimensions.width  = platform.get_window().get_width(),
--- a/samples/api/hello_triangle/hello_triangle.cpp
+++ b/samples/api/hello_triangle/hello_triangle.cpp
@@ -959,7 +959,7 @@ VkResult HelloTriangle::present_image(Context &context, uint32_t index)
 }
 
 /**
- * @brief Initializes the Vulkan frambuffers.
+ * @brief Initializes the Vulkan framebuffers.
  * @param context A Vulkan context with the render pass already set up.
  */
 void HelloTriangle::init_framebuffers(Context &context)
--- a/samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
+++ b/samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
@@ -76,30 +76,6 @@ bool HPPHelloTriangle::validate_extensions(const std::vector<const char *> &    
 }
 
 /**
- * @brief Validates a list of required layers, comparing it with the available ones.
- *
- * @param required A vector containing required layer names.
- * @param available A vk::LayerProperties object containing available layers.
- * @return true if all required extensions are available
- * @return false otherwise
- */
-bool HPPHelloTriangle::validate_layers(const std::vector<const char *> &       required,
-                                       const std::vector<vk::LayerProperties> &available)
-{
-	// inner find_if returns true if the layer was not found
-	// outer find_if returns true if none of the layers were not found, that is if all layers were found
-	return std::find_if(required.begin(),
-	                    required.end(),
-	                    [&available](auto layer) {
-		                    return std::find_if(available.begin(),
-		                                        available.end(),
-		                                        [&layer](auto const &lp) {
-			                                        return strcmp(lp.layerName, layer) == 0;
-		                                        }) == available.end();
-	                    }) == required.end();
-}
-
-/**
  * @brief Find the vulkan shader stage for a given a string.
  *
  * @param ext A string containing the shader stage name.
@@ -226,6 +202,10 @@ void HPPHelloTriangle::init_instance(Context &                        context,
 	active_instance_extensions.push_back(VK_MVK_MACOS_SURFACE_EXTENSION_NAME);
 #elif defined(VK_USE_PLATFORM_XCB_KHR)
 	active_instance_extensions.push_back(VK_KHR_XCB_SURFACE_EXTENSION_NAME);
+#elif defined(VK_USE_PLATFORM_XLIB_KHR)
+	active_instance_extensions.push_back(VK_KHR_XLIB_SURFACE_EXTENSION_NAME);
+#elif defined(VK_USE_PLATFORM_WAYLAND_KHR)
+	active_instance_extensions.push_back(VK_KHR_WAYLAND_SURFACE_EXTENSION_NAME);
 #elif defined(VK_USE_PLATFORM_DISPLAY_KHR)
 	active_instance_extensions.push_back(VK_KHR_DISPLAY_EXTENSION_NAME);
 #else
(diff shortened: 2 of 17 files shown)


samples/api/hello_triangle/hello_triangle.cpp
```diff
@@ -1070,9 +1068,7 @@ bool HelloTriangle::prepare(vkb::Platform &platform)
 {
 	init_instance(context, {VK_KHR_SURFACE_EXTENSION_NAME}, {});
 
-	vk_instance = std::make_unique<vkb::Instance>(context.instance);
-
-	context.surface = platform.get_window().create_surface(*vk_instance);
+	context.surface = platform.get_window().create_surface(context.instance);
```
TomAtkinsonArm: Nice, makes sense

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
```diff
@@ -0,0 +1,1004 @@
...
+ */
+vk::Result HPPHelloTriangle::present_image(Context &context, uint32_t index)
+{
+	vk::PresentInfoKHR present(context.per_frame[index].swapchain_release_semaphore, context.swapchain, index);
+	// Present swapchain image
+	return context.queue.presentKHR(present);
+}
+
+/**
+ * @brief Initializes the Vulkan frambuffers.
```
TomAtkinsonArm: typo: framebuffers

TomAtkinsonArm: Same on line 947 `hello_triangle.cpp`

TomAtkinsonArm (commented):
Port from `hello_triangle` LGTM

Experiencing a crash on S21 when retrieving VkSurface. Not sure why this is caused yet

TomAtkinsonArm:
> Experiencing a crash on S21 when retrieving VkSurface. Not sure why this is caused yet

Works now, error was due to own fault :)

gary-sweet (changes requested):
I'm seeing build errors for d2d platforms:

In file included from framework/platform/unix/unix_d2d_platform.cpp:28:
framework/platform/unix/direct_window.h:39:23: error: 'virtual VkSurfaceKHR_T* vkb::DirectWindow::create_surface(vkb::Instance&)' marked 'override', but does not override
  virtual VkSurfaceKHR create_surface(Instance &instance) override;
                       ^~~~~~~~~~~~~~
In file included from framework/platform/unix/direct_window.cpp:18:
framework/platform/unix/direct_window.h:39:23: error: 'virtual VkSurfaceKHR_T* vkb::DirectWindow::create_surface(vkb::Instance&)' marked 'override', but does not override
  virtual VkSurfaceKHR create_surface(Instance &instance) override;

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
asuessenbach: Resolved with https://github.com/KhronosGroup/Vulkan-Samples/pull/294/commits/1e497ba1a21aaf4e7c9b63ccbe9e7ce93915830d

asuessenbach:
The platform/unix override of `create_surface` is the only one that really uses the `vkb::Instance` argument, and not just the `VkInstance` from it (via `get_handle()`): the function `get_suitable_gpu` is called here.
One could either
1. replicate that functionality locally, or
2. extent the signature of `create_surface` to get a `VkPhysicalDevice` as a second argument, or
3. revert the changes on `create_surface` in this PR, and add some HPP-version of `vkb::Instance`,`vbk::PhysicalDevice`, as well as some HPP-version of every flavor of `vkb::Window` instead

While approach 1. would be very local, some code would be replicated and thus would exist redundantly (even though, it's very little code). One could maybe introduce `get_suitable_gpu` as a helper function, working on a `VkInstance`, instead.

Approach 2. would be pretty small, but would require some work on the other platforms where it is not needed. Moreover, the hello_triangle sample would either need to keep the usage of vkb::Instance or use its own version of `get_suitable_gpu`.

Finally, approach 3. would need some substantial work, replicating lots of code. But depending on the long-term goals of adding hpp-based samples, that work might be needed nevertheless.

So, any advice how to approach this?

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
```diff
@@ -0,0 +1,1004 @@
...
+		throw std::runtime_error("Required instance extensions are missing.");
+	}
+
+	std::vector<vk::LayerProperties> supported_validation_layers = vk::enumerateInstanceLayerProperties();
+
+	std::vector<const char *> requested_validation_layers(required_validation_layers);
+
+#ifdef VKB_VALIDATION_LAYERS
+	// Determine the optimal validation layers to enable that are necessary for useful debugging
+	std::vector<const char *> optimal_validation_layers = vkb::get_optimal_validation_layers(supported_validation_layers);
```
SaschaWillems: This doesn't compile for me when enabling validation layers. The `get_optimal_validation_layers` function takes a vector if `VkLayerProperties`, which is not compatible with the hpp's `vk::LayerProperties` array. 

I guess we need an additional overload for this to work.

SaschaWillems (changes requested):
Thanks for contributing this. Having up-to-date samples for Vulkan.hpp is a very welcome addition 👍 

Except for the compilation failure when enabling the validation layers, this looks good to me on Windows, but as with D2D, this fails for on Android.

SaschaWillems:
> I'm seeing build errors for d2d platforms:
> 
> In file included from framework/platform/unix/unix_d2d_platform.cpp:28:
> framework/platform/unix/direct_window.h:39:23: error: 'virtual VkSurfaceKHR_T* vkb::DirectWindow::create_surface(vkb::Instance&)' marked 'override', but does not override
> virtual VkSurfaceKHR create_surface(Instance &instance) override;
> ^~~~~~~~~~~~~~
> In file included from framework/platform/unix/direct_window.cpp:18:
> framework/platform/unix/direct_window.h:39:23: error: 'virtual VkSurfaceKHR_T* vkb::DirectWindow::create_surface(vkb::Instance&)' marked 'override', but does not override
> virtual VkSurfaceKHR create_surface(Instance &instance) override;

@gary-sweet : Maybe we can add a D2D target to the CI pipeline? This way we could easily catch build errors that only affect D2D builds.

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
asuessenbach: Resolved with https://github.com/KhronosGroup/Vulkan-Samples/pull/294/commits/3e62a9f5a339e539e4833c0594e0a8376991286d.

TomAtkinsonArm:
> Maybe we can add a D2D target to the CI pipeline? This way we could easily catch build errors that only affect D2D builds.

I can attempt this, I think last time I tried there where some dependencies that where not available for it to compile

gary-sweet:
> > Maybe we can add a D2D target to the CI pipeline? This way we could easily catch build errors that only affect D2D builds.
> 
> I can attempt this, I think last time I tried there where some dependencies that where not available for it to compile

I thought we already had this, but maybe it was only in the old gitlab version.

gary-sweet:
I'm now seeing:

framework/platform/unix/direct_window.cpp: In member function 'virtual vk::SurfaceKHR vkb::DirectWindow::create_surface(vk::Instance, vk::PhysicalDevice)':
framework/platform/unix/direct_window.cpp:373:40: error: 'mode_props' was not declared in this scope
  surface_create_info.displayMode     = mode_props.displayMode;
                                        ^~~~~~~~~~
framework/platform/unix/direct_window.cpp:373:40: note: suggested alternative: 'mode_t'
  surface_create_info.displayMode     = mode_props.displayMode;

gary-sweet:
I think you should be able to do a D2D build by forcing VKB_WSI_SELECTION STREQUAL to "D2D" in CMake (on Linux at least).

asuessenbach:
@gary-sweet Should be fixed now.

gary-sweet:
> @gary-sweet Should be fixed now.

Indeed it is - thanks.

I am seeing one slight oddity now that I can run it. When I terminate the app (using the ESC key for example), I see:
```
terminate called without an active exception
Aborted
```

I've run a bunch of the other samples and none of them exhibit this behaviour on shutdown.

asuessenbach:
@gary-sweet Strange thing!
Can you spot where that terminate comes from? Maybe some part of the DirectWindow that would need a transcoded version as well?

gary-sweet:
> @gary-sweet Strange thing!
> Can you spot where that terminate comes from? Maybe some part of the DirectWindow that would need a transcoded version as well?

All of the work in main() has completed before the message is printed. It appears when main() returns. The debugger isn't much help either, it spots the exception but the stack-trace says "[Unknown/Just-In-Time compiled code]"

The odd thing is that this is the only sample that has this behaviour - all the others seem ok, and they all use D2D on my platform.

asuessenbach:
@gary-sweet  Could you please give it another try?
I somehow missed to destroy the instance... and nobody complained.

gary-sweet:
> @gary-sweet Could you please give it another try?
> I somehow missed to destroy the instance... and nobody complained.

Yep. That fixes it. Thanks

gary-sweet (approved):

SaschaWillems (approved):
LGTM 👍

gpx1000 (approved):
LGTM;

TomAtkinsonArm (approved):

marty-johnson59:
Merging now that we have 4 approvals.  LMK if we need to revert.

mcj

PR #313, opened 2021-08-04
Ray Tracing Extended

gpx1000:
## Description

Render a simple scene showing the basics of ray tracing, including Ambient Occlusion, top and bottom level acceleration structure rebuilding for animation, multiple objects in a scene, and refraction. The sample loads the sponza gltf model, loads the base fire particle texture and creates geometry for one particle and one quad showing heat above fire then creates a bottom acceleration structure for each geometry type, then makes instances of those static geometries and multiple instances of the fire particle to achieve a particle effect.  It then demonstrates TLAS (Top Level Acceleration Structure) animation via updating the fire particle(s) and BLAS (Bottom Level Acceleration Structure) animation via updating the quad geometry.

## Tested Environments

- Windows 10 / Ubuntu Linux
- GeForce 3060 / GeForce 2080 / GeForce 1070
- Vulkan version: 1.2.176

## Sample Assets changes
depends upon flame texture found in https://github.com/KhronosGroup/Vulkan-Samples-Assets/pull/12

## General Checklist:

## Sample Checklist


shaders/khr_ray_tracing_extended/closesthit.rchit
```diff
@@ -0,0 +1,168 @@
...
+    if (render_mode == 4) { // global xyz
+      hitValue.color = vec4(heatmap(worldPt.x, -10, 10), 1);
+      return;
+    }
+    if ((objectType == 0 || objectType == 2)){
+      if (imageOffset >= 26){
+        return; // this shouldn't happen
+      }
+      // obtain texture coordinate
+      vec4 tex_value = texture(textures[imageOffset], texcoord);
```
cforfang: Should `imageOffset` be wrapped by `nonuniformEXT` here? Also, are derivatives valid -- or should this be using `textureLod`?

gpx1000: derivatives I believe are valid, or I at least have no reason to think otherwise.  This isn't using mipmaps for the main reason that I didn't want to make the sample too complicated.  Meaning if I add mipmaps, then why not add don't update things not in view?  or update the particle effect TLAS to use less particles if the LOD is far away etc.

I agree it should be nonuniformEXT.  I'll submit another commit with it corrected, thanks!

cforfang: Cheers :) If there's no mips, I would maybe argue using `textureLod(..., 0)` would be better -- as there's a risk people will copy this code as it _is_ sample code. I somewhat doubt derives are valid TBH, so my guess is this likely only works properly as long as there's only 1 mip.

gary-sweet (commented):
Builds ok for me, and runs with correct warnings about being unsupported.
I can't actually test the raytracing code though.

shaders/khr_ray_tracing_extended/closesthit.rchit
gpx1000: That's actually a very good point.  I'll switch to textureLod and leave a comment that spells out reasoning.

Thanks!

SaschaWillems:
Great looking sample. Runs fine on my RTX 2060.

One thing I noticed, is that the heat haze effect doesn't look correct when viewed from different angles. E.g. if you watch the fire from the other side, it looks like this:

image

It would also be great of alpha masking was applied on alpha masked materials:

image

Afair this would require use of an additional any hit shader.

gpx1000:
Thanks @SaschaWillems 

Good observations on refraction.  It kinda looks like the image is getting flipped on the screen shot.

For the alpha, I thought about the benefits of adding it, but worry about making the sample too complex to serve the purpose of highlighting AO, animation, and shadows.  I think we also need to update the textures the sponza model is using as I don't think it currently has the alpha mask; looking at the ktx files it seems none of them have the alpha baked in either.

SaschaWillems:
Refraction now looks fine from all angles.

I'm also more than fine with not having alpha, esp. with the model needing to be updated.

So this looks good to me now :)

samples/README.md
```diff
@@ -185,6 +185,14 @@ Uses the debug utilities extension to name and group Vulkan objects (command buf
...
 Render a basic scene using the official cross-vendor ray tracing extension. Shows how to setup all data structures required for ray tracing, including the bottom and top level acceleration structures for the geometry, the shader binding table and the ray tracing pipelines with shader groups for ray generation, ray hits, and ray misses. After dispatching the rays, the final result is copied to the swapchain image.<br/>
 
+### [Extended hardware accelerated ray tracing](./extensions/raytracing_extended)<br/>
+**Extensions**: [```VK_KHR_ray_tracing_pipeline```](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_ray_tracing_pipeline), [```VK_KHR_acceleration_structure```](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_acceleration_structure)<br/>
+Render Sponza with Ambient Oclusion.  Place a vase in center.  Generate a particle fire that
```
SaschaWillems: Small typo. Should read "occlusion".

samples/README.md
```diff
@@ -185,6 +185,14 @@ Uses the debug utilities extension to name and group Vulkan objects (command buf
...
 Render a basic scene using the official cross-vendor ray tracing extension. Shows how to setup all data structures required for ray tracing, including the bottom and top level acceleration structures for the geometry, the shader binding table and the ray tracing pipelines with shader groups for ray generation, ray hits, and ray misses. After dispatching the rays, the final result is copied to the swapchain image.<br/>
 
+### [Extended hardware accelerated ray tracing](./extensions/raytracing_extended)<br/>
+**Extensions**: [```VK_KHR_ray_tracing_pipeline```](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_ray_tracing_pipeline), [```VK_KHR_acceleration_structure```](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_acceleration_structure)<br/>
+Render Sponza with Ambient Oclusion.  Place a vase in center.  Generate a particle fire that 
+demonstrates the TLAS (Top Level Acceleration Structure) animation for the same underlining geometry.
```
SaschaWillems: Not a native speaker, but I think this should be "underlying" instead of "underlining"?

samples/extensions/raytracing_extended/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+get_filename_component(FOLDER_NAME ${CMAKE_CURRENT_LIST_DIR} NAME)
+get_filename_component(PARENT_DIR ${CMAKE_CURRENT_LIST_DIR} PATH)
+get_filename_component(CATEGORY_NAME ${PARENT_DIR} NAME)
+
+add_sample_with_tags(
+    ID ${FOLDER_NAME}
+    CATEGORY ${CATEGORY_NAME}
+    AUTHOR "Holochip Corporation"
+    NAME "Ray tracing extended"
+    DESCRIPTION "Extended example of Ray Tracing highlighting the Bottom and Top Acceleration Structure rebuild and AO.")
```
SaschaWillems: Bottom and Top **level** Acceleration Structure

samples/README.md
gpx1000: You're correct :)  it's updated now.  Good catch.

SaschaWillems (dismissed):

samples/extensions/raytracing_extended/raytracing_extended.cpp
```diff
@@ -0,0 +1,1450 @@
...
+{
+	return (value + alignment - 1) & ~(alignment - 1);
+}
+
+namespace
+{
+template <typename T>
+struct CopyBuffer
+{
+	std::vector<T> operator()(std::unordered_map<std::string, vkb::core::Buffer> &buffers, const char *bufferName)
```
JoseEmilio-ARM: style: snake_case for `bufferName` and `alreadyMapped`. Since this is also used by other samples (so far Ray Queries) maybe it could be moved to the framework, e.g. `common/utils.h`? Or would it be possible for them to use the `Buffer` class in `core/buffer.h`?

samples/extensions/raytracing_extended/raytracing_extended.cpp
```diff
@@ -0,0 +1,1450 @@
...
+	raytracing_pipeline_create_info.stageCount                   = static_cast<uint32_t>(shader_stages.size());
+	raytracing_pipeline_create_info.pStages                      = shader_stages.data();
+	raytracing_pipeline_create_info.groupCount                   = static_cast<uint32_t>(shader_groups.size());
+	raytracing_pipeline_create_info.pGroups                      = shader_groups.data();
+	raytracing_pipeline_create_info.maxPipelineRayRecursionDepth = 1;
+	raytracing_pipeline_create_info.layout                       = pipeline_layout;
+	VK_CHECK(vkCreateRayTracingPipelinesKHR(get_device().get_handle(), VK_NULL_HANDLE, VK_NULL_HANDLE, 1, &raytracing_pipeline_create_info, nullptr, &pipeline));
+}
+
+void RaytracingExtended::create_display_pipeline()
```
JoseEmilio-ARM: unused left-over?

samples/extensions/raytracing_extended/raytracing_extended.cpp
```diff
@@ -0,0 +1,1450 @@
...
+	submit_info.commandBufferCount = 1;
+	submit_info.pCommandBuffers    = &draw_cmd_buffers[current_buffer];
+	VK_CHECK(vkQueueSubmit(queue, 1, &submit_info, device->request_fence()));
+	device->get_fence_pool().wait();
+	ApiVulkanSample::submit_frame();
+}
+
+void RaytracingExtended::draw_gui()
+{
+	gui->show_options_window(
```
JoseEmilio-ARM: Currently when running the sample I cannot see the GUI, is that expected?

shaders/khr_ray_tracing_extended/raygen.rgen
```diff
@@ -0,0 +1,170 @@
...
+
+	vec4 origin = cam.viewInverse * vec4(0,0,0,1);
+	vec4 target = cam.projInverse * vec4(d.x, d.y, 1, 1) ;
+	vec4 direction = cam.viewInverse*vec4(normalize(target.xyz), 0) ;
+
+	float tmin = 0.001;
+	float tmax = 10000.0;
+
+	uint max_rays = maxRays;
+	max_rays = 100;
```
JoseEmilio-ARM: intended?

shaders/khr_ray_tracing_extended/raygen.rgen
```diff
@@ -0,0 +1,170 @@
...
+} cam;
+
+struct Payload
+{
+  vec4 color;
+  vec4 intersection; // {x, y, z, intersectionType}
+  vec4 normal; // {nx, ny, nz, distance}
+};
+
+vec3 heatmap(float value, float minValue, float maxValue)
```
JoseEmilio-ARM: not used in this shader

shaders/khr_ray_tracing_extended/raygen.rgen
```diff
@@ -0,0 +1,170 @@
...
+	// 0 = normal, 1 = shadow, 2 = AO
+	uint current_mode = 0;
+	float expectedDistance = -1;
+	for (uint i = 0; i < max_rays && current_mode < 100 && color.a < 0.95 && (color.r < 0.99 || color.b < 0.99 || color.g < 0.99); ++i)
+	{
+		traceRayEXT(topLevelAS, gl_RayFlagsOpaqueEXT, 0xff, 0, 0, 0, origin.xyz, tmin, direction.xyz, tmax, 0);
+		object_type = uint(hitValue.intersection.w);
+		const vec3 object_intersection_pt = hitValue.intersection.xyz;
+		const vec3 object_normal = hitValue.normal.xyz;
+		if (render_mode != 0)
```
JoseEmilio-ARM: would it possible to use defines that match enum values for `render_mode` and `object_type`, to improve readability? or maybe comments would be enough?

mklefrancois:
I tried it with a NVIDIA GeForce RTX 3060Ti and it works.
* Should there be a user interface to change the drawing mode? It does not show up in my test.
* Is it possible that there is a problem with the animation update? The buffers are updated with map/unmap, but I don't see how that protects against updating the buffer while drawing, unless I'm missing a wait/fence somewhere.
* Should the animation be created with `VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_UPDATE_BIT_KHR` to update it?
  * The BLAS update could then set both `srcAccelerationStructure` and `dstAccelerationStructure` to the same acceleration structure when it is recreated.
* Wouldn't it be a good idea to have the animation executed with a compute shader?

samples/extensions/raytracing_extended/raytracing_extended.cpp
gpx1000: until GUI's work in the framework, we took out the GUI here to resolve this.  However, will add it back when #334 is merged in.  Can do a separate PR for that addition.

gpx1000:
> I tried it with a NVIDIA GeForce RTX 3060Ti and it works.
> 
> * Should there be a user interface to change the drawing mode? It does not show up in my test.
> * Is it possible that there is a problem with the animation update? The buffers are updated with map/unmap, but I don't see how that protects against updating the buffer while drawing, unless I'm missing a wait/fence somewhere.
> * Should the animation be created with `VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_UPDATE_BIT_KHR` to update it?
>   
>   * The BLAS update could then set both `srcAccelerationStructure` and `dstAccelerationStructure` to the same acceleration structure when it is recreated.
> * Wouldn't it be a good idea to have the animation executed with a compute shader?

Thanks for reviewing @mklefrancois. 
* Until the framework supports GUI use in Ray Tracing pipelines #334 addresses that shortcoming.  We've removed the GUI all together.
* fixed
* That's not a bad idea.  Using the UPDATE instead of the BUILD to update the AS instead of rebuild it whole sale.  As I'm using this as the first ray tracing sample to use the framework's acceleration structure I'm going to opt for doing this in a future PR.  The framework will take a bit more thought to handle updates vs rebuilds. *** update: I just went ahead and fixed it.
* While I agree it's a good optimization to do all animation in a compute shader, I worry that it would add complexity to the sample in an area that doesn't highlight the general case of how to do animation with ray tracing pipelines.  Thus, while it's an excellent idea to do animation in a compute shader, it should probably be beyond the scope.

samples/extensions/raytracing_extended/raytracing_extended.cpp
gpx1000: GUI is removed until #334 is merged.

SaschaWillems (dismissed):

samples/extensions/raytracing_extended/raytracing_extended.h
```diff
@@ -21,7 +21,7 @@
 
 #pragma once
 
-#define USE_FRAMEWORK_ACCELERATION_STRUCTURE
+//#define USE_FRAMEWORK_ACCELERATION_STRUCTURE
```
JoseEmilio-ARM: why make it optional?

gpx1000: It's optional because it's the first sample to use the framework for acceleration structure.  I felt it's good in the first sample to show an equivalent how to build with and without the framework.  This can be both instructive and a nice validation of expectations.

mklefrancois (dismissed):
Tested once again and works fine

SaschaWillems (dismissed):

mklefrancois (dismissed):
Tested and works

marty-johnson59:
@aurolou, can we get your help to review this RT sample as well as PR #317?  These are ray tracing samples and we'd really like to make sure they support AMD before releasing them.  Thanks!

aurolou:
This sample throws me an OUT_OF_POOL_MEMORY error when allocating descriptor sets in raytracing_extended.cpp, line 884. The Pool Size struct is missing some entries. If fixed, the sample works fine, tested on a RX 6800 XT.

framework/core/acceleration_structure.h
```diff
@@ -56,16 +56,35 @@ class AccelerationStructure
...
+	                              VkDeviceSize                        vertex_stride,
+	                              uint32_t                            transform_offset              = 0,
+	                              VkFormat                            vertex_format                 = VK_FORMAT_R32G32B32_SFLOAT,
+	                              VkGeometryFlagsKHR                  flags                         = VK_GEOMETRY_OPAQUE_BIT_KHR,
+	                              uint64_t                            vertex_buffer_data_address    = 0,
+	                              uint64_t                            index_buffer_data_address     = 0,
+	                              uint64_t                            transform_buffer_data_address = 0);
 
 	/**
 	 * @brief Adds instance geometry to the acceleration structure (only valid for top level)
```
JoseEmilio-ARM, line 90: missing return documentation

JoseEmilio-ARM (dismissed):

JoseEmilio-ARM (approved):

SaschaWillems (approved):

PR #246, opened 2021-01-30
Add ray tracing related helper classes and functions to the framework

SaschaWillems:
## Description

This PR adds a new classes and minor changes to the framework to streamline development ray tracing samples. The aim of these changes is to reduce duplication of boiler plate code, as much of the ray tracing related functionality can be become quite verbose. For a basic ray tracing example, this saves around 300 lines of code.

New classes are added for:
- Acceleration structures
- Scratch buffers
- Shader binding tables

The scratch buffer and shader binding table classes are mostly just extended buffers with most of their properties like usage flags being preset in the classes to reduce boiler plate.

The acceleration structure class also hides away most of the boiler plate code and adds methods that simplify geometry setup and the build process.

Example:

```cpp
// Bottom level AS with multiple triangle geometries
bottom_level_acceleration_structure = std::make_unique<vkb::core::AccelerationStructure>(get_device(), VK_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL_KHR);
bottom_level_acceleration_structure->add_triangle_geometry(vertex_buffer, index_buffer, transform_data_buffer, 1, 2, sizeof(Vertex), 0);
bottom_level_acceleration_structure->add_triangle_geometry(vertex_buffer, index_buffer, transform_data_buffer, 1, 2, sizeof(Vertex), sizeof(VkTransformMatrixKHR));
bottom_level_acceleration_structure->add_triangle_geometry(vertex_buffer, index_buffer, transform_data_buffer, 1, 2, sizeof(Vertex), sizeof(VkTransformMatrixKHR) * 2);
bottom_level_acceleration_structure->build(queue);

// Top Level AS with single instance
top_level_acceleration_structure = std::make_unique<vkb::core::AccelerationStructure>(get_device(), VK_ACCELERATION_STRUCTURE_TYPE_TOP_LEVEL_KHR);
top_level_acceleration_structure->add_instance_geometry(instances_buffer, 1);
top_level_acceleration_structure->build(queue);
```

Changes to existing classes:
- Buffer

A function to retrieve the device address has been added to the buffer class. Ray tracing makes heavy use of device addresses, but device addresses are getting more and more common with newer Vulkan functionality, so this is more of a general addition. Note that this only works for buffers that have been created with the `VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT` usage flag.

Example:

```cpp
VkDeviceOrHostAddressConstKHR device_or_host_address{};
[...]

PR #317, opened 2021-08-07
Ray queries

gpx1000:
## Description

Render a simple scene showing the basics of ray queries, including dynamic shadows top and bottom level acceleration structure use with traditional vertex/fragment shaders. The sample loads the sponza gltf model, creates a bottom acceleration structure for geometry, then makes instances of the geometry. It then demonstrates traversing TLAS (Top Level Acceleration Structure) and BLAS (Bottom Level Acceleration Structure) from the fragment shader using ray queries to cast shadows from a moving light.

## Tested Environments
Windows 10 / Ubuntu Linux
GeForce 3060 / GeForce 2080 / GeForce 1070
Vulkan version: 1.2.176

## General Checklist:

## Sample Checklist


samples/README.md
```diff
@@ -181,6 +181,10 @@ Uses the debug utilities extension to name and group Vulkan objects (command buf
 
 - 🎓 [Using the debug utilities extension](./extensions/debug_utils/debug_utils_tutorial.md)
 
+### [Basic ray queries](./extensions/ray_queries)<br/>
+**Extensions**: [```VK_KHR_RAY_QUERY_EXTENSION_NAME```](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_ray_query), [```VK_KHR_acceleration_structure```](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_acceleration_structure) <br/>
```
JoseEmilio-ARM: We could use VK_KHR_ray_query here for consistency with other entries

samples/extensions/ray_queries/ray_queries.h
```diff
@@ -0,0 +1,109 @@
...
+
+	struct Model
+	{
+		std::vector<Vertex>                  vertices;
+		std::vector<std::array<uint32_t, 3>> indices;
+	} model;
+
+	struct AccelerationStructure
+	{
+		VkAccelerationStructureKHR         handle         = nullptr;
```
JoseEmilio-ARM: This causes the Android build to fail, fixed with:
```
--- a/samples/extensions/ray_queries/ray_queries.h
+++ b/samples/extensions/ray_queries/ray_queries.h
@@ -68,7 +68,7 @@ class RayQueries : public ApiVulkanSample

        struct AccelerationStructure
        {
-               VkAccelerationStructureKHR         handle         = nullptr;
+               VkAccelerationStructureKHR         handle         = VK_NULL_HANDLE;
                uint64_t                           device_address = 0;
                std::unique_ptr<vkb::core::Buffer> buffer         = nullptr;
        };
@@ -88,10 +88,10 @@ class RayQueries : public ApiVulkanSample
        void                                             create_top_level_acceleration_structure();
        void                                             create_bottom_level_acceleration_structure();

-       VkPipeline            pipeline{nullptr};
-       VkPipelineLayout      pipeline_layout{nullptr};
-       VkDescriptorSet       descriptor_set{nullptr};
-       VkDescriptorSetLayout descriptor_set_layout{nullptr};
+       VkPipeline            pipeline{VK_NULL_HANDLE};
+       VkPipelineLayout      pipeline_layout{VK_NULL_HANDLE};
+       VkDescriptorSet       descriptor_set{VK_NULL_HANDLE};
+       VkDescriptorSetLayout descriptor_set_layout{VK_NULL_HANDLE};

        void build_command_buffers() override;
        void create_uniforms();
```

shaders/ray_queries/ray_shadow.frag
```diff
@@ -0,0 +1,69 @@
...
+	const float distance  = sqrt(dot(direction, direction));
+
+	rayQueryEXT query;
+
+	// The following runs the actual ray query
+	// For performance, could use gl_RayFlagsTerminateOnFirstHitEXT, since we only need to know
+	// whether an intersection exists, and not necessarily any particular intersection
+	rayQueryInitializeEXT(query, topLevelAS, gl_RayFlagsTerminateOnFirstHitEXT, 0xFF, pos, tmin, direction.xyz, distance);
+	while (rayQueryProceedEXT(query))
+	{
```
JoseEmilio-ARM: Empty loops like this can be very expensive on some compilers, suggest using a single call to `rayQueryProceedEXT` here (no more are needed since we are using `gl_RayFlagsTerminateOnFirstHitEXT`)

samples/extensions/ray_queries/ray_queries.cpp
```diff
@@ -0,0 +1,713 @@
...
+		if (iter == buffers.cend())
+		{
+			return {};
+		}
+		auto &         buffer = iter->second;
+		std::vector<T> out;
+
+		const size_t sz = buffer.get_size();
+		out.resize(sz / sizeof(T));
+		const bool alreadyMapped = buffer.get_data() != nullptr;
```
JoseEmilio-ARM: style: snake_case for variable names

samples/extensions/ray_queries/ray_queries.h
```diff
@@ -0,0 +1,109 @@
...
+}        // namespace sg
+}        // namespace vkb
+
+class RayQueries : public ApiVulkanSample
+{
+  public:
+	RayQueries();
+	~RayQueries() override;
+	void request_gpu_features(vkb::PhysicalDevice &gpu) override;
+	void prepare_render_context() override;
```
JoseEmilio-ARM: no need to override `prepare_render_context`?

samples/extensions/ray_queries/ray_queries.cpp
```diff
@@ -0,0 +1,713 @@
...
+
+uint64_t RayQueries::get_buffer_device_address(VkBuffer buffer)
+{
+	VkBufferDeviceAddressInfoKHR buffer_device_address_info{};
+	buffer_device_address_info.sType  = VK_STRUCTURE_TYPE_BUFFER_DEVICE_ADDRESS_INFO;
+	buffer_device_address_info.buffer = buffer;
+	return vkGetBufferDeviceAddressKHR(device->get_handle(), &buffer_device_address_info);
+}
+
+void RayQueries::create_top_level_acceleration_structure()
```
JoseEmilio-ARM, line 231: would it be possible to re-use the helper classes introduced in https://github.com/KhronosGroup/Vulkan-Samples/pull/246 ? (@SaschaWillems would the other sample also need to be updated to use them?)

samples/extensions/ray_queries/ray_queries.cpp
```diff
@@ -0,0 +1,713 @@
...
+
+	VkAccelerationStructureBuildRangeInfoKHR acceleration_structure_build_range_info;
+	acceleration_structure_build_range_info.primitiveCount                                           = model.indices.size();
+	acceleration_structure_build_range_info.primitiveOffset                                          = 0;
+	acceleration_structure_build_range_info.firstVertex                                              = 0;
+	acceleration_structure_build_range_info.transformOffset                                          = 0;
+	std::vector<VkAccelerationStructureBuildRangeInfoKHR *> acceleration_build_structure_range_infos = {&acceleration_structure_build_range_info};
+
+	// Build the acceleration structure on the device via a one-time command buffer submission
+	// Some implementations may support acceleration structure building on the host (VkPhysicalDeviceAccelerationStructureFeaturesKHR->accelerationStructureHostCommands), but we prefer device builds
```
JoseEmilio-ARM: could we rephrase to `but we show device build here`?

shaders/ray_queries/ray_shadow.frag
```diff
@@ -0,0 +1,69 @@
...
+bool intersects_light(vec3 light_origin, vec3 pos)
+{
+	const float tmin = 0.01, tmax = 1000;
+	const vec3  direction = light_origin - pos;
+	const float distance  = sqrt(dot(direction, direction));
+
+	rayQueryEXT query;
+
+	// The following runs the actual ray query
+	// For performance, could use gl_RayFlagsTerminateOnFirstHitEXT, since we only need to know
```
JoseEmilio-ARM: can remove `could`?

samples/extensions/ray_queries/ray_queries.h
JoseEmilio-ARM: does not fail the CI though. There are other build warnings that also seem easy to fix: https://github.com/KhronosGroup/Vulkan-Samples/pull/317/checks?check_run_id=3270603265#step:5:366

samples/extensions/ray_queries/ray_queries.h
gpx1000: Ah yes, it does the same thing.  Makes sense.

samples/extensions/ray_queries/ray_queries.cpp
gpx1000: I didn't do this one yet, leaving it unresolved until I find time to research/implement using Sascha's helper class.

gpx1000: It looks like we don't have any code in any sample currently making use of @SaschaWillems helper class.  Sascha, please lemme know if you'd like me to help you make one or if you need this sample to be your guinea pig.

Unless Sascha asks for help, I'd like to respectfully allow Sascha to create the first sample using his class as it looks like he has planned and then in a future PR make use of it once any potential bugs are ironed out.  Just as an effort to not step on his toes ;-).

samples/extensions/ray_queries/ray_queries.cpp
JoseEmilio-ARM: thanks for addressing all of these so quickly! just noticed `bufferName` above too

samples/extensions/ray_queries/ray_queries.cpp
```diff
@@ -0,0 +1,708 @@
...
+
+		VkRect2D scissor = vkb::initializers::rect2D(static_cast<int32_t>(width), static_cast<int32_t>(height), 0, 0);
+		vkCmdSetScissor(draw_cmd_buffers[i], 0, 1, &scissor);
+
+		vkCmdBindPipeline(draw_cmd_buffers[i], VK_PIPELINE_BIND_POINT_GRAPHICS, pipeline);
+		vkCmdBindDescriptorSets(draw_cmd_buffers[i], VK_PIPELINE_BIND_POINT_GRAPHICS, pipeline_layout, 0, 1, &descriptor_set, 0, nullptr);
+		VkDeviceSize offsets[1] = {0};
+		vkCmdBindVertexBuffers(draw_cmd_buffers[i], 0, 1, vertex_buffer->get(), offsets);
+		vkCmdBindIndexBuffer(draw_cmd_buffers[i], index_buffer->get_handle(), 0, VK_INDEX_TYPE_UINT32);
+		vkCmdDrawIndexed(draw_cmd_buffers[i], static_cast<uint32_t>(model.indices.size()) * 3, 1, 0, 0, 0);
```
JoseEmilio-ARM, line 181: It seems that we are aggregating all geometry and dispatching a single draw call for the whole scene. Perhaps a more realistic use case could be to have a deferred rendering pipeline, to use the same draw calls to render both the scene and the shadows, since the ray queries can then be traced once per pixel during the lighting pass

shaders/ray_queries/ray_shadow.frag
```diff
@@ -0,0 +1,72 @@
...
+	const vec3  direction = light_origin - pos;
+	const float distance  = sqrt(dot(direction, direction));
+
+	rayQueryEXT query;
+
+	// The following runs the actual ray query
+	// For performance, use gl_RayFlagsTerminateOnFirstHitEXT, since we only need to know
+	// whether an intersection exists, and not necessarily any particular intersection
+	rayQueryInitializeEXT(query, topLevelAS, gl_RayFlagsTerminateOnFirstHitEXT, 0xFF, pos, tmin, direction.xyz, distance);
+	// The following is the cononical way of using ray Queries from the fragment shader when
```
JoseEmilio-ARM: `cononical` -> `canonical`

samples/extensions/ray_queries/ray_queries.cpp
gpx1000: I'm less certain about this, it's a single model.  A good optimization for most use cases is to reduce the number of draw calls and this is emblematic of that.  However, at the same time it would make sense to use space partitioning given that the model is a full scene one can traverse.  The complexity rises to do this yet doesn't refine the actual sample's use case.
Maybe a good compromise would be to add a model to the scene that explicitly could be a good use case for another draw call?  What are your thoughts?

JoseEmilio-ARM: I was more concerned about overdraw, since we might be unnecessarily running ray queries for occluded pixels. A deferred pass would limit it to one ray per pixel. I can see how this would add complexity to the sample though.

samples/extensions/ray_queries/ray_queries.cpp
SaschaWillems: Feel free to use and/or change them as and if you like ;) I was under the impression that the ray tracing reflection sample used them, but that seems to have changed for the reworked sample using buffer device address. So yes, they seem to be unused right now.

samples/extensions/ray_queries/ray_queries.h
```diff
@@ -0,0 +1,108 @@
...
+	void create_uniforms();
+	void load_scene();
+	void create_descriptor_pool();
+	void create_descriptor_sets();
+	void prepare_pipelines();
+	void update_uniform_buffers();
+	void draw();
+
+	uint32_t max_thread_count{1};
+	bool     enable_shadows{false};
```
SaschaWillems: These two variables are not used in the sample. Maybe some leftovers?

SaschaWillems:
Just gave this a spin and I'm happy with this. 

I have one suggestion though: Adding at least basic lighting to the scene. With the scene only being a single-colored mesh, it kinda looks odd if you're looking at a mostly shadowed area, with just two different grey values.

And will there be a markdown tutorial for this sample too?

samples/extensions/ray_queries/ray_queries.cpp
gpx1000: Overdraw is a compelling reason.  I have to admit to being torn as I think simple samples are better than complex ones to communicate the core idea.  However, a reasoning for ray queries to exist in a common pipeline would be to be able to make use of running them only in a deferred fragment pass so they are limited in scope to the number of pixels actually to be rendered.
I'm leaning towards saying this is beyond the scope of a basic introduction to ray queries as a core tenet of the extension is they can be used from any shader at any point in either the traditional or ray tracing pipelines.  However, a performance sample would make a lot of sense to demonstrate the reasoning why overdraw would be a potential issue.  Maybe a future sample should address this.  Do you have any thoughts on that approach?

gpx1000:
> Just gave this a spin and I'm happy with this.
> 
> I have one suggestion though: Adding at least basic lighting to the scene. With the scene only being a single-colored mesh, it kinda looks odd if you're looking at a mostly shadowed area, with just two different grey values.
> 
> And will there be a markdown tutorial for this sample too?

Thanks @SaschaWillems for the review
We added an AO per your request.

It's my hope that the code in the sample is a sufficient tutorial in this case as to where a markdown tutorial would be superfluous.  If we get a lot of questions about it, it'd make sense to add one in a future PR to make certain everything is explicit.  Otherwise, I'm inclined to recommend the code is simple enough to not benefit from an explicit tutorial.

samples/extensions/ray_queries/ray_queries.cpp
gpx1000: @SaschaWillems and @JoseEmilio-ARM I updated the #313  PR to optionally use the framework helper class.  There's a few bugs in the framework so I'd like to let that PR get merged first and come back in a future PR to do the same treatment to this sample if that's okay with you guys?

SaschaWillems:
> Thanks @SaschaWillems for the review
We added an AO per your request.

Thanks. That's much better :) One minor suggestion would be lowering the AO range a bit, as the effect looks a bit odd at first.

> It's my hope that the code in the sample is a sufficient tutorial in this case as to where a markdown tutorial would be superfluous. If we get a lot of questions about it, it'd make sense to add one in a future PR to make certain everything is explicit. Otherwise, I'm inclined to recommend the code is simple enough to not benefit from an explicit tutorial.

Sure, I'm fine with that too :)

SaschaWillems (dismissed):

shaders/ray_queries/ray_shadow.frag
```diff
@@ -0,0 +1,121 @@
...
+		for (uint k = 0; k < max_ao_each; ++k){
+			float theta =  0.5*(-3.14159 + 2 * 3.14159 * (float(k + 1) / float(max_ao_each + 2)));
+			float x = cos(phi) * sin(theta);
+			float y = sin(phi) * sin(theta);
+			float z = cos(theta);
+			vec3 direction = x * u + y * v + z * object_normal;
+
+			rayQueryEXT query;
+			rayQueryInitializeEXT(query, topLevelAS, gl_RayFlagsTerminateOnFirstHitEXT, 0xFF, object_point, tmin, direction.xyz, tmax);
+			while (rayQueryProceedEXT(query))
```
JoseEmilio-ARM: similar to `intersects_light` below, please avoid loop

samples/extensions/ray_queries/ray_queries.cpp
JoseEmilio-ARM: > a reasoning for ray queries to exist in a common pipeline would be to be able to make use of running them only in a deferred fragment pass so they are limited in scope to the number of pixels actually to be rendered

Yes exactly, I was going for a simple deferred pipeline like in some of the performance samples, to show how ray queries can be part of it. But also happy to let this sample focus on the use of the extension and address more realistic use-cases in future samples

samples/extensions/ray_queries/ray_queries.cpp
JoseEmilio-ARM: sounds good, maybe need to raise an issue to not forget

shaders/ray_queries/ray_shadow.frag
gpx1000: Do you have a suggestion for what to change it to?
I was under the impression that rayQueryProceedExt is meant to return false until it is finished traversing.  gl_RayFlagsTerminateOnFirstHitExp should hopefully ensure this is only called once.  If it's changed to a different flag, then the while loop might have a real cost but doesn't this combo result in a no cost loop that shows correct usage?

JoseEmilio-ARM: ah sorry, I meant to replace it with exactly the same as in lines 100-105 below, to call `rayQueryProceed` by itself rather than in an empty while loop

gpx1000: Sorry still a little confused, line 100 has an empty while loop as well.

Do you mean the nested for loop logic?  Move the while loop outside the for loop(s)?  This should run 9 times as written with a no-op essentially on the while loop as it's going to be a first hit.

aurolou (dismissed):

shaders/ray_queries/ray_shadow.frag
JoseEmilio-ARM: Hi, apologies for the delay. I mean the empty `while` loop, just for consistency and also because in this case a single call is enough and avoids compiler costs, as explained in the comments below in lines 100-105. Might not need to repeat the comments in both places though

```
	// The following is the canonical way of using ray Queries from the fragment shader when
	// there's more than one bounce or hit to traverse:
	// while (rayQueryProceedEXT(query)) { }
	// This sample has set flags to gl_RayFlagsTerminateOnFirstHitEXT which means that there
	// will never be a bounce and no need for an expensive while loop.  (i.e. we only need to call it once).
	rayQueryProceedEXT(query);
```

Happy to approve after this is resolved 🙏

samples/extensions/ray_queries/ray_queries.cpp
JoseEmilio-ARM: https://github.com/KhronosGroup/Vulkan-Samples/issues/367

JoseEmilio-ARM (approved):

aurolou (approved):

SaschaWillems (approved):
LGTM. Runs fine on my RTX 2060 :)

PR #320, opened 2021-08-09
WIP: More warnings as errors

TomAtkinsonArm:
## Description

As discussed on the call this PR will be used to determine what error level we want to conform to and the work needed to conform to said level

## General Checklist:

## Sample Checklist


framework/CMakeLists.txt
```diff
@@ -402,9 +402,9 @@ endif()
 if(${VKB_WARNINGS_AS_ERRORS})
     message(STATUS "Warnings as Errors Enabled")
     if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-        target_compile_options(${PROJECT_NAME} PRIVATE -Werror)
+        target_compile_options(${PROJECT_NAME} PRIVATE -Werror -Wall)
     elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-        target_compile_options(${PROJECT_NAME} PRIVATE /W3 /WX)
+        target_compile_options(${PROJECT_NAME} PRIVATE /Wall /WX)
```
gpx1000, line 414: Been a bit since last I checked on what compile flags MSVC supports, However, I thought Wall is clang/gcc only and msvc would use /W4

As there's going to be several warnings, would you like some help on the cleanup?  We can all use this branch after we get the plugins PR merged.

SaschaWillems: MSVC does indeed support /Wall. I'm not sure if we actually want to use that. Details on this can be found here: https://docs.microsoft.com/en-us/cpp/preprocessor/compiler-warnings-that-are-off-by-default?view=msvc-160

If we enable this for MSVC we'll probably need to explicitly disable different warnings, esp. with VS2019.

SaschaWillems:
Just did a quick run with /Wall on Visual Studio 2019, which results in 1727 warnings in the framework and failing the build due to an error in the Vulkan headers (!) in vulkan_core.h on line 2056. So we also probably need to exclude certain files from either warnings at all or /Wall.

framework/CMakeLists.txt
gpx1000: The trick we'll need to use is to specify that the compile options are private the the framework and samples so they don't also go into the dependencies.  That should cut down quite a bit.

Looking locally I'm seeing a fair number that can be auto resolved (clion has these tools... ;-)).  Mostly static analysis tools can get rid of very simple warnings.  However, I'm more concerned with the false positives.  We'll need to target very specific parts of the code with a macro that disables specific warnings as there will be several instances of unused variable that gets set in C++ but used in shaders.

TomAtkinsonArm:
Hey @gpx1000,

@cforfang had a quick look over the initial fixes and spotted our use of `LOG*()`

Currently in `common/logging.h` we end the definitions with ```;``` which causes the error when using the LOG definitions elsewhere. It would be better to continue to end our log calls with the semi colon and remove the semi colons from the definitions below.

```
#define LOGI(...) spdlog::info(__VA_ARGS__);
#define LOGW(...) spdlog::warn(__VA_ARGS__);
#define LOGE(...) spdlog::error("[{}:{}] {}", __FILENAME__, __LINE__, fmt::format(__VA_ARGS__));
#define LOGD(...) spdlog::debug(__VA_ARGS__);
```

If your in agreement I will do a quick find and replace to revert this specific change from your contribution.

gpx1000:
Hey @TomAtkinsonArm and @cforfang 

yep, that is a good idea.  I personally find it easier to read code when as a rule macros behave like regular function calls so I typically require similar solutions in my personal code (no ';' allowed to terminate a macro function).

TomAtkinsonArm:
@gpx1000 thanks for your input, Ill change that in my latest commit

TomAtkinsonArm:
Reverted the LOG stuff, combined some third parties into a single header (glm, maybe vulkan headers? not decided yet)

```
#	pragma warning(disable : 4820 4061 4625 4626 4355 5204 5220 5026 5027 4100 4324 4514 4365 5045 4263 4264 4389 4868 4623 4242 4244 5039)

// C4820 - 'bytes' bytes padding added after construct 'member_name'
// C4061 - enumerator 'identifier' in switch of enum 'enumeration' is not explicitly handled by a case label
// C4625 - 'derived class' : copy constructor was implicitly defined as deleted because a base class copy constructor is inaccessible or deleted
// C4626 - 'derived class' : assignment operator was implicitly defined as deleted because a base class assignment operator is inaccessible or deleted
// C4355 - 'this' : used in base member initializer list
// C5026 - 'type': move constructor was implicitly defined as deleted
// C5027 - 'type': move assignment operator was implicitly defined as deleted
// C4100 - unreferenced formal parameter
// C4324 - structure was padded due to alignment specifier
// C4514 - unreferenced inline function has been removed
// C4365 - conversion from x to y, signed/unsigned mismatch
// C5045 - Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified
// C4263 - 'void ApiVulkanSample::render(float)': member function does not override any base class virtual member function
// C4264 - 'void vkb::VulkanSample::render(vkb::CommandBuffer &)': no override available for virtual member function from base 'vkb::VulkanSample'; function is hidden
// C4389 - signed/unsigned mismatch
// C4868 - compiler may not enforce left-to-right evaluation order in braced initializer list
// C4623 - 'vkb::sg::AnimationChannel': default constructor was implicitly defined as deleted
// C4242 C4244 -  conversion from 'X' to 'Y', possible loss of data (occurs in algorithm annoyingly...)
// C5039 - pointer or reference to potentially throwing function passed to 'extern "C"' function under -EHc. Undefined behavior may occur if this function throws an exception (GLFW window callbacks)

// Concerns C4263, C4264, C5039
```

I disabled all these warnings to get the build to work. Some of the smaller things I fixed but these are the main ones that popped up. I think some of these are excessive (C4820, C4263, C4264...) and so would like further opinions before going through and fixing them. MSVC in general seems to include a lot of warnings at this level which seem to cause more issues than they solve...

Also some of these warnings occur in the standard library which is out of our control to fix. A solution to avoiding all of this is to wrap both third parties and the standard library with the enable/disable warnings macros.

TomAtkinsonArm:
depreciated. Framework v2.0 uses higher warning levels

PR #326, opened 2021-08-16
OpenCL interoperability sample

PavelRudko-ARM:
## Description

A new extensions sample and a tutorial that demonstrate interoperability with OpenCL.
Currently only data sharing using AHardwareBuffers is covered.
In the future we will add data sharing using host memory and synchronization primitives usage, once the necessary extensions are available.
The sample only works on Android and only if OpenCL is supported and cl_arm_import_memory extension is available.

## General Checklist:

## Sample Checklist


SaschaWillems:
Thanks for the sample. OpenCL interopt is a question I see a lot, so having a sample to show that is a welcome addition 👍 

But any hint on what device can be used to test this? The sample also requires a lot of Vulkan extensions, and none of the Android devices I own can run this.

samples/extensions/open_cl_interop/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+#
+ 
+# The OpenCL interoperability only works on Android
+get_filename_component(FOLDER_NAME ${CMAKE_CURRENT_LIST_DIR} NAME)
+get_filename_component(PARENT_DIR ${CMAKE_CURRENT_LIST_DIR} PATH)
+get_filename_component(CATEGORY_NAME ${PARENT_DIR} NAME)
+
+# OpenCL
+add_library(opencl INTERFACE)
+set(OPENCL_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/third_party/OpenCL-Headers/)
```
gpx1000: would it make sense to use FIND_OPENCL?  i.e. try something like this:
find_package(OpenCL);
target_link_libraries(target public OpenCL::OpenCL)

Then you wouldn't need to include OpenCL headers in this PR and if the OpenCL headers installed in the system are updated, CMake will use those (i.e. we won't have to worry about grabbing the latst OpenCL headers

gpx1000 (commented):
Thanks for this sample!  It'd be great if it worked on more platforms.

gpx1000 (commented):
Looks like the sample needs to get the latest from master.  Also need to add the sample to the CMakeLists.txt's sample list.

samples/extensions/open_cl_interop/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+        CATEGORY ${CATEGORY_NAME}
+        AUTHOR "Arm"
+        NAME "OpenCL Interoperability"
+        DESCRIPTION "Example showing sharing resources between OpenCL and Vulkan"
+        TAGS "arm"
+        LIBS opencl
+        FILES
+            open_cl_functions.inl
+            open_cl_utils.h
+            open_cl_utils.cpp)
```
gpx1000, line 35: Is this the complete list of files that we want the compiler to compile?  How does open_cl_interop.cpp get compiled?

oddhack (commented):
I don't really have the knowledge to review this - sorry.

samples/extensions/open_cl_interop/CMakeLists.txt
PavelRudko-ARM: I tried to use the same approach as in `Vulkan-Samples/third_party/CMakeLists.txt`.  All the dependencies are included (even vulkan itself) so that the developers don't need to install anything extra. 
If OpenCL is not present for some reason (for example, not supported on the host platform, but available on the target device), then the headers will be missing.

PavelRudko-ARM:
> Looks like the sample needs to get the latest from master. Also need to add the sample to the CMakeLists.txt's sample list.

I have rebased to the current master. 
The samples are not listed in CMakeLists.txt explicitly. Instead, all the sample directories are included automatically:

```
foreach(SUB_DIR ${SUB_DIRS})
    add_sub_dirs(${CMAKE_CURRENT_SOURCE_DIR}/${SUB_DIR})
endforeach()
```

There is also a list for putting the samples in a certain order, but I'm not sure this one needs to be included there (open_gl_interop, for example, is not in this list).

samples/extensions/open_cl_interop/CMakeLists.txt
PavelRudko-ARM: It is included automatically. You can find the following lines in `bldsys/cmake/sample_helper.cmake`:
```
set(SRC_FILES
    ${TARGET_ID}.h
    ${TARGET_ID}.cpp
)

# Append extra files if present
if (TARGET_FILES)
    list(APPEND SRC_FILES ${TARGET_FILES})
endif()
```

PavelRudko-ARM:
> Thanks for the sample. OpenCL interopt is a question I see a lot, so having a sample to show that is a welcome addition 👍
> 
> But any hint on what device can be used to test this? The sample also requires a lot of Vulkan extensions, and none of the Android devices I own can run this.

Yes, unfortunately the sample is currently limited to Arm devices (due to `cl_arm_import_memory`), but it would be nice to extend it later for other platforms (which will probably involve some #ifdef's).

I have personally tested on the following devices (both have Mali GPUs):

- Samsung S21
- Huawei Mate40 Pro

kpet:
@PavelRudko-ARM Have you considered adding the OpenCL headers as a submodule rather than as a copy? It would be cleaner and make future updates and testing easier.

samples/extensions/open_cl_interop/CMakeLists.txt
gpx1000: It's probably best to include them as a submodule within the third_party folder of the framework so they're exposed to more potential samples in addition it'd be easier to keep it up to date.
However, something I'd like to recommend is to show use of system based inclusion such as the find_package.  If not found (i.e. in cmake after find_package OPENCL_NOTFOUND would be what the variable is set as).  Then fall back to the one in the third_party submodule.
The idea behind this approach is to encourage consumers of the sample to learn best practices and not embed a snap shot in time but rather the package which gets auto-updated with the system.

samples/extensions/open_cl_interop/open_cl_interop.h
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+#pragma once
+
+#include "api_vulkan_sample.h"
+#include "rendering/render_pipeline.h"
+#include "scene_graph/components/camera.h"
+
+#include <android/hardware_buffer.h>
+#include <android/hardware_buffer_jni.h>
```
gpx1000: these includes can be moved into an appropriate if macro check to ensure on Android.

Also, use forward declare of struct AHardwareBuffer; in the header then include these in the CPP to allow for less tightly coupling of header files.

samples/extensions/open_cl_interop/open_cl_interop.h
```diff
@@ -0,0 +1,100 @@
...
+
+struct CLData;
+
+class OpenCLInterop : public ApiVulkanSample
+{
+  public:
+	OpenCLInterop();
+	~OpenCLInterop();
```
gpx1000: please add override or virtual to a virtual desctuctor (prefer override).

samples/extensions/open_cl_interop/open_cl_interop.h
```diff
@@ -0,0 +1,100 @@
...
+
+	virtual bool prepare(vkb::Platform &platform) override;
```
gpx1000: either virtual or override, not both.  prefer override.

samples/extensions/open_cl_interop/open_cl_interop.cpp
```diff
@@ -0,0 +1,622 @@
...
+{
+	cl_context       context{nullptr};
+	cl_device_id     device_id{nullptr};
+	cl_command_queue command_queue{nullptr};
+	cl_program       program{nullptr};
+	cl_kernel        kernel{nullptr};
+	cl_mem           image{nullptr};
+};
+
+OpenCLInterop::OpenCLInterop()
```
gpx1000, line 44: index_count should be initialized here.  I also recommend using initializer lists for zoom, title etc.

samples/extensions/open_cl_interop/open_cl_interop.cpp
```diff
@@ -0,0 +1,622 @@
...
+		render_pass_begin_info.framebuffer = framebuffers[i];
+
+		VK_CHECK(vkBeginCommandBuffer(draw_cmd_buffers[i], &command_buffer_begin_info));
+
+		vkCmdBeginRenderPass(draw_cmd_buffers[i], &render_pass_begin_info, VK_SUBPASS_CONTENTS_INLINE);
+
+		VkViewport viewport = vkb::initializers::viewport((float) width, (float) height, 0.0f, 1.0f);
+		vkCmdSetViewport(draw_cmd_buffers[i], 0, 1, &viewport);
+
+		VkRect2D scissor = vkb::initializers::rect2D(width, height, 0, 0);
```
gpx1000: width/height are uint32_t static_cast them from the int32_t type they are here for warning removal.

samples/extensions/open_cl_interop/open_cl_interop.cpp
```diff
@@ -0,0 +1,622 @@
...
+		vkCmdSetScissor(draw_cmd_buffers[i], 0, 1, &scissor);
+
+		vkCmdBindDescriptorSets(draw_cmd_buffers[i], VK_PIPELINE_BIND_POINT_GRAPHICS, pipeline_layout, 0, 1, &descriptor_set, 0, NULL);
```
gpx1000: NULL -> nullptr

samples/extensions/open_cl_interop/open_cl_interop.cpp
```diff
@@ -0,0 +1,622 @@
...
+	        //	Fragment shader: layout (binding = 1) uniform sampler2D samplerColor;
+	        vkb::initializers::write_descriptor_set(
+	            descriptor_set,
+	            VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER,        // The descriptor set will use a combined image sampler (sampler and image could be split)
+	            1,                                                // Shader binding point 1
+	            &image_descriptor)                                // Pointer to the descriptor image for our texture
+	    };
+
+	vkUpdateDescriptorSets(get_device().get_handle(), vkb::to_u32(write_descriptor_sets.size()),
+	                       write_descriptor_sets.data(), 0, NULL);
```
gpx1000: NULL -> nullptr

samples/extensions/open_cl_interop/open_cl_interop.cpp
```diff
@@ -0,0 +1,622 @@
...
+	    VK_DYNAMIC_STATE_SCISSOR};
+
+	VkPipelineDynamicStateCreateInfo dynamic_state =
+	    vkb::initializers::pipeline_dynamic_state_create_info(
+	        dynamic_state_enables.data(),
+	        vkb::to_u32(dynamic_state_enables.size()),
+	        0);
+
+	// Load shaders
+	std::array<VkPipelineShaderStageCreateInfo, 2> shader_stages;
```
gpx1000: shader_stages isn't initialized.

samples/extensions/open_cl_interop/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+get_filename_component(PARENT_DIR ${CMAKE_CURRENT_LIST_DIR} PATH)
+get_filename_component(CATEGORY_NAME ${PARENT_DIR} NAME)
+
+# OpenCL
+add_library(opencl INTERFACE)
+set(OPENCL_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/third_party/OpenCL-Headers/)
+target_sources(opencl INTERFACE ${OPENCL_INCLUDE_DIR}/CL/opencl.h)
+target_include_directories(opencl INTERFACE ${OPENCL_INCLUDE_DIR})
+
+if(ANDROID)
```
gpx1000, line 23: took me a bit to figure this out.  the problem with this path of limiting inclusion of the add_sample_with_tags for only one platform is if you're on a different platform and try to run the sample, or list samples, the folder discovery logic is still finding this sample.  In order to keep things working with other samples as normal, let's put the project specialization within the code of the sample itself.
This will promote both the idea of future work on making work with other platforms and allow for the system to correctly say "extension not found" rather than failing out in a unique way.

samples/extensions/open_cl_interop/open_cl_interop.cpp
```diff
@@ -0,0 +1,622 @@
...
+	vkDestroyDescriptorSetLayout(device->get_handle(), descriptor_set_layout, nullptr);
+	vkDestroyFence(device->get_handle(), rendering_finished_fence, nullptr);
+	vkDestroySampler(device->get_handle(), shared_texture.sampler, nullptr);
+	vkDestroyImageView(device->get_handle(), shared_texture.view, nullptr);
+	vkDestroyImage(device->get_handle(), shared_texture.image, nullptr);
+	vkFreeMemory(device->get_handle(), shared_texture.memory, nullptr);
+
+	clReleaseMemObject(cl_data->image);
+	clReleaseContext(cl_data->context);
+	delete cl_data;
```
gpx1000: recommend testing that cl_data is not nullptr as its creation is in prepare and not the initializer (in other words even on supported platforms, this *could* cause a crash bug on exit).

gpx1000 (changes requested):
unfortunately, I don't have a device that would support testing this further.  I've tried and gotten as far as I can running it on unsupported devices.  Code review wise, just need a few changes and then I'll give an approval based upon the code without being able to run it locally.

r-potter:
> I have personally tested on the following devices (both have Mali GPUs):
> 
> * Samsung S21
> * Huawei Mate40 Pro

@PavelRudko-ARM Was this a stock (consumer) S21? 

My S21 fails to dlopen libOpenCL.so due to restrictions on where a shared object can be dlopen'd from.

samples/extensions/open_cl_interop/CMakeLists.txt
PavelRudko-ARM: Moved OpenCL headers to a submodule.

PavelRudko-ARM:
> @PavelRudko-ARM Have you considered adding the OpenCL headers as a submodule rather than as a copy? It would be cleaner and make future updates and testing easier.

I have moved OpenCL Headers to a submodule. It's indeed cleaner and this way we don't need to add OpenCL files to `.copyrightignore `and `.snakeignore`.

samples/extensions/open_cl_interop/open_cl_interop.cpp
PavelRudko-ARM: I have added `index_count ` initialization in the header file (to keep it consistent with the way other class members are initialized). We cannot use initializer lists with `zoom `or `title `as they are members of the base class (ApiVulkanSample).

samples/extensions/open_cl_interop/CMakeLists.txt
PavelRudko-ARM: The problem is, AHardwareBuffer does not exist on other platforms. Even if I forward declare it and add platform checks for `#include <android/hardware_buffer.h>` as suggested above, it will look slightly odd on Windows. 
I just used the same approach as in `open_gl_interop` sample.
If the check is performed during CMake stage, the project is not included at all. Sorry, I'm not sure what you mean by "folder discovery logic".

samples/extensions/open_cl_interop/open_cl_interop.h
PavelRudko-ARM: Done (override).

samples/extensions/open_cl_interop/open_cl_interop.h
PavelRudko-ARM: Removed `virtual`.

samples/extensions/open_cl_interop/open_cl_interop.cpp
PavelRudko-ARM: Done.

samples/extensions/open_cl_interop/open_cl_interop.cpp
PavelRudko-ARM: Done.

samples/extensions/open_cl_interop/open_cl_interop.cpp
PavelRudko-ARM: Done.

samples/extensions/open_cl_interop/open_cl_interop.cpp
PavelRudko-ARM: Fixed.

samples/extensions/open_cl_interop/open_cl_interop.cpp
PavelRudko-ARM: Done.

PavelRudko-ARM:
> > I have personally tested on the following devices (both have Mali GPUs):
> > 
> > * Samsung S21
> > * Huawei Mate40 Pro
> 
> @PavelRudko-ARM Was this a stock (consumer) S21?
> 
> My S21 fails to dlopen libOpenCL.so due to restrictions on where a shared object can be dlopen'd from.

Hi! Yes, it's Samsung S21 5G with stock Android. I've just tested it after a rebase to the latest version of `master` and it works.
Here are my ` build.gradle` settings:

        compileSdk 31
        ndkVersion '23.0.7599858'

        defaultConfig {
            minSdk 24
            targetSdk 31
        ...

r-potter:
> Hi! Yes, it's Samsung S21 5G with stock Android. I've just tested it after a rebase to the latest version of `master` and it works. Here are my ` build.gradle` settings:

Thanks. I'll try to investigate further this week and see if I can figure out what is causing the discrepancy

samples/extensions/open_cl_interop/CMakeLists.txt
gpx1000: What I meant by the folder discovery logic is that if you try to run the sample on an unsupported platform it will currently crash out rather than print a sensible message and exit.  This is caused by CMake discovering the folder but not finding any symbols to load, so it passes the logic that discovers samples, but fails to find anything to run.
Ideally, allow for it to compile these files, but drop a "unsupported platform" assertion error in the startup so it replicates what happens on a platform that doesn't have an extension for example.

PavelRudko-ARM: If `add_sample_with_tags `or `add_sample` is not called for the sample in CMakeLists.txt, it is not added to `std::vector<SampleInfo> samples` in apps.cpp, so there is no way to run it and it will not be listed by `vulkan_samples samples`.
It is also not included in the build (I don't see the open_cl_interop project in the Visual Studio solution, for example).

gpx1000: So it looks like I was completely wrong.  I was getting a crash bug when I ran this in linux with "sample open_cl_interop" but now am getting it completely not validating that as a sample (i.e. it works).  Okay, am happy with this path.

SaschaWillems:
Finally found a device with a Mali G71 that supports all the required Vulkan and OpenCL extensions. The sample starts, but displays a black quad for me, which doesn't look right.

In logcat, I do see an error about importing the OpenCL memory:
```
10-18 19:06:59.018: E/VulkanSamples(28623): [error] [_cl_interop/open_cl_interop.cpp:562] Cannot import OpenCL memory, error code: -64.
```

gpx1000 (dismissed):

gpx1000:
I'm approving based upon the code changes that I can review.  Unfortunately, I'm unable to test on this platform.  Thus, all other reviewers please note that this code is untested by me.

PavelRudko-ARM:
> Finally found a device with a Mali G71 that supports all the required Vulkan and OpenCL extensions. The sample starts, but displays a black quad for me, which doesn't look right.
> 
> In logcat, I do see an error about importing the OpenCL memory:
> 
> ```
> 10-18 19:06:59.018: E/VulkanSamples(28623): [error] [_cl_interop/open_cl_interop.cpp:562] Cannot import OpenCL memory, error code: -64.
> ```

Hello! Based on the error, `cl_import_memory_arm` extension is present, but Android Hardware Buffers are not supported. It's very likely because another extension 'cl_arm_import_memory_android_hardware_buffer' is not available. I have added checks for extensions availability on the platform, so now the error message should be more meaningful. Could you try it again, please?

Also, previously I was loading `clImportMemoryARM` function pointer using dlsym. Although the symbols might be exposed by libOpenCL.so, it's not supposed to be so. I have changed loading extension functions, now they are loaded using `clGetExtensionFunctionAddressForPlatform` (correct way).

gpx1000:
I got this working on my pixel 6.  I can now approve.

However, it might be beneficial to put together some documentation for end users that might want to use this.
I'd suggest instructions on how to get libOpenCL.so to be found as Google doesn't support direct access to it.
/system/etc/public.libraries.txt <-- if you do see libOpenCL.so in your /system/vendor/lib but don't find it in the public libraries  text file, then you must grab the library with adb pull.  Then you need to add it to the apk by adding it to the jnilibs.  Next you need to update the path on line 30 of Open_cl_utils.cpp to the exploded apk library directory.

Doing all that will get any phone that has the requisite capabilities to work.  Another option we should consider is to write a CMake tool or script that will do these things automatically for end users.

gpx1000 (dismissed):

samples/extensions/open_cl_interop/open_cl_utils.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "open_cl_functions.inl"
+
+static void *handle = nullptr;
+
+cl_platform_id load_opencl()
+{
+	static const std::vector<std::string> possible_libraries{"libOpenCL.so", "libGLES_mali.so", "libmali.so"};
+	for (const auto &possible_library : possible_libraries)
+	{
+		handle = dlopen("libOpenCL.so", RTLD_LAZY | RTLD_LOCAL);
```
gpx1000: While this change is likely not needed, possible_library isn't used anywhere in the for loop; so dlopen libOpenCL.so is called 3 times.
I'd recommend reversing this change to just the single libOpenCL.so and allow ld to determine what dependencies it needs to load from loading libOpenCL.so.

PavelRudko-ARM: Agree. I have reverted this change, thanks for your attentiveness!

gpx1000 (approved):

gpx1000 (approved):

marty-johnson59:
Merging this vender-specific extension - LMK if issues.  Thx

PR #330, opened 2021-08-23
bindless resources demonstration

gpx1000:
@wlohry thanks for the effort on this!

## Description

This sample demonstrates how to reduce CPU usage by offloading draw call generation and frustum culling to the GPU.

assets PR https://github.com/KhronosGroup/Vulkan-Samples-Assets/pull/13 required for this sample

## Tested Environments
Windows 10 / Ubuntu Linux
GeForce 3060 / GeForce 2080 / GeForce 1070
Vulkan version: 1.2.176

## General Checklist:

- [X] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)

## Sample Checklist

- [X] For new samples, I have added a paragraph with a summary to the appropriate chapter in the [samples readme](./../samples/README.md)


shaders/bindless/bindless.vert
```diff
@@ -0,0 +1,43 @@
...
+layout(set = 0, binding = 2) uniform GlobalUniform
+{
+	mat4 view;
+	mat4 proj;
+}
+global_uniform;
+
+layout(location = 0) out vec4 o_pos;
+layout(location = 1) out vec2 o_uv;
+layout(location = 2) out float o_texture_index;
```
cforfang: Could this be `out uint` (in VS) and `flat in uint` (in FS) instead of casting to float and back?

shaders/bindless/cull_address.comp
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+	return true;
+}
+
+void main()
+{
+	uint             id         = gl_GlobalInvocationID.x;
+	ModelInformation model      = model_buffer.arr[id];
+	bool             is_visible = check_is_visible(global_uniform.proj * global_uniform.view, vec3(model.x, model.y, model.z), model.r);
```
cforfang: Would it make sense to pre-multiply `global_uniform.proj * global_uniform.view` on CPU instead of in every compute thread?

shaders/bindless/bindless.vert
```diff
@@ -0,0 +1,43 @@
...
+layout(location = 4) in uint texture_index;
+
+layout(set = 0, binding = 2) uniform GlobalUniform
+{
+	mat4 view;
+	mat4 proj;
+}
+global_uniform;
+
+layout(location = 0) out vec4 o_pos;
```
cforfang: Unused?

cforfang:
These are texture-bleeding artifacts I think? They don't go away with nearest filtering, so probably an issue when generating the textures?
image
image

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,925 @@
...
+		vkDestroySampler(device->get_handle(), sampler, VK_NULL_HANDLE);
+	}
+}
+
+void BindlessResources::request_gpu_features(vkb::PhysicalDevice &gpu)
+{
+	if (gpu.get_features().multiDrawIndirect)
+	{
+		gpu.get_mutable_requested_features().multiDrawIndirect = VK_TRUE;
+		m_supports_mci                                         = true;
```
cforfang: `m_supports_mci` -> `m_supports_mdi`?

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,925 @@
...
+	auto staging_model_buffer  = std::make_unique<vkb::core::Buffer>(get_device(), model_buffer_size, VK_BUFFER_USAGE_TRANSFER_SRC_BIT, VMA_MEMORY_USAGE_CPU_TO_GPU);
+
+	// We will store the GPU commands in the indirect call buffer
+	constexpr auto default_indirect_flags = VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_STORAGE_BUFFER_BIT | VK_BUFFER_USAGE_INDIRECT_BUFFER_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT;
+	auto           indirect_flags         = default_indirect_flags;
+	if (m_supports_buffer_device)
+	{
+		indirect_flags |= VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT;
+	}
+	indirect_call_buffer = std::make_unique<vkb::core::Buffer>(get_device(), models.size() * sizeof(VkDrawIndexedIndirectCommand), indirect_flags, VMA_MEMORY_USAGE_GPU_ONLY);
```
cforfang: This buffer is created with `sharingMode VK_SHARING_MODE_EXCLUSIVE`, however from the code it might look like the culling compute dispatches and rendering could run on different queue families. Should this either be `VK_SHARING_MODE_CONCURRENT`, or there needs to be ownership transfer as part of the compute -> graphics handover?

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,925 @@
...
+	{
+		build_command_buffers();
+		m_requires_rebuild = false;
+	}
+
+	draw();
+
+	update_scene_uniform();
+
+	run_cull();
```
cforfang: The rendering done in `draw()` above reads the indirect arguments. Is there currently anything synchronizing the work in `run_cull()` which _writes_ the indirect arguments, i.e. preventing a WAR hazard?

(There is a wait after the dispatch in `run_cull()`, preventing RAW hazard. However I also wonder if there should be a pipeline barrier there to guarantee the written data is made available?)

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,925 @@
...
+
+		VkImageMemoryBarrier image_barrier = vkb::initializers::image_memory_barrier();
+		image_barrier.srcAccessMask        = 0;
+		image_barrier.dstAccessMask        = 0;
+		image_barrier.image                = texture.image->get_handle();
+		image_barrier.subresourceRange     = subresource_range;
+		image_barrier.oldLayout            = VK_IMAGE_LAYOUT_UNDEFINED;
+		image_barrier.newLayout            = VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL;
+
+		// The semaphore takes care of srcStageMask.
```
cforfang: Which semaphore does this refer to?

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,925 @@
...
+	if (render_mode == RenderMode::GPU)
+	{
+		bind(gpu_cull_pipeline, gpu_cull_pipeline_layout, gpu_cull_descriptor_set);
+	}
+	else
+	{
+		bind(device_address_pipeline, device_address_pipeline_layout, device_address_descriptor_set);
+	}
+
+	vkCmdDispatch(cmd, models.size(), 1, 1);
```
cforfang: This would effectively dispatch N work-groups consisting of 1 thread, would it? Would it make more sense, performance wise, to dispatch e.g. N/64 workgroups of 64 threads -- so 64-wide warps can be fully filled?

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,925 @@
...
+		region.imageSubresource.layerCount     = 1;
+		region.imageExtent                     = image->get_extent();
+		region.bufferOffset                    = offsets[0][0];
+
+		texture_cmd.copy_buffer_to_image(*data_buffer, *texture.image, {region});
+		texture_cmd.end();
+
+		auto &queue = device->get_queue_by_flags(VK_QUEUE_GRAPHICS_BIT, 0);
+		queue.submit(texture_cmd, device->request_fence());
+		device->get_fence_pool().wait();
```
cforfang: Should probably add `device->get_fence_pool().reset();` after this, else the request-pool grows in size for each mesh. This shouldn't _really_ be a problem -- would just result in a bunch of already-waited-on-fenced being redundantly passed to WaitForFences -- however, without the reset, I'm getting occasional DEVICE_LOST on my RTX2060 on Windows.

cforfang:
I didn't see any mention of licensing for the assets; should that be mentioned somewhere?

SaschaWillems:
> I didn't see any mention of licensing for the assets; should that be mentioned somewhere?

See https://github.com/KhronosGroup/Vulkan-Samples-Assets/pull/13. We already discussed that on a different communication channel, and yes, the license for those assets needs to be added there too.

gpx1000:
> These are texture-bleeding artifacts I think? They don't go away with nearest filtering, so probably an issue when generating the textures?
> image
> image

Interesting find.  I'm honestly not sure if it's the model as intended or indeed texture bleeding.  I'll investigate further.

**Update
The problem is in the blender converting to gltf and multiple textures as per requirement.  The conversion resulted in the model not having correctly lined up textures.  We're working on reexporting the model to create a fix.

samples/performance/bindless/bindless.cpp
gpx1000: The synchronization should be corrected explicitly now.  Draw has the device wait idle (that I'd love to take out of the framework), the memory barrier and the fence in culling should be correctly sufficient; please double check.

gpx1000:
> > I didn't see any mention of licensing for the assets; should that be mentioned somewhere?
> 
> See [KhronosGroup/Vulkan-Samples-Assets#13](https://github.com/KhronosGroup/Vulkan-Samples-Assets/pull/13). We already discussed that on a different communication channel, and yes, the license for those assets needs to be added there too.

Attribution should now exist in the README properly

gpx1000:
> **Update
> The problem is in the blender converting to gltf and multiple textures as per requirement. The conversion resulted in the model not having correctly lined up textures. We're working on reexporting the model to create a fix.

This should now be fixed by the latest assets

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,955 @@
...
+		auto &texture_cmd = device->get_command_pool().request_command_buffer();
+		texture_cmd.begin(VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT, VK_NULL_HANDLE);
+
+		VkImageSubresourceRange subresource_range = {VK_IMAGE_ASPECT_COLOR_BIT, 0, 1, 0, 1};
+		subresource_range.baseMipLevel            = 0;
+		subresource_range.levelCount              = texture.n_mip_maps;
+
+		VkImageMemoryBarrier image_barrier = vkb::initializers::image_memory_barrier();
+		image_barrier.srcAccessMask        = 0;
+		image_barrier.dstAccessMask        = 0;
```
cforfang: `dstAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT`

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,955 @@
...
+		image_barrier.image                = texture.image->get_handle();
+		image_barrier.subresourceRange     = subresource_range;
+		image_barrier.oldLayout            = VK_IMAGE_LAYOUT_UNDEFINED;
+		image_barrier.newLayout            = VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL;
+
+		vkCmdPipelineBarrier(texture_cmd.get_handle(), VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, VK_PIPELINE_STAGE_ALL_COMMANDS_BIT, 0, 0, nullptr, 0, nullptr, 1, &image_barrier);
```
cforfang: `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT` -> `VK_PIPELINE_STAGE_TRANSFER_BIT` -- might as well be specific here.

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,955 @@
...
+		image->create_vk_image(*device);
+		Texture texture;
+		texture.n_mip_maps = uint32_t(image->get_mipmaps().size());
+		texture.image      = std::make_unique<vkb::core::Image>(*device,
+                                                           image->get_extent(),
+                                                           image->get_format(),
+                                                           VK_IMAGE_USAGE_SAMPLED_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT,
+                                                           VMA_MEMORY_USAGE_GPU_ONLY,
+                                                           VK_SAMPLE_COUNT_1_BIT,
+                                                           texture.n_mip_maps,
```
cforfang: Only mip 0 is uploaded below I think -- could one make this `1` and/or assert `texture.n_mip_maps == 1`?

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,955 @@
...
+
+		VkImageMemoryBarrier image_barrier = vkb::initializers::image_memory_barrier();
+		image_barrier.srcAccessMask        = 0;
+		image_barrier.dstAccessMask        = 0;
+		image_barrier.image                = texture.image->get_handle();
+		image_barrier.subresourceRange     = subresource_range;
+		image_barrier.oldLayout            = VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL;
+		image_barrier.newLayout            = VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL;
+
+		vkCmdPipelineBarrier(cmd.get_handle(), VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, VK_PIPELINE_STAGE_ALL_COMMANDS_BIT, 0, 0, nullptr, 0, nullptr, 1, &image_barrier);
```
cforfang: `VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT` -> `VK_PIPELINE_STAGE_TRANSFER_BIT`. Could minimize `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT` to `VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT` as well.

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,955 @@
...
+	auto &cmd = device->get_command_pool().request_command_buffer();
+	cmd.begin(VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT, VK_NULL_HANDLE);
+	for (auto &&texture : textures)
+	{
+		VkImageSubresourceRange subresource_range = {VK_IMAGE_ASPECT_COLOR_BIT, 0, 1, 0, 1};
+		subresource_range.baseMipLevel            = 0;
+		subresource_range.levelCount              = texture.n_mip_maps;
+
+		VkImageMemoryBarrier image_barrier = vkb::initializers::image_memory_barrier();
+		image_barrier.srcAccessMask        = 0;
```
cforfang: `srcAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT`

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,955 @@
...
+		image_barrier.dstAccessMask        = 0;
```
cforfang: `dstAccessMask = VK_ACCESS_SHADER_READ_BIT`

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,955 @@
...
+				}
+			}
+			model.bounding_sphere = BoundingSphere(pts);
+			models.emplace_back(std::move(model));
+		}
+	}
+
+	auto &cmd = device->get_command_pool().request_command_buffer();
+	cmd.begin(VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT, VK_NULL_HANDLE);
+	for (auto &&texture : textures)
```
cforfang: Worth it to do 1x `vkCmdPipelineBarrier` with Nx `VkImageMemoryBarrier`, instead of Nx `vkCmdPipelineBarrier` with 1x `VkImageMemoryBarrier`?

samples/performance/bindless/bindless.cpp
```diff
@@ -0,0 +1,955 @@
...
+	auto &memory_cmd = device->get_command_pool().request_command_buffer();
+	memory_cmd.begin(VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT, VK_NULL_HANDLE);
+
+	VkBufferMemoryBarrier memory_barrier = vkb::initializers::buffer_memory_barrier();
+	memory_barrier.buffer                = scene_uniform_buffer->get_handle();
+	memory_barrier.srcAccessMask         = VK_ACCESS_MEMORY_WRITE_BIT;
+	memory_barrier.dstAccessMask         = VK_ACCESS_MEMORY_READ_BIT;
+	memory_barrier.size                  = scene_uniform_buffer->get_size();
+
+	vkCmdPipelineBarrier(memory_cmd.get_handle(), VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, VK_PIPELINE_STAGE_HOST_BIT, 0, 0, VK_NULL_HANDLE, 1, &memory_barrier, 0, VK_NULL_HANDLE);
```
cforfang: Not sure I understand what this is doing -- add a comment maybe?

samples/performance/bindless/bindless.cpp
cforfang: > memory barrier and the fence in culling

Hm, is there supposed to be a memory barrier in `run_gpu_cull()`? Don't think I see it.

SaschaWillems (changes requested):
I'm getting two validation layer messages at startup using SDK Version 1.2.189:

```[error] [framework\core\instance.cpp:41] 1165064310 - VUID-VkGraphicsPipelineCreateInfo-layout-00756: Validation Error: [ VUID-VkGraphicsPipelineCreateInfo-layout-00756 ] Object 0: handle = 0x16ee604fae0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x45717876 | Shader expects at least 256 descriptors for binding 0.1 but only 225 provided The Vulkan spec states: layout must be consistent with all shaders specified in pStages (https://vulkan.lunarg.com/doc/view/1.2.189.0/windows/1.2-extensions/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-layout-00756)```

```[error] [framework\core\instance.cpp:41] 1028204675 - VUID-VkShaderModuleCreateInfo-pCode-04147: Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-04147 ] Object 0: handle = 0x16ee604fae0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x3d492883 | vkCreateShaderModule(): The SPIR-V Extension (SPV_EXT_physical_storage_buffer) was declared, but none of the requirements were met to use it. The Vulkan spec states: If pCode declares any of the SPIR-V extensions listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied (https://vulkan.lunarg.com/doc/view/1.2.189.0/windows/1.2-extensions/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-04147)```

And also a bunch of validation layer errors at instance destruction with pipelines, pipeline layouts and descriptor sets not being freed.

**Comment by SaschaWillems** [comment]
It also looks like the "Freeze culling" option isn't implemented yet. There is a toggle for it. but `m_freeze_cull` is never evaluated.

**Comment by gpx1000** [comment]
Message From Will regarding the remaining validation error:

> There are two SPV extensions: SPV_EXT_physical_storage_buffer and SPV_KHR_physical_storage_buffer. The current compiler outputs SPV_EXT*, but then the code has to use VK_EXT_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME, but this causes a lot of other problems within the code (e.g. vma allocator).
> Alternatively, if the SPV_KHR_physical_storage_buffer and VK_KHR_BUFFER_DEVICE_ADDRESS is used, then no validation layer messages will be produced. However, I can't figure out how to get the compiler to use the SPV_KHR* extension rather than SPV_EXT* one.
> 

I gave an effort to see if I could get the compiler to output the KHR extension but it doesn't look like that's happening.  Anyone have suggestions?

**Comment by SaschaWillems** [comment]
It looks like this behavior has been fixed by this PR: https://github.com/KhronosGroup/glslang/pull/2588

Taking a quick look at my local copy of our repo, it seems we're on an older glslang version.

So I guess we can fix this by updating the glslang submodule on our repo.

**Comment by gpx1000** [comment]
Thanks Sascha that does indeed fix it.

File: samples/performance/bindless/bindless.cpp (hunk 17 above)
Comment (gpx1000) [reply samples/performance/bindless/bindless.cpp]: this has been reworked

File: samples/performance/bindless/bindless.cpp (hunk 16 above)
Comment (gpx1000) [reply samples/performance/bindless/bindless.cpp]: I don't think there's a significant difference to the two in this particular case but will defer to your judgement if you think it's worth making the sync change.  It should be correct sync in either solution, I'm unaware of the performance or tutorial benefit of switching the two for a small sample like this?

**Comment by SaschaWillems** [comment]
The freeze culling option now works as expected 👍 

But it seems like the sample is leaking memory. I've been running for a few minutes to test everything, and noticed memory steadily increasing in VS (same for release/debug) mode with validation disabled:

image

I did some memory snapshots, and it seems like the app is leaking several objects on each frame:

image

Snapshot 1:
image

Snapshot 4:
image

**Comment by gpx1000** [comment]
Good catch @SaschaWillems !

gpu_cull was indeed leaking a command buffer.  When testing with LSAN, I'm still seeing 
SUMMARY: LeakSanitizer: 225403 byte(s) leaked in 791 allocation(s)

However, after spending the day tracking all of these down, each of those allocations are either false positives or buried deep in the framework and are incurred no matter what the sample does.  We should do a cleanup pass, but most of the significant leaks found look like false positives (i.e. storing shared_ptr inside a map structure accounts for several).
Do me a favor and retest to confirm that you're not seeing anymore odd memory leaks specific to this sample.  I'll try to find some time to sanitize the framework in a future PR.

**Comment by SaschaWillems** [comment]
Did a rerun, and as you noticed, most of those leaks are caused by the framework. As this is not the sample's fault and needs fixing in the framework I'm happy with the sample :)

P.S. : Sorry for closing, clicked the wrong button.

**Review by SaschaWillems — DISMISSED** [review]

**Review by gary-sweet — CHANGES_REQUESTED** [review]
I'm seeing a driver assert about accessing samplers out-of-range.

It would appear that the sample is not respecting maxPerStageDescriptorSamplers (64 in our case).

**Comment by cforfang** [comment]
> I'm seeing a driver assert about accessing samplers out-of-range.
> 
> It would appear that the sample is not respecting maxPerStageDescriptorSamplers (64 in our case).

Ah, good find. I suppose the fix is to enable update after bind, to get the guaranteed min limit of maxPerStageDescriptorUpdateAfterBindSamplers = 500000?

**Comment by gary-sweet** [comment]
> Ah, good find. I suppose the fix is to enable update after bind, to get the guaranteed min limit of maxPerStageDescriptorUpdateAfterBindSamplers = 500000?

Might be better to just test that maxPerStageDescriptorSamplers is large enough for this sample and flag it as unsupported if it isn't. If the sample doesn't strictly require update-after-bind it seems a shame to require it to workaround a limits issue.

**Comment by cforfang** [comment]
> Might be better to just test that maxPerStageDescriptorSamplers is large enough for this sample and flag it as unsupported if it isn't. If the sample doesn't strictly require update-after-bind it seems a shame to require it to workaround a limits issue.

Well, it _is_ a bindless sample :) The only way to get the guaranteed support is to actually use the new feature provided by descriptor_indexing -- so from that angle doing this makes sense to me.

**Comment by cforfang** [comment]
Actually I forgot this sample _doesn't_ actually use descriptor_indexing, which is potentially a bit confusing given the references to 'bindless'. But anyway that does maybe mean it makes less sense to add this as a new dependency just to be able to bind more textures/samplers. 

Maybe the way to go is to:
- Check this particular limit -- and any others that may apply. Maybe verify against the Vulkan min specs by using the devsim layer or similar?
- Check against shaderSampledImageArrayDynamicIndexing feature and any others that apply.

One could maybe make an argument that it could make sense to simple depend on descriptor_indexing being available, though. That would guarantee the implementation provides all needed functionality here.

**Comment by gary-sweet** [comment]
> One could maybe make an argument that it could make sense to simple depend on descriptor_indexing being available, though. That would guarantee the implementation provides all needed functionality here.

Personally, I'd prefer not to depend on descriptor indexing if possible (since we don't actually support that yet in our driver).

**Comment by JoseEmilio-ARM** [comment]
> > One could maybe make an argument that it could make sense to simple depend on descriptor_indexing being available, though. That would guarantee the implementation provides all needed functionality here.
> 
> Personally, I'd prefer not to depend on descriptor indexing if possible (since we don't actually support that yet in our driver).

I agree with @cforfang (an ARM colleague), for it to be 'bindless' (indexing textures dynamically at runtime) the sample would need to use the descriptor indexing extension.

@gpx1000 I think it would be more appropriate to rename this sample. 'Multi-draw indirect' would be less ambiguous than 'Bindless', what do yo think?

**Comment by gpx1000** [comment]
I'm okay with renaming the sample.  The fact that adding in descriptor indexing would limit the broader support holds a lot of weight with me when it comes to not wanting to add it.  We can rename it for accuracy, maintain the fact that it works in a broad range of devices, and be a very trivial stepping stone one would need to do to get further into bindless territory.
That path makes a lot of sense to me as it does provide a good example of how to do bindless with the benefits of bindless while being more descriptively named due to not implementing descriptor indexing.
What do you guys think?
I'm certainly open to adding in descriptor indexing and getting full bindless as an alternative path but not dropping support isn't a bad thing.

**Review by JoseEmilio-ARM — DISMISSED** [review]

**Comment by SaschaWillems** [comment]
Looking good, but I do get one validation layer error with SDK 1.2.189:

```
[error] [framework\core\instance.cpp:41] 1028204675 - VUID-VkShaderModuleCreateInfo-pCode-04147: Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-04147 ] Object 0: handle = 0x1c13c768e80, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x3d492883 | vkCreateShaderModule(): The SPIR-V Extension (SPV_EXT_physical_storage_buffer) was declared, but none of the requirements were met to use it. The Vulkan spec states: If pCode declares any of the SPIR-V extensions listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied (https://vulkan.lunarg.com/doc/view/1.2.189.0/windows/1.2-extensions/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-04147)
```

I think the sample needs to enable the `bufferDeviceAddress` feature of `VkPhysicalDeviceBufferDeviceAddressFeatures` to silence this.

But other than that, the sample works and looks great.

gpx1000:
Hi Sascha,
I think that error is caused by GLSL being an older version.  The one updated in this PR should resolve to SPV_EXT*.  If you can verify that if you're on the same GLSL that this PR has this error goes away it would be appreciated.

SaschaWillems:
Yeah, thanks for pointing me in that direction. Switching branches didn't seem to update the submodules, but after a manual update, the false validation errors are gone :)

SaschaWillems (dismissed):

mklefrancois (dismissed):
Looking good to me, runs correctly on an NVIDIA GeForce GTX 3060 Ti

SaschaWillems:
The sample behaves oddly for me when running a release build: It starts without any errors, but all I can see is the clear color (I have set it to a dark blue to verify this). This happens randomly, but ~ 8 out of 10 times. And it doesn't seem to properly resize when changing the window size:

image

The UI also doesn't show up, and there are no validation errors.

Sometimes I can get it to draw the scene by wildly resizing, pressing keys and making the window inactive and active again, but most of the time it just looks like that when run as a release build.

I haven't had much time to debug this, but I can imagine that the problem is being caused by uninitialized memory, though I wasn't able to pinpoint this. I am suspecting the `cpu_commands` count at draw time being zero but that is also the case in debug mode.

Btw. I find the naming of `cpu_commands` a bit misleading, as it's also the source of the draw commands when GPU culling is used. So maybe we could rename that one to make it a bit easier to follow?

gpx1000:
Hmm... @SaschaWillems, How'd you get it to behave like that?  I'm not seeing anything that would make an issue happen from Release mode.

SaschaWillems:
Just running it in release mode from Visual Studio. But I'm using a Vulkan developer driver, so it may not be the application's fault. If no one else sees this problem I'm fine with approving, cause otherwise the sample looks and works just fine.

SaschaWillems:
As discussed on Slack: This seems to be framework related problem that also happens with other samples. Will do a separate issue for that.

The sample itself looks good to me :)

SaschaWillems (dismissed):

gary-sweet (commented):
It correctly reports as not supported due to only having 64 samplers, but that's as much as I can test.

JoseEmilio-ARM (dismissed):

JoseEmilio-ARM (dismissed):

marty-johnson59:
Merging per discussion on Samples call - 3 prior reviews with some minor changes..

PR #334, opened 2021-08-28
Allow ray tracing samples to use the ImGui overlay

### LINKED ISSUE #331: Make ImGui overlay usable in ray tracing samples
As discussed on today's call, we want to make the ImGui overlay available to ray tracing samples so they can e.g. add option toggles. Right now this is not possible, as the ImGui overlay requires a render pass (which is not required for ray tracing). This is a tracking issue for documentation purposes, a PR to make this work will follow.

SaschaWillems:
## Description

This PR adds new functions and fixes to the api sample class to make the ImGui overlay usable in ray tracing samples for e.g. displaying information and options.

image

It also fixes a bug with changing swap chain image usage flags after it's initial creation, which wasn't visible up until now as no ray tracing sample made use of a render pass. The initial function to do that didn't recreate certain resources that were required by the render pass. This is now properly handled by the ```update_swapchain_image_usage_flags``` in the ```ApiVulkanSample``` class, so the code for this is slightly changed:

```cpp
// This sample copies the ray traced output to the swap chain image, so we need to enable the required image usage flags
const std::set<VkImageUsageFlagBits> image_usage_flags = {VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_IMAGE_USAGE_TRANSFER_DST_BIT};
update_swapchain_image_usage_flags(image_usage_flags);
```

In addition to the above , ray tracing samples that want to display the ImGui overlay need to do the following:

1. Update the render pass to "don't clear" in order to keep the ray traced output:

```cpp
update_render_pass_flags(RenderPassCreateFlags::ColorAttachmentLoad);
```

2. Add a render pass for the ImGui draw at the end of the command buffer:

```cpp
VkClearValue clear_values[2];
clear_values[0].color        = {{0.0f, 0.0f, 0.2f, 0.0f}};
clear_values[1].depthStencil = {0.0f, 0};

VkRenderPassBeginInfo render_pass_begin_info    = vkb::initializers::render_pass_begin_info();
render_pass_begin_info.renderPass               = render_pass;
render_pass_begin_info.framebuffer              = framebuffers[i];
render_pass_begin_info.renderArea.extent.width  = width;
render_pass_begin_info.renderArea.extent.height = height;
render_pass_begin_info.clearValueCount          = 2;
render_pass_begin_info.pClearValues             = clear_values;

vkCmdBeginRenderPass(draw_cmd_buffers[i], &render_pass_begin_info, VK_SUBPASS_CONTENTS_INLINE);
draw_ui(draw_cmd_buffers[i]);
vkCmdEndRenderPass(draw_cmd_buffers[i]);
```

I have updated both the ray tracing basic and ray tracing reflection sample to show the ImGui overlay using these changes, so future samples can simply copy the setup from these samples.

Fixes #331

## General Checklist:


framework/api_vulkan_sample.cpp
```diff
@@ -737,6 +750,91 @@ void ApiVulkanSample::setup_render_pass()
...
+	subpass_description.pResolveAttachments     = nullptr;
+
+	// Subpass dependencies for layout transitions
+	std::array<VkSubpassDependency, 2> dependencies;
+
+	dependencies[0].srcSubpass      = VK_SUBPASS_EXTERNAL;
+	dependencies[0].dstSubpass      = 0;
+	dependencies[0].srcStageMask    = VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT;
+	dependencies[0].dstStageMask    = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
+	dependencies[0].srcAccessMask   = VK_ACCESS_MEMORY_READ_BIT;
```
cforfang: Should this be WRITE_BIT instead, as the (following) render-pass reads will read what was written by the vkCmdCopyImage transfer -- i.e. previous access was a write?

samples/extensions/ray_tracing_reflection/ray_tracing_reflection.cpp
```diff
@@ -903,6 +903,26 @@ void RaytracingReflection::build_command_buffers()
 		                      VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL,
 		                      VK_IMAGE_LAYOUT_GENERAL,
 		                      subresource_range);
+
+		/*
+			Start a new render pass to draw the UI overlay on top of the ray traced image
```
cforfang, line 908: Worth adding a comment which says dependencies/barriers/layout-transitions are handled by sub-pass dependencies? Else it might look like a pipeline-barrier is missing. (If doing this, best to change `raytracing_basic` as well.)

framework/api_vulkan_sample.cpp
SaschaWillems: Can you please open an issue for that? I have copied this from the default render pass setup, and this is not the scope of this PR.

samples/extensions/ray_tracing_reflection/ray_tracing_reflection.cpp
SaschaWillems: I'm not sure if such a comment would make sense in that place. If so, we would need to add it in all samples where implicit transitions are used.

cforfang: Fair enough; I wasn't sure if this is something the samples are already explicit about or not :)

framework/api_vulkan_sample.cpp
SaschaWillems: That's true. The WRITE_BIT makes sense in here. Thanks for noticing.

gpx1000 (approved):
Works well here. LGTM

TomAtkinsonArm (approved):
LGTM

mklefrancois:
Look all good to me

mklefrancois (approved):
This is all fine

PR #357, opened 2021-09-20
Introduce new Vulkan-Hpp-based sample, transcoded from dynamic_uniform_buffers.

asuessenbach:
## Description

Introduces a new Vulkan-Hpp-based sample, which is a transcoded version of the dynamic_uniform_buffers sample.
Additionally, adds classes vkb::core::HPPBuffer, vkb::core::HPPDevice, and framework::hpp_api_vulkan_sample, as wrappers around vkb::core::Buffer, vkb::Device, and framework::api_vulkan_sample, respectively, providing a Vulkan-Hpp-based interface to those classes.
Build tested on Windows.

## General Checklist:

## Sample Checklist

### DIFF
--- a/framework/core/hpp_buffer.cpp
+++ b/framework/core/hpp_buffer.cpp
@@ -1,0 +1,58 @@
+/* Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 the "License";
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <core/hpp_buffer.h>
+
+namespace vkb
+{
+namespace core
+{
+HPPBuffer::HPPBuffer(vkb::core::HPPDevice &   device,
+                     vk::DeviceSize           size,
+                     vk::BufferUsageFlags     buffer_usage,
+                     VmaMemoryUsage           memory_usage,
+                     VmaAllocationCreateFlags flags) :
+    Buffer(device, static_cast<VkDeviceSize>(size), static_cast<VkBufferUsageFlags>(buffer_usage), memory_usage, flags)
+{}
+
+void HPPBuffer::flush() const
+{
+	vkb::core::Buffer::flush();
+}
+
+vk::Buffer HPPBuffer::get_handle() const
+{
+	return static_cast<vk::Buffer>(vkb::core::Buffer::get_handle());
+}
+
+vk::DeviceSize HPPBuffer::get_size() const
+{
+	return static_cast<vk::DeviceSize>(vkb::core::Buffer::get_size());
+}
+
+void HPPBuffer::update(uint8_t const *data, size_t size, size_t offset)
+{
+	vkb::core::Buffer::update(data, size, offset);
+}
+
+void HPPBuffer::update(void *data, size_t size, size_t offset)
+{
+	vkb::core::Buffer::update(data, size, offset);
+}
+
+}        // namespace core
+}        // namespace vkb
--- a/framework/core/hpp_buffer.cpp
+++ b/framework/core/hpp_buffer.cpp
@@ -26,7 +26,11 @@ HPPBuffer::HPPBuffer(vkb::core::HPPDevice &   device,
                      vk::BufferUsageFlags     buffer_usage,
                      VmaMemoryUsage           memory_usage,
                      VmaAllocationCreateFlags flags) :
-    Buffer(device, static_cast<VkDeviceSize>(size), static_cast<VkBufferUsageFlags>(buffer_usage), memory_usage, flags)
+    Buffer(*reinterpret_cast<vkb::Device *>(&device),
+           static_cast<VkDeviceSize>(size),
+           static_cast<VkBufferUsageFlags>(buffer_usage),
+           memory_usage,
+           flags)
 {}
 
 void HPPBuffer::flush() const
--- a/framework/core/hpp_buffer.cpp
+++ b/framework/core/hpp_buffer.cpp
@@ -26,7 +26,7 @@ HPPBuffer::HPPBuffer(vkb::core::HPPDevice &   device,
                      vk::BufferUsageFlags     buffer_usage,
                      VmaMemoryUsage           memory_usage,
                      VmaAllocationCreateFlags flags) :
-    Buffer(device, static_cast<VkDeviceSize>(size), static_cast<VkBufferUsageFlags>(buffer_usage), memory_usage, flags)
+    Buffer(reinterpret_cast<vkb::Device &>(device), static_cast<VkDeviceSize>(size), static_cast<VkBufferUsageFlags>(buffer_usage), memory_usage, flags)
 {}
 
 void HPPBuffer::flush() const
--- a/framework/core/hpp_buffer.cpp
+++ b/framework/core/hpp_buffer.cpp
@@ -26,7 +26,7 @@ HPPBuffer::HPPBuffer(vkb::core::HPPDevice &   device,
                      vk::BufferUsageFlags     buffer_usage,
                      VmaMemoryUsage           memory_usage,
                      VmaAllocationCreateFlags flags) :
-    Buffer(reinterpret_cast<vkb::Device &>(device), static_cast<VkDeviceSize>(size), static_cast<VkBufferUsageFlags>(buffer_usage), memory_usage, flags)
+    Buffer(device, static_cast<VkDeviceSize>(size), static_cast<VkBufferUsageFlags>(buffer_usage), memory_usage, flags)
 {}
 
 void HPPBuffer::flush() const
--- a/framework/core/hpp_device.cpp
+++ b/framework/core/hpp_device.cpp
@@ -25,5 +25,10 @@ vk::Device HPPDevice::get_handle() const
 {
 	return static_cast<vk::Device>(Device::get_handle());
 }
+
+HPPDevice::operator vkb::Device &()
+{
+	return *this;
+}
 }        // namespace core
 }        // namespace vkb
--- a/framework/core/hpp_device.cpp
+++ b/framework/core/hpp_device.cpp
@@ -1,0 +1,29 @@
+/* Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 the "License";
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <core/hpp_device.h>
+
+namespace vkb
+{
+namespace core
+{
+vk::Device HPPDevice::get_handle() const
+{
+	return static_cast<vk::Device>(Device::get_handle());
+}
+}        // namespace core
+}        // namespace vkb
--- a/framework/core/hpp_device.cpp
+++ b/framework/core/hpp_device.cpp
@@ -21,9 +21,5 @@ namespace vkb
 {
 namespace core
 {
-vk::Device HPPDevice::get_handle() const
-{
-	return static_cast<vk::Device>(Device::get_handle());
-}
 }        // namespace core
 }        // namespace vkb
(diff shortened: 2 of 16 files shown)


asuessenbach:
Unfortunately, I get two failing checks here, but I can't get what to do to get them resolved.
Any help is very much appreciated!

SaschaWillems:
Both failures look like problems with the build system. The Android build is currently broken, and will be fixed by #348.

The copyright header also fails in an odd way before actually getting to the files. I get the same error when trying to run it locally. Maybe @TomAtkinsonArm can help with this.

TomAtkinsonArm:
image

Copyright check is failing because of the characters used in your surname. Removing these the check continues. Which codec are these characters from? `io.open(filename, "r+", encoding="utf-8") ` changing `utf-8` to a different codec should fix this issue https://docs.python.org/3/library/codecs.html

gpx1000:
Something weird is going on.  I'm getting a frozen screen in a debug build.  I first thought it was validation layers causing it but turn those off and it's still stuck.  Release build works fine here.  Could someone see if they're seeing the same thing?

asuessenbach:
> image
> 
> Copyright check is failing because of the characters used in your surname. Removing these the check continues. Which codec are these characters from? `io.open(filename, "r+", encoding="utf-8") ` changing `utf-8` to a different codec should fix this issue https://docs.python.org/3/library/codecs.html

Ah yes, that's silly. I should not use those characters in source code!
But, in fact, I'm not sure what I should use as the Copyright owner here (you know: I'm not a lawyer): is it me? Or NVIDIA? Or Khronos? Or something completely different?

asuessenbach:
> Something weird is going on. I'm getting a frozen screen in a debug build. I first thought it was validation layers causing it but turn those off and it's still stuck. Release build works fine here. Could someone see if they're seeing the same thing?

I don't see any issues, running on Win10 in debug mode.
The release build, though, seems to get frozen here. Switching focus to the command window and back to the render window makes it run as expected! This happens with both this hpp-based version and the original C-based version (dynamic_uniform_buffers), which makes me wonder if the reason for this behavior is outside of this sample.

TomAtkinsonArm:
> But, in fact, I'm not sure what I should use as the Copyright owner here (you know: I'm not a lawyer): is it me? Or NVIDIA? Or Khronos? Or something completely different?

Yep, it can be very confusing at times!! The updated copyright looks fine to me. Here it should be which entity is submitting the contribution. For example, if the work is carried out and endorsed by Nvidia, then Nvidia is the owner of the contribution. If you are contributing as an independent (like Sascha) then the copyright owner will be yourself.

SaschaWillems (dismissed):
LGTM. Thank you very much for your contribution :)

I only have a very small suggestion: It's now possible to add shaders used by a sample to the project file, so they can be opened directly from the IDE. To do this, you add them via `SHADER_FILES_GLSL` in your samples CMakeLists. It would then look like this:

```
add_sample_with_tags(
    ID ${FOLDER_NAME}
    CATEGORY ${CATEGORY_NAME}
    AUTHOR "Sascha Willems"
    NAME "HPP Dynamic uniform buffers"
    DESCRIPTION "Demonstrates the use of dynamic offsets into one single uniform buffers for rendering multiple objects, using vulkan.hpp"
    SHADER_FILES_GLSL
        "dynamic_uniform_buffers/base.vert"
        "dynamic_uniform_buffers/base.frag")
```

asuessenbach:
@SaschaWillems That's a great improvement! Added the SHADER_FILES_GLSL to the CMakeLists.txt in this PR.

SaschaWillems (dismissed):

framework/core/hpp_buffer.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+HPPBuffer::HPPBuffer(vkb::core::HPPDevice &   device,
+                     vk::DeviceSize           size,
+                     vk::BufferUsageFlags     buffer_usage,
+                     VmaMemoryUsage           memory_usage,
+                     VmaAllocationCreateFlags flags) :
+    Buffer(device, static_cast<VkDeviceSize>(size), static_cast<VkBufferUsageFlags>(buffer_usage), memory_usage, flags)
+{}
+
+void HPPBuffer::flush() const
```
gpx1000, line 36: hides the super class flush function and that function is not marked virtual.  (All functions in this file suffer from the same problem)

framework/core/hpp_device.cpp
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+#include <core/hpp_device.h>
+
+namespace vkb
+{
+namespace core
+{
+vk::Device HPPDevice::get_handle() const
```
gpx1000: vkb::core::device is not a virtual class, need to add virtual to be able to override it.

framework/hpp_api_vulkan_sample.h
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * @brief wrapper class for use with vulkan.hpp in the samples of this framework
+ *
+ * See vkb::VulkanSample for documentation
+ */
+class HPPApiVulkanSample : public ApiVulkanSample
+{
+  public:
+	virtual bool prepare(vkb::Platform &platform) override;
```
gpx1000: no need for virtual AND override; use just override is correct.

framework/hpp_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,96 @@
...
+bool HPPApiVulkanSample::prepare(vkb::Platform &platform)
+{
+	if (!ApiVulkanSample::prepare(platform))
+	{
+		return false;
+	}
+
+	// initialize function pointers for C++-bindings
+	static vk::DynamicLoader  dl;
+	PFN_vkGetInstanceProcAddr vkGetInstanceProcAddr =
```
gpx1000: two things here: vkGetInstanceProcAddr shadows a previous global declaration.  Also when the type is specified in the template, no need to use anything other than auto.

framework/hpp_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+size_t HPPApiVulkanSample::get_command_buffers_count() const
+{
+	return draw_cmd_buffers.size();
+}
+
+vkb::core::HPPDevice &HPPApiVulkanSample::get_device() const
+{
+	return static_cast<vkb::core::HPPDevice &>(*device);
```
gpx1000: when down casting it's better to use dynamic_cast instead of static_cast.

samples/api/hpp_dynamic_uniform_buffers/hpp_dynamic_uniform_buffers.cpp
```diff
@@ -0,0 +1,480 @@
...
+ * contains all matrices for the objects in the scene.
+ *
+ * The used descriptor type vk::DescriptorType::eUniformBufferDynamic then allows to set a dynamic
+ * offset used to pass data from the single uniform buffer to the connected shader binding point.
+ */
+
+#include <benchmark_mode/benchmark_mode.h>
+#include <hpp_dynamic_uniform_buffers.h>
+
+HPPDynamicUniformBuffers::HPPDynamicUniformBuffers()
```
gpx1000, line 32: doesn't initialize index_count or dynamic_alignment member variables.

samples/api/hpp_dynamic_uniform_buffers/hpp_dynamic_uniform_buffers.cpp
```diff
@@ -0,0 +1,480 @@
...
+}
+
+void HPPDynamicUniformBuffers::aligned_free(void *data)
+{
+#if defined(_MSC_VER) || defined(__MINGW32__)
+	_aligned_free(data);
+#else
+	free(data);
+#endif
+}
```
gpx1000, line 77: I know aligned_alloc/aligned_free is copied from the already existing sample.  However, it would be nice to do something a bit more cross platform.  Unfortunately Windows has problems with the STL aligned_alloc so it's probably better to use aligned_storage and aligned.
Fixing this isn't a requirement to do in this sample, but if you're willing, it would be nice.
in any case, both of these functions can be made static.

samples/api/hpp_dynamic_uniform_buffers/hpp_dynamic_uniform_buffers.cpp
```diff
@@ -0,0 +1,480 @@
...
+{
+	// Update at max. 60 fps
+	animation_timer += delta_time;
+	if ((animation_timer + 0.0025 < (1.0f / 60.0f)) && (!force))
+	{
+		return;
+	}
+
+	// Dynamic ubo with per-object model matrices indexed by offsets in the command buffer
+	uint32_t  dim = static_cast<uint32_t>(pow(OBJECT_INSTANCES, (1.0f / 3.0f)));
```
gpx1000: can switch to auto as the specification of the cast gives the type.

samples/api/hpp_dynamic_uniform_buffers/hpp_dynamic_uniform_buffers.cpp
```diff
@@ -0,0 +1,480 @@
...
+	for (uint32_t x = 0; x < dim; x++)
+	{
+		for (uint32_t y = 0; y < dim; y++)
+		{
+			for (uint32_t z = 0; z < dim; z++)
+			{
+				uint32_t index = x * dim * dim + y * dim + z;
+
+				// Aligned offset
+				glm::mat4 *model_mat = (glm::mat4 *) (((uint64_t) ubo_data_dynamic.model + (index * dynamic_alignment)));
```
gpx1000: can use auto as the specification is giving in the cast.

samples/api/hpp_dynamic_uniform_buffers/hpp_dynamic_uniform_buffers.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+				// Update rotations
+				rotations[index] += animation_timer * rotation_speeds[index];
+
+				// Update matrices
+				glm::vec3 pos = glm::vec3(-((dim * offset.x) / 2.0f) + offset.x / 2.0f + x * offset.x,
+				                          -((dim * offset.y) / 2.0f) + offset.y / 2.0f + y * offset.y,
+				                          -((dim * offset.z) / 2.0f) + offset.z / 2.0f + z * offset.z);
```
gpx1000: dim and x y z should be cast to float from their current type of uint32_t

gpx1000 (changes requested):
The only changes in the list that I would highly recommend is the virtual on base class changes.  Too many bugs can crop up from overriding member functions implicitly.

Also, for some reason, debug started working when I grabbed your latest.  If you could, please do a normal commit/push instead of force push or modifying a previous commit, git wasn't updating for me and you can see that the history only has one commit for this PR rather than the actual history.

framework/core/hpp_buffer.cpp
asuessenbach: HPPBuffer is derived protectively from vkb::core::Buffer. That is, technically, the functions in HPPBuffer do not hide those of vkb::core::Buffer, since they are not part of HPPBuffer's public interface.
To be more explicit, I could change that to private inheritance or even make vkb::core::Buffer a private member of HPPBuffer, making it an implementation-only detail.

framework/hpp_api_vulkan_sample.cpp
asuessenbach: Only if vkb::Device was polymorphic.

framework/core/hpp_device.cpp
asuessenbach: Similar to HPPBuffer/vkb::core::Buffer, HPPDevice (now) derives protectively from vkb::Device. So no need to make vkb::Device virtual, as HPPDevice does not override anything from it.

samples/api/hpp_dynamic_uniform_buffers/hpp_dynamic_uniform_buffers.cpp
asuessenbach: I've made those two functions static.
But what do you mean with 

> it would be nice to do something a bit more cross platform.

How should that be "fixed"?

asuessenbach:
Regarding all the changes on code that was a direct copy from the dynamic_uniform_buffers sample: should that be changed accordingly as well?

samples/api/hpp_dynamic_uniform_buffers/hpp_dynamic_uniform_buffers.cpp
gpx1000: an ideal fix in my opinion would be std::aligned_storage / std::aligned use such that there's no need for macro platform separated code.
Here's a quick search on SO:
https://stackoverflow.com/questions/32133203/what-can-i-use-instead-of-stdaligned-alloc-in-ms-visual-studio-2013

gpx1000:
> Regarding all the changes on code that was a direct copy from the dynamic_uniform_buffers sample: should that be changed accordingly as well?

I think it'd be great thing to do if you're willing.  However, not required.  Thanks again for the work!

samples/api/hpp_dynamic_uniform_buffers/hpp_dynamic_uniform_buffers.cpp
asuessenbach: I think, I'll leave that for others to handle.

framework/core/hpp_device.cpp
asuessenbach: I've changed HPPDevice to its bar minimum now, making it a wrapper class that simply hides all the vkb::Device stuff.

gary-sweet (approved):

gpx1000 (approved):

SaschaWillems (approved):

marty-johnson59:
Merging on 3 approvals

PR #400, opened 2021-11-23
Explicitily set window focus

### LINKED ISSUE #399: Some samples won't properly update/render after starting in release mode
While testing the samples in release mode, I noticed some of the samples no longer properly rendering after starting them. E.g. the AFBC sample shows the first frame, and upon resize it's not updated and does not show the expected animation:

image
(white border = initial window size after startup, black border appears in the resized area of the window)

Resizing the window multiple times make the sample work as expected.

This is on Windows 10 and happens on two completely different machines .One with an Nvidia RTX 2060 with the public 472.69 Vulkan Developer drivers and one with an Intel UHD 630 iGPU.

This does only seem to happen in release builds and mostly with samples that load more complex assets. So this could be a timing issue or may be caused by not properly initialized variables.

TomAtkinsonArm:
## Description

Explicitly set window focus to true so that the application begins rendering on start.

Currently the first frame renders but rendering is then halted until the window is perceived as focused again. For GLFW this takes multiple resizes or min/maximizing the window. Setting focus to true to begin with resolves this issue for me

Fixes #399 

## General Checklist:

## Sample Checklist

### DIFF
--- a/framework/api_vulkan_sample.h
+++ b/framework/api_vulkan_sample.h
@@ -413,8 +413,6 @@ class ApiVulkanSample : public vkb::VulkanSample
 		bool middle = false;
 	} mouse_buttons;
 
-	// true if application has focused, false if moved to background
-	bool focused = false;
 	struct TouchPos
 	{
 		int32_t x;
--- a/framework/platform/platform.h
+++ b/framework/platform/platform.h
@@ -175,7 +175,7 @@ class Platform
 	bool               fixed_simulation_fps{false};    /* Delta time should be fixed with a fabricated value */
 	float              simulation_frame_time = 0.016f; /* A fabricated delta time */
 	bool               process_input_events{true};     /* App should continue processing input events */
-	bool               focused;                        /* App is currently in focus at an operating system level */
+	bool               focused{true};                  /* App is currently in focus at an operating system level */
 	bool               close_requested{false};         /* Close requested */
 
   private:


TomAtkinsonArm:
@SaschaWillems can you check if this resolves the initial issue?

If it works it may be a better fix to force GLFW to focus the window on creation

SaschaWillems:
This does indeed fix my issue. I can no longer reproduce the bug mentioned in #399 :)

gpx1000 (approved):

SaschaWillems (approved):

MattGuerrette:
Unfortunately this will not solve situations where the application will lose focus explicitly on launch. For example, when running these samples using the [Radeon Developer Panel](https://github.com/GPUOpen-Tools/radeon_developer_panel), the samples cannot be captured unless the sample app regains focus as the panel will focus itself to the top level when a client application (Vulkan app or DX12 app) connects, stealing the focus from the samples.

Could there be a build option or runtime option implemented that removes entirely the check for application focus?

TomAtkinsonArm:
@MattGuerrette I agree, the current implementation doesn't work well with what the samples is trying to achieve. Especially when considering the developer tools use case.

If I recall correctly, stopping rendering when loosing focus was implemented so that in demos phones has a chance to cool down in between long runs without having to close the app. Closing the app would mean we need to wait for assets to re load etc.

That said, it makes sense to allow the user to decide what "optimizations" should be used. I prefer runtime options over compile time options especially when it is something as small as this. Let me see if I can add a quick plugin to toggle this

TomAtkinsonArm:
Interesting, I also needed to bump Volk and Vulkan headers to 1.2.198 to build on windows on my intel laptop... This has not been caught in the CI

pdaniell-nv:
I've noticed this same problem with the latest Khronos sample code. Any chance we can get this fix merged soon?

gary-sweet (approved):
Looks fine to me, but I note that there are still some unaddressed comments.

PR #403, opened 2021-11-29
Add sample readme files

marty-johnson59:
## Description

This PR adds a README.md file to each sample (currently just copied from the README descriptions on the main samples folder) and updates the CONTRIBUTING.md file to include adding a README.md file to future samples folders.

Fixes #<issue number>

## General Checklist:


SaschaWillems:
Looking good 👍 

One thing we should check though, is if a sample has a readme that's not named readme.md and then maybe just rename that. See e.g. the buffer device address sample at https://github.com/KhronosGroup/Vulkan-Samples/tree/4af5a031c5078c64739914413641b7a4c1bd0251/samples/extensions/buffer_device_address

It already has a tutorial in markdown format, but the tutorial is called `buffer_device_address_tutorial.md`, so it won't show up as the default readme in that folder. I think it may be better to just rename that file instead of adding a stub readme.

gpx1000:
Might make sense to add a CICD automated task to ensure that a 'readme.md' file is found in each sample directory.

marty-johnson59:
@SaschaWillems, wondering if tutorial.md files should just be referenced by the new README.md files (I tweaked the buffer_device_address to do this)?  There were several other samples with tutorials that I treated that way (missed this one tho).  LMK which works better - I'm fine with either

marty-johnson59:
@gpx1000, good idea.  Beyond my current skill set tho:(.  Where do I add that?

TomAtkinsonArm:
> Might make sense to add a CICD automated task to ensure that a 'readme.md' file is found in each sample directory

I'm wondering what Unix commands can be used for this? It is probably easier to write a quick python script for this?

gpx1000:
https://github.com/KhronosGroup/Vulkan-Samples/blob/master/.github/workflows/check.yml

The basic idea is it's a software defined workflow, which means that the jobs in CI/CD get defined by the software, but also get updated/created etc by the software.  In github, these jobs are typically found in a common location the hidden folder .github contains them.  Workflows contains the definitions of the workflows themselves.  check.yml for this project contains the non-build checks (i.e. it doesn't need to create a build to ensure that it works).

If you look in the above, other jobs, you'll notice that mostly it's a linux system running which calls into python3 some scripts we have also in the source code.  However, don't have to be python that runs, just a program that fails on problems thankfully Linux has such programs.  NB: the path to the workspace is in the environment variable: GITHUB_WORKSPACE You could use bash for cross platform examples like this:
[ ! -f $GITHUB_WORKSPACE/foo ] && echo "$FILE does not exist."

However, the problem with this simplistic solution is it wont auto loop through the samples to find the right thing, so how about a simple bash script that will:
the_file=README.md
for file in $GITHUB_WORKSPACE/samples/{api,extensions,performance}/"$the_file"; do 
   if [[ -e $file ]]; then
      printf '%s found in %s:\n' "$the_file" "${file%/*}"
   else
     exit 1
   fi
done

Note this was off the top of my head so very much not tested ;-)

gpx1000:
The only bad thing about another python script is that docker file build was such a sore spot of how long it takes.  Simple bash change means no new docker build required ;-)

TomAtkinsonArm:
> The only bad thing about another python script is that docker file build was such a sore spot of how long it takes. Simple bash change means no new docker build required ;-)

Very true, I never got round to finishing my docker improvements, I think they are on my fork somewhere but haven't had chance to get back to them

SaschaWillems:
@marty-johnson59 I think renaming the existing tutorial to README.md is the better option. Esp. in regards to generating documentation or a github page, this would directly show the actual tutorial instead of having the views go through a readme stub to find the actual tutorial.

marty-johnson59:
Hi @SaschaWillems, before I go too far down this path, note that renaming all the tutorials to README.md will break the links in the main samples directory README file.  I can go back and change all those to point to the appropriate README files instead, but wanted to make sure this is the desired path first...is there value in retaining the existing tutorial file names?  Are there any other places that link to those tutorials files?  Thx

SaschaWillems:
I don't think there is any value in retaining the existing file names. Renaming the links should be fine.

gpx1000 (approved):

SaschaWillems (approved):

PR #410, opened 2021-12-07
Change vkb::HPPVulkanSample from a facade over vkb::VulkanSample to a self-contained class using vulkan.hpp

asuessenbach:
## Description

Replacing this part of the framework shows how to manage some more resources using vulkan.hpp.

In order to do so, a couple of new facade classes and functions are introduced in `common/hpp_strings.h`, `common/hpp_utils.h`, `core/hpp_command_buffer.h`, `core/hpp_image_view.h`, `hpp_gui.h`, `platform/hpp_application.h`, `platform/hpp_window.h`, `rendering/hpp_render_frame.h`, `rendering/hpp_render_pipeline.h`, `rendering/hpp_render_target.h`, and `stats/hpp_stats.h`

The vulkan.hpp-based classes and facades in `common/hpp_vk_common.h`, `core/hpp_device.h`, `core/hpp_instance.h`, `core/hpp_physical_device.h`, `core/hpp_swapchain.h`, `hpp_api_vulkan_sample.[cpp|h]`, `hpp_gltf_loader.h`, `platform/hpp_platform.h`, `rendering/hpp_render_context.h`, `api/hpp_dynamic_uniform_buffers/hpp_dynamic_uniform_buffers.[cpp|h]`, `api/hpp_hello_triangle/hpp_hello_triangle.[cpp|h]`, and `api/hpp_texture_loading/hpp_texture_loading.[cpp|h]` needed some adjustments and additions.

In addition to that, some consts needed to be spread over the framework in `core/command_buffer.[cpp|h]` and `platform/platform.[cpp|h]`.

And, finally, the dependency of vkb::Window and its derivatives on vulkan.hpp is removed.

I have tested it on Window10 on nvidia HW.

## General Checklist:

## Sample Checklist

### DIFF
--- a/framework/common/hpp_strings.h
+++ b/framework/common/hpp_strings.h
@@ -25,7 +25,10 @@ namespace vkb
 {
 namespace common
 {
-const std::string to_string(vk::Extent2D const &extent)
+/**
+ * @brief facade helper functions around the functions in common/strings.h, providing a vulkan.hpp-based interface
+ */
+std::string to_string(vk::Extent2D const &extent)
 {
 	return vkb::to_string(static_cast<VkExtent2D const &>(extent));
 }
--- a/framework/common/hpp_strings.h
+++ b/framework/common/hpp_strings.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+/* Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
--- a/framework/common/hpp_strings.h
+++ b/framework/common/hpp_strings.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2021, Arm Limited and Contributors
+/* Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -15,38 +15,19 @@
  * limitations under the License.
  */
 
-#include "headless_window.h"
+#pragma once
+
+#include <common/strings.h>
+
+#include <vulkan/vulkan.hpp>
 
 namespace vkb
 {
-HeadlessWindow::HeadlessWindow(const Window::Properties &properties) :
-    Window(properties)
+namespace common
 {
-}
-
-VkSurfaceKHR HeadlessWindow::create_surface(Instance &instance)
+const std::string to_string(vk::Extent2D const &extent)
 {
-	return VK_NULL_HANDLE;
+	return vkb::to_string(static_cast<VkExtent2D const &>(extent));
 }
-
-VkSurfaceKHR HeadlessWindow::create_surface(VkInstance, VkPhysicalDevice)
-{
-	return VK_NULL_HANDLE;
-}
-
-bool HeadlessWindow::should_close()
-{
-	return closed;
-}
-
-void HeadlessWindow::close()
-{
-	closed = true;
-}
-
-float HeadlessWindow::get_dpi_factor() const
-{
-	// This factor is used for scaling UI elements, so return 1.0f (1 x n = n)
-	return 1.0f;
-}
+}        // namespace common
 }        // namespace vkb
--- a/framework/common/hpp_utils.h
+++ b/framework/common/hpp_utils.h
@@ -21,6 +21,9 @@
 
 #include <rendering/hpp_render_context.h>
 
+/**
+ * @brief facade helper functions around the functions in common/utils.h, providing a vulkan.hpp-based interface
+ */
 namespace vkb
 {
 namespace common
--- a/framework/common/hpp_utils.h
+++ b/framework/common/hpp_utils.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+/* Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
--- a/framework/common/hpp_utils.h
+++ b/framework/common/hpp_utils.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2021, Arm Limited and Contributors
+/* Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -15,38 +15,28 @@
  * limitations under the License.
  */
 
-#include "headless_window.h"
+#pragma once
+
+#include <common/utils.h>
+
+#include <rendering/hpp_render_context.h>
 
 namespace vkb
 {
-HeadlessWindow::HeadlessWindow(const Window::Properties &properties) :
-    Window(properties)
+namespace common
 {
+void screenshot(vkb::rendering::HPPRenderContext &render_context, const std::string &filename)
+{
+	vkb::screenshot(reinterpret_cast<vkb::RenderContext &>(render_context), filename);
 }
 
-VkSurfaceKHR HeadlessWindow::create_surface(Instance &instance)
+namespace graphs
 {
-	return VK_NULL_HANDLE;
+bool generate_all(vkb::rendering::HPPRenderContext &context, sg::Scene &scene)
+{
+	return vkb::graphs::generate_all(reinterpret_cast<vkb::RenderContext &>(context), scene);
 }
+}        // namespace graphs
 
-VkSurfaceKHR HeadlessWindow::create_surface(VkInstance, VkPhysicalDevice)
-{
-	return VK_NULL_HANDLE;
-}
-
-bool HeadlessWindow::should_close()
-{
-	return closed;
-}
-
-void HeadlessWindow::close()
-{
-	closed = true;
-}
-
-float HeadlessWindow::get_dpi_factor() const
-{
-	// This factor is used for scaling UI elements, so return 1.0f (1 x n = n)
-	return 1.0f;
-}
+}        // namespace common
 }        // namespace vkb
--- a/framework/common/hpp_vk_common.h
+++ b/framework/common/hpp_vk_common.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+/* Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
--- a/framework/common/hpp_vk_common.h
+++ b/framework/common/hpp_vk_common.h
@@ -25,7 +25,7 @@ namespace vkb
 namespace common
 {
 /**
- * @brief facade helper functions and structs around the functions and structs in common/vk_commmon, providing a vulkan.hpp-based interface
+ * @brief facade helper functions and structs around the functions and structs in common/vk_common, providing a vulkan.hpp-based interface
  */
 struct HPPImageMemoryBarrier
 {
--- a/framework/common/hpp_vk_common.h
+++ b/framework/common/hpp_vk_common.h
@@ -25,69 +25,63 @@ namespace vkb
 namespace common
 {
 /**
- * @brief facade helper functions around the functions in common/vk_commmon, providing a vulkan.hpp-based interface
+ * @brief facade helper functions and structs around the functions and structs in common/vk_commmon, providing a vulkan.hpp-based interface
  */
-vk::Format get_suitable_depth_format(vk::PhysicalDevice             physical_device,
-                                     bool                           depth_only                 = false,
-                                     const std::vector<vk::Format> &depth_format_priority_list = {
-                                         vk::Format::eD32Sfloat, vk::Format::eD24UnormS8Uint, vk::Format::eD16Unorm})
+struct HPPImageMemoryBarrier
 {
-	std::vector<VkFormat> dfpl;
-	dfpl.reserve(depth_format_priority_list.size());
-	for (auto const &format : depth_format_priority_list)
-	{
-		dfpl.push_back(static_cast<VkFormat>(format));
-	}
-	return static_cast<vk::Format>(vkb::get_suitable_depth_format(physical_device, depth_only, dfpl));
+	vk::PipelineStageFlags src_stage_mask{vk::PipelineStageFlagBits::eBottomOfPipe};
+
+	vk::PipelineStageFlags dst_stage_mask{vk::PipelineStageFlagBits::eTopOfPipe};
+
+	vk::AccessFlags src_access_mask{};
+
+	vk::AccessFlags dst_access_mask{};
+
+	vk::ImageLayout old_layout{vk::ImageLayout::eUndefined};
+
+	vk::ImageLayout new_layout{vk::ImageLayout::eUndefined};
+
+	uint32_t old_queue_family{VK_QUEUE_FAMILY_IGNORED};
+
+	uint32_t new_queue_family{VK_QUEUE_FAMILY_IGNORED};
+};
+
+inline int32_t get_bits_per_pixel(vk::Format format)
+{
+	return vkb::get_bits_per_pixel(static_cast<VkFormat>(format));
 }
 
-bool is_depth_only_format(vk::Format format)
+inline vk::Format get_suitable_depth_format(vk::PhysicalDevice             physical_device,
+                                            bool                           depth_only                 = false,
+                                            const std::vector<vk::Format> &depth_format_priority_list = {
+                                                vk::Format::eD32Sfloat, vk::Format::eD24UnormS8Uint, vk::Format::eD16Unorm})
 {
-	return format == vk::Format::eD16Unorm || format == vk::Format::eD32Sfloat;
+	return static_cast<vk::Format>(
+	    vkb::get_suitable_depth_format(physical_device, depth_only, reinterpret_cast<std::vector<VkFormat> const &>(depth_format_priority_list)));
 }
 
-bool is_depth_stencil_format(vk::Format format)
+inline bool is_depth_only_format(vk::Format format)
 {
-	return format == vk::Format::eD16UnormS8Uint || format == vk::Format::eD24UnormS8Uint || format == vk::Format::eD32SfloatS8Uint ||
-	       is_depth_only_format(format);
+	return vkb::is_depth_only_format(static_cast<VkFormat>(format));
 }
 
-vk::ShaderModule load_shader(const std::string &filename, vk::Device device, vk::ShaderStageFlagBits stage)
+inline bool is_depth_stencil_format(vk::Format format)
+{
+	return vkb::is_depth_stencil_format(static_cast<VkFormat>(format));
+}
+
+inline vk::ShaderModule load_shader(const std::string &filename, vk::Device device, vk::ShaderStageFlagBits stage)
 {
 	return vkb::load_shader(filename, device, static_cast<VkShaderStageFlagBits>(stage));
 }
 
-vk::ImageLayout map_descriptor_type_to_image_layout(vk::DescriptorType descriptor_type, vk::Format format)
-{
-	switch (descriptor_type)
-	{
-		case vk::DescriptorType::eCombinedImageSampler:
-		case vk::DescriptorType::eInputAttachment:
-			if (is_depth_stencil_format(format))
-			{
-				return vk::ImageLayout::eDepthStencilReadOnlyOptimal;
-			}
-			else
-			{
-				return vk::ImageLayout::eShaderReadOnlyOptimal;
-			}
-			break;
-		case vk::DescriptorType::eStorageImage:
-			return vk::ImageLayout::eGeneral;
-			break;
-		default:
-			return vk::ImageLayout::eUndefined;
-			break;
-	}
-}
-
-void set_image_layout(vk::CommandBuffer         command_buffer,
-                      vk::Image                 image,
-                      vk::ImageLayout           old_layout,
-                      vk::ImageLayout           new_layout,
-                      vk::ImageSubresourceRange subresource_range,
-                      vk::PipelineStageFlags    src_mask = vk::PipelineStageFlagBits::eAllCommands,
-                      vk::PipelineStageFlags    dst_mask = vk::PipelineStageFlagBits::eAllCommands)
+inline void set_image_layout(vk::CommandBuffer         command_buffer,
+                             vk::Image                 image,
+                             vk::ImageLayout           old_layout,
+                             vk::ImageLayout           new_layout,
+                             vk::ImageSubresourceRange subresource_range,
+                             vk::PipelineStageFlags    src_mask = vk::PipelineStageFlagBits::eAllCommands,
+                             vk::PipelineStageFlags    dst_mask = vk::PipelineStageFlagBits::eAllCommands)
 {
 	vkb::set_image_layout(command_buffer,
 	                      image,
--- a/framework/core/hpp_command_buffer.h
+++ b/framework/core/hpp_command_buffer.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+/* Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
--- a/framework/core/hpp_command_buffer.h
+++ b/framework/core/hpp_command_buffer.h
@@ -1,0 +1,51 @@
+/* Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 the "License";
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <core/command_buffer.h>
+
+#include <common/hpp_vk_common.h>
+#include <core/hpp_image_view.h>
+
+namespace vkb
+{
+namespace core
+{
+/**
+ * @brief facade class around vkb::CommandBuffer, providing a vulkan.hpp-based interface
+ *
+ * See vkb::CommandBuffer for documentation
+ */
+class HPPCommandBuffer : protected vkb::CommandBuffer
+{
+  public:
+	using vkb::CommandBuffer::end_render_pass;
+
+	vk::CommandBuffer get_handle() const
+	{
+		return static_cast<vk::CommandBuffer>(vkb::CommandBuffer::get_handle());
+	}
+
+	void image_memory_barrier(const vkb::core::HPPImageView &image_view, const vkb::common::HPPImageMemoryBarrier &memory_barrier) const
+	{
+		vkb::CommandBuffer::image_memory_barrier(reinterpret_cast<vkb::core::ImageView const &>(image_view),
+		                                         reinterpret_cast<vkb::ImageMemoryBarrier const &>(memory_barrier));
+	}
+};
+}        // namespace core
+}        // namespace vkb
--- a/framework/core/hpp_debug.h
+++ b/framework/core/hpp_debug.h
@@ -75,18 +75,16 @@ class HPPDebugUtils : protected vkb::DebugUtils
 	virtual void set_debug_name(vk::Device     device,
 	                            vk::ObjectType object_type,
 	                            uint64_t       object_handle,
-	                            const char *   name) const    = 0;
+	                            const char *   name) const                                                           = 0;
 	virtual void set_debug_tag(vk::Device     device,
 	                           vk::ObjectType object_type,
 	                           uint64_t       object_handle,
 	                           uint64_t       tag_name,
 	                           const void *   tag_data,
-	                           size_t         tag_data_size) const = 0;
-	virtual void
-	             cmd_begin_label(vk::CommandBuffer command_buffer, const char *name, glm::vec4 color = {}) const = 0;
-	virtual void cmd_end_label(vk::CommandBuffer command_buffer) const                                           = 0;
-	virtual void
-	    cmd_insert_label(vk::CommandBuffer command_buffer, const char *name, glm::vec4 color = {}) const = 0;
+	                           size_t         tag_data_size) const                                                        = 0;
+	virtual void cmd_begin_label(vk::CommandBuffer command_buffer, const char *name, glm::vec4 color = {}) const  = 0;
+	virtual void cmd_end_label(vk::CommandBuffer command_buffer) const                                            = 0;
+	virtual void cmd_insert_label(vk::CommandBuffer command_buffer, const char *name, glm::vec4 color = {}) const = 0;
 };
 
 /**
--- a/framework/core/hpp_debug.h
+++ b/framework/core/hpp_debug.h
@@ -1,0 +1,254 @@
+/* Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 the "License";
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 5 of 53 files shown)


asuessenbach:
Any idea anyone what's wrong with the failing Android build?

SaschaWillems:
Thanks for the contribution :). Looks good to me so far, though I'd like to take a deeper look at this before approving.

gary-sweet (changes requested):
The direct-2-display changes don't work for me. I get a segv when running hpp_hello_triangle.

It looks like the vkGetPhysicalDeviceDisplayPropertiesKHR function pointer is null.

asuessenbach:
@gary-sweet I see... I need to add some direct-2-display specific code (initialize volk) to hpp_hello_triangle, then.
Is `VK_USE_PLATFORM_DISPLAY_KHR` the correct define to check for that platform?

gary-sweet:
Yes, VK_USE_PLATFORM_DISPLAY_KHR is the one.

gary-sweet (dismissed):
Thanks. That fixes the issue for me.

asuessenbach:
Is there anything I can do to get someone to review this?

SaschaWillems:
I'll try to review it this weekend :)

framework/common/hpp_strings.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <vulkan/vulkan.hpp>
+
+namespace vkb
+{
+namespace common
+{
+const std::string to_string(vk::Extent2D const &extent)
+{
+	return vkb::to_string(static_cast<VkExtent2D const &>(extent));
```
gpx1000, line 33: I'm confused about the need for this?  can't we use vkb::to_string as needed?  or even using namespace?

framework/common/hpp_utils.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <rendering/hpp_render_context.h>
+
+namespace vkb
+{
+namespace common
+{
+void screenshot(vkb::rendering::HPPRenderContext &render_context, const std::string &filename)
+{
+	vkb::screenshot(reinterpret_cast<vkb::RenderContext &>(render_context), filename);
```
gpx1000, line 33: same as with the to_string and the generate_all below.  What are we gaining from this?

framework/common/hpp_strings.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <common/strings.h>
+
+#include <vulkan/vulkan.hpp>
+
+namespace vkb
+{
+namespace common
+{
+const std::string to_string(vk::Extent2D const &extent)
```
gpx1000: const not needed in the return type.  It is const qualified at the top level which means it wouldn't effect const correctness and could reduce readability.

framework/common/hpp_vk_common.h
```diff
@@ -25,69 +25,63 @@ namespace vkb
 namespace common
 {
 /**
- * @brief facade helper functions around the functions in common/vk_commmon, providing a vulkan.hpp-based interface
+ * @brief facade helper functions and structs around the functions and structs in common/vk_commmon, providing a vulkan.hpp-based interface
```
gpx1000: vk_commmon has one too many 'm' characters.

framework/common/hpp_vk_common.h
```diff
@@ -25,69 +25,63 @@ namespace vkb
...
+
+	vk::AccessFlags dst_access_mask{};
+
+	vk::ImageLayout old_layout{vk::ImageLayout::eUndefined};
+
+	vk::ImageLayout new_layout{vk::ImageLayout::eUndefined};
+
+	uint32_t old_queue_family{VK_QUEUE_FAMILY_IGNORED};
+
+	uint32_t new_queue_family{VK_QUEUE_FAMILY_IGNORED};
```
gpx1000, line 46: old_queue_family /new_queue_family don't appear to be used yet although given that they are used in hte ImageMemoryBarrier struct it makes sense to leave them?

framework/common/hpp_vk_common.h
```diff
@@ -25,69 +25,63 @@ namespace vkb
...
-bool is_depth_stencil_format(vk::Format format)
+inline bool is_depth_only_format(vk::Format format)
 {
-	return format == vk::Format::eD16UnormS8Uint || format == vk::Format::eD24UnormS8Uint || format == vk::Format::eD32SfloatS8Uint ||
-	       is_depth_only_format(format);
+	return vkb::is_depth_only_format(static_cast<VkFormat>(format));
 }
 
-vk::ShaderModule load_shader(const std::string &filename, vk::Device device, vk::ShaderStageFlagBits stage)
+inline bool is_depth_stencil_format(vk::Format format)
```
gpx1000, line 68: function isn't used yet; it makes sense to keep it as it's used in the vk_common.h file which this emulates I think?

framework/core/hpp_command_buffer.h
```diff
@@ -0,0 +1,51 @@
...
+ * @brief facade class around vkb::CommandBuffer, providing a vulkan.hpp-based interface
+ *
+ * See vkb::CommandBuffer for documentation
+ */
+class HPPCommandBuffer : protected vkb::CommandBuffer
+{
+  public:
+	using vkb::CommandBuffer::end_render_pass;
+
+	vk::CommandBuffer get_handle() const
```
gpx1000, line 39: this hides a non-virtual get_handle() from VulkanResource::get_handle.

framework/core/hpp_debug.h
```diff
@@ -0,0 +1,252 @@
...
+{
+/**
+ * @brief facade class around vkb::DebugUtils, providing a vulkan.hpp-based interface
+ *
+ * See vkb::DebugUtils for documentation
+ */
+class HPPDebugUtils : protected vkb::DebugUtils
+{
+  public:
+	virtual void set_debug_name(VkDevice     device,
```
gpx1000: virtual is redundant when the function is already marked with override.  (same through out this file).

framework/core/hpp_debug.h
```diff
@@ -0,0 +1,252 @@
...
+		set_debug_tag(static_cast<vk::Device>(device),
+		              static_cast<vk::ObjectType>(object_type),
+		              object_handle,
+		              tag_name,
+		              tag_data,
+		              tag_data_size);
+	}
+
+	virtual void
+	    cmd_begin_label(VkCommandBuffer command_buffer, const char *name, glm::vec4 color = {}) const override
```
gpx1000: default parameters aren't great practice on overridden functions because the value of the default will depend upon the static type of the object while the virtual function call will depend upon the dynamic type of the object. (clang tidy will complain correctly on this).  Same else where in this file.

framework/core/hpp_device.h
```diff
@@ -35,8 +36,19 @@ namespace core
...
 	using vkb::Device::get_memory_allocator;
 
+	HPPDevice(vkb::core::HPPPhysicalDevice &              gpu,
+	          vk::SurfaceKHR                              surface,
+	          std::unique_ptr<vkb::core::HPPDebugUtils> &&debug_utils,
+	          std::unordered_map<const char *, bool>      requested_extensions = {}) :
+	    vkb::Device(reinterpret_cast<vkb::PhysicalDevice &>(gpu),
+	                static_cast<VkSurfaceKHR>(surface),
+	                std::unique_ptr<vkb::DebugUtils>(reinterpret_cast<vkb::DebugUtils *>(debug_utils.release())),
+	                requested_extensions)
```
gpx1000: requested_extensions is passed by value and only copied once, use std::move() on it to avoid unnecessary copies

framework/core/hpp_swapchain.h
```diff
@@ -36,6 +36,11 @@ class HPPSwapchain : protected vkb::Swapchain
 		return static_cast<vk::Result>(vkb::Swapchain::acquire_next_image(image_index, image_acquired_semaphore, fence));
 	}
 
+	const vk::Extent2D &get_extent() const
```
gpx1000, line 39: get_extent hides a non-virtual function from Swapchain class

framework/hpp_api_vulkan_sample.h
```diff
@@ -84,7 +86,7 @@ class HPPApiVulkanSample : public vkb::HPPVulkanSample
 	/// Stores the swapchain image buffers
 	std::vector<HPPSwapchainBuffer> swapchain_buffers;
 
-	virtual void create_render_context(vkb::Platform &platform) override;
+	virtual void create_render_context(vkb::platform::HPPPlatform const &platform) override;
```
gpx1000: virtual or override not both, and prefer override.

framework/hpp_gui.h
```diff
@@ -0,0 +1,95 @@
...
+	{
+		return reinterpret_cast<vkb::HPPDrawer &>(Gui::get_drawer());
+	}
+
+	void prepare(const vk::PipelineCache pipeline_cache, const vk::RenderPass render_pass, const std::vector<vk::PipelineShaderStageCreateInfo> &shader_stages)
+	{
+		std::vector<VkPipelineShaderStageCreateInfo> pssci;
+		for (auto const &ci : shader_stages)
+		{
+			pssci.push_back(*reinterpret_cast<VkPipelineShaderStageCreateInfo const *>(&ci));
```
gpx1000, line 81: push_back called inside a loop.  It's probably better to pre-allocate what you need, then assign values in the loop.

framework/hpp_gui.h
```diff
@@ -0,0 +1,95 @@
...
+	{
+		Gui::draw(command_buffer);
+	}
+
+	void draw(vkb::core::HPPCommandBuffer &command_buffer)
+	{
+		Gui::draw(reinterpret_cast<vkb::CommandBuffer &>(command_buffer));
+	}
+
+	vkb::HPPDrawer &get_drawer()
```
gpx1000, line 70: this hides a non-virtual from GUI

framework/hpp_gui.h
```diff
@@ -0,0 +1,95 @@
...
+		}
+		Gui::prepare(pipeline_cache, render_pass, pssci);
+	}
+
+	void show_top_window(const std::string &app_name, const vkb::stats::HPPStats *stats = nullptr, DebugInfo *debug_info = nullptr)
+	{
+		Gui::show_top_window(app_name, reinterpret_cast<vkb::Stats const *>(stats), debug_info);
+	}
+
+	bool update_buffers()
```
gpx1000, line 91: this hides a non-virtual from GUI

framework/hpp_vulkan_sample.cpp
```diff
@@ -0,0 +1,552 @@
...
+			for (auto script : scripts)
+			{
+				script->input_event(input_event);
+			}
+		}
+	}
+
+	if (input_event.get_source() == EventSource::Keyboard)
+	{
+		const auto &key_event = static_cast<const KeyInputEvent &>(input_event);
```
gpx1000, line 400: avoid using static_cast to downcast from a base class.

framework/hpp_vulkan_sample.cpp
```diff
@@ -0,0 +1,552 @@
...
+		if (key_event.get_action() == KeyAction::Down &&
+		    (key_event.get_code() == KeyCode::PrintScreen || key_event.get_code() == KeyCode::F12))
+		{
+			vkb::common::screenshot(*render_context, "screenshot-" + get_name());
+		}
+
+		if (key_event.get_code() == KeyCode::F6 && key_event.get_action() == KeyAction::Down)
+		{
+			if (!vkb::common::graphs::generate_all(get_render_context(), *scene.get()))
```
gpx1000: redundant get called on smart pointer (just remove the get and it'll work).

framework/hpp_vulkan_sample.cpp
```diff
@@ -0,0 +1,552 @@
...
+			if (auto camera_node = camera->get_node())
+			{
+				const glm::vec3 &pos = camera_node->get_transform().get_translation();
+				get_debug_info().insert<field::Vector, float>("camera_pos", pos.x, pos.y, pos.z);
+			}
+		}
+	}
+}
+
+void HPPVulkanSample::set_viewport_and_scissor(vkb::core::HPPCommandBuffer const &command_buffer, const vk::Extent2D &extent) const
```
gpx1000: function can be made static.

framework/hpp_vulkan_sample.cpp
```diff
@@ -0,0 +1,552 @@
...
+}
+
+void HPPVulkanSample::set_api_version(uint32_t requested_api_version)
+{
+	api_version = requested_api_version;
+}
+
+void HPPVulkanSample::request_gpu_features(vkb::core::HPPPhysicalDevice &gpu)
+{
+	// To be overriden by sample
```
gpx1000: two 'd's' in overridden.

framework/hpp_vulkan_sample.h
```diff
@@ -17,71 +17,252 @@
...
-		}
-		render_context = platform.create_render_context(*device, surface, spl);
-	}
+	HPPVulkanSample() = default;
 
-	vkb::core::HPPDevice const &get_device() const
-	{
-		return *reinterpret_cast<vkb::core::HPPDevice const *>(&*device);
-	}
+	virtual ~HPPVulkanSample();
```
gpx1000: let's prefer override or final to virtual on child classes.

framework/hpp_vulkan_sample.h
```diff
@@ -17,71 +17,252 @@
...
 
-	vkb::core::HPPInstance const &get_instance() const
-	{
-		return *reinterpret_cast<vkb::core::HPPInstance const *>(&*instance);
-	}
+	/**
+	 * @brief Additional sample initialization
+	 */
+	virtual bool prepare(vkb::platform::HPPPlatform &platform) override;
```
gpx1000: virtual or override not both.  Prefer override (everywhere in this file).

framework/hpp_vulkan_sample.h
```diff
@@ -17,71 +17,252 @@
...
+	/**
+	 * @brief Main loop sample events
+	 */
+	virtual void update(float delta_time) override;
 
-	vkb::rendering::HPPRenderContext &get_render_context() const
-	{
-		return *reinterpret_cast<vkb::rendering::HPPRenderContext *>(&*render_context);
-	}
+	virtual bool resize(const uint32_t width, const uint32_t height) override;
```
gpx1000: width and height are both const qualified, this only has effect in the function definitions so not needed here.

framework/hpp_vulkan_sample.h
```diff
@@ -17,71 +17,252 @@
...
+	Configuration configuration{};
+
+	/**
+	 * @brief Sets whether or not the first graphics queue should have higher priority than other queues.
+	 * Very specific feature which is used by async compute samples.
+	 * Needs to be called before prepare().
+	 * @param enable If true, present queue will have prio 1.0 and other queues have prio 0.5.
+	 * Default state is false, where all queues have 0.5 priority.
+	 */
+	void set_high_priority_graphics_queue_enable(bool enable)
```
gpx1000, line 250: set_high_priority_graphics_queue_enable isn't used but is used in the replaced class.  assumption is this is to be addressed.

framework/hpp_vulkan_sample.h
```diff
@@ -17,71 +17,252 @@
...
+
+	/**
+	 * @brief Starts the render pass, executes the render pipeline, and then ends the render pass
+	 * @param command_buffer The command buffer to record the commands to
+	 * @param render_target The render target that is being drawn to
+	 */
+	virtual void draw_renderpass(vkb::core::HPPCommandBuffer &command_buffer, vkb::rendering::HPPRenderTarget const &render_target) const;
+
+	/**
+	 * @brief Triggers the render pipeline, it can be overriden by samples to specialize their rendering logic
```
gpx1000: add a 'd' to overridden.

framework/hpp_vulkan_sample.h
```diff
@@ -17,71 +17,252 @@
...
+	 * @brief Get additional sample-specific instance layers.
+	 *
+	 * @return Vector of additional instance layers. Default is empty vector.
+	 */
+	virtual std::vector<const char *> const &get_validation_layers() const;
+
+	/**
+	 * @brief Get sample-specific instance extensions.
+	 *
+	 * @return Map of instance extensions and wether or not they are optional. Default is empty map.
```
gpx1000: whether needs an 'h'.

framework/hpp_vulkan_sample.h
```diff
@@ -17,71 +17,252 @@
...
+	 * @brief Get sample-specific device extensions.
+	 *
+	 * @return Map of device extensions and whether or not they are optional. Default is empty map.
+	 */
+	std::unordered_map<const char *, bool> const &get_device_extensions() const;
+
+	/**
+	 * @brief Add a sample-specific device extension
+	 * @param extension The extension name
+	 * @param optional (Optional) Wether the extension is optional
```
gpx1000: whether needs an 'h.'

framework/hpp_vulkan_sample.h
```diff
@@ -17,71 +17,252 @@
...
+	 */
+	void add_device_extension(const char *extension, bool optional = false);
+
+	/**
+	 * @brief Add a sample-specific instance extension
+	 * @param extension The extension name
+	 * @param optional (Optional) Wether the extension is optional
```
gpx1000: whether needs an 'h'

framework/hpp_vulkan_sample.h
```diff
@@ -17,71 +17,252 @@
...
+	virtual void create_render_context(vkb::platform::HPPPlatform const &platform);
+
+	/** 
+	 * @brief Override this to customise the creation of the swapchain and render_context
+	 */
+	virtual void prepare_render_context();
+
+	/**
+	 * @brief Resets the stats view max values for high demanding configs
+	 *        Should be overriden by the samples since they
```
gpx1000: overridden needs a d

framework/hpp_vulkan_sample.h
```diff
@@ -17,71 +17,252 @@
...
+	 * Default state is false, where all queues have 0.5 priority.
+	 */
+	void set_high_priority_graphics_queue_enable(bool enable)
 	{
-		// To be overriden by sample
+		high_priority_graphics_queue = enable;
 	}
+
+  private:
+	/** @brief Set of device extensions to be enabled for this example and wether they are optional (must be set in the derived constructor) */
```
gpx1000: whether needs a h

framework/platform/glfw_window.h
```diff
@@ -38,7 +38,7 @@ class GlfwWindow : public Window
 
 	virtual VkSurfaceKHR create_surface(Instance &instance) override;
 
-	virtual vk::SurfaceKHR create_surface(vk::Instance instance, vk::PhysicalDevice physical_device) override;
+	virtual VkSurfaceKHR create_surface(VkInstance instance, VkPhysicalDevice physical_device) override;
```
gpx1000: virtual or override but not both and prefer override

framework/platform/headless_window.h
```diff
@@ -41,7 +41,7 @@ class HeadlessWindow : public Window
 	 * @brief A direct window doesn't have a surface
 	 * @returns nullptr
 	 */
-	virtual vk::SurfaceKHR create_surface(vk::Instance instance, vk::PhysicalDevice physical_device) override;
+	virtual VkSurfaceKHR create_surface(VkInstance instance, VkPhysicalDevice physical_device) override;
```
gpx1000: virtual or override but not both, prefer override.

framework/platform/hpp_application.h
```diff
@@ -0,0 +1,53 @@
...
+ */
+class HPPApplication : public vkb::Application
+{
+  public:
+	vkb::platform::HPPPlatform &get_platform()
+	{
+		return *reinterpret_cast<vkb::platform::HPPPlatform *>(platform);
+	}
+
+	virtual bool prepare(vkb::Platform &platform) final
```
gpx1000: virtual or final, prefer final

framework/platform/hpp_platform.h
```diff
@@ -32,9 +35,25 @@ namespace platform
...
+	    create_render_context(vkb::core::HPPDevice &device, vk::SurfaceKHR surface, const std::vector<vk::SurfaceFormatKHR> &surface_format_priority) const
+	{
+		return std::unique_ptr<vkb::rendering::HPPRenderContext>(reinterpret_cast<vkb::rendering::HPPRenderContext *>(
+		    create_render_context(reinterpret_cast<vkb::Device &>(device),
+		                          static_cast<VkSurfaceKHR>(surface),
+		                          reinterpret_cast<std::vector<VkSurfaceFormatKHR> const &>(surface_format_priority))
+		        .release()));
+	}
+
+	vkb::platform::HPPWindow &get_window()
```
gpx1000, line 51: get_window hides a non-virtual function in Platform.

framework/rendering/hpp_render_context.h
```diff
@@ -35,6 +38,17 @@ class HPPRenderContext : protected vkb::RenderContext
 	using vkb::RenderContext::get_render_frames;
 	using vkb::RenderContext::handle_surface_changes;
 	using vkb::RenderContext::has_swapchain;
+	using vkb::RenderContext::prepare;
+
+	vkb::core::HPPCommandBuffer &begin(vkb::CommandBuffer::ResetMode reset_mode = vkb::CommandBuffer::ResetMode::ResetPool)
```
gpx1000, line 43: begin hides a non-virtual function.

framework/rendering/hpp_render_context.h
```diff
@@ -35,6 +38,17 @@ class HPPRenderContext : protected vkb::RenderContext
...
+	{
+		return reinterpret_cast<vkb::core::HPPCommandBuffer &>(vkb::RenderContext::begin(reset_mode));
+	}
+
+	vkb::rendering::HPPRenderFrame &get_active_frame()
```
gpx1000, line 48: get_active_frame hides a non-virtual function.

framework/rendering/hpp_render_frame.h
```diff
@@ -0,0 +1,42 @@
...
+{
+/**
+ * @brief facade class around vkb::RenderFrame, providing a vulkan.hpp-based interface
+ *
+ * See vkb::RenderFrame for documentation
+ */
+class HPPRenderFrame : protected vkb::RenderFrame
+{
+  public:
+	vkb::rendering::HPPRenderTarget &get_render_target()
```
gpx1000, line 36: get_render_target hides a non-virtual function

framework/rendering/hpp_render_target.h
```diff
@@ -0,0 +1,45 @@
...
+{
+/**
+ * @brief facade class around vkb::RenderTarget, providing a vulkan.hpp-based interface
+ *
+ * See vkb::RenderTarget for documentation
+ */
+class HPPRenderTarget : protected vkb::RenderTarget
+{
+  public:
+	const vk::Extent2D &get_extent() const
```
gpx1000, line 34: get_extent hides a non-virtual function

framework/rendering/hpp_render_target.h
```diff
@@ -0,0 +1,45 @@
...
+	{
+		return reinterpret_cast<vk::Extent2D const &>(vkb::RenderTarget::get_extent());
+	}
+
+	std::vector<vkb::core::HPPImageView> const &get_views() const
```
gpx1000, line 39: get_views hides a non-virtual function

framework/stats/hpp_stats.h
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * See vkb::Stats for documentation
+ */
+class HPPStats : protected vkb::Stats
+{
+  public:
+	using vkb::Stats::resize;
+	using vkb::Stats::update;
+
+	HPPStats(vkb::rendering::HPPRenderContext &render_context, size_t buffer_size = 16) :
```
gpx1000: constructors that are callable with a single parameter need to be marked explicit.

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
```diff
@@ -15,13 +15,10 @@
  * limitations under the License.
  */
 
-#include "hpp_hello_triangle.h"
+#include <hpp_hello_triangle.h>
```
gpx1000: personal preference however, unless it breaks something which it shouldn't; this should be reversed (hpp_hello_traingle.h) is found at the same folder as the cpp file.

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
```diff
@@ -268,7 +274,7 @@ void HPPHelloTriangle::init_instance(Context &                        context,
  * @param context A Vulkan context with an instance already set up.
  * @param platform The platform the application is being run on
  */
-void HPPHelloTriangle::select_physical_device_and_surface(Context &context, vkb::Platform &platform)
+void HPPHelloTriangle::select_physical_device_and_surface(Context &context, vkb::platform::HPPPlatform &platform)
```
gpx1000: while we're at it (although I'm trying to limit my comments to only new changes these are easy to fix).  Could we rename the context parameter so it doesn't shadow the member variable context?  Also this function is safe to be made static.

gpx1000 (commented):
I limited my review to changes only.  Some day we need to whole sale fix all of the warnings.  These changes work on a linux device here and look good to me!  Thanks for the work!

framework/common/hpp_strings.h
asuessenbach: This function (as well as all the other hpp-facade classes and functions) is just there to make the really vulkan.hpp-based classes look like they're in a completely vulkan.hpp-based environment: call that function without the need to explicitly cast the arguments to some non-vk-namespace stuff.

framework/common/hpp_utils.h
asuessenbach: Same answer as above. I just realized that the facade-comment is missing in these two files.

framework/common/hpp_strings.h
asuessenbach: I just replicated the exact same interface (besides the vk-types) as the wrapped function, to make sure this function can be used in the exact same manner.
Will adjust this, together with the return type of vkb::to_string().

framework/common/hpp_vk_common.h
asuessenbach: I think, it makes sense to keep them.
The facade-classes/structs should be "as similar" as possible to the original classes/structs.

framework/rendering/hpp_render_target.h
asuessenbach: That was my intention.
The facade-classes are supposed to return vk-namespace stuff.

framework/rendering/hpp_render_target.h
asuessenbach: Same answer as above.

framework/stats/hpp_stats.h
asuessenbach: This is again (besides the vulkan.hpp-based types) the exact same interface as the wrapped constructor.
Adjusted in both hpp_stats.h and stats.h.

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
asuessenbach: Good point!
I'll remove that parameter, as it's already available by the member variable.

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
asuessenbach: Ok, will revert that.

framework/common/hpp_vk_common.h
asuessenbach: Good catch.

framework/common/hpp_vk_common.h
asuessenbach: Agreed.

framework/core/hpp_command_buffer.h
asuessenbach: Again, by intention. This function should return vk-namespace stuff.

framework/core/hpp_debug.h
asuessenbach: Will remove the virtuals.

framework/core/hpp_debug.h
asuessenbach: Will remove the default on this function.
But I think, the other defaults in this file should be kept as they are (or the corresponding defaults in core/debug.h should be changed as well; I don't know how many adjustments that would provoke throughout the code base).

framework/core/hpp_device.h
asuessenbach: Good point. Will do so.
Or would it be better to pass requested_extensions by const reference?

framework/core/hpp_swapchain.h
asuessenbach: For the same reason as above.

framework/hpp_api_vulkan_sample.h
asuessenbach: Will remove virtuals.

framework/hpp_gui.h
asuessenbach: Or to reserve the needed space, and still push_back.
No need for a counter variable, then.

framework/hpp_gui.h
asuessenbach: Same reason as above.

framework/hpp_gui.h
asuessenbach: Here, a simple using `vkb::Gui::update_buffers` did not work, as there's also a private `update_buffers` in `vkb::Gui`. Could not successfully compile then.

framework/hpp_vulkan_sample.cpp
asuessenbach: This line of code is an exact copy from vulkan_sample.cpp.
My understanding was, that the if-clause makes sure that the input_event in fact is a KeyInputEvent.

framework/hpp_vulkan_sample.cpp
asuessenbach: Will remove it (in both hpp_vulkan_sample.cpp and vulkan_sample.cpp)

framework/hpp_vulkan_sample.cpp
asuessenbach: Will make it static (in both, HPPVulkanSample and VulkanSample).

framework/hpp_vulkan_sample.cpp
asuessenbach: ?

framework/hpp_vulkan_sample.h
asuessenbach: Adjusted in both hpp_vulkan_sample.h and vulkan_sample.h

framework/hpp_vulkan_sample.h
asuessenbach: Adjusted in both hpp_vulkan_sample.h and vulkan_sample.h.

framework/platform/hpp_platform.h
asuessenbach: As noted above, by intention.

framework/rendering/hpp_render_context.h
asuessenbach: By intention.

framework/rendering/hpp_render_context.h
asuessenbach: By intention

framework/rendering/hpp_render_frame.h
asuessenbach: By intention.

asuessenbach:
Does anybody know how to resolve that failing Quality Check?

framework/common/hpp_strings.h
gpx1000: makes sense to me.

gpx1000:
What it's having a problem with is the doxygen syntax.

in stats_common.h at line 31 it's calling a reference to the Stats object.  That doesn't look like it's resolving anymore.  My guess is one of the c++ includes is including this file which is making doxygen think that Stats isn't defined in the namespace it has knowledge of.  You might need to put vkb:: infront of Stats?  I haven't tested any of this so I don't know for sure and am operating totally off gut instinct.

asuessenbach:
Changing from `Stats` -> `vkb::Stats` resolved the first warning of the Quality Checks.
But how would I resolve the other two, complaining about `size_t` and `uint32_t`? Adding namespace `std::` does not help here.

SaschaWillems:
I'm actually wondering why the @ref is there. Afaik @ref can be used to link to other sections of the documentation (e.g. for custom types), but I'm not sure if it can be used on intrinsic types. I think you can safely remove the @ref and it should work.

gpx1000:
I think @SaschaWillems  is correct.  it would be better to remove the ref directive from intrinsic types.  My guess is the problem is going to stem from size_t and uint32_t are defined as part of the std namespace in C++ as a type and in C which naturally doesn't have the std namespace, they are a normal POC type.  So doxygen is going to get confused when they are included in either direction (i.e. adding std:: will work in C++ to fix it but will break C).  The "real" fix if want to maintain the ref function is to set one vs the other depending upon language type with a check.  Probably the easiest way to "fix" it if we want to keep the ref is to copy/rename the file to .hpp extension and only include it in C++ files.  I think that's overkill for this problem so let's nuke the ref.  It's safe to assume people reading the doc knows what size_t and uint32_t are.
NB: a different solution would be to switch to a true POC type that both C and C++ share.  i.e. unsigned int for both.  It technically doesn't mean the same thing, but that would also fix it.

asuessenbach:
To rephrase my comment from two weeks ago: Is there anything I can do to get this PR approved?

gary-sweet (approved):

gpx1000 (approved):

SaschaWillems (approved):

PR #442, opened 2022-04-12
Change vkb::core::HPPPhysicalDevice and vkb::core::HPPInstance from facades over vkb::PhysicalDevice and vkb::Instance, respectively, to self-contained classes using vulkan.hpp

asuessenbach:
## Description

Replacing this part of the framework shows how to manage some more resources using vulkan.hpp.

In order to do so, minor adjustments in vkb::core::HPPDevice and framework/hpp_vulkan_sample.cpp were needed.

I have tested it on Window10 on nvidia HW.

## General Checklist:

### DIFF
--- a/framework/CMakeLists.txt
+++ b/framework/CMakeLists.txt
@@ -278,7 +278,9 @@ set(CORE_FILES
     core/acceleration_structure.cpp
     core/shader_binding_table.cpp
     core/vulkan_resource.cpp
-    core/hpp_device.cpp)
+    core/hpp_device.cpp
+    core/hpp_instance.cpp
+	core/hpp_physical_device.cpp)
 
 set(PLATFORM_FILES
     # Header Files
--- a/framework/core/hpp_device.cpp
+++ b/framework/core/hpp_device.cpp
@@ -169,7 +169,7 @@ HPPDevice::HPPDevice(vkb::core::HPPPhysicalDevice &              gpu,
 	{
 		vk::QueueFamilyProperties const &queue_family_property = queue_family_properties[queue_family_index];
 
-		vk::Bool32 present_supported = gpu.is_present_supported(surface, queue_family_index);
+		vk::Bool32 present_supported = gpu.get_handle().getSurfaceSupportKHR(queue_family_index, surface);
 
 		for (uint32_t queue_index = 0U; queue_index < queue_family_property.queueCount; ++queue_index)
 		{
--- a/framework/core/hpp_instance.cpp
+++ b/framework/core/hpp_instance.cpp
@@ -1,0 +1,458 @@
+/* Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 the "License";
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <core/hpp_instance.h>
+
+#include <common/logging.h>
+#include <core/hpp_physical_device.h>
+#include <volk.h>
+
+namespace vkb
+{
+namespace
+{
+#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
+VKAPI_ATTR VkBool32 VKAPI_CALL debug_utils_messenger_callback(VkDebugUtilsMessageSeverityFlagBitsEXT message_severity, VkDebugUtilsMessageTypeFlagsEXT message_type,
+                                                              const VkDebugUtilsMessengerCallbackDataEXT *callback_data,
+                                                              void *                                      user_data)
+{
+	// Log debug messge
+	if (message_severity & VK_DEBUG_UTILS_MESSAGE_SEVERITY_WARNING_BIT_EXT)
+	{
+		LOGW("{} - {}: {}", callback_data->messageIdNumber, callback_data->pMessageIdName, callback_data->pMessage);
+	}
+	else if (message_severity & VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT)
+	{
+		LOGE("{} - {}: {}", callback_data->messageIdNumber, callback_data->pMessageIdName, callback_data->pMessage);
+	}
+	return VK_FALSE;
+}
+
+static VKAPI_ATTR VkBool32 VKAPI_CALL debug_callback(VkDebugReportFlagsEXT flags, VkDebugReportObjectTypeEXT /*type*/,
+                                                     uint64_t /*object*/, size_t /*location*/, int32_t /*message_code*/,
+                                                     const char *layer_prefix, const char *message, void * /*user_data*/)
+{
+	if (flags & VK_DEBUG_REPORT_ERROR_BIT_EXT)
+	{
+		LOGE("{}: {}", layer_prefix, message);
+	}
+	else if (flags & VK_DEBUG_REPORT_WARNING_BIT_EXT)
+	{
+		LOGW("{}: {}", layer_prefix, message);
+	}
+	else if (flags & VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT)
+	{
+		LOGW("{}: {}", layer_prefix, message);
+	}
+	else
+	{
+		LOGI("{}: {}", layer_prefix, message);
+	}
+	return VK_FALSE;
+}
+#endif
+
+bool validate_layers(const std::vector<const char *> &       required,
+                     const std::vector<vk::LayerProperties> &available)
+{
+	for (auto layer : required)
+	{
+		bool found = false;
+		for (auto &available_layer : available)
+		{
+			if (strcmp(available_layer.layerName, layer) == 0)
+			{
+				found = true;
+				break;
+			}
+		}
+
+		if (!found)
+		{
+			LOGE("Validation Layer {} not found", layer);
+			return false;
+		}
+	}
+
+	return true;
+}
+}        // namespace
+
+namespace core
+{
+std::vector<const char *> get_optimal_validation_layers(const std::vector<vk::LayerProperties> &supported_instance_layers)
+{
+	std::vector<std::vector<const char *>> validation_layer_priority_list =
+	    {
+	        // The preferred validation layer is "VK_LAYER_KHRONOS_validation"
+	        {"VK_LAYER_KHRONOS_validation"},
+
+	        // Otherwise we fallback to using the LunarG meta layer
+	        {"VK_LAYER_LUNARG_standard_validation"},
+
+	        // Otherwise we attempt to enable the individual layers that compose the LunarG meta layer since it doesn't exist
+	        {
+	            "VK_LAYER_GOOGLE_threading",
+	            "VK_LAYER_LUNARG_parameter_validation",
+	            "VK_LAYER_LUNARG_object_tracker",
+	            "VK_LAYER_LUNARG_core_validation",
+	            "VK_LAYER_GOOGLE_unique_objects",
+	        },
+
+	        // Otherwise as a last resort we fallback to attempting to enable the LunarG core layer
+	        {"VK_LAYER_LUNARG_core_validation"}};
+
+	for (auto &validation_layers : validation_layer_priority_list)
+	{
+		if (validate_layers(validation_layers, supported_instance_layers))
+		{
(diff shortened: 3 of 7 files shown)


gpx1000 (dismissed):
LGTM

SaschaWillems (dismissed):
LGTM

gary-sweet (changes requested):
I'm seeing some build errors with this change:

```
/usr/bin/ld: ../framework/lib/Debug/x86_64/libframework.a(hpp_device.cpp.o): in function `vk::PhysicalDevicePerformanceQueryFeaturesKHR& vkb::core::HPPPhysicalDevice::request_extension_features<vk::PhysicalDevicePerformanceQueryFeaturesKHR>()':
/build/gsweet/vc5_git/thirdparty_vulkan/Vulkan-Samples/framework/core/hpp_physical_device.h:113: undefined reference to `vk::PhysicalDevicePerformanceQueryFeaturesKHR::structureType'
/usr/bin/ld: /build/gsweet/vc5_git/thirdparty_vulkan/Vulkan-Samples/framework/core/hpp_physical_device.h:123: undefined reference to `vk::PhysicalDevicePerformanceQueryFeaturesKHR::structureType'
/usr/bin/ld: /build/gsweet/vc5_git/thirdparty_vulkan/Vulkan-Samples/framework/core/hpp_physical_device.h:126: undefined reference to `vk::PhysicalDevicePerformanceQueryFeaturesKHR::structureType'
/usr/bin/ld: ../framework/lib/Debug/x86_64/libframework.a(hpp_device.cpp.o): in function `vk::PhysicalDeviceHostQueryResetFeatures& vkb::core::HPPPhysicalDevice::request_extension_features<vk::PhysicalDeviceHostQueryResetFeatures>()':
/build/gsweet/vc5_git/thirdparty_vulkan/Vulkan-Samples/framework/core/hpp_physical_device.h:113: undefined reference to `vk::PhysicalDeviceHostQueryResetFeatures::structureType'
/usr/bin/ld: /build/gsweet/vc5_git/thirdparty_vulkan/Vulkan-Samples/framework/core/hpp_physical_device.h:123: undefined reference to `vk::PhysicalDeviceHostQueryResetFeatures::structureType'
/usr/bin/ld: /build/gsweet/vc5_git/thirdparty_vulkan/Vulkan-Samples/framework/core/hpp_physical_device.h:126: undefined reference to `vk::PhysicalDeviceHostQueryResetFeatures::structureType'
collect2: error: ld returned 1 exit status
```

gary-sweet:
The build errors appear to only happen in debug builds.

asuessenbach:
@gary-sweet Strange thing!
What compiler are you using?
Could you please verify, that you have this in your vulkan_structs.hpp:
```
  struct PhysicalDevicePerformanceQueryFeaturesKHR
  {
    using NativeType = VkPhysicalDevicePerformanceQueryFeaturesKHR;

    static const bool                                  allowDuplicate = false;
    static VULKAN_HPP_CONST_OR_CONSTEXPR StructureType structureType  = StructureType::ePhysicalDevicePerformanceQueryFeaturesKHR;
```
I don't see, where that could be different in release and debug builds. And I don't get any errors with MSVC 2022, neither release nor debug.
Could you slightly modify that struct in vulkan_structs.hpp, by
- replacing that `VULKAN_HPP_CONST_OR_CONSTEXPR` in front of the `StructureType` by a simple `const`?
- adding `VULKAN_HPP_NAMESPACE::` in front of the two `StructureType`s in that line?
Does that change anything?

gary-sweet:
This is happening on Linux on both cross-compiled gcc 8.3 and native gcc 9.4.
I'll do some more digging to see if I can find the cause.

asuessenbach:
@gary-sweet Please try a static_cast to the call in hpp_physical_device.h, on line 113, like this:
`		auto extension_features_it = extension_features.find(static_cast<vk::StructureType>(HPPStructureType::structureType));
`

gary-sweet:
> @gary-sweet Please try a static_cast to the call in hpp_physical_device.h, on line 113, like this: ` auto extension_features_it = extension_features.find(static_cast<vk::StructureType>(HPPStructureType::structureType)); `

No change sadly.

asuessenbach:
@gary-sweet Does it change anything, if you're compiling with -std=c++17 ?
Then, that `structureType` would be a `constexpr`, instead of just a `const`.

asuessenbach:
@gary-sweet And another thing I'd like you to try:
```
		vk::StructureType structureType         = HPPStructureType::structureType;
		auto              extension_features_it = extension_features.find(structureType);
```

gary-sweet:
> @gary-sweet Does it change anything, if you're compiling with -std=c++17 ? Then, that `structureType` would be a `constexpr`, instead of just a `const`.

No difference with c++17

gary-sweet:
> @gary-sweet And another thing I'd like you to try:
> 
> ```
> 		vk::StructureType structureType         = HPPStructureType::structureType;
> 		auto              extension_features_it = extension_features.find(structureType);
> ```

Also no change I'm afraid.
I'll keep digging here.

asuessenbach:
@gary-sweet Just to be sure...
The things I suggested to fix the build failures were supposed to fix just the first one out of three (identical issues). I'm sure you realized that, but I just wanted to make sure.
(You see: I'm really running out of ideas!)

gary-sweet:
> @gary-sweet Just to be sure... The things I suggested to fix the build failures were supposed to fix just the first one out of three (identical issues). I'm sure you realized that, but I just wanted to make sure. (You see: I'm really running out of ideas!)

Yes, I was aware, thanks. I'm OOO for a couple of days, but will circle back to this when I return.

gary-sweet:
Is someone else able to try a debug build of this on Linux?
I get the same result when building with clang as I do with various versions of gcc.

Would be good to know if it's just a problem locally, or if others can reproduce.

gary-sweet:
Well, I'm mystified. I can only think that we're hitting a compiler or linker bug.

If I add methods to return `structureType` instead of accessing it directly in `request_extension_features(`) then it links ok.

This works for example:

```
  struct PhysicalDeviceHostQueryResetFeatures
  {
    using NativeType = VkPhysicalDeviceHostQueryResetFeatures;

    static const bool                                  allowDuplicate = false;
    static VULKAN_HPP_CONST_OR_CONSTEXPR StructureType structureType =
      StructureType::ePhysicalDeviceHostQueryResetFeatures;

    static VULKAN_HPP_CONST_OR_CONSTEXPR StructureType ST() { return structureType; }
    ....
```

SaschaWillems:
This builds fine for me on my Ubuntu 18 based system with a (seemingly) outdated GCC 7.5.
 
P.S. : Build times on Linux are horrible. A clean build took ~ 45 Minutes on my Laptop :(

gary-sweet:
> This builds fine for me on my Ubuntu 18 based system with a (seemingly) outdated GCC 7.5.
> 
> P.S. : Build times on Linux are horrible. A clean build took ~ 45 Minutes on my Laptop :(

And that was definitely a debug build?

SaschaWillems:
I ran the following command:

```cmake --build build/linux --config Debug --target vulkan_samples -- -j4```

So I think it should've been a debug build. Or is there any other way I need to force that?

gary-sweet:
> I ran the following command:
> 
> `cmake --build build/linux --config Debug --target vulkan_samples -- -j4`
> 
> So I think it should've been a debug build. Or is there any other way I need to force that?

So, I ran that very command myself and it linked ok. Setting the CMAKE_BUILD_TYPE explicitly to Debug in cmake-gui and then running the same command fails however. I don't think your command is actually building debug for some reason.

gary-sweet:
> > I ran the following command:
> > `cmake --build build/linux --config Debug --target vulkan_samples -- -j4`
> > So I think it should've been a debug build. Or is there any other way I need to force that?
> 
> So, I ran that very command myself and it linked ok. Setting the CMAKE_BUILD_TYPE explicitly to Debug in cmake-gui and then running the same command fails however. I don't think your command is actually building debug for some reason.

This probably explains why:

"If the build tool isn't multi-config (like gcc), then the --config argument is ignored. Instead the build type is set via the CMAKE_BUILD_TYPE CMake variable; i.e. it's set when running CMake, not when running the build tool."

SaschaWillems:
You're right. Setting CMAKE_BUILD_TYPE to Debug the build fails for me too with the same message you posted above.

gpx1000:
Yep, it fails here too.  It *looks* like it's having problems figuring out what StructureType is as a type.  Not sure why, it'd take a bit more debugging to figure out why for me.  However, I have a silly stupid solution that works here.
at line 104 change the signature to pass in the StructureType:
HPPStructureType &request_extension_features(vk::StructureType type)

then use type everywhere in the function (line 113, 123, 126).

Finally in hpp_device line 102-103 add in vk::PhysicalDevicePerformanceQueryFeaturesKHR::structureType and vk::PhysicalDeviceHostQueryResetFeatures::structureType to the passed in parameters.

Doing that it will build/link fine.  It's about as silly, stupid workaround as it gets, but hey, why work too hard?

asuessenbach:
> HPPStructureType &request_extension_features(vk::StructureType type)

Essentially, that should work. But it would remove the advantage you've got using vulkan.hpp.
After all, passing in the structure type to get via both, the template argument and the function argument is redundant. In vulkan.h, you don't have any connection between the actual structure type and corresponding value of type VkStructureType. But in vulkan.hpp, you have that. And therefore, such a redundant calling pattern feels wrong.

gary-sweet:
> > HPPStructureType &request_extension_features(vk::StructureType type)
> 
> Essentially, that should work. But it would remove the advantage you've got using vulkan.hpp. After all, passing in the structure type to get via both, the template argument and the function argument is redundant. In vulkan.h, you don't have any connection between the actual structure type and corresponding value of type VkStructureType. But in vulkan.hpp, you have that. And therefore, such a redundant calling pattern feels wrong.

Wrapping access to `structureType` in a static method also worked for me. That doesn't require passing any redundant arguments. `structureType` could be made private to ensure the accessor was always used. It's obviously not trivial, but nothing else I tried worked.

gary-sweet:
> > > HPPStructureType &request_extension_features(vk::StructureType type)
> > 
> > 
> > Essentially, that should work. But it would remove the advantage you've got using vulkan.hpp. After all, passing in the structure type to get via both, the template argument and the function argument is redundant. In vulkan.h, you don't have any connection between the actual structure type and corresponding value of type VkStructureType. But in vulkan.hpp, you have that. And therefore, such a redundant calling pattern feels wrong.
> 
> Wrapping access to `structureType` in a static method also worked for me. That doesn't require passing any redundant arguments. `structureType` could be made private to ensure the accessor was always used. It's obviously not trivial, but nothing else I tried worked.

Steven's suggestion is obviously much simpler and doesn't affect the hpp api's at all, just the framework, so I'd probably be leaning towards that personally.

gpx1000:
I'm doing this from my phone as it's 1am here, but I had a thought.
Someone try inlining the template function.  I know it sounds silly, but I
have a feeling it has to do with the optimizations are inlining
automatically while the debug version isn't.  This results in violating the
one definition rule.

Other options include explicit template specialization.  Either should not
involve changes to the underlining header files.

gary-sweet:
> I'm doing this from my phone as it's 1am here, but I had a thought. Someone try inlining the template function. I know it sounds silly, but I have a feeling it has to do with the optimizations are inlining automatically while the debug version isn't. This results in violating the one definition rule. Other options include explicit template specialization. Either should not involve changes to the underlining header files.

No luck with inlining request_extension_features() sadly - same link errors.

gary-sweet:
>Other options include explicit template specialization.

That isn't possible in this case since these are class methods, which aren't at namespace scope. I believe C++17 would be needed for that (we are stuck on C++11 for a while yet).

asuessenbach:
I've created a very downsized version of this problem (https://godbolt.org/z/qfc3a9YTr), reproducing this compilation error.
But with that "repro", two of the approaches I suggested above would resolve that error.
Maybe someone could try to extend that snippet such that those approaches don't help anymore?

gary-sweet:
> I've created a very downsized version of this problem (https://godbolt.org/z/qfc3a9YTr), reproducing this compilation error. But with that "repro", two of the approaches I suggested above would resolve that error. Maybe someone could try to extend that snippet such that those approaches don't help anymore?

Well, I know I tried option one before and still got failures, but having tried it again it appears to work just fine. No idea how I managed to screw that up the first time around, but apologies to everyone that has spent any time on this since then!

At least we have a trivial workaround though.

gary-sweet:
Here's an explanation of the issue that Andreas found:

https://stackoverflow.com/questions/5391973/undefined-reference-to-static-const-int

gary-sweet (approved):
Thanks, LGTM

asuessenbach:
Well, after this compile issue (hopefully) is ruled out, we should consider to have some debug builds as part of the CI checks !

gpx1000 (approved):
good sluething.  Glad this worked out.

LGTM!

SaschaWillems (approved):

asuessenbach:
Now, that there are three approvals on this PR... may I just merge it?

TomAtkinsonArm:
> Now, that there are three approvals on this PR... may I just merge it?

@marty-johnson59

TomAtkinsonArm (approved):
LGTM

PR #457, opened 2022-05-24
Bump to C++17

### LINKED ISSUE #454: Bump C++ Version to C++17
It seems that most compilers support most of C++17 features including Android. There are lots of benefits of moving to C++17 beyond just language features. We can replace parts of the project with better external libraries.

For instance, one pain point has been the scene graph. We can replace this with just the spatial tree (transform, AABB, parent -> n child). The component store can be backed by EnTT which provides efficient views over components in a given registry and is much more robust than our own component store implementation. EnTT would also allow for samples to define their own components to attach to the spatial tree which may help simplify samples for those reading the source code, especially if these samples require some bespoke behavior which is not in the framework.

This is one example, I'm sure others can think of many more improvements where C++17 can help improve the project.

The decision boils down to the implications of bumping C++ versions. What platforms or devices are effected by the change? Which users may not be able to compile the project? There are definitely many more questions to answer here.

It would be great to get everyone's opinion on this

TomAtkinsonArm:
## Description

Bumps the C++ version to 17. Those that are effected by this should test compilation on their specific platforms.

Fixes #454 

## General Checklist:

## Sample Checklist


asuessenbach:
Regarding the build failures: They seem to be caused by hpp_device.cpp and hpp_physical_device.cpp.
I don't see, why there are differences between C++14 and C++17, but to resolve them, one would either need to include `<../third_party/spdlog/include/spdlog/fmt/bundled/ostream.h>`, which seems a bit odd, or slightly modify hpp_device.cpp and hpp_physical_device.cpp, like this:
Replace
`	LOGI("Selected GPU: {}", gpu.get_properties().deviceName);`
with
`	LOGI("Selected GPU: {}", gpu.get_properties().deviceName.data());`
(Plus the same change at two other spots)
Modifiying the two files would probably be ok, but I'd really like to know why that's different to C++14.

asuessenbach:
Further investigations revealed: the spdlog has problems with std::string_view. And for C++17, vulkan.hpp provides a cast-operator from vk::ArrayWrapper1D to std::string_view.
Maybe updating that spdlog (which seems to be pretty old) might help as well, and be the better approach.

TomAtkinsonArm:
Hey @asuessenbach, those changes worked. I havent tested locally after the rebase so the CI may still fail. Ill revisit this in an hour

Issue #458, opened 2022-05-24
Goal: Dynamic Runtime Samples


TomAtkinsonArm:
As the project will become more modular over time it makes sense to reintroduce single sample compilation. This was scrapped a while back due to the setup of the framework and build system. This time around we should approach it in a different way to make sure that compilation time stay as fast as possible.

- Each Sample should compile as a runtime dynamic library which should be loaded and linked against
- Each platform should have a sample launcher which is driven by a `samples.json` file
- `samples.json` should contain all information about a sample and also the location or name of the dynamic library to load for a given sample
- Desktop platforms may benefit from an ImGUI based samples launcher

The benefits of this approach is that there is no cyclical dependency between compiling samples, launchers or the framework components. Each sample can be compiled individually with the only other dependency needed to run the sample being the launcher. Launchers can be extremely light weight and can take advantage off code already written and tested in framework modules.

Samples will no longer be classes or be tied to any `vkb::Platform` lifecycle. They should be able to be composed off any components they deem fit and run in a way which best demonstrates their lessons.

> This is purely pseudo code of a sample.cpp file

```c++
#include <components/events/events.hpp>
#include <components/windows/window.hpp>
#include <components/assts/models.hpp>
#include <components/input/manager.hpp>
#include <components/sample/defines.hpp>

static std::shared_ptr<SampleContext> sample_context;

sample::Result prepare(sample::Context context) {

   auto default_window =  windows::_default(context);

   input::Manager input;

   input.subscribe(default_window); // < subcribe to input events (cursor pos, key input)

   events::EventBus bus;

   bus.subscribe(context); // < possibly control a samples execution from external sources
   bus.subscribe(default_window); // < subscribe to window events (resize, move, close)

  // Init some vulkan stuff
  sample_context = {};

   bus.on<events::Update>([&](float delta_time){
      // Do some update stuff

     if (input.get_key<input::KeyCode::A>() == input::KeyAction::Down) {
        // move forward?
    }
   });

   bus.on<events::Draw>([&](){
      // ready to draw again
   })

  auto result = bus.run();

   return // do some error handling
}

sample::Result destroy() {
  // clean up code
  // might not be needed if we treat prepare as a main replacement
}
```

Each sample can choose how to configure its own running environment which allows for super simple samples or more robust samples. The current frameworks can be made into helper libraries which can be used or not. No need to compile the render graph for a hello world sample.

TomAtkinsonArm:
[...]

PR #462, opened 2022-05-27
Virtual File System

TomAtkinsonArm:
## Description

A small improvement on the previous virtual file system PR.

Added the RE2 library for better regex, the API is easier to use and it removes some of the compiler specific issues in C++ regex

**StackError**
- Hopefully a more descriptive error mechanism
- tracks filenames and line usage in a more controllable way than the current Vulkan Samples
- Eventually we should be able to produce detailed error stacks

**Strings**
- the start of replacing the strings header

**VFS**
- Errors moved from vfs::status to StackError
- Added a test - hope to add more in the future

**Testing**
- Adds catch2
- Adds basic unit tests
- Adds `vkb__register_test`
- Compile target `vkb_tests` bundles all tests into the same compile target
- Adds CI unit test workflow


CMakeLists.txt
```diff
@@ -19,46 +19,29 @@ cmake_minimum_required(VERSION 3.12)
...
+include(test_helpers)
+
+# options
+set(VKB_WARNINGS_AS_ERRORS ON CACHE BOOL "Enable Warnings as Errors")
+set(VKB_VALIDATION_LAYERS OFF CACHE BOOL "Enable validation layers for every application.")
+set(VKB_VALIDATION_LAYERS_GPU_ASSISTED OFF CACHE BOOL "Enable GPU assisted validation layers for every application.")
+set(VKB_VULKAN_DEBUG ON CACHE BOOL "Enable VK_EXT_debug_utils or VK_EXT_debug_marker if supported.")
+set(VKB_BUILD_SAMPLES ON CACHE BOOL "Enable generation and building of Vulkan best practice samples.")
+set(VKB_BUILD_TESTS OFF CACHE BOOL "Enable generation and building of Vulkan best practice tests.")
+set(VKB_WSI_SELECTION "XCB" CACHE STRING "Select WSI target (XCB, XLIB, WAYLAND, D2D)")
```
gpx1000, line 40: As we're making changes to the location, think we can switch from set to option for optional variables?  The advantage is in cmake 3.13 and up option wont create a separate second variable.  Also, we can use dependent options.

CMakeLists.txt
```diff
@@ -19,46 +19,29 @@ cmake_minimum_required(VERSION 3.12)
...
-
-if(VKB_BUILD_SAMPLES)
-    # Add vulkan samples
-    add_subdirectory(samples)
-endif()
+vkb__enable_testing()
 
-# Add vulkan app (runs all samples)
-add_subdirectory(app)
+add_subdirectory(components)
```
gpx1000: I don't remember where, but I've seen warnings in the past for no new line at end of documents.

gpx1000 (commented):
FWIW, we might be able to use the C++17 stl Filesystem classes and save a bit of platform dependent effort here.  However, in general I really like this approach you have working thus far.
Lemme know if you need any help.

CMakeLists.txt
TomAtkinsonArm: Sounds like a good edition. I prefer these variables in the main CMake file as it is easier to find all the options that you can use with a library. I expect in third parties for them to be in the main CMake, so it makes sense to follow this standard here

CMakeLists.txt
TomAtkinsonArm: Ah yeah, Im not sure why this is a big issue but it is an easy fix

gpx1000: Yeah, it's really not a big issue, just seen the warnings and now am a diligent new line at end of file devotee..

CMakeLists.txt
gpx1000: I whole-heartedly agree for the same reasons you're saying.  One of the subtle objections I have with using set instead of option is it doesn't really denote what's safe for end users to change on a whim in 3rd party libraries.

TomAtkinsonArm:
> FWIW, we might be able to use the C++17 stl Filesystem classes and save a bit of platform dependent effort here. However, in general I really like this approach you have working thus far.
Lemme know if you need any help.

If you have chance please add or groom tickets on the Vulkan Samples Framework Improvements board. Currently its mainly broad statements of what we may want but it would be great to add more and prioritise which areas need to be worked on - which are bad ideas and which are good....

For now i've just been working on cleaning up some of the peripheral stuff. If you would like to make an issue discussing how the C++17 filesystem can simplify this PR then id be all for it @gpx1000!

CMakeLists.txt
TomAtkinsonArm: Should work with option now. I will let the CI decide

CMakeLists.txt
TomAtkinsonArm: Added a new line

TomAtkinsonArm:
Force pushed the test framework to framework/v2.0 so that I can use it on other branches. It works as expected as it has been used in multiple PRs now. This should be one of the last force pushes to the branch.

TomAtkinsonArm:
test framework added to `framework/v2.0`. rebases and other fixes applied to this PR

SaschaWillems:
Wanted to give this a spin, bud oddly my project structure looks completely different using this PR, and no samples are generated that I could run:

image

What I did:
- Checkout this PR
- Delete old build/windows folder
- Re-run cmake
- Open project in VS2019

Looking at the CI it does seem to have the same problem.

Am I missing something?

TomAtkinsonArm:
> Wanted to give this a spin, bud oddly my project structure looks completely different using this PR, and no samples are generated that I could run

This is the correct behaviour. Making the framework changes propagate between every sample would be too much work and counter productive for reviewers (larger diffs) lots of manual testing.

The framework/v2.0 branch has no samples or framework folder.

The aim is to write tests and code for the parts of the framework which we deem need the most attention. The CI should then validate that the contribution works on a given system with the desired behaviour. Code should build and the tests should pass.  The workflows for this are Build Project and Unit Test Project.

The end goal for a sample is shown here #458. The new framework should be more of a sandbox of components that samples can be made from.

In terms of Vulkan, most will stay the same with some exterior API changes to simplify things. This PR #471 shows how I write unit tests against Vulkan usage. Extremely experimental but a promising example.

Although simple, PRs #469 and #470 show this process working fairly well.

If you would like to build and run the unit tests locally you can use:

```
cmake --build "build/windows" --target vkb_tests --config Release -j 2
```

and 

```
ctest --output-on-failure --test-dir "build/windows" -C Release
```

You can also run the test executables directly to see the catch2 output

SaschaWillems:
Thanks for the clarification. I think I get it now ;)

This builds fine for me with VS2019 on Windows, and all tests pass. So this looks good to me.

SaschaWillems (approved):

TomAtkinsonArm:
Merging with 1 approval

PR #464, opened 2022-05-30
WIP: Scene Graph, glTF loader and Image Loaders

### LINKED ISSUE #435: FI: Scene Graph
### What is the Problem?

The Scene Graph is a Tree containing positional Nodes which can have 1 or more components attached to them. The node hierarchy suffers from following the glTF structure to the letter. The issue with this is that not all models are made equal. Accounting for all possible combinations of glTF resources can make code less readable which is not great for an educational project.

### What is the proposed Solution?

Remove sub-meshes - Our implementation of sub-meshes cause havoc with no real benefit
Allow for Scene Optimization Passes - This would be similar to a compile pass where you traverse the tree and can cull or optimize nodes as you see fit.
Allow for custom component backing stores - One size fits all is never a great solution. We should be able to replace how scenes save components
Scene Graph -> Scene View - Currently it is hard to merge multiple models into the same scene due to its ownership of resources and its assumption that only one model will be loaded.

### Examples

TBC

### Issues

- Adjusting the scene may have repercussions on all samples

### LINKED ISSUE #437: FI: Image Loaders/Encoders/Decoders
### What is the Problem?

Image loading is baked into the Scene Graph Image Component constructor. This makes it difficult to use image loaders/encoders and decoders by them selves.

### What is the proposed Solution?

Each image third party should be wrapped by a common image loader interface and removed from the Scene Graph.

Images should be placed into the Scene Graph in their usable state and the Graph itself should not control how an image is mutated.

Removing image libraries from the framework will make the creation of platform specific asset pipelines easier in the future. Assets stored in samples should be stored in a raw optimized format and later preprocessed at compile time to be optimal for specific platforms.

A more complex multipurpose image decoder could be created from the existence of lower level decoders and encoders. This would be a stretch goal and is not necessary as part of this improvement

### Examples

```cpp
Data data;
fs.read_file("some_image.png", &data)

Image image;

PNGLoader loader;
loader.read_from_memory(data, &image);

Image encoded_image;

ASTCEncoder encoder;
encoder.encode_to(VK_ASTC_FORMAT, image, &encoded_image);

ASTCDecoder decoder;
decoder.decode_image(encoded_image, &image);
```

TomAtkinsonArm:
## Description

The previous scene graph required a lot of maintenance and was not flexible to an individual samples needs. It also allocated components using heap memory. This is ok for large amounts of data but for small components can quickly become inefficient. We also required implementing the component class which restricted how components could be used and what a component was.

On top of this, asset loading was done using RAII - in itself not a huge bad but this meant reusing asset loading logic was awkward. It is also harder to test the asset loading logic if it is buried inside a scene graph component.

### Problems
- Scene graph too restrictive
- Asset loading hidden inside scene graph components
- Heap allocations everywhere
- Hard to add new components for a given scenario
- glTF loader too confusing

### Solutions
- Replace scene graph with a simple node hierarchy
- Nodes are created with a backing store (Registry)
- Backing store using EnTT - Lots of active development, simplifies the rest of the scene graph a lot compared to the previous version, components are POD, we can perform updates using systems.
- glTF loader simplified

### TO DO
- More tests are needed, I am not happy with the current amount, tests should highlight most of the initial bugs associated with asset loading
- Readme for each component folder discussing techniques and decisions, possibly links to other learning resources
- Clean up PR description

closes #435
closes #437


TomAtkinsonArm:
test framework added to `framework/v2.0`. rebases and other fixes applied to this PR

PR #469, opened 2022-05-31
Event Channels

### LINKED ISSUE #468: FI: Event Channels
**What is the Problem?**

To reach #458 we will need some form of customizable event system which is used by all event emitting classes.

Channel communication may be the best solution here as channels can store a backlog of events whilst they are not being queried. Event channels can replace input events, surface events, user defined events and we can even emit events through proxy sources - AI active agents?

**Examples**

Example API usage.

```cpp
struct EventType {
  uint32_t value;
};

EventChannel<EventType> channel;

channel.push(EventType{12U});

// get the next event
auto event = channel.next();

// drain all events and only pay attention to the last - resizing surface, latest cursor position etc
auto last = channel.drain();
```

TomAtkinsonArm:
## Description

Small PR to add event channels to Vulkan Samples.

The idea is these can be used in other components to communicate events. Channels if mutex locked can also be used to push tasks to worker threads. A sample could use these to defer computation to another thread and then update the scene graph when data is returned on an output channel.

The first use case for these will be an `EventBus` scenario, window events and input events.

Fixes #468

Includes Unit Tests

### DIFF
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -1,0 +1,1 @@
+add_subdirectory(events)
--- a/components/events/CMakeLists.txt
+++ b/components/events/CMakeLists.txt
@@ -1,0 +1,10 @@
+add_library(components__events INTERFACE)
+target_include_directories(components__events INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/include)
+
+vkb__register_tests(
+  NAME "events_tests"
+  SRC
+    tests/channel.test.cpp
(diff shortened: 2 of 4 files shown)


TomAtkinsonArm:
@gpx1000 this should have probably been the proof of concept PR 😆

Really small one, implements a channel idea. tests should have full coverage so if the CI passes this should be a valid contribution without building and running

gpx1000 (approved):

PR #470, opened 2022-06-01
Input Manager

### LINKED ISSUE #433: FI: Input Manager
**What is the Problem?**

Vulkan Samples uses a `process_input` hook to propagate input events around the project. At the beginning this was a clean design as there was a strict control flow. Over time this has caused several components to reinvent the wheel when it comes to input state tracking and it has stopped new components from being created as there is extra overhead designing a state tracking system.

Add an input manager which processes all I/O events and hows a central state tracking mechanism. Other parts of the project will then request for this state manager and query the current state of key presses, cursor position etc.

**Examples**

```cpp
if(input.get_key_pressed<KeyCode::A>()) {
    std::cout << "A pressed\n";
}
```

**Issues**

- May introduce a singleton which means that the input manager can not be easily changed
- This would replace a large amount of the input handling logic which may cause regressions
- Would require Event channel implementation

TomAtkinsonArm:
## Description

Input logic is tracked in lots of different areas of the current Vulkan Samples framework. This class aims to act as a sole container for input logic. The touch logic is a little lacking for now but this can be improved when revisiting touch events.

Full coverage with unit tests. If the CI passes this contribution should only need a code review

Contains changes made in #469 

fixes #433


SaschaWillems (approved):
Sorry for the late review. LGTM. Compiles fine on windows and all tests pass :)

gary-sweet (approved):

PR #471, opened 2022-06-01
WIP: Vulkan Instance Builder

### LINKED ISSUE #466: FI: Vulkan Instance
**What is the Problem?**

Samples will need to create a Vulkan Instance. The usual route of passing vectors of layers and extensions to toggle is not very dynamic and tends to lead to bad code in the future. It would be great to have some form of InstanceBuilder to simplify this process.

Add an instance builder class which has some helpful API patterns to simplify the process of enabling and disabling layers and extensions.

**Examples**

Example API usage. Allows a sample to have full control over instance creation with the least amount of code.

```cpp
InstanceBuilder builder;

builder
    .target_vulkan_version(1, 2, 0);

builder
    .enable_layer_if_exists("VK_LAYER_NAME")
    .enable_layer_if_exists("VK_LAYER_NAME_1")
    .enable_layer_if_exists("VK_LAYER_NAME_2")
    .enable_layer_or_error("VK_LAYER_NAME_3");

builder
   // maintain a library of common features that samples need to enable - default_sample_features_func
   .apply(some_builder_mutation_func_declared_somewhere_else)
   // some custom mutation function
   .apply([](InstanceBuilder& builder) {
      builder.enable_extension<SomeExtension>([](const VkExtensionFeatures& supported_features, VkExtensionFeatures* requested_features){
          // check for device support and enable features
         // we need some useful error handling here StackError?

      });
   });

Instance instance = builder.build();
```

TomAtkinsonArm:
## Description

Adds an instance builder to allow a sample to completely customize how a vulkan instance is constructed.

Also adds a vulkan test framework so that we can intercept vulkan calls and make sure that our API usage is as expected. I've never been able to do Vulkan TDD before but I think this solution makes it viable. It works by overriding global volk `PFN_vkFunction` declarations so that we can intercept vulkan calls and validate our actual vulkan usage.

This method of testing will never replace OOB testing against an actual driver. But at least with good test coverage we can confidently say that a specific Vulkan component should work as expected against the Vulkan specification.

fixes #466

### DIFF
--- a/bldsys/cmake/test_helpers.cmake
+++ b/bldsys/cmake/test_helpers.cmake
@@ -26,7 +26,7 @@ add_custom_target(vkb_tests)
 function(vkb__register_tests)
     set(options)  
     set(oneValueArgs NAME)
-    set(multiValueArgs SRC LIBS)
+    set(multiValueArgs SRC LINK_LIBS INCLUDE_DIRS)
 
     if(NOT ((CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND BUILD_TESTING) OR VKB_BUILD_TESTS))
         return() # testing not enabled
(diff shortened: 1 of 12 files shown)


TomAtkinsonArm:
@gpx1000 thoughts on the Vulkan unit tests here?

PR #473, opened 2022-06-05
Fix tmax in the ray_queries sample

KonstantinSeurer:
https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#ray-intersection-candidate-determination:

We need to set tmax to 1 because the length of direction already accounts for the distance to the light.

### DIFF
--- a/shaders/ray_queries/ray_shadow.frag
+++ b/shaders/ray_queries/ray_shadow.frag
@@ -87,14 +87,13 @@ bool intersects_light(vec3 light_origin, vec3 pos)
 {
 	const float tmin = 0.01, tmax = 1000;
 	const vec3  direction = light_origin - pos;
-	const float distance  = sqrt(dot(direction, direction));
 
 	rayQueryEXT query;
 
 	// The following runs the actual ray query
 	// For performance, use gl_RayFlagsTerminateOnFirstHitEXT, since we only need to know
 	// whether an intersection exists, and not necessarily any particular intersection
-	rayQueryInitializeEXT(query, topLevelAS, gl_RayFlagsTerminateOnFirstHitEXT, 0xFF, pos, tmin, direction.xyz, distance);
+	rayQueryInitializeEXT(query, topLevelAS, gl_RayFlagsTerminateOnFirstHitEXT, 0xFF, pos, tmin, direction.xyz, 1.0);
 	// The following is the canonical way of using ray Queries from the fragment shader when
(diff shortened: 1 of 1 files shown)


JoseEmilio-ARM:
> https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#ray-intersection-candidate-determination:
> 
> We need to set tmax to 1 because the length of direction already accounts for the distance to the light.

Thanks for pointing this out! However, not sure I follow, is `tmax = 1` a special case? Could you please quote the part of the spec that you are referring to? In this case, we are using ray query with `gl_RayFlagsTerminateOnFirstHitEXT`, so do not really need to track `tmax`. I agree than rather than `distance` maybe it could be `tmax = 1000` (variable which is currently [unused](https://github.com/KhronosGroup/Vulkan-Samples/pull/473/files#diff-458b20d617252a9a8c02d63fdf59dec32e54f5c6b5472ef7407655d21674965fR88))

KonstantinSeurer:
There is a section that defines t to be the z coordinate in ray space divided by the length of the unnormalized ray direction vector. So if we want to trace from some world_pos to some light_pos and out direction is light_pos - world_pos, tmax has to be one since the value of t for light_pos would be 1.

gpx1000:
Good catch on this.  Jose is right that the first hit usage means it's not really need to track tmax.  However, I wonder if it would be clearer and thus a better sample if we did switch tmax to be what's used here and delete distance?  Setting tmax to 1.0 seems to be correct too.

KonstantinSeurer:
The point is that we either need to normalize the direction vector and set tmax to distance or keep the direction vector the way it currently is and set tmax to 1.0. I is correct that we do not care about t because we are using `gl_RayFlagsTerminateOnFirstHitEXT` but we need tmax to be correct in order to discard any potential intersection past the light.

JoseEmilio-ARM:
Yes sorry, got confused, this is a positional light so we need a shorter `tmax`. Happy with either solution, did you have a chance to test it on a couple of platforms?

KonstantinSeurer:
Setting `tmax` to 1 is probably the more elegant/optimal solution as it avoids calculating the length. I only tested it on AMD with RADV and I don't have a ray tracing capable Nvidia GPU.

SaschaWillems (approved):
LGTM and works fine on Windows 10 with my NV RTX 2060.

gpx1000 (approved):

PR #477, opened 2022-06-18
Display sample name instead of ID in UI

SaschaWillems:
## Description

Samples currently only display the ID (set in the CMakeLists.txt) in the UI. This may be cryptic, e.g.:

image

This PR fixes that and displays the actual name instead:

image

## General Checklist:

### DIFF
--- a/framework/platform/platform.cpp
+++ b/framework/platform/platform.cpp
@@ -369,7 +369,7 @@ bool Platform::start_app()
 
 	active_app = requested_app_info->create();
 
-	active_app->set_name(requested_app_info->id);
+	active_app->set_name(requested_app_info->name);
 
 	if (!active_app)
 	{
--- a/app/apps/apps.h
+++ b/app/apps/apps.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2020-2021, Arm Limited and Contributors
+/* Copyright (c) 2020-2022, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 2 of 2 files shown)


asuessenbach:
I like this approach.
But wouldn't it be cleaner to remove the name from SampleInfo, then?

gpx1000 (approved):

framework/platform/platform.cpp
```diff
@@ -369,7 +369,7 @@ bool Platform::start_app()
 
 	active_app = requested_app_info->create();
 
-	active_app->set_name(requested_app_info->id);
```
TomAtkinsonArm, line 372: We might want some method on AppInfo which returns the id if the name is empty?

Something like:

```
active_app->set_name(requested_app_info->human_readable_name());
```

TomAtkinsonArm (changes requested):
LGTM. Suggested a nice to have. If we know for a fact that every sample has a name then this nice to have can be ignored

framework/platform/platform.cpp
asuessenbach: I think it would be better if the name in AppInfo was mandatory.

gpx1000: I tend to agree that AppInfo should be mandatory.  We'll need to ensure all legacy samples are updated for it and it makes its way into acceptance criteria for new sample submissions.  An automated check would be helpful from our CI/CD

marty-johnson59:
This will no longer be an issue once the updated framework (current WIP) is complete.  Closing...can revisit if not fully addressed with the new FW

PR #482, opened 2022-07-07
Add Warnings As Errors

### LINKED ISSUE #478: Framework v2.0 warning as error level
Looking at #320, would it be useful to start framework v2.0 at a specific warning level? I remember windows being quite verbose when bumping the level but this may be a good time to do it

TomAtkinsonArm:
## Description

Adds warnings as errors.

Adds `vkb__register_component` to guarantee consistent library creation. Warnings are set at a target level which should make sure third party warnings are ignored. If this is not the case then we will need to create warning wrappers similar to the master branch.

Fixes #478

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,7 @@ include(global_config)
 include(sample_helper)
 include(check_atomic)
 include(test_helpers)
+include(component_helpers)
 
 # options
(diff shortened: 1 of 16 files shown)


TomAtkinsonArm:
@asuessenbach if this merges first I can fix the folder issue that you found in the other PR in a new PR

SaschaWillems (approved):
LGTM and compiles fine with Visual Studio 2019.

PR #483, opened 2022-07-07
 Image Loaders

### LINKED ISSUE #437: FI: Image Loaders/Encoders/Decoders

TomAtkinsonArm:
## Description

Previously, Image loaders where RAII based. This becomes quite confusing when loading assets as you must choose the correct constructor. Also, not all loaders work the same so you need to have edge cases here and there. Its also confusing to read and understand. Decoding an image would also be handled by constructors...

In this PR image loaders are separate entities which follow a simple interface. If an image loader needs to be configured in a specific way then the loaders constructor should take options. An image is now a simple struct of data and can be passed between multiple loaders.

Each image loader has its own test cases and takes assets from the `tests` folder. These tests will never be 100% coverage when compression and other factors are involved, but they should act as a decent base line for image loading. ASTC is not completely finished but this may be attempted at a later date. Currently we only require the decoding aspect.

Fixes #437


asuessenbach:
>Previously, Image loaders where RAII based.

I would prefer to stay with RAII.

When I try to use your code (gh pr checkout 483), my VS solution seems to be completely scrambled. Am I missing something to get CMake do-the-right-thing™?
image

TomAtkinsonArm:
> I would prefer to stay with RAII.

Makes sense when the asset is loaded into memory. Images can be RAII.

But the process of loading the image should not be RAII, makes code overly complicated and doesn't have a clear separation of concerns. Also harder to test the loading code and harder to build more things from it.

The aim is to have an asset pipeline so these loaders will be reused in smaller executables which don't care about anything else in the project.

> When I try to use your code (gh pr checkout 483), my VS solution seems to be completely scrambled. Am I missing something to get CMake do-the-right-thing™?

Ah im using VSCode so the folders are just normal for me. This is a CMake fix. I believe its this sort of thing `set_property(TARGET volk PROPERTY FOLDER "ThirdParty")`.

asuessenbach:
> Ah im using VSCode so the folders are just normal for me.

I think, I don't get what you mean... where's the vulkan_samples gone? And, more important, where would I find the files from your PR, e.g. components/assets/include/components/image.cpp?

TomAtkinsonArm:
`.cpp` are under `components/<component>/src`

For this PR https://github.com/TomAtkinsonArm/Vulkan-Samples/tree/images/components/assets/src/images

I bundled asset loaders into the asset component as the theme is loading external assets. Alternatively, we could have a component for each type of loader

PR #485, opened 2022-07-07
Event Bus

TomAtkinsonArm:
## Description

An observer is added to the event bus through attach(observer). Once attached, an observer can register event listeners (each<EventType>(), last<EventType>()) and request ChannelSender<EventTypes>. Each step of the EventBus calls update() on its observers. Which in turn allows an observer to submit events to the bus. After this, the bus then processes all event callbacks with a stream of events.

The combination of these actions will allow for inter component communication without any hard links. This allows samples to create and organize components in anyway they deem fit. We can also create sets of default components used in a sample.

### Example

```c++
void main() {
    auto window = std::make_shared<Window>();
    auto input_manager = std::make_shared<InputManager>();

    EventBus bus;

    bus

        // recieves window events on process by calling window->update();
        .attach(window) 

        // consumes window events on process as input_manager->attach(bus) registered event callbacks
        .attach(input_manager); 
    
   // A custom user defined event callback which closes the app on an escape key press
   // These callbacks can be used to communicate between core components and reduce the manually pluming needed in each sample
    {
        auto sender = bus.request_sender<AppCloseEvent>();

        bus.each<KeyPress>([auto press_sender = std::move(sender)](const KeyPress& event){
            if (event.key == KeyCode::ESC) {
                press_sender->push(AppCloseEvent{
                    .force = true;
                });
            }
        });
    }
    
    bus.process();
}
```

Works towards #458 

We can also use this to synchronize application stages such as `SurfaceReady` , `SceneReady` or `CloseSample`.  Samples can design custom application flows for their own needs using events. We can build on top of this to add pipelines - a sequential set of events which are automatically fired (`Init`, `Update`, `Draw`)

### DIFF
--- a/components/events/src/event_bus.cpp
+++ b/components/events/src/event_bus.cpp
@@ -1,0 +1,73 @@
+/* Copyright (c) 2022, Arm Limited and Contributors
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 the "License";
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "components/events/event_bus.hpp"
+
+#include <cassert>
+
+namespace components
+{
+namespace events
+{
+EventBus &EventBus::attach(std::weak_ptr<EventObserver> &&observer)
+{
+	for (auto it = m_observers.begin(); it != m_observers.end(); it++)
+	{
+		if (same_ptr(*it, observer))
+		{
+			return *this;
+		}
+	}
+
+	if (auto shared_observer = observer.lock())
+	{
+		shared_observer->attach(*this);
+		m_observers.push_back(shared_observer);
+	}
+
+	return *this;
+}
+
+void EventBus::process()
+{
+	for (auto it = m_observers.begin(); it != m_observers.end(); it++)
+	{
+		if ((*it).expired())
+		{
+			m_observers.erase(it);
+			it++;
+			continue;
+		}
+
+		if (auto shared = it->lock())
+		{
+			shared->update();
+		}
+	}
+
+	for (auto &it : m_each_callbacks)
+	{
+		it.second->process_each();
+	}
+
+	for (auto &it : m_last_callbacks)
+	{
+		it.second->process_last();
+	}
+}
+}        // namespace events
+}        // namespace components
--- a/components/events/src/event_bus.cpp
+++ b/components/events/src/event_bus.cpp
@@ -25,15 +25,7 @@ namespace events
 {
 EventBus &EventBus::attach(std::weak_ptr<EventObserver> &&observer)
 {
-	for (auto it = m_observers.begin(); it != m_observers.end(); it++)
-	{
-		auto exists = same_ptr(*it, observer);
-		assert(!exists && "attempting to attach an existing observer");
-		if (exists)
-		{
-			return *this;
-		}
-	}
+	assert((std::find_if(m_observers.begin(), m_observers.end(), [this, observer](const auto &obs) { return same_ptr(obs, observer); }) == m_observers.end()) && "attempting to attach an existing observer");
 
 	if (auto shared_observer = observer.lock())
 	{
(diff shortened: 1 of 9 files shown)


components/events/src/event_bus.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include <cassert>
+
+namespace components
+{
+namespace events
+{
+EventBus &EventBus::attach(std::weak_ptr<EventObserver> &&observer)
+{
+	for (auto it = m_observers.begin(); it != m_observers.end(); it++)
```
asuessenbach: Is adding the same observer multiple times supposed to be valid? Or should you just assert that it's not yet attached?

components/events/src/event_bus.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+void EventBus::process()
+{
+	for (auto it = m_observers.begin(); it != m_observers.end(); it++)
+	{
+		if ((*it).expired())
+		{
+			m_observers.erase(it);
+			it++;
```
asuessenbach: Calling `erase(it)` invalidates the iterator!
Just do `it = m_observers.erase(it);` instead, and adjust the loop accordingly

asuessenbach (commented):
Some brief comments:
- The `Channel` member functions `receiver()` and `sender()` actually create something... should be named `create_receiver()` and `create_sender()`, then?
- The class `ChannelReceiver` seems to be designed for multi-threaded usage (it holds a `std::mutex`), but separate operations `has_next()` and `next()` doesn't fit well to multi-threading. Maybe a single function `std::optional<Type> next()` would be better. Also, `size()` might tell you something, that's outdated a nanosecond later.

TomAtkinsonArm:
> Some brief comments:
> 
> The `Channel` member functions `receiver()` and `sender()` actually create something... should be named `create_receiver()` and `create_sender()`, then?

agreed and changed!

> The class `ChannelReceiver` seems to be designed for multi-threaded usage (it holds a `std::mutex`), but separate operations `has_next()` and `next()` doesn't fit well to multi-threading. Maybe a single function `std::optional<Type> next()` would be better. Also, `size()` might tell you something, that's outdated a nanosecond later.

Agreed and changed! `size()` is only useful for testing. I don't think we should be able to use it in actual code. For now I have wrapped it in a test ifdef but there may be a better way of dealing with this.

components/events/src/event_bus.cpp
TomAtkinsonArm: Adding the same observer object multiple times is technically valid behavior as we can just ignore the addition. However, I have added an assertion as this may indicate something else is wrong in the event bus usage.

asuessenbach: Maybe a better assert, without explicit loop:
`assert((std::find_if(m_observers.begin(), m_observers.end(), [observer](auto const & obs){return same_ptr(obs, observer);}) == m_observers.end()) && "attempting...");`

components/events/src/event_bus.cpp
```diff
@@ -48,8 +45,7 @@ void EventBus::process()
 	{
 		if ((*it).expired())
 		{
-			m_observers.erase(it);
-			it++;
+			it = m_observers.erase(it);
```
asuessenbach, line 45: Here, `it` is already advanced to the next observer.
That is, you should move the `it++` from the loop to the bottom of its body:
```
for ( auto it = m_observers.begin(); it != m_observers.end();)
{
    ....
    ++it;
}
```

TomAtkinsonArm: Ill need to add a test for this

TomAtkinsonArm: Yep, this was the correct fix and the test confirms this! see `expire an observer before process` in `events_bus.test.cpp`

TomAtkinsonArm:
@asuessenbach happy to approve this?

gpx1000 (approved):
great direction!

SaschaWillems:
This looks good and compiles fine for me. But I'm unsure on how to run the tests for it. I see that there are tests for the event bus and input manager, but I can't find them in the solution. There is a vkb_tests project in the VS solution that I can compile, but I can't run it.

asuessenbach:
I have the same issue as Sascha: vbk_tests doesn't run anything.
Moreover: all the test projects should probably be moved into a new directory.
And: Where's the vulkan_samples and all the samples and the framework gone? And none of your new files can be found in the solution.

TomAtkinsonArm:
`vbk_tests` is a proxy target used to target all available tests.

In this branch the following test executables are available:
- `common_tests`
- `events_tests`
- `virtual_file_system_tests`

`events_tests` covers channels and the Event Bus

These tests can be run individually, or all together. To run a test individually, run the test executable from the projects root directory, this is similar to running `vulkan_samples`. To run all together you may use CTest `ctest --output-on-failure --test-dir=<your build dir>`

> all the test projects should probably be moved into a new directory.

This is partially cleaned up in #496 

> Where's the vulkan_samples and all the samples and the framework gone? And none of your new files can be found in the solution.

Currently this has been removed as I wanted to add the framework back with good test coverage and some simplified classes and a new build system. This stuff will be added back at a later date. The closest replacement for `vulkan_samples` can be found in #493

TomAtkinsonArm:
All Visual Studio folder related issues are being addressed in #496, for now these issues should be ignored in PRs that are not rebased on #496. You can use the CMake target view or folder view in Visual Studio to see the actual folder structure and all source files.

TomAtkinsonArm:
Merging for now as the final comments are addressed in newer PRs

PR #487, opened 2022-07-08
Event Pipelines

TomAtkinsonArm:
## Description

Builds on-top of #485.

The Event Bus allows for communication between components and the firing of one off events. It does not allow for a Sample to continually fire and react to a given sequence of events.

Pipelines aim to allow stages to continually run without the need for an Event Observer to submit the event manually. This allows a sample to have a very compact view of its execution flow. Samples can now have completely different execution flows.

### Example

```c++
// hello world

struct Context{
    // ... Global Vulkan Handles
};

struct FrameContext {
    // ... per frame Vulkan handles
};

class FrameManager : public EventObserver {
    // ... React to window updates, recreate the Swapchain, recreate frame data
};

std::shared_ptr<ExitOnEscape> exit_on_escape() {
    // ... default exit behaviour defined some where else in the project
    static std::shared_ptr<ExitOnEscape> observer = std::make_shared<ExitOnEscape>();
    return observer;
}

// ... project defined update stage
struct UpdateEvent {
    float delta_time;
};

class Update : public TypedPipelineStage<UpdateEvent>;

// ... project defined draw stage
struct DrawEvent {
    CommandPool& pool;
};

class Draw : public TypedPipelineStage<DrawEvent>;

void sample_main(void* platform_context) {
    auto window = windows::_default(platform_context, "Hello Triangle");

    Context context{};

    // ... init Vulkan stuff

    auto frame_manager = std::make_shared<FrameManager>(context);
    
    EventPipeline pipeline;

    pipeline
      .attach(window)
      .attach(frame_manager)
      .attach(exit_on_escape())
      .then(std::make_shared<Update>())
      .then(std::make_shared<Draw>());

    // ... step pipeline
    while(true) {
        // process will eventually have error handling
        // its likely that this statement will be
        // return pipeline.run();
        // where run() internally has a loop and return on error or close
        pipeline.process(); 
    }
}
```


SaschaWillems:
Technically this looks great, but I'm not sure if we want to use the term pipeline for this. I fear that people could get confused with having graphics pipelines and also pipelines for event related stuff?

TomAtkinsonArm:
> Technically this looks great, but I'm not sure if we want to use the term pipeline for this. I fear that people could get confused with having graphics pipelines and also pipelines for event related stuff?

@SaschaWillems `EventPipeline`?

components/events/src/pipelines.cpp
```diff
@@ -0,0 +1,68 @@
+/* Copyright (c) 2022, Arm Limited and Contributors
```
asuessenbach: This file should be renamed to event_pipeline.cpp

components/events/include/components/events/event_pipeline.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+class EventPipeline : public EventBus
+{
+  public:
+	EventPipeline()          = default;
+	virtual ~EventPipeline() = default;
+
+	EventPipeline &once(std::unique_ptr<EventPipelineStage> &&stage);
+
+	EventPipeline &then(std::unique_ptr<EventPipelineStage> &&stage);
```
asuessenbach: `add_once` and `add_always`, maybe?

TomAtkinsonArm:
Rebased on events_bus

components/events/include/components/events/event_pipeline.hpp
TomAtkinsonArm: Changed and added a comment.

Ill start to use this pattern more often as it is much more obvious on what is happening. Similar to the `create_` in a different PR

SaschaWillems:
Yes, I'm fine with `EventPipeline`. That's a clear distinction between this and Vulkan's pipeline. The PR looks fine to me, but just as with the event bus PR I can't seem to find or run the tests for this in VS2019.

TomAtkinsonArm:
> The PR looks fine to me, but just as with the event bus PR I can't seem to find or run the tests for this in VS2019.

Try `ctest --output-on-failure --test-dir=<build/folder>` also configure using `-DVKB_BUILD_TESTS=ON`

SaschaWillems:
Tried running that, but it just tells me that no tests were found. Even when building with `-DVKB_BUILD_TESTS=ON`. Maybe we can take a look at this in today's call.

TomAtkinsonArm:
In the CI we use:

```
cmake -H"." -B"build/windows" -DVKB_BUILD_TESTS=ON
cmake --build "build/windows" --target vkb_tests --config Debug --parallel
ctest --output-on-failure --test-dir "build/windows" -C Debug
```

This also works locally for me

TomAtkinsonArm:
Rebased @SaschaWillems @asuessenbach is this ready to be approved?

TomAtkinsonArm:
Comments resolved, merging for now. The concept will evolve over time and more than likely be replaced in the near future if it does not fit its desired use case.

PR #488, opened 2022-07-08
Platform Context and CUSTOM_MAIN

TomAtkinsonArm:
## Description

In VFS there is mention of a  `void* context`. This context should contain all information available for a given platform so that components can have specialized implementations for each primary OS. This PR introduces very basic contexts for each OS.

There is a likely hood that we want to reuse cross platform components in many executables. `CUSTOM_MAIN` aims to reduce duplication of a cross platform main.

### Example

```
CUSTOM_MAIN(context)
{
	// _default is a temporary pattern, this is likely to be replaced in the future

	auto window = windows::_default(context, "A Window");        // < window platform default
	auto fs     = vfs::_default(context);                        // < virtual fs platform default
	auto camera = cameras::_default(context);                    // < get a devices camera AR?
}
```

### DIFF
--- a/bldsys/cmake/test_helpers.cmake
+++ b/bldsys/cmake/test_helpers.cmake
@@ -74,7 +74,11 @@ function(vkb__register_tests_no_catch2)
         message(FATAL_ERROR "One or more source files must be added to vkb__register_tests")
     endif()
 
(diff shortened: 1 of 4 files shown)


TomAtkinsonArm:
Superseded by #490

PR #490, opened 2022-07-08
Dynamic Runtime Samples

TomAtkinsonArm:
## Description

There is a high chance that we want to reuse cross platform components in many executables. `CUSTOM_MAIN` aims to reduce duplication of a cross platform main.

```
CUSTOM_MAIN(context)
{
	// _default is a temporary pattern, this is likely to be replaced in the future

	auto window = windows::_default(context, "A Window");        // < window platform default
	auto fs     = vfs::_default(context);                        // < virtual fs platform default
	auto camera = cameras::_default(context);                    // < get a devices camera AR?
}
```

Adds `sample_main(context)`. Works similar to a normal `main` function except it hides the cross platform support.

The minimum code needed for a sample is

```
#include "components/platform/sample_main.hpp"

#include <iostream>

int sample_main(const components::PlatformContext *context)
{
	return true;
}
```

Later a Sample Launcher will read library paths and sample info from a JSON file. Allow the user to select a sample to run and then run the sample by loading the dynamic library and calling  `sample_main`. We can extend this to allow for CLI configuration and more.

### DIFF
--- a/components/platform/include/components/platform/dl.hpp
+++ b/components/platform/include/components/platform/dl.hpp
@@ -17,7 +17,7 @@
 
 #pragma once
 
-#include <string>
+#include <string_view>
 
 #define EXPORT
 #define IMPORT
@@ -42,7 +42,7 @@ namespace dl
  * @param name library name
  * @return std::string OS library name
  */
-std::string os_library_name(const std::string &name);
+std::string os_library_name(const std::string_view &name);
 
 /**
  * @brief Open a dynamic library
@@ -50,7 +50,7 @@ std::string os_library_name(const std::string &name);
  * @param library_path library path
  * @return void* library handle
  */
-void *open_library(const char *library_path);
+void *open_library(const std::string_view &library_path);
 
 /**
  * @brief Load a function ptr from a library
@@ -59,7 +59,7 @@ void *open_library(const char *library_path);
  * @param function_name function name
  * @return void* function ptr
  */
-void *load_function(void *library, const char *function_name);
+void *load_function(void *library, const std::string_view &function_name);
 
 /**
  * @brief Load a function ptr from a library and cast to a specific type
@@ -70,7 +70,7 @@ void *load_function(void *library, const char *function_name);
  * @return PFN function pointer
  */
 template <typename PFN>
-PFN load_function(void *library, const char *function_name)
+PFN load_function(void *library, const std::string_view &function_name)
 {
 	return reinterpret_cast<PFN>(load_function(library, function_name));
 }
--- a/components/platform/include/components/platform/dl.hpp
+++ b/components/platform/include/components/platform/dl.hpp
@@ -36,12 +36,39 @@ namespace components
 {
 namespace dl
 {
+/**
+ * @brief Convert a library name to an OS specific library name
+ * 
+ * @param name library name
+ * @return std::string OS library name
+ */
 std::string os_library_name(const std::string &name);
 
+/**
+ * @brief Open a dynamic library
+ * 
+ * @param library_path library path
+ * @return void* library handle
+ */
 void *open_library(const char *library_path);
 
+/**
+ * @brief Load a function ptr from a library
+ * 
+ * @param library library handle
+ * @param function_name function name
+ * @return void* function ptr
+ */
 void *load_function(void *library, const char *function_name);
 
+/**
+ * @brief Load a function ptr from a library and cast to a specific type
+ * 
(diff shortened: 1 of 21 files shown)


components/platform/include/components/platform/dl.hpp
```diff
@@ -0,0 +1,78 @@
...
+ */
+std::string os_library_name(const std::string &name);
+
+/**
+ * @brief Open a dynamic library
+ * 
+ * @param library_path library path
+ * @return void* library handle
+ */
+void *open_library(const char *library_path);
```
asuessenbach: Mixing `std::string` and `char *` in one interface should be done with a good reason.
I think, you should consistently use the one or the other type here.

components/platform/include/components/platform/dl.hpp
```diff
@@ -0,0 +1,78 @@
+/* Copyright (c) 2022, Arm Limited and Contributors
```
asuessenbach, line 1: Are those files here really supposed to be at `components/platform/include/components/platform`, or just at `components/platform/include` ?

components/platform/tests/dummy_sample.cpp
```diff
@@ -0,0 +1,37 @@
...
+{
+#if defined(_WIN32)
+	if (dynamic_cast<components::WindowsContext *>(context) == nullptr)
+#elif defined(__ANDROID__)
+	if (dynamic_cast<components::AndroidContext *>(context) == nullptr)
+#elif defined(__APPLE__) || defined(__MACH__)
+	if (dynamic_cast<components::MacOSXContext *>(context) == nullptr)
+#elif defined(__linux__)
+	if (dynamic_cast<components::UnixContext *>(context) == nullptr)
+#endif
```
asuessenbach: Shouldn't such platform specific stuff be hidden by the PlatformContext approach?
Maybe just call some context->is_valid() or so?

components/platform/include/components/platform/dl.hpp
TomAtkinsonArm: Theres two layers to this.

`components/platform/` contains all files for the `platform` component. Everything in this folder is for the platform component and only the platform component.

`components/platform/include/components/platform` preserves the include path in source `#include <components/platform/platform.hpp>`, most IDEs should collapse this to act like a single folder.

For now components are small so it looks a little overkill. However, when these components inevitably grow in size, this structure should scale with that growth.

I'm open to suggestions on how to improve this but I am hoping to merge the current open MRs before trying reorganizations of source files.

components/platform/include/components/platform/dl.hpp
TomAtkinsonArm: replaced with `std::string_view` as these functions do not need to own the underlying strings memory

components/platform/tests/dummy_sample.cpp
TomAtkinsonArm: added `const WindowsContext * cast() const`. Platform contexts should only be mutable in the function they are defined. Any thing other than this they are read only. The idea is that they just contain helpers and utilities needed for specific platforms

TomAtkinsonArm: you would then use a platform with `if(const auto* platform = context->cast()) { ...platform specific functions }`

TomAtkinsonArm:
Closed #488 as this PR includes its changes. Updated description

components/platform/include/components/platform/dl.hpp
gpx1000: The only real downside potential is the character limit in windows on file paths.  I'm fine with the path in this format as it makes the most sense to me.  However, we might consider renaming them to something else to avoid confusion.  Personally this is one I don't care about one way or the other.

components/platform/include/components/platform/platforms/android.hpp
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	android_app *android_app;
+
+  private:
+	std::vector<std::string> m_arguments;
+};
+}        // namespace components
+
+// TODO: get arguments from bundle
```
gpx1000, line 48: Are there any arguments that are actually worth getting for our samples?

gpx1000 (approved):

components/platform/include/components/platform/platforms/android.hpp
TomAtkinsonArm: Better than passing things statically. But I may deal with this using a different mechanism in the future (a native app wrapper for any sample)

PR #491, opened 2022-07-08
Headless and GLFW Windows

TomAtkinsonArm:
## Description

Adds Headless and GLFW windows back to the project. Events are forwarded though the Event Bus interface.

No need to pass Platform anymore as the Event Bus acts as the communication mechanism.

Headless and GLFW tests included. GLFW tests will not work in the CI so these require `VKB_TEST_GLFW` locally. All GLFW tests pass locally. We could add a small compile time executable to automatically trigger GLFW tests if a GPU is available.

### DIFF
--- a/bldsys/cmake/component_helpers.cmake
+++ b/bldsys/cmake/component_helpers.cmake
@@ -44,7 +44,7 @@ function(vkb__register_component)
     if(TARGET_SRC) # Create static library
         message("ADDING STATIC: vkb__${TARGET_NAME}")
 
-        add_library("vkb__${TARGET_NAME}" STATIC ${TARGET_SRC} ${TARGET_HEADERS})
+        add_library("vkb__${TARGET_NAME}" STATIC ${TARGET_SRC})
 
         if(TARGET_LINK_LIBS)
             target_link_libraries("vkb__${TARGET_NAME}" PUBLIC ${TARGET_LINK_LIBS})
--- a/components/events/include/components/events/event_types.hpp
+++ b/components/events/include/components/events/event_types.hpp
@@ -152,8 +152,8 @@ struct KeyEvent
 
 struct CursorPositionEvent
 {
-	uint32_t pos_x;
-	uint32_t pos_y;
+	float x;
+	float y;
 };
 
 enum class TouchAction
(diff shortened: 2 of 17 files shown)


bldsys/cmake/component_helpers.cmake
```diff
@@ -16,21 +16,35 @@
...
     if(TARGET_NAME STREQUAL "")
         message(FATAL_ERROR "NAME must be defined in vkb__register_tests")
     endif()
 
-    if(TARGET_SRC) ## Create static library
+    if(TARGET_SRC) # Create static library
         message("ADDING STATIC: vkb__${TARGET_NAME}")
 
-        add_library("vkb__${TARGET_NAME}" STATIC ${TARGET_SRC})
+        add_library("vkb__${TARGET_NAME}" STATIC ${TARGET_SRC} ${TARGET_HEADERS})
```
gpx1000: Why add headers to the target explicitly?  Can't we just set the include directory?

components/events/include/components/events/event_types.hpp
```diff
@@ -152,8 +152,8 @@ struct KeyEvent
 
 struct CursorPositionEvent
 {
-	uint32_t pos_x;
-	uint32_t pos_y;
+	float x;
+	float y;
 };
```
gpx1000, line 157: I'm debating suggesting making this a vec3 for xyz.  The use case would be lightfield or XR development where some input is in 3d. However, this is a very niche corner case so probably best left to the sample to implement when needed.

components/vfs/src/root_file_system.cpp
```diff
@@ -198,13 +216,15 @@ StackErrorPtr RootFileSystem::enumerate_files(const std::string &file_path, cons
...
 
-	*files = files_with_extension;
+	*files = std::move(files_with_extension);
 
 	return nullptr;
 }
 
 StackErrorPtr RootFileSystem::enumerate_files_recursive(const std::string &file_path, const std::string &extension, std::vector<std::string> *files)
 {
+	assert(files);
```
gpx1000: it looks like this will crash if it's run in release mode when asserts are presumably no-ops.  We might want to preemptively put in more graceful logic here.

gpx1000 (approved):

components/events/include/components/events/event_types.hpp
TomAtkinsonArm: let's add a specific event type for that when we get to it. We can then adjust the components used in the samples to also support XR controllers

components/vfs/src/root_file_system.cpp
TomAtkinsonArm: I will be revisiting the FS error handling to match the exception error handling that @asuessenbach suggested. I will fix this there

PR #492, opened 2022-07-09
Encoding API

TomAtkinsonArm:
## Description

Encoding data to and from a serialized form will allow the project to use config files and communication protocols with ease. Here is a simple API used to serialize structs. For now I test basic data types. A test should be written for each serialize and de-serialize specialization written in the future - these can be local to the implementation (not in the encoding folder).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -59,3 +59,6 @@
 [submodule "third_party/re"]
 	path = third_party/re
 	url = https://github.com/google/re2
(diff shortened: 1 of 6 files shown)


TomAtkinsonArm (commented):
~typeid fix in tests~

turns out using the web editor from a PR is not what you expect... good for reviewing though

TomAtkinsonArm:
outdated

PR #493, opened 2022-07-09
CLI Samples launcher + Vulkan Hello Triangle

TomAtkinsonArm:
## Description

Blocked by #485 #487 #488 #490

Large diff as this contains commits from several branches

Currently only works on Ubuntu with XLIB enabled as I have not added every window and surface type yet

```
loading sample configs
candidate: /build/linux/cmd/samples_launcher/samples.json
selected: /build/linux/cmd/samples_launcher/samples.json
looking for: libsample__hello_sample.so - not found
looking for: libsample__hello_triangle.so
candidate: ./build/linux/samples/hello_triangle/lib/Debug/x86_64/libsample__hello_triangle.so
selecting: ./build/linux/samples/hello_triangle/lib/Debug/x86_64/libsample__hello_triangle.so


Available Samples: (1/2)

id:             hello_triangle
name:           Hello Triangle
description:    Vulkan Hello Triangle
compile target: sample__hello_triangle

Hello Triangle!
```

`samples.json` is maintained by CMake automatically and contains metadata for all registered samples. Samples may be available or not, we cant easily track which have been built previously and which haven't. The Sample Launcher deduces available samples by searching the current working directory for platform specific runtime libraries for a given sample. Available samples are then displayed in a list.

The code sample shows the launcher being run with `hello_triangle`. The runtime library is found as stated in the lookup log. The launcher then loads this sample and runs its `sample_main`.

Screenshot from 2022-07-09 18-44-13

Screenshot from 2022-07-09 18-41-06

After a little bit more work I ported Hello Triangle

Screenshot from 2022-07-09 22-51-06

# Solved Problems

### Compilation

Single samples can be compiled by using the `sample__<sample_name>` target. All samples can be compiled with `vkb__samples`. `samples_launcher` is only compiled when one or more samples are selected for compilation. A user should never need to manually compile `samples_launcher`.

As the framework is now split into components a sample and the launcher will only compile the targets needed. For instance, compiling the current samples yields sub second compile times. There is also no link step so serialized link times in our current builds will not occur in this version of the framework.

### Samples

Samples no longer need to follow strict patterns for creation. Requirements for a sample are:

- A file defining `sample_main`
- `vkb__register_sample` - sample build target
- `vkb__register_sample_descriptor` - sample metadata

From here samples can come in any shape or size. The most appropriate execution flow can be used for each sample. We may want to define a component which acts as a base sample to reduce boilerplate, but these building blocks do not need to be the same for all samples.

# TODO

- [ ] - condense launcher code
- [ ] - add a default samples build test which makes sure the launcher can find all samples when building with `vkb__samples`


bldsys/cmake/component_helpers.cmake
```diff
@@ -32,6 +32,8 @@ function(vkb__register_component)
 
         add_library("vkb__${TARGET_NAME}" STATIC ${TARGET_SRC})
 
+        set_property(TARGET "vkb__${TARGET_NAME}" PROPERTY POSITION_INDEPENDENT_CODE ON) # PIC required for shared lib linking
```
gpx1000, line 49: it is probably better to use set_compile_options("vkb_${TARGET_NAME}" PRIVATE POSITION_INDEPENDENT_CODE ON) 
NB: prior to CMake 3.19, -fPIC didn't mean that pie is also by default mandated in cmake other than in Android builds.  I think we're mandating a recent enough version for this to not be an issue.  Just listing here as a just incase.

bldsys/cmake/sample_helper.cmake
```diff
@@ -17,159 +17,57 @@
...
 
-    # Add shaders to project group
-    if (SHADERS_GLSL)
-        source_group("\\Shaders" FILES ${SHADERS_GLSL})
+        if(TARGET_INCLUDE_DIRS)
+            target_include_directories("sample__${TARGET_NAME}" PUBLIC ${TARGET_INCLUDE_DIRS})
+        endif()
+        
+        if(MSVC)
+            target_compile_options("sample__${TARGET_NAME}" PRIVATE /W4 /WX)
```
gpx1000, line 63: Dunno if you get any warnings in MSVC when compiling with this due to /W3 and /Md being set automatically D9025 might crop up.  However, just in case it happens:
https://cmake.org/cmake/help/v3.15/policy/CMP0092.html
https://cmake.org/cmake/help/v3.15/policy/CMP0091.html
add:
if (POLICY CMP0091)
  cmake_policy(SET CMP0091 NEW)
endif ()
if (POLICY CMP0092)
  cmake_policy(SET CMP0092 NEW)
endif ()

bldsys/cmake/component_helpers.cmake
```diff
@@ -16,21 +16,37 @@
...
     if(TARGET_NAME STREQUAL "")
         message(FATAL_ERROR "NAME must be defined in vkb__register_tests")
     endif()
 
-    if(TARGET_SRC) ## Create static library
+    if(TARGET_SRC) # Create static library
         message("ADDING STATIC: vkb__${TARGET_NAME}")
 
-        add_library("vkb__${TARGET_NAME}" STATIC ${TARGET_SRC})
+        add_library("vkb__${TARGET_NAME}" STATIC ${TARGET_SRC} ${TARGET_HEADERS})
```
gpx1000, line 47: think this is the same as in 490?  Not sure of the advantage of adding headers explicitly to the cmake target.

bldsys/cmake/sample_helper.cmake
```diff
@@ -17,159 +17,57 @@
...
-        TAGS 
-            "any"
-        FILES
-            ${SRC_FILES}
-        LIBS
-            ${TARGET_LIBS}
-        SHADER_FILES_GLSL
-            ${TARGET_SHADER_FILES_GLSL})
+    get_property(DESCRIPTORS GLOBAL PROPERTY SAMPLE_DESCRIPTORS)
+    list(APPEND DESCRIPTORS "{\"id\":\"${TARGET_ID}\",\"name\":\"${TARGET_NAME}\", \"description\": \"${TARGET_DESCRIPTION}\", \"library_name\": \"${TARGET_LIB}\"}")
```
gpx1000, line 30: could potentially use jq or https://github.com/sbellus/json-cmake or some such to make this a usable thing at the cmake level. however, this works.

TomAtkinsonArm: Does this allow for creating json structs in cmake? Or just parsing?

TomAtkinsonArm:
@SaschaWillems @gary-sweet @asuessenbach can you please take a look at this PR when you get a chance? Looking for general feedback on how these components fit together and how samples can be written. The idea is to remove all the constraints on samples so that they can be more flexible in the future.

An example of this is `sample_main` which is all that is required to create a sample. Or `EventPipeline` which allows a sample to define its own lifecycle. "Frameworks" can be extracted into a series of default configurations for these components as most components are configurable or extendable

asuessenbach:
First of all: the framework compiles without any issues with VS2022.
But when I start the `samples_launcher`, I get the following message:
```
loading sample configs
[vfs/std_filesystem.cpp:208] path does not exist: C:\git\Vulkan-Samples\build\VS2022_64\cmd\samples_launcher\build
```
And, indeed, that path does not exist. I found the file `samples.json` directly in the `samples_launcher` directory; there's no build directory there.
For a thorough review, I might need some more time. After all, this is pretty big.

cmd/samples_launcher/include/config.hpp
```diff
@@ -0,0 +1,42 @@
...
+	std::string library_name;
+};
+
+struct Config
+{
+	std::vector<Sample> samples;
+};
+
+// Load a config from a binary buffer
+components::StackErrorPtr load_config_from_json(const std::vector<uint8_t> &data, Config *config);
```
asuessenbach, line 41: I think, we agreed to switch to `std::exception`s here, instead of `components::StackErrorPtr`?
(Funnily, `components::StackErrorPtr` derives from `std::exception`, but is not used by the `throw`/`catch` mechanism of exceptions!)

cmd/samples_launcher/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+	std::string    library_path;
+};
+
+CUSTOM_MAIN(context)
+{
+	auto &fs = vfs::_default(context);
+
+	std::cout << "loading sample configs\n";
+
+	std::vector<std::string> files;
```
asuessenbach, line 42: Maybe rename `files` to `config_files`, or so?

cmd/samples_launcher/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+	std::vector<AvailableSample> available_samples;
+	available_samples.reserve(samples_config.samples.size());
+
+	for (auto &sample : samples_config.samples)
+	{
+		auto library_name = dl::os_library_name(sample.library_name);
+
+		std::cout << "looking for: " << library_name;
+
+		std::vector<std::string> candidates;
```
asuessenbach, line 104: Do you expect more than one candidate per sample? But you just use the first one.
Maybe the search in the sample_files could then be done by `std::find_if`:
```
		auto sample_it = std::find_if(sample_files.begin(),
		                              sample_files.end(),
		                              [&library_name](auto const &sample_file) { return vfs::helpers::get_file_name(sample_file) == library_name; });
		if (sample_it == sample_files.end())
		{
			std::cout << " - not found\n";
			continue;
		}

```

cmd/samples_launcher/src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+	auto args = context->arguments();
+
+	if (args.size() == 0)
+	{
+		std::cout << "no sample select\n";
+		return EXIT_SUCCESS;
+	}
+
+	for (auto &info : available_samples)
```
asuessenbach, line 152: Maybe use `std::find_if`, again?
```
  auto sample_it =
	    std::find_if(available_samples.begin(), available_samples.end(), [&args](auto &sample_info) { return sample_info.sample.id == args[0]; });
	if (sample_it == available_samples.end())
	{
		std::cout << "Missing sample " << args[0] << "\n";
	}

```

components/encoding/include/components/encoding/json.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+using json = nlohmann::json;
+
+/*
+ * WARNING: serializing floats using nlohmann JSON does not work as expected. USE DOUBLES!
+ */
+
+#define HANDLE_JSON_ERROR_START() \
+	try                           \
+	{
```
asuessenbach, line 37: I would not hide scope opening in a macro. Functions like marshall_json below look somewhat awkward then.
Maybe just some HANDLE_JSON_ERROR representing the catch block would be enough?

cmd/samples_launcher/include/config_marshalers.hpp
```diff
@@ -0,0 +1,162 @@
...
+	json j;
+	j["id"]           = sample.id;
+	j["name"]         = sample.name;
+	j["description"]  = sample.description;
+	j["library_name"] = sample.library_name;
+	return j;
+}
+
+template <>
+class JsonMarshaler<config::Sample>
```
asuessenbach, line 49: The termini `Marshaler` and `UnMarshaler` are a bit alien to me.
Maybe something like `Serializer`/`Deserializer` or `Encoder`/`Decoder` would be a bit more descriptive?

components/events/src/input_manager.cpp
```diff
@@ -59,18 +66,18 @@ void InputManager::process(ChannelReceiverPtr<TouchEvent> &events)
...
 		}
 
 		auto &touch = it->second;
 
-		if (event.pos_x != touch.position.x || event.pos_y != touch.position.y)
+		if (event.x != touch.position.x || event.y != touch.position.y)
 		{
-			touch.position.x = event.pos_x - touch.position.x;
-			touch.position.y = event.pos_y - touch.position.y;
+			touch.position.x = event.x - touch.position.x;
```
asuessenbach, line 76: What is this supposed to do?
It's unconditionally overwritten next.

components/events/tests/event_bus.test.cpp
```diff
@@ -0,0 +1,348 @@
...
+	virtual ~TestEventBus() = default;
+
+	inline size_t observer_count() const
+	{
+		return m_observers.size();
+	}
+
+	inline size_t unobserved_each_event_count() const
+	{
+		size_t unprocessed_events{0};
```
asuessenbach, line 42: Maybe
```
  return std::accumulate(
    m_each_callbacks.begin(), m_each_callbacks.end(), 0, []( size_t acc, auto const & callback ) { return acc + callback.second->queue_size(); } );
```
And similar for the other count functions.

components/events/tests/event_pipeline.test.cpp
```diff
@@ -0,0 +1,313 @@
...
+	    .add_always(std::make_unique<TestStage>(execution_index, 3))
+	    .add_always(std::make_unique<TestStage>(execution_index, 4));
+
+	// assertions inside TestStage
+	pipeline.process();
+
+	// on a second pass we expect only the "then" stages to be processed
+	execution_index = 2;
+
+	pipeline.process();
```
asuessenbach, line 233: I don't see how the correct order is verified. Could you please elaborate a bit?

components/events/tests/event_pipeline.test.cpp
```diff
@@ -0,0 +1,313 @@
...
+	    });
+
+	pipeline.process();
+	pipeline.process();
+	pipeline.process();
+	pipeline.process();
+
+	REQUIRE(one_execution_count == 1);
+	REQUIRE(two_execution_count == 4);
+	REQUIRE(three_execution_count == 4);
```
asuessenbach, line 285: I see, the number of calls of each event is checked, but how is the correct order verified?

components/platform/include/components/platform/dl.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#define EXPORT
+#define IMPORT
+
+#if defined(_WIN32)
+#	undef EXPORT
+#	undef IMPORT
+#	define EXPORT __declspec(dllexport)
+#	define IMPORT __declspec(dllimport)
+#endif
```
asuessenbach, line 30: Maybe better to read:
```
#if defined(_WIN32)
#  define EXPORT __declspec(dllexport)
#  define IMPORT __declspec(dllimport)
#else
#  define EXPORT
#  define IMPORT
#endif
```

components/platform/include/components/platform/dl.hpp
```diff
@@ -0,0 +1,81 @@
...
+const char *os_library_prefix();
+const char *os_library_postfix();
+
+/**
+ * @brief Convert a library name to an OS specific library name
+ *
+ * @param name library name
+ * @return std::string OS library name
+ */
+std::string os_library_name(const std::string &name);
```
asuessenbach, line 48: Mixing functions taking a `const std::string &` with functions taking a `const char *` in the same header seems counter-intuitive. I would stay with one of them, keeping the same level of abstraction.

components/platform/include/components/platform/platform.hpp
```diff
@@ -0,0 +1,196 @@
...
+extern "C"
+{
+	namespace components
+	{
+	/**
+	 * @brief A base context used for platform detection. Components or functions that consume this context can use it to create platform specific functionality.
+	 */
+	struct PlatformContext
+	{
+		virtual std::vector<std::string> arguments() const = 0;
```
asuessenbach, line 32: Why do you make this function pure virtual?
Probably all Context types just hold the arguments in a vector, that is it could be as well be moved into `PlatformContext`. Then, you don't even need that accessor function.

components/platform/include/components/platform/platform.hpp
```diff
@@ -0,0 +1,196 @@
...
+	 *      }
+	 *
+	 *      ... other code
+	 *
+	 *      return EXIT_SUCCESS;
+	 *  }
+	 *
+	 * @return int status code
+	 */
+	int platform_main(components::PlatformContext *);
```
asuessenbach, line 54: Shouldn't `platform_main` take a `const &`, or maybe a `const *`?

components/platform/src/dl.cpp
```diff
@@ -0,0 +1,93 @@
...
+	return ".so";
+#endif
+}
+
+std::string os_library_name(const std::string &name)
+{
+	std::stringstream lib_name;
+	lib_name << os_library_prefix() << name << os_library_postfix();
+	return lib_name.str();
+}
```
asuessenbach, line 60: No need to use a `std::stringstream` here. Simply
`	return os_library_prefix() + name + os_library_postfix();
`

components/vfs/src/root_file_system.cpp
```diff
@@ -256,19 +274,12 @@ StackErrorPtr RootFileSystem::enumerate_folders_recursive(const std::string &fil
...
-				all_dirs.emplace(full_path);
-				dirs_to_visit.emplace(full_path);
-			}
-		}
+		dirs_to_visit.insert(dirs_to_visit.end(), dirs.begin(), dirs.end());
+		all_dirs.insert(all_dirs.end(), dirs.begin(), dirs.end());
 	}
 
-	*folders = std::vector<std::string>{all_dirs.begin(), all_dirs.end()};
+	std::vector<std::string> all_folders{all_dirs.begin(), all_dirs.end()};
```
asuessenbach, line 281: Why do you copy `all_dirs` into `all_folders` here?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+	if (error)
+	{
+		throw std::runtime_error{"failed to initialize temporary file directory path"};
+	}
+}
+
+bool StdFSFileSystem::folder_exists(const std::string &file_path)
+{
+	return std::filesystem::is_directory(m_base_path += file_path);
```
asuessenbach, line 70: You probably wanted to use `+`, instead of `+=` here?
Then, the function could be `const`.

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,256 @@
...
+}
+
+bool StdFSFileSystem::file_exists(const std::string &file_path)
+{
+	return std::filesystem::is_regular_file(m_base_path += file_path);
```
asuessenbach, line 75: As above: `+`, instead of `+=`; and `const`.

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,256 @@
...
+}
+
+StackErrorPtr StdFSFileSystem::read_chunk(const std::string &file_path, const size_t offset, const size_t count, std::shared_ptr<Blob> *blob)
```
asuessenbach, line 78: Could this function be const?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,256 @@
...
+	{
+		return StackError::unique("failed to read chunk from file: " + full_path.string(), "vfs/std_filesystem.cpp", __LINE__);
+	}
+
+	*blob = std::move(w_blob);
+
+	return nullptr;
+}
+
+size_t StdFSFileSystem::file_size(const std::string &file_path)
```
asuessenbach, line 124: Could be const?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+	if (!std::filesystem::exists(full_path))
+	{
+		return 0;
+	}
+
+	return static_cast<size_t>(std::filesystem::file_size(full_path));
+}
+
+StackErrorPtr StdFSFileSystem::write_file(const std::string &file_path, const void *data, size_t size)
```
asuessenbach, line 136: Could be const?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	if (!stream.write(reinterpret_cast<const char *>(data), size))
+	{
+		return StackError::unique("failed to write to file: " + full_path.string(), "vfs/std_filesystem.cpp", __LINE__);
+	}
+	return nullptr;
+}
+
+StackErrorPtr StdFSFileSystem::enumerate_files(const std::string &dir, std::vector<std::string> *files)
```
asuessenbach, line 159: Could be const?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,256 @@
...
+			temp_files.push_back(helpers::sanitize(file));
+		}
+	}
+
+	*files = std::move(temp_files);
+
+	return nullptr;
+}
+
+StackErrorPtr StdFSFileSystem::enumerate_folders(const std::string &dir, std::vector<std::string> *folders)
```
asuessenbach, line 197: Could be const?

components/windows/src/glfw.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+GLFWWindow::GLFWWindow(const std::string &title, const Extent &initial_extent) :
+    Window{},
+    m_title{title},
+    m_extent{initial_extent},
+    m_callback_helper{std::make_unique<GLFWCallbackHelper>(*this)}
+{
+	// TODO: GLFW_X11_XCB_VULKAN_SURFACE
+
+	if (!glfwInit())
```
asuessenbach, line 347: Is `glfwTerminate` called somewhere?

components/windows/src/glfw.cpp
```diff
@@ -0,0 +1,479 @@
...
+GLFWWindow::~GLFWWindow()
+{
+	glfwDestroyWindow(m_handle);
+}
+
+void GLFWWindow::set_extent(const Extent &extent)
+{
+	glfwSetWindowSize(m_handle, (int) extent.width, (int) extent.height);
+	m_extent = extent;
+	if (m_content_rect_sender)
```
asuessenbach, line 388: Are the checks against the sender needed?
Or should that just be an assertion?

samples/hello_triangle/src/context.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+	for (VkImageView image_view : context.swapchain_image_views)
+	{
+		vkDestroyImageView(context.device, image_view, nullptr);
+	}
+
+	if (context.swapchain != VK_NULL_HANDLE)
+	{
+		vkDestroySwapchainKHR(context.device, context.swapchain, nullptr);
+		context.swapchain = VK_NULL_HANDLE;
```
asuessenbach, line 52: You should either reset all teared-down members here, or none.

samples/hello_triangle/src/device.cpp
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	uint32_t device_extension_count;
+	VK_CHECK(vkEnumerateDeviceExtensionProperties(context.gpu, nullptr, &device_extension_count, nullptr));
+	std::vector<VkExtensionProperties> device_extensions(device_extension_count);
+	VK_CHECK(vkEnumerateDeviceExtensionProperties(context.gpu, nullptr, &device_extension_count, device_extensions.data()));
+
+	if (!validate_extensions(required_device_extensions, device_extensions))
+	{
+		throw std::runtime_error("Required device extensions are missing, will try without.");
```
asuessenbach, line 80: What do you mean with "will try without" ?

samples/hello_triangle/src/instance.cpp
```diff
@@ -0,0 +1,199 @@
...
+	{
+		LOGI("Validation Layer: Information: {}: {}", layer_prefix, message);
+	}
+	return VK_FALSE;
+}
+#endif
+
+bool validate_extensions(const std::vector<const char *> &         required,
+                         const std::vector<VkExtensionProperties> &available)
+{
```
asuessenbach, line 51: Maybe:
```
	// inner find_if returns true if the extension was not found
	// outer find_if returns true if none of the extensions were not found, that is if all extensions were found
	return std::find_if(required.begin(),
	                    required.end(),
	                    [&available](auto required_extension) {
		                    return std::find_if(available.begin(),
		                                        available.end(),
		                                        [&required_extension](auto const &available_extension) {
			                                        return strcmp(available_extension.extensionName, required_extension) == 0;
		                                        }) == available.end();
	                    }) == required.end();

```

samples/hello_triangle/src/instance.cpp
```diff
@@ -0,0 +1,199 @@
...
+		}
+	}
+
+	return true;
+}
+
+bool validate_layers(const std::vector<const char *> &     required,
+                     const std::vector<VkLayerProperties> &available)
+{
+	for (auto extension : required)
```
asuessenbach, line 76: Maybe:
```
	// inner find_if returns true if the layer was not found
	// outer find_if returns true if none of the layers was not found, that is if all layers were found
	return std::find_if(required.begin(),
	                    required.end(),
	                    [&available](auto required_layer) {
		                    return std::find_if(available.begin(),
		                                        available.end(),
		                                        [&required_layer](auto const &available_layer) {
			                                        return strcmp(available_layer.layerName, required_layer) == 0;
		                                        }) == available.end();
	                    }) == required.end();

```

samples/hello_triangle/src/instance.cpp
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	return true;
+}
+
+void init_instance(Context &context, const std::vector<const char *> &required_instance_extensions, const std::vector<const char *> &required_validation_layers)
+{
+	LOGI("Initializing vulkan instance.");
+
+	if (volkInitialize())
```
asuessenbach, line 101: Maybe:
`if (volkInitialize() != VK_SUCCESS)`

samples/hello_triangle/src/swapchain.cpp
```diff
@@ -0,0 +1,390 @@
...
+	}
+	else
+	{
+		if (format_count == 0)
+		{
+			throw std::runtime_error("Surface has no formats.");
+		}
+
+		format.format = VK_FORMAT_UNDEFINED;
+		for (auto &candidate : formats)
```
asuessenbach, line 106: Some `std::find_if` here?

asuessenbach (commented):
On Windows using VS, I had to set the Working Directory to the root of the repository in order to find everything. I don't see that mentioned anywhere.

PR #495, opened 2022-07-10
STD Filesystem VFS

### LINKED ISSUE #467: FI: Use std::filesystem
#462 is a great implementation that depends upon platform specific code for filesystem classes (i.e. windowsFileSystem, unixFileSystem class etc).  I would suggest that we can switch to using std::filesystem to provide a common class that takes away the need for platform dependent code.

TomAtkinsonArm:
## Description

Replaces OS specific filesystem implementations with std filesystem

This is significantly slower when it comes to enumerating files or folders. Likely due to the conversion between path and string. For now I have added a TODO for future reference


components/vfs/include/components/vfs/helpers.hpp
```diff
@@ -44,32 +44,41 @@ std::string get_directory(const std::string &path);
...
 std::vector<std::string> get_directory_parts(const std::string &path);
 
+/**
+ * @brief Get a paths directory and then split its path into an array of individual parts
+ *        e.g /temp/folder/file.txt -> ["temp", "folder"]
+ *
+ * @param path A path
+ * @return std::vector<std::string> A vector of topological directory parts
+ */
+std::vector<std::string> get_individual_directory_parts(const std::string &path);
```
asuessenbach: Maybe something like `tokenize_path` would be a more concise name?

components/vfs/src/helpers.cpp
```diff
@@ -75,6 +76,41 @@ std::vector<std::string> get_directory_parts(const std::string &path)
...
 
+std::vector<std::string> get_individual_directory_parts(const std::string &path)
+{
+	auto dir_path = get_directory(path);
+
+	std::string accum;
+	accum.reserve(dir_path.size());
+
+	std::vector<std::string> dirs;
+	dirs.reserve(dir_path.size());
```
asuessenbach: Is it reasonable to reserve space for that many strings?

components/vfs/src/helpers.cpp
```diff
@@ -75,6 +76,41 @@ std::vector<std::string> get_directory_parts(const std::string &path)
...
+
+	for (char c : dir_path)
+	{
+		if (c == '/' && accum.size() > 0)
+		{
+			if (accum.size() > 0)
```
asuessenbach: You can't get here with `accum.size() <= 0`.

components/vfs/src/root_file_system.cpp
```diff
@@ -226,7 +226,7 @@ StackErrorPtr RootFileSystem::enumerate_files_recursive(const std::string &file_
 		all_files.reserve(folder_files.size());
```
asuessenbach: Should be `all_files.reserve(all_files.size() + folder_files.size());` ?

components/vfs/src/root_file_system.cpp
```diff
@@ -226,7 +226,7 @@ StackErrorPtr RootFileSystem::enumerate_files_recursive(const std::string &file_
...
 		for (auto &file : folder_files)
 		{
-			all_files.push_back(folder + file);
+			all_files.push_back(file);
 		}
```
asuessenbach: simply `all_files.insert(all_files.end(), folder_files.begin(), folder_files.end());` ?

components/vfs/src/root_file_system.cpp
```diff
@@ -238,7 +238,6 @@ StackErrorPtr RootFileSystem::enumerate_files_recursive(const std::string &file_
 StackErrorPtr RootFileSystem::enumerate_folders_recursive(const std::string &file_path, std::vector<std::string> *folders)
```
asuessenbach: Shouldn't the first argument be named `folder_path`, or so?

components/vfs/src/root_file_system.cpp
```diff
@@ -226,7 +226,7 @@ StackErrorPtr RootFileSystem::enumerate_files_recursive(const std::string &file_
 		all_files.reserve(folder_files.size());
 		for (auto &file : folder_files)
 		{
-			all_files.push_back(folder + file);
+			all_files.push_back(file);
 		}
 	}
```
asuessenbach, line 249: instead of `*files = all_files;`: `assert(files); files->swap(all_files);` to prevent copying all the data

components/vfs/src/root_file_system.cpp
```diff
@@ -258,12 +257,11 @@ StackErrorPtr RootFileSystem::enumerate_folders_recursive(const std::string &fil
 
 		for (std::string &dir : dirs)
 		{
-			std::string full_path = front + dir;
-			auto        it        = all_dirs.find(full_path);
+			auto it = all_dirs.find(dir);
 			if (it == all_dirs.end())
 			{
```
asuessenbach: `dirs` already hold full paths? How can that `dir` be already part of the set of `all_dirs`?

components/vfs/src/root_file_system.cpp
```diff
@@ -285,6 +283,18 @@ void RootFileSystem::make_directory(const std::string &file_path)
...
+bool RootFileSystem::remove(const std::string &path)
+{
+	std::string adjusted_path;
+	auto        fs = find_file_system(path, &adjusted_path);
+	if (!fs)
+	{
+		return false;
+	}
+
+	return fs->remove(path);
```
asuessenbach: `fs->remove(adjusted_path)` ?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+	if (!stream.read(reinterpret_cast<char *>(w_blob->buffer.data()), size))
+	{
+		return StackError::unique("failed to read file contents: " + full_path.string(), "vfs/std_filesystem.cpp", __LINE__);
+	}
+
+	*blob = w_blob;
+
+	return nullptr;
+}        // namespace vfs
```
asuessenbach: `// namespace vfs` ?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,292 @@
...
+	if (!blob)
+	{
+		return nullptr;
+	}
+
+	auto full_path = get_full_path(m_base_path, file_path);
+
+	if (!std::filesystem::exists(full_path))
+	{
+		return StackError::unique("file does not exist: " + full_path.string(), "vfs/std_filesystem.cpp", __LINE__);
```
asuessenbach, line 89: use `__FILE__ `instead of `"vfs/std_filesystem.cpp"` ?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,292 @@
...
+	return std::filesystem::is_directory(m_base_path += file_path);
+}
+
+bool StdFSFileSystem::file_exists(const std::string &file_path)
+{
+	return std::filesystem::is_regular_file(m_base_path += file_path);
+}
+
+StackErrorPtr StdFSFileSystem::read_file(const std::string &file_path, std::shared_ptr<Blob> *blob)
+{
```
asuessenbach, line 79: implement by just calling `read_chunk` with offset == 0 ?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,292 @@
...
+		if (std::filesystem::is_regular_file(entry.status()))
+		{
+			// TODO: switching between std::path and std::string is inefficient
+			auto file = entry.path().string();
+			file      = file.substr(m_base_path.string().size());
+			temp_files.push_back(helpers::sanitize(file));
+		}
+	}
+
+	*files = temp_files;
```
asuessenbach: `file->swap( temp_files );` ?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,292 @@
...
+		if (std::filesystem::is_directory(entry.status()))
+		{
+			// TODO: switching between std::path and std::string is inefficient
+			auto directory = entry.path().string();
+			directory      = directory.substr(m_base_path.string().size());
+			temp_folders.push_back(helpers::sanitize(directory));
+		}
+	}
+
+	*folders = temp_folders;
```
asuessenbach: `folders->swap( temp_folders );` ?

components/vfs/src/std_filesystem.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+	std::vector<std::string> temp_files{};
+
+	for (const auto &entry : std::filesystem::directory_iterator(full_path))
+	{
+		if (std::filesystem::is_regular_file(entry.status()))
+		{
+			// TODO: switching between std::path and std::string is inefficient
+			auto file = entry.path().string();
+			file      = file.substr(m_base_path.string().size());
```
asuessenbach: `m_base_path.string().size()` could at least be pulled out of the loop.

components/vfs/include/components/vfs/filesystem.hpp
```diff
@@ -121,14 +121,15 @@ class FileSystem
 	virtual size_t        file_size(const std::string &file_path)                                                                        = 0;
 	virtual StackErrorPtr write_file(const std::string &file_path, const void *data, size_t size)                                        = 0;
 	virtual void          make_directory(const std::string &path)                                                                        = 0;
+	virtual bool          remove(const std::string &path)                                                                                = 0;
 
 	virtual StackErrorPtr enumerate_files(const std::string &file_path, std::vector<std::string> *files)     = 0;
```
asuessenbach, line 127: Why are you using `std::vector<std::string> *`, instead of `std::vector<std::string> &` ?
Or maybe return a `std::variant<StackErrorPtr, std::vector<std::string>>`, instead, to make it obvious, that the enumerated data is not appended to the vector?

The question holds for all the other functions, where you pass in a pointer instead of a reference, as well.

components/vfs/src/helpers.cpp
TomAtkinsonArm: Im not sure we want to add a part if the accumulation here is smaller or equal to 0? Here we are only adding a path part if its length is 1+?

components/vfs/src/root_file_system.cpp
TomAtkinsonArm: this would be a better name

components/vfs/src/helpers.cpp
asuessenbach: You check for `accum.size() > 0` just two lines above... one of those checks should be enough.

components/vfs/src/std_filesystem.cpp
TomAtkinsonArm: `__FILE__` is an absolute path. In some cases this can be hundreds of characters long. It is also not consistent across platforms so there is no easy way to shorten this. `"vfs/std_filesystem.cpp"` this style is not correct in terms of filesystem path but will still be easy to read from a log output.

components/vfs/src/std_filesystem.cpp
TomAtkinsonArm: done

components/vfs/src/std_filesystem.cpp
TomAtkinsonArm: used `std::move`

components/vfs/src/std_filesystem.cpp
TomAtkinsonArm: used `std::move`

components/vfs/include/components/vfs/filesystem.hpp
TomAtkinsonArm: I have always preferred the C style output parameters over pass by reference. From the usage it is more obvious which elements are being mutated.

`enumarate_files(path, &files)` compared to `enumarate_files(path, files)`

Here it is clear to see that `files` will be mutated in some way. This is mainly due to the fact I have had strict requirements on older C++ versions where variant was not yet part of the std lib. If we want to use more variants and have Go style returns of multiple variables then I would be happy to change the API.

Will `std::variant<StackErrorPtr, std::vector<std::string>>` incur more copies?

components/vfs/src/root_file_system.cpp
TomAtkinsonArm: I now add to the queue and perform an insert

components/vfs/src/root_file_system.cpp
TomAtkinsonArm: using `std::move` added more assertions on output parameters

TomAtkinsonArm: See discussion below on possible use of variants

components/vfs/include/components/vfs/filesystem.hpp
gary-sweet: It's a fairly common idiom to pass const params by reference and mutable params by pointer. As you say, it's obvious from the call-site that a parameter may be changed by the call. I don't personally see any issue with it since it's what I also do.

components/vfs/src/helpers.cpp
TomAtkinsonArm: it is not

components/vfs/include/components/vfs/helpers.hpp
TomAtkinsonArm: much better name, thank you!

components/vfs/src/helpers.cpp
TomAtkinsonArm: Ah yeah missed that

TomAtkinsonArm: fixed in latest commit

components/vfs/include/components/vfs/filesystem.hpp
asuessenbach: When you pass a mutable parameter by pointer, you have to check against nullptr. When you pass it by reference you don't need to.
Either way, you don't know if the passed parameter is reset, or just modified (here, it could as well be that the enumerated files are appended to the vector of files); when you return something, it's clearly new.

TomAtkinsonArm: So we would be going form this style:

```c++
Thing something;

if (auto err = action(&something)) {
    // we hit an error
    fix_this(err);
}

// no error
something.do();
```

to this style:

```c++
auto result = action(); // returns std::variant<StackErrorPtr, Thing>

if (auto err = std::get<StackErrorPtr>(result)) {
    // we hit an error
    fix_this(err);
}

// no error
Thing  something = std::get<Thing>(result);
```

in practice the change would look more like this:

```c++
template <typename Type>
using StackErrorVariant = std::variant<StackErrorPtr, Type>

virtual StackErrorVariant<std::vector<std::string>> enumerate_files(const std::string &file_path) = 0; 
```

Thoughts @SaschaWillems @gpx1000 @gary-sweet ?

asuessenbach: Would switching from those StackErrorPtr to exceptions be an option here? Would simplify the function even more:
`std::vector<std::string> enumerate_files( const std::string &file_path) = 0;`

components/vfs/src/root_file_system.cpp
```diff
@@ -256,19 +274,16 @@ StackErrorPtr RootFileSystem::enumerate_folders_recursive(const std::string &fil
...
-			std::string full_path = front + dir;
-			auto        it        = all_dirs.find(full_path);
-			if (it == all_dirs.end())
-			{
-				all_dirs.emplace(full_path);
-				dirs_to_visit.emplace(full_path);
-			}
+			dirs_to_visit.push(dir);
 		}
+
```
asuessenbach: One might think about changing `dirs_to_visit` from `std::queue` to `std::deque`, allowing to insert all the elements from `dirs` at the end of `dirs_to_visit` with one call?

components/vfs/src/root_file_system.cpp
```diff
@@ -223,25 +243,23 @@ StackErrorPtr RootFileSystem::enumerate_files_recursive(const std::string &file_
...
 
 	return nullptr;
 }
 
-StackErrorPtr RootFileSystem::enumerate_folders_recursive(const std::string &file_path, std::vector<std::string> *folders)
+StackErrorPtr RootFileSystem::enumerate_folders_recursive(const std::string &folder_path, std::vector<std::string> *folders)
 {
+	assert(folders);
+
 	std::set<std::string> all_dirs;
```
asuessenbach: As you don't need to find in `all_dirs` anymore, there's no need to use a `std::set` here, you can just use a `std::vector`.

components/vfs/src/helpers.cpp
```diff
@@ -88,11 +88,8 @@ std::vector<std::string> tokenize_path(const std::string &path)
...
-			if (accum.size() > 0)
-			{
-				dirs.emplace_back(accum);
-				accum.clear();
-			}
+			dirs.emplace_back(accum);
+			accum.clear();
 		}
 
 		if (c != '/')
```
asuessenbach, line 89: Maybe a bit better to read:
```
if (c != '/')
{
    ...
}
else if (accum.size() > 0)
{
    ...
}
```

components/vfs/include/components/vfs/filesystem.hpp
TomAtkinsonArm: It might, but exceptions just move the verbosity up a level. For now this interface works and so it may be best to keep it until there are major hurdles caused by it. At the end of the day this is all just syntactic sugar and none of the options would not change the behavior of the system.

Any further opinions on this @gpx1000 @SaschaWillems @gary-sweet? It would make sense to find a common ground on error handling sooner rather than later, but should be addressed in a new PR

components/vfs/src/root_file_system.cpp
TomAtkinsonArm: Good call!

components/vfs/include/components/vfs/filesystem.hpp
gpx1000: I personally prefer pointers to references.  It's easier for me to see who actually originates the memory in a passing context and where it must be created/deleted.  However, that's just easier for older guys like me that remember the C days.  The reference method and new tooling makes this more of a spaces/tabs argument.  Both work equally well and if anything, points are in favor of pass by reference due to the fact that pointers are way easier to have security problems and get lost causing memory leaks etc.

As for the exceptions instead of StackErrorPtr, I'm a fan of exceptions.  I like programs to break when they hit real problems that put them into a non-working state.  Break early break often is usually the path towards highlighting errors sooner in the dev process.  That said, I agree with Tom that in this instance, I don't see it providing much new functionality.  It would add to the size of the executable and is about 16 times slower to switch to them for no appreciable benefit (not that anyone really cares about exe size anymore, and why care about CPU performance when you want to throw an error).  This one I'm more ambivalent about.

asuessenbach: > but exceptions just move the verbosity up a level

I don't really understand this. What do you mean?

TomAtkinsonArm: The function definition looks cleaner when using exceptions but handling the error is more verbose with try/catch compared to the `if` style. It is also not clear by the function definition that it throws an error. We can get lazy and have multiple functions which throw inside the same try/catch statement, it might be recoverable earlier in the process.

asuessenbach: > It is also not clear by the function definition that it throws an error

That's not exactly true... as this function uses a non-const std::vector, it might throw anyways, no matter how other errors might be communicated.

> We can get lazy and have multiple functions which throw inside the same try/catch statement, it might be recoverable earlier in the process.

Again: what do you mean with that?
Besides that: I don't see why try/catch should be more verbose than error code checking. I think, the opposite is true.

TomAtkinsonArm: Ok, so we have hit a little bit of a stale mate. Both techniques have pros and cons but essentially do the same thing. As it is not the purpose of this PR to address error handling, are you happy to approve this PR as is?

We should either raise a ticket which moves this discussion there or you can create a proposal PR for how we should handle errors going forward. I'm not opposed to exceptions, simply using the method that I have found works best for me in the past. This is not the PR that should address it :)

The only real requirement in terms of errors is that we have a consistent error handling mechanism across the code base so that errors are predictable :)

asuessenbach: > The only real requirement in terms of errors is that we have a consistent error handling mechanism across the code base so that errors are predictable :)

Totally agree!

asuessenbach:
Well, the PR actually LGTM.
I have a few questions/issues on the surroundings, though. Please sorry, if that feels too extensive:
- what's those `class Blob` and `class StdBlob` in filesystem.hpp for?
  especially Blob::empty(): can there be the case that size() gives some positive value, but data() is nullptr?
  If some abstraction for a blob is needed, maybe something like `using Blob = std::vector<uint8_t>;` would be sufficient?
- The argument to the functions `folder_exists` should be name `folder_path`
- Why is the blob passed into `read_chunk` and `readFile` a `std::sharedPtr<Blob> *`, not just a `Blob *`
- Why is `FileSystem::readFile` virtual? Why does `RootFileSystem` override it?
- Why are those three `enumerate`-functions members of `RootFileSystem`, not `FileSystem`?
- Most of the functions of `FileSystem` and `RootFileSystem` could be const.
- In helpers.cpp, `get_file_name`: the second argument to `substr` is the a count, which defaults to `npos`. That is, as you want to get all the rest right of '/', you can omit it. Any maybe you should add a comment that `std::string::npos == -1`, so you get the complete string if there's no '/' in `path`.
- In helpers.cpp, `sanitize`, maybe the `remove trailing /` part could be coded like this:
```
if (!sanitized.empty() && sanitized.back() == '/')
{
    sanitized.pop_back();
}
```
- In root_file_system.cpp, `StdBlob::data()` could be implemented as `return static_cast<const void *>(buffer.data());`
- In root_file_system.cpp, `RootFileSystem::enumerate_files()`, you should `*files = std::move(all_files);` if the extension is empty. And you might use `helpers::get_file_extension()` in the loop there.
- In root_file_system.cpp, `mount`, maybe like this:
```
auto it = std::find_if(m_mounts.begin(), m_mounts.end(), [&file_path](auto const & mount){ return mount.first == file_path; });
if ( it != m_mounts.end())
{
    it->second = file_system;
}
else
{
    m_mounts.push_back({file_path, file_system});
}
```
- In root_file_system.cpp, unmount, maybe like this:
```
auto it = std::find_if(m_mounts.begin(), m_mounts.end(), [&file_path](auto const & mount){ return mount.first == file_path; });
if ( it != m_mounts.end())
{
    m_mounts.erase(it);
}
```
- In std_filesystem.cpp, `folder_exists `and `file_exists`: do you really want to add `file_path` to the `m_base_path`?

TomAtkinsonArm:
Merging so that #506 may be applied on top

PR #496, opened 2022-07-11
Visual Studio Folders

### LINKED ISSUE #486: Framework v2.0 Visual Studio broken folder layout
There is a broken folder layout in Visual Studio. Once, #482 merges the facilities to fix this across the entire project will be in place.

Original mention:

> I would prefer to stay with RAII.
> 
> When I try to use your code (gh pr checkout 483), my VS solution seems to be completely scrambled. Am I missing something to get CMake do-the-right-thing™? image

TomAtkinsonArm:
## Description

An attempt at fixing the visual studio folder structure. When using the VS CMake View I get:

image

When using the folder view instead of the CMake target view I get:

image

There are two issues here. One, we have to manually add header files - not a deal breaker as we do this in the main branch. Secondly, interface libraries are not shown in the solution view. This is a deal breaker as common is a header only interface library.

I'm open to suggestions on how to improve this but as I am not a Visual Studio user I am not sure what to expect from their solution explorer? @SaschaWillems @asuessenbach any suggestions?

### DIFF
--- a/bldsys/cmake/component_helpers.cmake
+++ b/bldsys/cmake/component_helpers.cmake
@@ -16,10 +16,22 @@
 
 add_custom_target(vkb_components)
 
+function(vkb__register_headers)
+    set(options)
+    set(oneValueArgs FOLDER OUTPUT)
+    set(multiValueArgs HEADERS)
+
+    cmake_parse_arguments(TARGET "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+    set(TEMP ${TARGET_HEADERS})
+    list(TRANSFORM TEMP PREPEND "${CMAKE_CURRENT_SOURCE_DIR}/include/components/${TARGET_FOLDER}/")
+    set(${TARGET_OUTPUT} ${TEMP} PARENT_SCOPE)
+endfunction()
+
 function(vkb__register_component)
     set(options)
     set(oneValueArgs NAME)
-    set(multiValueArgs SRC LINK_LIBS INCLUDE_DIRS)
+    set(multiValueArgs SRC HEADERS LINK_LIBS INCLUDE_DIRS)
 
     cmake_parse_arguments(TARGET "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
@@ -27,10 +39,10 @@ function(vkb__register_component)
         message(FATAL_ERROR "NAME must be defined in vkb__register_tests")
     endif()
 
-    if(TARGET_SRC) ## Create static library
+    if(TARGET_SRC) # Create static library
         message("ADDING STATIC: vkb__${TARGET_NAME}")
 
-        add_library("vkb__${TARGET_NAME}" STATIC ${TARGET_SRC})
+        add_library("vkb__${TARGET_NAME}" STATIC ${TARGET_SRC} ${TARGET_HEADERS})
 
         if(TARGET_LINK_LIBS)
             target_link_libraries("vkb__${TARGET_NAME}" PUBLIC ${TARGET_LINK_LIBS})
@@ -39,13 +51,15 @@ function(vkb__register_component)
         if(TARGET_INCLUDE_DIRS)
             target_include_directories("vkb__${TARGET_NAME}" PUBLIC ${TARGET_INCLUDE_DIRS})
         endif()
-        
+
         if(MSVC)
             target_compile_options("vkb__${TARGET_NAME}" PRIVATE /W4 /WX)
         else()
             target_compile_options("vkb__${TARGET_NAME}" PRIVATE -Wall -Wextra -Wpedantic -Werror)
         endif()
-    else() ## Create interface library
+
+        set_property(TARGET vkb__${TARGET_NAME} PROPERTY FOLDER "components/${TARGET_NAME}")
+    else() # Create interface library
         message("ADDING INTERFACE: vkb__${TARGET_NAME}")
 
         add_library("vkb__${TARGET_NAME}" INTERFACE)
@@ -58,7 +72,6 @@ function(vkb__register_component)
             target_include_directories("vkb__${TARGET_NAME}" INTERFACE ${TARGET_INCLUDE_DIRS})
         endif()
     endif()
-
 
     add_dependencies(vkb_components "vkb__${TARGET_NAME}")
 endfunction()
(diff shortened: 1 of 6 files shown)


gpx1000:
FWIW, VS can now read cmake projects directly rather than having to run the generator step first.  https://docs.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=msvc-170  It's still on my TODO list to add documentation to the guide about how to use IDE's for Vulkan with CMake.  This is a good use case on thoughts gathering.

TomAtkinsonArm:
> VS can now read cmake projects directly

I'll need to test this with Interface libraries. Seems to only show static or shared libraries and only headers if its added to the target. Not the biggest fan of its integration but anything is better than nothing. Its more likely that I've made a mistake somewhere and I've configured it wrong.

gpx1000:
VS2019 is much improved in my opinion on the state of msvc cmake integration.  It's no where near as well thought out as CLion for integration; however, it works.  The view I usually use with msvc in cmake projects is the project view instead of the target view which gets the headers and everything on the folder structure the same as before.
You can edit the cmake project settings in the json files and add a task there to get msvc to run a header discovery process which "works" but I've had limited success getting those json files to auto discover from clone (usually have to have some by hand setup to make that work).

SaschaWillems:
I'm not a Visual Studio power user, but for me the structure now looks almost fine. The only problem I see is that all the test projects are on the root of the solution structure. Is it possible to move them to their own folders?

Current structure for me on VS2019:

image

Structure for me on VS2019 for this branch:

image

bldsys/cmake/component_helpers.cmake
```diff
@@ -39,13 +51,15 @@ function(vkb__register_component)
...
-        
+
         if(MSVC)
             target_compile_options("vkb__${TARGET_NAME}" PRIVATE /W4 /WX)
         else()
             target_compile_options("vkb__${TARGET_NAME}" PRIVATE -Wall -Wextra -Wpedantic -Werror)
         endif()
-    else() ## Create interface library
+
+        set_property(TARGET vkb__${TARGET_NAME} PROPERTY FOLDER "components/${TARGET_NAME}")
```
asuessenbach: Using just "components" as folder would remove the additional hierarchy in the solution.

bldsys/cmake/component_helpers.cmake
```diff
@@ -39,13 +51,15 @@ function(vkb__register_component)
...
+    else() # Create interface library
         message("ADDING INTERFACE: vkb__${TARGET_NAME}")
 
         add_library("vkb__${TARGET_NAME}" INTERFACE)
```
asuessenbach: Adding `${TARGET_HEADERS}` after the keyword `INTERFACE` would list the headers in that interface library.

bldsys/cmake/component_helpers.cmake
```diff
@@ -59,6 +73,5 @@ function(vkb__register_component)
         endif()
```
asuessenbach, line 73: Adding
`        set_property(TARGET vkb__${TARGET_NAME} PROPERTY FOLDER "components")`
would list this interface library in the solution

asuessenbach (commented):
I'm far from a CMake power user, but changing `components/common/CMakeLists.txt` like this:
```
vkb__register_headers(
    FOLDER
        common
    HEADERS
        stack_error.hpp
        strings.hpp
    OUTPUT
      EVENTS_HEADERS
)

vkb__register_component(
    NAME common
  HEADERS
    ${EVENTS_HEADERS}
  INCLUDE_DIRS
    ${CMAKE_CURRENT_SOURCE_DIR}/include
)
```
would, together with the suggested changes in `bldsys/cmake/component_helpers.cmake`, add the interface library `common` to the solution.
Moreover, the extraneous folders in the `components` folder would be gone.

Besides that, `spdlog_headers_for_ide` should be moved into the ThirdParty folder, somehow.
And the projects `vkb_components` and `vbk_tests`, which both just hold the top-level `CMakeLists.txt`, should probably removed.

TomAtkinsonArm:
Apply @asuessenbach suggestions, and moving some more third parties around, I now have this

image

bldsys/cmake/component_helpers.cmake
```diff
@@ -59,6 +73,7 @@ function(vkb__register_component)
...
     endif()
 
+    set_property(TARGET vkb__${TARGET_NAME} PROPERTY FOLDER "components/${TARGET_NAME}")
```
asuessenbach: That is, you want to keep that additional folder layer? Do you expect more projects under `common`, `events`, or `vfs`?

bldsys/cmake/component_helpers.cmake
```diff
@@ -16,21 +16,35 @@
 
 add_custom_target(vkb_components)
 
+set_property(TARGET vkb_components PROPERTY FOLDER "components")
```
asuessenbach, line 19: What is that custom target `vkb_components` for? Is it just to have one central target for all components?
(The same question holds with `vkb_tests`.)

asuessenbach (commented):
Shouldn't that `spdlog_headers_for_ide` project be moved into third_party, somehow? Or is that needed to be a top-level project?

Besides that, I don't see any test projects (yet).
image

bldsys/cmake/component_helpers.cmake
TomAtkinsonArm: forgot to push my commit. This has been flattened

bldsys/cmake/component_helpers.cmake
TomAtkinsonArm: `vkb_components` and `vkb_tests` are proxy targets used to group tests and components.

`vkb_tests` is useful for CI and local testing

`vkb_components` will become more useful when you only want to test build issues in components and not every sample. It can be easier to work up the stack slowly than the compiler spit out a bunch of nonsense which is hard to read

TomAtkinsonArm:
> Shouldn't that spdlog_headers_for_ide project be moved into third_party, somehow? Or is that needed to be a top-level project?

`spdlog_headers_for_ide` does not exist in this branch, make sure to update your submodules

> Besides that, I don't see any test projects (yet).

make sure to use `-DVKB_BUILD_TESTS=ON` at config time

asuessenbach:
LGTM.

PR #502, opened 2022-07-20
Move definition of VKB_DEBUG from header (error.h) to cmake (global_options.cmake)

asuessenbach:
## Description

Additionally remove all the now unnecessary includes of error.h
Compiles and runs fine with MSVC.

## General Checklist:


bldsys/cmake/global_options.cmake
```diff
@@ -52,3 +52,5 @@ add_definitions(-DROOT_PATH_SIZE=${ROOT_PATH_SIZE})
 
 set(CMAKE_C_FLAGS_DEBUG   "-DDEBUG=0 ${CMAKE_C_FLAGS_DEBUG}")
 set(CMAKE_CXX_FLAGS_DEBUG "-DDEBUG=0 ${CMAKE_CXX_FLAGS_DEBUG}")
+
+add_compile_definitions($<$<CONFIG:DEBUG>:VKB_DEBUG>)
```
TomAtkinsonArm: We should add compile definitions per target `target_compile_definitions`

https://cmake.org/cmake/help/latest/command/target_compile_definitions.html

asuessenbach: Why should we do that?
VKB_DEBUG probably is a some global setting, that should not be defined per target.

gpx1000:
Because if it's defined by target and public than all targets that inherit
get it too.  So it still goes everywhere we want it.  It makes things
cleaner when we don't want/need it on 3rd party libraries that might have
something defined with it for whatever reason that modifies behavior.
It's named with vkb at the front to protect from that eventuality, but if
it's not even in the define list when compiler is working on the library
then it's not even a chance of happening.

asuessenbach:
Ok, sounds reasonable... will move that define into sample_helper.cmake and make it per target then.

gary-sweet (approved):

TomAtkinsonArm (approved):

Issue #505, opened 2022-07-26
Error Handling


TomAtkinsonArm:
We should use this thread to further discuss error handling.

Currently, we return an error like so:

```
if(auto err = some_function()) {
  // handle error
}
```

@asuessenbach would you be able to make a PR showing how we can use exceptions?

asuessenbach:
I have ported the filesystem files from `StackErrorPtr` to exceptions, by essentially just replacing them with a `std::runtime_error` and adjust their usage.
One could argue, though, whether, for example, trying to read a chunk of data from a non-existent file, or trying to read beyond the file size actually are runtime errors or in fact programmer's errors. (I think, we can assume the files used here are constant!)
But in either case, it should show the principal usage.

TomAtkinsonArm:
This issue is left open to track the remaining error handling comments.

We would like to use a macro to help simplify throwing errors. @gpx1000 suggestion.

> Does it make sense to make this a macro and use something like:
#define ERROR_STR(...) (( FILE ":" LINE " " VA_ARGS))

We would also like to remove the usage of hard coded file paths. Preferably cutting down the file path length generated by `__FILE__` as much as possible. @gary-sweet suggestion

tomadamatkinson:
Resolved in future framework v2.0 changes

PR #506, opened 2022-07-27
Replace StackErrorPtr in filesystem by exceptions

asuessenbach:
## Description

This PR shows how exceptions could be used in the new framework -> see #505.

All functions in vfs/filesystem.hpp that previously returned a StackErrorPtr have been changed to throw an exception on failure.
Besides that, the non-modifying functions have been marked as const, and some enumeration functions have been moved from RootFileSystem to FileSystem.

Note that
- in order to get the virtual_file_system_tests succeed, I had to manually adjust the base path `cwd` in default.cpp. That test failed in my environment even before any of my changes.
- all test were done on Windows; android_aasset_manager.cpp is just edited, not compiled.

### DIFF
--- a/components/vfs/include/components/vfs/filesystem.hpp
+++ b/components/vfs/include/components/vfs/filesystem.hpp
@@ -28,81 +28,6 @@ namespace components
 namespace vfs
 {
 /**
- * @brief Binary Blob container
- */
-class Blob
-{
-  public:
-	Blob()          = default;
-	virtual ~Blob() = default;
-
-	/**
-	 * @brief Query size of blob
-	 *
-	 * @return size_t size of blob in bytes
-	 */
-	virtual size_t size() const = 0;
-
-	/**
-	 * @brief Ptr to blob data
-	 *
-	 * @return const void* ptr to the start of the blob data
-	 */
-	virtual const void *data() const = 0;
-
-	/**
-	 * @brief Check if the blob contains data or not
-	 */
-	inline bool empty() const
-	{
-		return size() == 0 || data() == nullptr;
-	}
-
-	/**
-	 * @brief Convert blob to std vector of uint8_t
-	 *
-	 * @return std::vector<uint8_t> std vector of data
-	 */
-	inline std::vector<uint8_t> binary() const
-	{
-		if (empty())
-		{
-			return {};
-		}
-
-		const uint8_t *bin_ptr = static_cast<const uint8_t *>(data());
-		return {bin_ptr, bin_ptr + size()};
-	}
-
-	/**
-	 * @brief Convert blob to ascii string
-	 *
-	 * @return std::string ascii string
-	 */
-	inline std::string ascii() const
-	{
-		auto bin = binary();
-		return std::string{bin.begin(), bin.end()};
-	}
-};
-
-/**
- * @brief Blob implementation using std::vector<uint8_t>
- *
- */
-class StdBlob final : public Blob
-{
-  public:
-	StdBlob()          = default;
-	virtual ~StdBlob() = default;
-
-	virtual size_t      size() const override;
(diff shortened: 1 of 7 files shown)


TomAtkinsonArm:
Looking at the test cases this has convinced me that this is the way to go!

Shall we merge #495 and then address your further comments either here or in smaller PRs? If the tests pass here i'm happy for this to be merged on top of 495 :)

components/vfs/include/components/vfs/filesystem.hpp
```diff
@@ -138,35 +68,31 @@ class RootFileSystem : public FileSystem
...
-	virtual size_t        file_size(const std::string &file_path) override;
-	virtual StackErrorPtr write_file(const std::string &file_path, const void *data, size_t size) override;
-	virtual StackErrorPtr enumerate_files(const std::string &file_path, std::vector<std::string> *files) override;
-	virtual StackErrorPtr enumerate_folders(const std::string &file_path, std::vector<std::string> *folders) override;
-	virtual void          make_directory(const std::string &path) override;
-
-	StackErrorPtr enumerate_files(const std::string &file_path, const std::string &extension, std::vector<std::string> *files);
-	StackErrorPtr enumerate_files_recursive(const std::string &file_path, const std::string &extension, std::vector<std::string> *files);
-	StackErrorPtr enumerate_folders_recursive(const std::string &file_path, std::vector<std::string> *folders);
+	virtual bool                     folder_exists(const std::string &folder_path) const override;
```
gpx1000, line 71: It's a small thing.  But worth pointing out should keep override and remove virtual.

components/vfs/src/android_aasset_manager.cpp
```diff
@@ -145,34 +116,31 @@ StackErrorPtr AndroidAAssetManager::read_chunk(const std::string &file_path, con
...
+	return blob;
 }
 
-size_t AndroidAAssetManager::file_size(const std::string &file_path)
+size_t AndroidAAssetManager::file_size(const std::string &file_path) const
 {
 	if (!asset_manager)
 	{
-		return 0;
+		throw std::runtime_error("vfs/android_aasset_manager.cpp line" + std::to_string(__LINE__) + "AAsset Manager not initialized");
```
gpx1000, line 143: Does it make sense to make this a macro and use something like:
#define ERROR_STR(...) (( __FILE__ ":" __LINE__ " "  __VA_ARGS__))

Something to make the repeated tasks like listing out the file name a little less required and only capture the actual error response is what I'm looking for.  Can change the above as needed.

gpx1000:
I like this direction on the whole.  I think it's good to take this after #495

components/vfs/src/android_aasset_manager.cpp
TomAtkinsonArm: This should be addressed in a new PR, I will create an issue to follow up on this.

components/vfs/include/components/vfs/filesystem.hpp
TomAtkinsonArm: This is mainly my fault. Although its not needed, i've gotten into the habit of adding virtual to all virtual functions. We should address this in a new PR for the entire branch.

components/vfs/src/android_aasset_manager.cpp
gary-sweet: I'm not a fan of having the filename hard-wired like this. When someone wants to change the filename at some point in the future, it's going to be a little irritating to have to change all these too.

__FILE__ seems like a perfectly reasonable thing to use here, and hiding that inside a macro seems fine to me. If you really don't want the full path to appear in the output then maybe do some processing on the path to strip the front off.

TomAtkinsonArm:
Merged! issues can be addressed in later PRs see #505

PR #513, opened 2022-08-04
direct-to-display window improvements

gary-sweet:
## Description

Rather than just taking the first plane of the first display we now gather a list of valid candidates and pick the best fit.

An additional display 'FullscreenStretch' mode has been added. This uses VkDisplayPresentInfoKHR from the VK_KHR_display_swapchain extension, when available, to specify separate src and dst rectangles for the surface and display. This enables you to have a 640x360 surface stretched fullscreen onto the display for example, rather than just appearing in a corner.

In standard or stretch mode, with direct-to-display, specifying --width & --height controls the surface size.
In fullscreen or borderless modes, the surface is the same size as the display regardless of the width and height settings. The width and height are used to select the best display/plane combination however.

## General Checklist:

Note: Built and tested only on Linux

### DIFF
--- a/framework/hpp_api_vulkan_sample.cpp
+++ b/framework/hpp_api_vulkan_sample.cpp
@@ -485,10 +485,9 @@ void HPPApiVulkanSample::submit_frame()
 			present_info.setWaitSemaphores(semaphores.render_complete);
 		}
 
-		vk::DisplayPresentInfoKHR disp_present_info{};
+		vk::DisplayPresentInfoKHR disp_present_info;
 		if (device->is_extension_supported(VK_KHR_DISPLAY_SWAPCHAIN_EXTENSION_NAME) &&
-		    platform->get_window().get_display_present_info(reinterpret_cast<VkDisplayPresentInfoKHR *>(&disp_present_info),
-		                                                    extent.width, extent.height))
+		    get_platform().get_window().get_display_present_info(&disp_present_info, extent.width, extent.height))
 		{
 			// Add display present info if supported and wanted
 			present_info.setPNext(&disp_present_info);
--- a/framework/hpp_api_vulkan_sample.cpp
+++ b/framework/hpp_api_vulkan_sample.cpp
@@ -485,6 +485,15 @@ void HPPApiVulkanSample::submit_frame()
 			present_info.setWaitSemaphores(semaphores.render_complete);
 		}
 
+		vk::DisplayPresentInfoKHR disp_present_info{};
+		if (device->is_extension_supported(VK_KHR_DISPLAY_SWAPCHAIN_EXTENSION_NAME) &&
+		    platform->get_window().get_display_present_info(reinterpret_cast<VkDisplayPresentInfoKHR *>(&disp_present_info),
+		                                                    extent.width, extent.height))
+		{
+			// Add display present info if supported and wanted
+			present_info.setPNext(&disp_present_info);
+		}
+
 		// Shows how to filter an error code from a vulkan function, which is mapped to an exception but should be handled here!
 		vk::Result present_result;
 		try
--- a/framework/platform/hpp_window.h
+++ b/framework/platform/hpp_window.h
@@ -36,6 +36,7 @@ class HPPWindow : private vkb::Window
 	using vkb::Window::create_surface;
 	using vkb::Window::get_extent;
 	using vkb::Window::get_window_mode;
+	using vkb::Window::get_display_present_info;
 
 	vk::SurfaceKHR create_surface(vkb::core::HPPInstance &instance)
 	{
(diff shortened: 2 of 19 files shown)


framework/hpp_api_vulkan_sample.cpp
```diff
@@ -485,6 +485,15 @@ void HPPApiVulkanSample::submit_frame()
 			present_info.setWaitSemaphores(semaphores.render_complete);
 		}
 
+		vk::DisplayPresentInfoKHR disp_present_info{};
```
asuessenbach: You don't need to initialize the vk-structs with `{}`, as they have a default constructor.

framework/hpp_api_vulkan_sample.cpp
```diff
@@ -485,6 +485,15 @@ void HPPApiVulkanSample::submit_frame()
...
+		if (device->is_extension_supported(VK_KHR_DISPLAY_SWAPCHAIN_EXTENSION_NAME) &&
+		    platform->get_window().get_display_present_info(reinterpret_cast<VkDisplayPresentInfoKHR *>(&disp_present_info),
+		                                                    extent.width, extent.height))
+		{
```
asuessenbach, line 491: Would be great, if you could add some wrapper function to platform/hpp_window.h, like this:
```
	bool get_display_present_info(vk::DisplayPresentInfoKHR* info, uint32_t src_width, uint32_t src_height) const
	{
		return get_display_present_info(reinterpret_cast<VkDisplayPresentInfoKHR *>(info), src_width, src_height);
	}
```
to keep this file free from reinterpreting to VK-structs.

framework/platform/unix/direct_window.cpp
```diff
@@ -244,76 +244,193 @@ VkSurfaceKHR DirectWindow::create_surface(Instance &instance)
...
+                                      std::max(properties.extent.width, best.caps.minSrcExtent.width));
+		image_extent.height = std::min(best.caps.maxSrcExtent.height,
+		                               std::max(properties.extent.height, best.caps.minSrcExtent.height));
 	}
 
-	VkExtent2D image_extent;
-	image_extent.width  = mode_props.parameters.visibleRegion.width;
-	image_extent.height = mode_props.parameters.visibleRegion.height;
+	// Calculate the display DPI
+	dpi = 25.4f * best.display_props.physicalResolution.width / best.display_props.physicalDimensions.width;
```
asuessenbach: Magic number `25.4f`? What is it?
Replace it with some define or const?

framework/hpp_api_vulkan_sample.cpp
gary-sweet: Done

framework/hpp_api_vulkan_sample.cpp
gary-sweet: Done

framework/platform/unix/direct_window.cpp
gary-sweet: Done

framework/platform/hpp_window.h
```diff
@@ -46,6 +46,11 @@ class HPPWindow : private vkb::Window
 	{
 		return static_cast<vk::SurfaceKHR>(create_surface(static_cast<VkInstance>(instance), static_cast<VkPhysicalDevice>(physical_device)));
 	}
+
+	bool get_display_present_info(vk::DisplayPresentInfoKHR *info, uint32_t src_width, uint32_t src_height) const
+	{
+		return vkb::Window::get_display_present_info(reinterpret_cast<VkDisplayPresentInfoKHR *>(info), src_width, src_height);
```
asuessenbach: Unfortunately, you should not explicitly call `vkb::Window::get_display_present_info`, as that would really just call that function.
Please add
`	using vkb::Window::get_display_present_info;`
to the using-section in this file, and remove the explicit `vkb::Window::` here to really get the actual override called.

gary-sweet: Thanks. That was a subtlety that had passed me by the first time around. Following what's actually going on with the class hierarchies is a lot trickier when they are reinterpret_cast'ed to other things.

asuessenbach:
Besides the fact, that I have not checked the actual functionality (just checked on Windows where it is not used): LGTM now!

asuessenbach (approved):

gary-sweet:
I suspect that no-one else will actually be able to test the direct-to-display changes, so this may be a code-review only change.

gpx1000 (approved):

framework/vulkan_sample.cpp
```diff
@@ -131,6 +131,8 @@ bool VulkanSample::prepare(Platform &platform)
 
 	// Getting a valid vulkan surface from the platform
 	surface = platform.get_window().create_surface(*instance);
+	if (!surface)
+		throw std::runtime_error("Failed to create window surface.");
```
yangfengzzz, line 135: Headless failed

gary-sweet: This is a merged PR. If you believe there is a problem somewhere, please raise a new issue.

PR #514, opened 2022-08-10
Use <vulkan/vulkan.hpp> as precompiled header.

asuessenbach:
## Description

Use <vulkan/vulkan.hpp> as precompiled header with all the hpp-based samples and the framework, expecting build times to be reduced.

Tested on Windows 11.

## General Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.16)
 
 project(vulkan_samples)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2021, Arm Limited and Contributors
+# Copyright (c) 2020-2022, Arm Limited and Contributors
 #
 # SPDX-License-Identifier: Apache-2.0
 #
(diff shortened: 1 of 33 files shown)


SaschaWillems:
Just gave this a spin with VS2019 on Windows 10, and it doesn't compile for me. I get the same errors as in the CI build:

```fatal error C1010: unexpected end of file while looking for precompiled header. Did you forget to add '#include "pch.h"' to your source?```

gpx1000:
It does indeed work in Linux.  I'm wondering if there's a bug in the CMake version being used.  target_precompile_headers wasn't introduced until CMake 3.16 and our base minimum version is 3.12.  The windows build not using /fi to avoid needing to include the precompiled header is probably to do with not having a > CMake 3.16 designation maybe?  Although it might be a bug elsewhere in CMake as 3.24 looks like the version that CICD uses.

asuessenbach:
That's strange, I don't have that issue on Windows, using CMake 3.23 or CMake 3.24.
For testing, I'll remove the call to vulkan_samples_pch in framework/CMakeLists.txt, to see if there's some interference.

asuessenbach:
It seems, moving to CMake version 3.16 worked!

gary-sweet (commented):
This change works ok for me, but it appears to also be adding a new sample (hpp_hlsl_shaders) which isn't mentioned anywhere in the description.

asuessenbach:
> but it appears to also be adding a new sample (hpp_hlsl_shaders) which isn't mentioned anywhere in the description.

That sample is already in... It's just the case that I rebased my changes after that PR was merged, and then had it in the list of commits... Don't know how to prevent such a noisance :(

SaschaWillems (approved):
This compiles fine for me now with VS2019. And build times seem to be at least a bit faster :)

gpx1000 (approved):

gary-sweet (approved):

PR #518, opened 2022-09-01
Update and standardise sRGB usage

### LINKED ISSUE #516: Inconsistent and incorrect usage of color-spaces across samples
Correct use of sRGB surfaces and textures seems to be a rare thing. The samples sadly haven't fared any better.
It's quite easy to get things wrong and end up with a result that looks close to being correct, which is why it's sometimes hard to know that things are wrong in the first place.

Low-dynamic-range displays always expect to receive sRGB encoded data.

Here's my understanding of what the GPU hardware does w.r.t. color-spaces:

> Shader code expects to operate in linear space, so all the calculations can be done linearly.
> - Input textures that are UNORM are assumed to be in linear space already and the values within will get used by shaders directly.
> - Input textures that are sRGB are assumed to be non-linear and the values will be automatically converted to linear by the GPU when used by a shader.
> 
> The output values of the shader will be in linear space.
> - If the output surface is UNORM, the linear shader output will be copied directly to the surface.
> - If the output surface is sRGB, the linear shader output will be automatically converted to sRGB and then copied to the surface.

If you have a UNORM swapchain surface (which will be sent to a display which expects sRGB), you need to manually do the sRGB conversion at the end of your fragment shader. That is currently done in the basisu sample.

We really want all swapchain surfaces to be sRGB, so the GPU does the automatic conversion. Right now, all the performance samples correctly choose an sRGB swapchain, but the API samples all choose a UNORM swapchain.

The reason the API samples don't look terrible is that the texture data on disk is actually sRGB encoded - most image authoring tools save in sRGB. Those sRGB disk images are read into UNORM textures without any conversion, so we have UNORM textures containing sRGB data. The shaders operate on that sRGB data linearly (as no conversion is done by the GPU) and then write (slightly wrong) sRGB data to the UNORM surface. The result is incorrect, because non-linear data was processed linearly, but to the naked eye is close enough that you don't generally notice.

So we have `sRGB-in-UNORM -> linear-shader-processing -> sRGB-in-UNORM -> display`

The correct solution is to load the sRGB texture data from disk into an sRGB image. The GPU will then correctly convert that to linear for shader processing. With an sRGB swapchain surface the data will be converted back to sRGB ready for display.

`sRGB-in-sRGB -> auto-GPU-conv-to-linear ->linear-shader-processing -> auto-GPU-conv-to-sRGB ->sRGB-in-sRGB -> display`

_Note: this only applies to texture data that really contains color data. Normal maps and such should always be kept in linear space (i.e. loaded into UNORM images)._

I've made all these code changes locally, and will push a PR that addresses all these issues once it's passed internal review.

gary-sweet:
## Description

Most displays expect to receive sRGB encoded data. This change ensures that all of the samples request sRGB swapchain images. The GPU hardware will then ensure that the linear result of the fragment shader is automatically converted to non-linear sRGB when written to the swapchain.

The performance samples were already selecting sRGB as the priority format, but the API samples were not.

I've removed the UNORM formats from the swapchain format priority lists as it doesn't make much sense to have a UNORM swapchain image unless you want to manually convert to sRGB at the end of your fragment shaders.

Textures that are in UNORM formats are treated as already being linearly encoded and enter the shaders unaltered. SRGB texture values are converted by the GPU into linear space before being seen by shaders. This ensures that the shaders are doing
computations on linear data at all times.

Most image authoring tools will write the data encoded in sRGB. Some file formats don't record a colorspace within them though. In the cases where the disk file doesn't know the encoding the code now assumes that the data is sRGB encoded if, and only if, that image is going to be used as a color texture. For that reason, I needed to pass the intended content of an image down to the loaders. For image formats that know their colorspace, this is preserved.

The gltf loader has been modified slightly. The gltf spec is very clear about which images must be encoded as sRGB and which have to be linear. That knowledge is now in the loader, so the texture images are given the correct format during loading if it's not known from the image file.

I've modified some of the clear colors in the API samples to convert them into linear space. Previously they were effectively in sRGB space since the swapchain was unorm. This ensures that the clear color looks close to how it did previously.

After moving to sRGB output I noticed that the 'hdr' and 'dynamic_rendering' samples had some odd 'always white' areas. Debugging tracked the cause of this down to our h/w handling the clamping of Nan/Inf slightly differently for sRGB than unorms. I traced the root cause of Nan/Inf being produced by these fragment shaders for these samples and fixed it.

To summarise, the samples should now be doing this:

For textures that contain color data:
`sRGB texture -> GPU-conv -> linear shading -> GPU-conv -> sRGB swapchain -> display`

For textures that contain linear data (e.g. normals):
`unorm texture -> linear shading -> GPU-conv -> sRGB swapchain -> display`

Fixes #516

## General Checklist:

- [x] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions] - **only tested on Linux**
- [x] I have tested every sample to ensure everything runs correctly - **every sample that my h/w supports**

### DIFF
--- a/framework/gltf_loader.cpp
+++ b/framework/gltf_loader.cpp
@@ -1,5 +1,5 @@
-/* Copyright (c) 2018-2021, Arm Limited and Contributors
- * Copyright (c) 2019-2021, Sascha Willems
+/* Copyright (c) 2018-2022, Arm Limited and Contributors
+ * Copyright (c) 2019-2022, Sascha Willems
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -318,6 +318,18 @@ inline void upload_image_to_gpu(CommandBuffer &command_buffer, core::Buffer &sta
 		command_buffer.image_memory_barrier(image.get_vk_image_view(), memory_barrier);
 	}
 }
+
+static inline bool texture_needs_srgb_colorspace(const std::string &name)
+{
+	// The gltf spec states that the base and emissive textures MUST be encoded with the sRGB
+	// transfer function. All other texture types are linear.
+	if (name == "baseColorTexture" || name == "emissiveTexture")
+		return true;
+
+	// metallicRoughnessTexture, normalTexture & occlusionTexture must be linear
+	return false;
+}
+
 }        // namespace
 
 std::unordered_map<std::string, bool> GLTFLoader::supported_extensions = {
@@ -583,7 +595,12 @@ sg::Scene GLTFLoader::load_scene(int scene_index)
 			{
 				std::string tex_name = to_snake_case(gltf_value.first);
 
-				material->textures[tex_name] = textures.at(gltf_value.second.TextureIndex());
+				vkb::sg::Texture *tex = textures.at(gltf_value.second.TextureIndex());
+
+				if (texture_needs_srgb_colorspace(gltf_value.first))
+					tex->get_image()->coerce_format_to_srgb();
+
+				material->textures[tex_name] = tex;
 			}
 		}
 
@@ -593,7 +610,12 @@ sg::Scene GLTFLoader::load_scene(int scene_index)
 			{
 				std::string tex_name = to_snake_case(gltf_value.first);
 
-				material->textures[tex_name] = textures.at(gltf_value.second.TextureIndex());
+				vkb::sg::Texture *tex = textures.at(gltf_value.second.TextureIndex());
+
+				if (texture_needs_srgb_colorspace(gltf_value.first))
+					tex->get_image()->coerce_format_to_srgb();
+
+				material->textures[tex_name] = tex;
 			}
 		}
 
@@ -1257,7 +1279,7 @@ std::unique_ptr<sg::Image> GLTFLoader::parse_image(tinygltf::Image &gltf_image) 
 	{
 		// Load image from uri
 		auto image_uri = model_path + "/" + gltf_image.uri;
-		image          = sg::Image::load(gltf_image.name, image_uri);
+		image          = sg::Image::load(gltf_image.name, image_uri, vkb::sg::Image::Unknown);
 	}
 
 	// Check whether the format is supported by the GPU
--- a/framework/gltf_loader.cpp
+++ b/framework/gltf_loader.cpp
@@ -327,6 +327,7 @@ static inline bool texture_needs_srgb_colorspace(const std::string &name)
 		return true;
 
 	// metallicRoughnessTexture, normalTexture & occlusionTexture must be linear
+	assert(name == "metallicRoughnessTexture" || name == "normalTexture" || name == "occlusionTexture");
 	return false;
 }
 
--- a/framework/gui.cpp
+++ b/framework/gui.cpp
@@ -1,5 +1,5 @@
-/* Copyright (c) 2018-2021, Arm Limited and Contributors
- * Copyright (c) 2019-2021, Sascha Willems
+/* Copyright (c) 2018-2022, Arm Limited and Contributors
+ * Copyright (c) 2019-2022, Sascha Willems
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -106,6 +106,7 @@ Gui::Gui(VulkanSample &sample_, const Window &window, const Stats *stats,
 	ImGuiStyle &style = ImGui::GetStyle();
 
 	// Color scheme
+	style.Colors[ImGuiCol_WindowBg]         = ImVec4(0.0f, 0.0f, 0.0f, 0.94f);
 	style.Colors[ImGuiCol_TitleBg]          = ImVec4(1.0f, 0.0f, 0.0f, 0.6f);
 	style.Colors[ImGuiCol_TitleBgActive]    = ImVec4(1.0f, 0.0f, 0.0f, 0.8f);
 	style.Colors[ImGuiCol_MenuBarBg]        = ImVec4(1.0f, 0.0f, 0.0f, 0.4f);
--- a/framework/gui.cpp
+++ b/framework/gui.cpp
@@ -106,7 +106,7 @@ Gui::Gui(VulkanSample &sample_, const Window &window, const Stats *stats,
 	ImGuiStyle &style = ImGui::GetStyle();
 
 	// Color scheme
-	style.Colors[ImGuiCol_WindowBg]         = ImVec4(0.0f, 0.0f, 0.0f, 0.94f);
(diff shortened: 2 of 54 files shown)


framework/gltf_loader.cpp
```diff
@@ -318,6 +318,18 @@ inline void upload_image_to_gpu(CommandBuffer &command_buffer, core::Buffer &sta
...
 }
+
+static inline bool texture_needs_srgb_colorspace(const std::string &name)
+{
+	// The gltf spec states that the base and emissive textures MUST be encoded with the sRGB
+	// transfer function. All other texture types are linear.
+	if (name == "baseColorTexture" || name == "emissiveTexture")
+		return true;
+
+	// metallicRoughnessTexture, normalTexture & occlusionTexture must be linear
```
asuessenbach, line 329: I would add an assertion here, verifying that `name` really is one of the other expected textures.

framework/hpp_api_vulkan_sample.cpp
```diff
@@ -158,9 +158,7 @@ bool HPPApiVulkanSample::resize(const uint32_t, const uint32_t)
 
 void HPPApiVulkanSample::create_render_context(vkb::platform::HPPPlatform const &platform)
 {
-	auto surface_priority_list = std::vector<vk::SurfaceFormatKHR>{{vk::Format::eB8G8R8A8Unorm, vk::ColorSpaceKHR::eSrgbNonlinear},
-	                                                               {vk::Format::eR8G8B8A8Unorm, vk::ColorSpaceKHR::eSrgbNonlinear},
-	                                                               {vk::Format::eR8G8B8A8Srgb, vk::ColorSpaceKHR::eSrgbNonlinear},
+	auto surface_priority_list = std::vector<vk::SurfaceFormatKHR>{{vk::Format::eR8G8B8A8Srgb, vk::ColorSpaceKHR::eSrgbNonlinear},
```
asuessenbach, line 163: Maybe add the same comment as in ApiVulkanSample::create_render_context?

asuessenbach (commented):
Besides two very minor things, it looks good to me.
The terrain_tessellation looks very different, though. Is it "more correct" than before?

gary-sweet:
> The terrain_tessellation looks very different, though. Is it "more correct" than before?

Unfortunately, that's one of the samples I can't run on my h/w. Can you attach a screenshot of before and after perhaps?

framework/gltf_loader.cpp
gary-sweet: Done

framework/hpp_api_vulkan_sample.cpp
gary-sweet: Done

gary-sweet:
I do need to update the clear colors, as it seems I've messed some of them up somehow.
I've also managed to run the terrain_tessellation locally, so I'm looking at those differences too.

gary-sweet:
> I've also managed to run the terrain_tessellation locally, so I'm looking at those differences too.

I believe the new result is actually correct, but I'm going to tweak the example slightly to adjust the texture blending layer boundaries so that it looks more similar to the original.  When you're using the wrong color space in the shaders, as the original code was doing, you're performing linear calculations on non-linear data which will affect the texture blending being done.

I'll upload this fix and the correct clear colors tomorrow morning.

gary-sweet:
> I'll upload this fix and the correct clear colors tomorrow morning.

Done

gary-sweet:
@SaschaWillems The particle_rgba.ktx file used by compute_nbody (and others) looks like a greyscale luminance map for the particles. I get quite different results when I mark that image as 'Color' vs 'Other'. 'Color' will treat the file as being in sRGB, 'Other'  will treat is as linear (which is how it was effectively being treated previously).

This is the result when treated as linear (which is roughly what we got before):
grad_marked_as_other

And this is what I get if treated as sRGB:
grad_marked_as_color

Just wondering which of these was the intent, or which you would prefer now?

SaschaWillems:
@gary-sweet: I'd prefer the the sRGB one. In the linear ones the gradients don't look right and a bit over-satured.

P.S. : Thanks for this PR. Proper use of colorspaces is something lots of Vulkan developers struggle with, and having it correct in our repo would be great 👍

framework/gui.cpp
```diff
@@ -106,6 +106,7 @@ Gui::Gui(VulkanSample &sample_, const Window &window, const Stats *stats,
 	ImGuiStyle &style = ImGui::GetStyle();
 
 	// Color scheme
+	style.Colors[ImGuiCol_WindowBg]         = ImVec4(0.005f, 0.005f, 0.005f, 0.94f);
```
SaschaWillems, line 109: Minor suggestion: We should also adjust the border color, as the default one results in a bright, almost white, very noticeable border.

This should look good and close to how it was before:

```cpp
style.Colors[ImGuiCol_Border] = ImVec4(0.02f, 0.02f, 0.02f, 0.94f);
```

gary-sweet: I'm now officially confused. I don't see a border at all.
The code has `style.WindowBorderSize = 0.0f;` which I assume should remove it.

gary-sweet:
> I'd prefer the the sRGB one.

Ok - that's what is currently done in this PR.

framework/gui.cpp
SaschaWillems: Right. I somehow hat borders enabled for a local branch. Forgot we disabled them for this repo.

gary-sweet: Ok - that explains it

asuessenbach (dismissed):

framework/scene_graph/components/image.cpp
```diff
@@ -68,6 +68,81 @@ bool is_astc(const VkFormat format)
 	        format == VK_FORMAT_ASTC_12x12_SRGB_BLOCK);
 }
 
+static VkFormat maybe_coerce_to_srgb(VkFormat fmt)
```
TomAtkinsonArm, line 76: A small comment here would be helpful for future reference

TomAtkinsonArm (changes requested):
LGTM - Compute nbody looks much better now. Code looks good :)

framework/scene_graph/components/image.cpp
gary-sweet: Done

TomAtkinsonArm (commented):
LGTM - thanks for the extra comment

asuessenbach (approved):

gpx1000 (approved):

SaschaWillems (approved):
LGTM.

Some samples now obviously look different, but most probably just looked wrong before this change.

As already noted e.g. the compute nbody sample looks better now :)

PR #519, opened 2022-09-10
Add timestamp query sample

### LINKED ISSUE #515: Add new sample for timestamp queries
A new sample should be added that shows how to use timestamp queries, that can be used to fetch execution times for given shader stages / workloads from the GPU.

SaschaWillems:
## Description

This PR adds a new sample and tutorial for using timestamp queries with Vulkan.

Sample was tested on an nvidia RTX 2060 with Windows 10 and a Google Pixel 4a on Android 13.

*Note*: The sample comes with a tutorial (https://github.com/KhronosGroup/Vulkan-Samples/blob/b10c9ed4be5fb2697fdfa68390ec1c911bd7a250/samples/api/timestamp_queries/README.md). Timestamp queries have lots of limitations, and I tried to explain them as good as I could. It would be great if people doing the reviews could also check the tutorial. Any feedback on it is appreciated.

Fixes #515

## General Checklist:

## Sample Checklist

### DIFF
--- a/samples/api/timestamp_queries/README.md
+++ b/samples/api/timestamp_queries/README.md
@@ -29,24 +29,37 @@ Vulkan offers several [query types](https://registry.khronos.org/vulkan/specs/1.
 
 This provides your application with a mechanism to time the execution of commands on the GPU. As with the other query types, a query pool is then used to either directly fetch or copy over the results to the host.
 
-## A note on timestamp queries
+## A few important notes on timestamp queries
 
-It's important to know that timestamp queries differ greatly from how timing can be done on the CPU with e.g. the high performance counter on Windows. This is mostly due to how a GPU's dispatches, overlaps and finishes work across different stages of the pipeline. So while in theory you can specify any pipeline stage at which the timestamp should be written, a lot of stage combinations and orderings won't give meaningful result.
+It's important to know that timestamp queries differ greatly from how timing can be done on the CPU with e.g. the high performance counter. This is mostly due to how a GPU's dispatches, overlaps and finishes work across different stages of the pipeline. So while in theory you can specify any pipeline stage at which the timestamp should be written, a lot of stage combinations and orderings won't give meaningful result. This also means that you you can't compare timestamps taken on different queues.
 
-So while it may sound reasonable to e.g. write timestamps for the vertex and fragment shader stage one after another, it will not give useful results.
+So while it may may sound reasonable to write timestamps for the vertex and fragment shader stage directly one after another, that will usually not return meaningful results due to how the GPU works.
 
 And so for this example, we take the same approach as some popular CPU/GPU profilers by only using the top and bottom stages of the pipeline. This combination is known to give proper timing results on most GPUs.
 
 ## Checking for support
 
-Not all devices support timestamp queries, so before using them we need to make sure that they can be used. This differs slightly from checking other features with a simple `VkBool`. Here we need to check if the `timestampPeriod` limit of the physical device is greater than zero. If that's the case, timestamp queries are supported:
+Not all GPUs support timestamp queries, so before using them we need to make sure that they can be used. This differs slightly from checking other features with a simple `VkBool`. Here we need to check if the `timestampPeriod` limit of the physical device is greater than zero. If that's the case, timestamp queries are supported:
 
 ```cpp
-// Check if the selected device supports timestamps. A value of zero means no support.
 VkPhysicalDeviceLimits device_limits = device->get_gpu().get_properties().limits;
 if (device_limits.timestampPeriod == 0)
 {
-    throw std::runtime_error{"The selected device does not support timestamp queries!"};
+	throw std::runtime_error{"The selected device does not support timestamp queries!"};
+}
+```
+
+Another limit we need to check is `timestampComputeAndGraphics`. If this is `VK_TRUE`, all graphics and compute pipelines support timestamp queries and the above check is sufficient. If not, we need to check if the queue we want to use supports timestamps:
+
+```cpp
+if (!device_limits.timestampComputeAndGraphics)
+{
+	// Check if the graphics queue used in this sample supports time stamps
+	VkQueueFamilyProperties graphics_queue_family_properties = device->get_suitable_graphics_queue().get_properties();
+	if (graphics_queue_family_properties.timestampValidBits == 0)
+	{
+		throw std::runtime_error{"The selected graphics queue family does not support timestamp queries!"};
+	}
 }
 ```
 
@@ -94,22 +107,26 @@ vkCmdResetQueryPool(draw_cmd_buffers[i], query_pool_timestamps, 0, static_cast<u
 
 ## Writing time stamps
 
-// @todo: talk about pipeline stages
+Unlike getting CPU side timing information that can be queried immediately, with GPU time stamps we need to tell the implementation inside a command buffer when/where to write timestamps instead. The results are then fetched afterwards (see below).
 
-// @todo: When vkCmdWriteTimestamp is submitted to a queue, it defines an execution dependency on
-commands that were submitted before it, and writes a timestamp to a query pool.
+This is done inside the command buffer with `vkCmdWriteTimestamp`. This function will request a timestamp to be written from the GPU for a certain pipeline stage and write that value to memory.
 
-If an implementation is unable to detect completion and latch the timer
-immediately after stage has completed, it may instead do so at any logically later
-stage
+The most interesting part of calling this function is the `pipelineStage` argument. As noted earlier, it's theoretically possible to use any pipeline stage in here, not all pipeline stages will net proper results due to how GPUs overlap work. It's also important to note that not all implementations are able to latch timers at all pipeline stages (e.g. if they don't have hardware that maps to a given stage) and may return timers at a later pipeline stage instead.
+ 
+Calling this function also defines an execution dependency similar to a barrier on all commands that were submitted before it.
 
 ```cpp
 vkCmdWriteTimestamp(draw_cmd_buffers[i], VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, query_pool_timestamps, 0);
 // Do some work
+for (int i = 0; i < draw_call_count; i++) {
+	vkCmdDraw(...);
+}
 vkCmdWriteTimestamp(draw_cmd_buffers[i], VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT, query_pool_timestamps, 1);
 ```
 
-@todo: maybe insert image with pipeline stages from top to bottom
+To measure GPU times for the draw calls(s) we first tell the GPU to write a timestamp at the `VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT` pipeline stage. This is not a real pipeline stage (as in e.g. the vertex or fragment stages) but a special constant that tells the GPU to wait for all commands currently in flight to finish. This ensures, that we get a timestamp right before starting on the draw calls we want to measure. This timestamp will then be the base for calculating a delta time.
+
+The second timestamp is written at the `VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT` pipeline stage. Once again this is not a real pipeline stage, but it tells the GPU to wait for all work to be finished. 
 
 ## Getting the results
 
@@ -119,8 +136,6 @@ Reading back the results can be done in two ways:
 - Get the results after the command buffer has finished executing using `vkGetQueryPoolResults`
 
 For our sample we'll use option two (see `get_time_stamp_results`):
-
-@todo: another option is to do a bufer copy in the command buffer, but that requires a dedicated buffer.
 
 ```cpp
 vkQueueSubmit();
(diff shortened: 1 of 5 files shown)


samples/api/timestamp_queries/README.md
```diff
@@ -0,0 +1,184 @@
...
+```cpp
+vkCmdWriteTimestamp(draw_cmd_buffers[i], VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, query_pool_timestamps, 0);
+// Do some work
+for (int i = 0; i < draw_call_count; i++) {
+	vkCmdDraw(...);
+}
+vkCmdWriteTimestamp(draw_cmd_buffers[i], VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT, query_pool_timestamps, 1);
+```
+
+To measure GPU times for the draw calls(s) we first tell the GPU to write a timestamp at the `VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT` pipeline stage. This is not a real pipeline stage (as in e.g. the vertex or fragment stages) but a special constant that tells the GPU to wait for all commands currently in flight to finish. This ensures, that we get a timestamp right before starting on the draw calls we want to measure. This timestamp will then be the base for calculating a delta time.
```
cforfang: > [TOP is] a special constant that tells the GPU to wait for all commands currently in flight to finish

This seems wrong -- TOP here means wait for all previous commands to finish the TOP stage, which in turns means... well, nothing, I think, since the TOP stage has no observable effects?

If you want to make sure all previous commands are _done_ executing, shouldn't this be BOTTOM?

samples/api/timestamp_queries/timestamp_queries.cpp
```diff
@@ -0,0 +1,1018 @@
...
+	query_pool_info.queryType = VK_QUERY_TYPE_TIMESTAMP;
+	// Set the no. of queries in this pool
+	query_pool_info.queryCount = static_cast<uint32_t>(time_stamps.size());
+	VK_CHECK(vkCreateQueryPool(device->get_handle(), &query_pool_info, nullptr, &query_pool_timestamps));
+}
+
+void TimestampQueries::get_time_stamp_results()
+{
+	// The number of timestamps changes if the bloom pass is disabled
+	uint32_t count = bloom ? time_stamps.size() : time_stamps.size() - 2;
```
asuessenbach: You should static_cast<uint32_t> here.

samples/api/timestamp_queries/README.md
asuessenbach: I think, the usage of TOP and BOTTOM here is correct.
But the wording might be misleading: the GPU doesn't wait for anything. Writing a timestamp does not imply some waitIdle, or so.
Instead, with TOP, the timestamp is written as soon as all previous commands were taken by the GPU's command processor (not necessarily started yet, though!), and with BOTTOM, the timestamp is written as soon as all previous commands were done.

samples/api/timestamp_queries/README.md
```diff
@@ -0,0 +1,184 @@
...
+- See the License for the specific language governing permissions and
+- limitations under the License.
+-
+-->
+
+# Timestamp queries
+
+This tutorial, along with the accompanying example code, shows how to use timestamp queries to measure timings on the GPU.
+
+The sample, based on the HDR one, does multiple render passes and will use timestamp queries to get GPU timings for the different render passes. This is done by writing GPU timestamps at certain points within a command buffer. These can then be read on the host and used for light profiling and to e.g. improve performance where needed.
```
asuessenbach: Instead of "light profiling" maybe something like "rough profiling" or "approximate profiling" would be a better fit?

samples/api/timestamp_queries/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Introduction
+
+Vulkan offers several [query types](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#queries) that allow you to query different types of information from the GPU. One such query type is the [timestamp query](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#queries-timestamps).
+
+This provides your application with a mechanism to time the execution of commands on the GPU. As with the other query types, a query pool is then used to either directly fetch or copy over the results to the host.
+
+## A few important notes on timestamp queries
+
+It's important to know that timestamp queries differ greatly from how timing can be done on the CPU with e.g. the high performance counter. This is mostly due to how a GPU's dispatches, overlaps and finishes work across different stages of the pipeline. So while in theory you can specify any pipeline stage at which the timestamp should be written, a lot of stage combinations and orderings won't give meaningful result. This also means that you you can't compare timestamps taken on different queues.
```
asuessenbach: "So while in theory..." -> "So while technically..." ?
That is, it's not only in theory, that you can specify any pipeline stage. The bits to do so are available. But, as you say, it's of very limited value.

samples/api/timestamp_queries/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+So while it may may sound reasonable to write timestamps for the vertex and fragment shader stage directly one after another, that will usually not return meaningful results due to how the GPU works.
+
+And so for this example, we take the same approach as some popular CPU/GPU profilers by only using the top and bottom stages of the pipeline. This combination is known to give proper timing results on most GPUs.
```
asuessenbach: "proper timing results" -> "proper approximate timing results" ?

samples/api/timestamp_queries/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+------
+
+## Writing time stamps
+
+Unlike getting CPU side timing information that can be queried immediately, with GPU time stamps we need to tell the implementation inside a command buffer when/where to write timestamps instead. The results are then fetched afterwards (see below).
+
+This is done inside the command buffer with `vkCmdWriteTimestamp`. This function will request a timestamp to be written from the GPU for a certain pipeline stage and write that value to memory.
+
+The most interesting part of calling this function is the `pipelineStage` argument. As noted earlier, it's theoretically possible to use any pipeline stage in here, not all pipeline stages will net proper results due to how GPUs overlap work. It's also important to note that not all implementations are able to latch timers at all pipeline stages (e.g. if they don't have hardware that maps to a given stage) and may return timers at a later pipeline stage instead.
```
asuessenbach: as above: "theoretically" -> "technically" ?
"... will net proper results..." ?

samples/api/timestamp_queries/README.md
asuessenbach: As already said: the GPU doesn't wait for anything here! It's just that the timestamp is written right after all previous commands are digested by the GPU's command processor. That's why it's just an approximate measurement: some work from previous commands might "tunnel through" into the measurement interval.

samples/api/timestamp_queries/README.md
```diff
@@ -0,0 +1,184 @@
...
+// Do some work
+for (int i = 0; i < draw_call_count; i++) {
+	vkCmdDraw(...);
+}
+vkCmdWriteTimestamp(draw_cmd_buffers[i], VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT, query_pool_timestamps, 1);
+```
+
+To measure GPU times for the draw calls(s) we first tell the GPU to write a timestamp at the `VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT` pipeline stage. This is not a real pipeline stage (as in e.g. the vertex or fragment stages) but a special constant that tells the GPU to wait for all commands currently in flight to finish. This ensures, that we get a timestamp right before starting on the draw calls we want to measure. This timestamp will then be the base for calculating a delta time.
+
+The second timestamp is written at the `VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT` pipeline stage. Once again this is not a real pipeline stage, but it tells the GPU to wait for all work to be finished.
```
asuessenbach: No wait!

samples/api/timestamp_queries/README.md
```diff
@@ -0,0 +1,184 @@
...
+- Copy the results into a `VkBuffer` inside the command buffer using `vkCmdCopyQueryPoolResults`
+- Get the results after the command buffer has finished executing using `vkGetQueryPoolResults`
+
+For our sample we'll use option two (see `get_time_stamp_results`):
+
+```cpp
+vkQueueSubmit();
+...
+// The number of timestamps changes if the bloom pass is disabled
+uint32_t count = bloom ? time_stamps.size() : time_stamps.size() - 2;
```
asuessenbach, line 144: static_cast<uint32_t> here

samples/api/timestamp_queries/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+Most arguments are straightforward, e.g. where the data will be copied to (the `time_stamps` vector). The important part here are the `VK_QUERY_RESULT_` flags used here.
+
+`VK_QUERY_RESULT_64_BIT` will tell the api that we want to get the results as 64 bit values. Without this flag, we would only get 32 bit values. And since timestamp queries can operate in nanoseconds, only using 32 bits could result into an overflow. E.g. if your device has a `timestampPeriod` of 1, so that one increment in the result maps to exactly one nanosecond, with 32 bit precision you'd run into such an overflow after only about 0.43 seconds.
+
+The `VK_QUERY_RESULT_WAIT_BIT` bit then tells the api to wait for all results to be available. So when using this flag the values written to our `time_stamps` vector is guaranteed to be available after calling `vkGetQueryPoolResults`. This is fine for our use-case where we want to immediately access the results, but may introduce unnecessary stalls in other scenarios. An alternative is using the `VK_QUERY_RESULT_WITH_AVAILABILITY_BIT`, which will let you poll the availability of the results and defer writing new timestamps until the results are available.
```
asuessenbach: Maybe you should elaborate a bit more on how to use the WITH_AVAILABILITY flag.
Any maybe we could add an alternative code path to the sample, where that flag is used?

samples/api/timestamp_queries/timestamp_queries.cpp
```diff
@@ -0,0 +1,1018 @@
...
+
+#include "timestamp_queries.h"
+
+#include "scene_graph/components/sub_mesh.h"
+
+TimestampQueries::TimestampQueries()
+{
+	title = "Timestamp queries";
+	// This sample uses vkCmdResetQueryPool to reset the timestamp query pool on the host, which requires VK_EXT_host_query_reset or Vulkan 1.2
+	add_device_extension(VK_EXT_HOST_QUERY_RESET_EXTENSION_NAME);
```
asuessenbach, line 30: Maybe just add this extension with vulkan version less than 1.2?

asuessenbach (commented):
Overall, the sample looks good to me.
You can nicely see how the time for the actual scene rendering varies with the different geometries, while the scene display and the second bloom path vary with window size, and switching the skybox on and off influences the scene rendering and the display, but not the bloom path.

samples/api/timestamp_queries/README.md
cforfang: But for the purpose of timing individual workloads, don't you want the two time-stamps to reflect (a) when the previous work ended, and (b) when the work you want to time ended?

A timestamp in TOP can legally be taken immediately after the vkQueueSubmit -- in the same way that e.g. an image transitions with srcStage = TOP can legally be executed immediately after vkQueueSubmit. Furthermore, _all_ timestamps with srcStage = TOP can be taken _at the same time_ immediately after vkQueueSubmit, which does not seem like the intent here. (The sample code takes a timestamp with srcStage=TOP prior to every render-pass -- we definitely don't want these taken at roughly the same time immediately after submit, but that is legal per the spec as far as I can tell.)

(That said, it's true that on many GPUs the command processor will not re-order work in this way, so in practice on such GPUs, this is not a problem. But there are also implementations _without_ a global command processor, which do re-order (or more generally, split the logically single command stream into multiple independent streams executing independently except for when implicit or explicit execution dependencies force synchronization between them). It's these I mostly have in mind here, so the sample is likely to work as intended everywhere.)

That's why it seems like src=BOTTOM followed by src=BOTTOM makes the most sense here, so that you're more likely to get:
- Timestamp before the work starts (after previous work is done), and
- Timestamp after the work is done

samples/api/timestamp_queries/timestamp_queries.cpp
SaschaWillems: The framework will always require a 1.0 instance unless a sample requests something different. So for this sample, adding the extension is a requirement that can't be changed.

samples/api/timestamp_queries/README.md
SaschaWillems: Yeah, that makes more sense. Will change the wording here.

samples/api/timestamp_queries/README.md
SaschaWillems: Will change that. The net makes no sense here, will replace it with "yield".

samples/api/timestamp_queries/README.md
asuessenbach: I think, it kind of depends on what you want to measure here... Because of the parallel work scheduling of modern GPUs,
* with TOP to BOTTOM, you might have some work from previous commands on your bill, but can be sure to measure all your work.
* with BOTTOM to BOTTOM, you can be sure to not have any work from previous commands on your bill, but might have missed some of your work.

That is, with modern GPUs, it's hard to have a "Timestamp before the work starts (after previous work is done)", because there might be no point in time that fits that criterium. You actually described exactly that situation in your parenthesized paragraph.
The only way to be sure, that described point in time exists, would be to waitIdle right before the first timestamp. And then it shouldn't matter if it's TOP or BOTTOM.

samples/api/timestamp_queries/timestamp_queries.cpp
asuessenbach: > The framework will always require a 1.0 instance unless a sample requests something different.

That's probably true, and will ever be valid.
But I think, it's better to have restrictions in code, not in comments (even though, it's not really a restriction here: you can get the extension, even though you might be running on 1.3). So a check of `vkEnumerateInstanceVersion` against `VK_MAKE_API_VERSION( 0, 1, 2, 0 )` would also make it clear for the casual reader.

samples/api/timestamp_queries/README.md
cforfang: I'm not sure I understand this concern:
> with BOTTOM to BOTTOM, you can be sure to not have any work from previous commands on your bill, but might have missed some of your work.

How could some of the render-pass work not be included given BOTTOM is the last stage in the pipeline and the spec for vkCmdWriteTimestamp says:
```The first synchronization scope includes all commands that occur earlier in submission order. The synchronization scope is limited to operations on the pipeline stage specified by pipelineStage.```

I'd think all work in the render-pass must be included in the timestamp's first sync scope here, as for something to _not_ be included it would have to execute _after_ the BOTTOM stage -- which is not possible? (That said, a possible concern is that _other_ work, after the render-pass, could be done _before_ the timestamp (so long as the timestamp's BOTTOM stage execute before the other works' BOTTOM stage). To avoid this, one could e.g. emit a BOTTOM -> TOP pipeline barrier directly after the barrier, however.)

cforfang: Ah, right, I think I see what you mean now :) The BOTTOM timestamp _before_ the render-pass is not guaranteed to be taken before the render-pass starts. What do you think about adding a pipe-barrier as a workaround for this?

In summary:
- Timestamp in BOTTOM (i.e. after all previous work)
- BOTTOM -> TOP barrier (guarantees timestamp is taken before the upcoming render-pass starts)
- Render-pass
- Timestamp in BOTTOM (i.e. after all previous work)
- Optionally another BOTTOM -> TOP barrier here to make sure the previous timestamp is taken prior to any more work starting (but can skip this if there is a BOTTOM -> TOP barrier before the upcoming work anyway)
- .. repeat for each render-pass ..

asuessenbach: Yes, with such a barrier, you would make sure that the timestamp is taken after all previous work and before the work to be measured. And it shouldn't matter then, whether you start at BOTTOM or at TOP.
But of course, you would stall the GPU.
@SaschaWillems: after all this discussion here, it seems worthwhile to have a few words on this topic in the tutorial as well.

gary-sweet:
Sample works fine for me. Looks like there are a few changes requested though, so will hold off on the approval.

samples/api/timestamp_queries/README.md
asuessenbach: ... and maybe such barriers can be used to get more accurate timing results on the single stages as well?
I think, that's definitely worth to be tried.

samples/api/timestamp_queries/timestamp_queries.cpp
SaschaWillems: While it's a perfectly good point, we don't do that in any of our samples. So if you're fine with it, I'd leave it the way it currently is implemented.

samples/api/timestamp_queries/README.md
SaschaWillems: Will fix this.

samples/api/timestamp_queries/README.md
SaschaWillems: I have rephrased this.

samples/api/timestamp_queries/README.md
SaschaWillems: I have rephrased this.

samples/api/timestamp_queries/README.md
SaschaWillems: I have added approximate.

samples/api/timestamp_queries/README.md
SaschaWillems: Adding an alternative code path to the sample probably won't work. Using the availability flag you'd need to not write timestamps in the command buffer until the results are available. Since our samples pre-built the command buffers, this isn't easy to implement.

I'll elaborate a bit on this in the readme.

SaschaWillems: I added some more details on that flag along with a code example, please see https://github.com/KhronosGroup/Vulkan-Samples/blob/dcfb2f6cb446ad38375d3e67e392d8074c2340f7/samples/api/timestamp_queries/README.md#getting-the-results

samples/api/timestamp_queries/README.md
SaschaWillems: Thanks everyone. That's a valuable discussion :)

But I'd rather not add barriers in between.  I just took a look how tracy, a very popular CPU/GPU profiler, uses timestamp queries and it seems to be at least similar to what I'm doing. With the exception that tracy always only uses the BOTTOM stage bit. But no barriers.

As noted in the other comments, I have changed to wording in the tutorial to make sure that there are no GPU waits.

gary-sweet (approved):

asuessenbach (approved):

gpx1000 (approved):

PR #528, opened 2022-09-29
Add new descriptor management strategy: create descriptors directly; …

### LINKED ISSUE #527: descriptor_management sample gives misleading results
### Premise
In the README file of descriptor_management sample, it is stated that the sample shows how slow _"simplest descriptor management scheme"_ is and that introducing a descriptor set cache used in the framework _"has a dramatic effect on frame time"_. The sample itself allows setting **Descriptor set caching** to **Enabled** or **Disabled**. One might conclude that this sample demonstrates the performance gain that can be achieved by introducing a descriptor set cache into a Vulkan app.

### Problem
By looking into VS profiler when the cache is disabled and then looking in the code I noticed that all choosing the **Disabled** option
does is it clears descriptors set cache and resets pools for the current frame. The cache is still queried and filled.
The sample doesn't show the performance difference between recreating descriptor sets every frame and using descriptor set cache.
It shows the difference between 
getting descriptor sets from filled descriptor cache
and 
querying empty cache, recreating descriptor sets, putting them into the cache, and clearing the cache.
Obviously, the first option is always faster, it's the same, but less work.

Simply put, **Descriptor set caching** is never truly **Disabled** in this sample.

### Motivation
Implementing descriptor set cache in an engine/renderer is a tradeoff between saving time on creating cached objects (here: descriptor sets) and introducing cache overhead (hashing, comparing, copying, possibly allocating memory).

DmitryTsyganov:
…use it in descriptor_management. (#527)

* Add vkb::DescriptorManagementStrategy enum
* Add new descriptor management strategy: recreating descriptor sets every frame
* Use new strategy in descriptor_management

## Description

This pull request changes the descriptor_management sample in a way that when **Disabled** is chosen for **Descriptor set caching**, using descriptor cache is fully avoided and the descriptor sets are allocated and updated directly.

This allows users to get a more realistic estimation of how much performance can implementing descriptor set cache save.

The difference between the results can be significant. Here are the results from my realme XT (Android 11):
Original:
image
image
After my changes:
image

Fixes #527

## General Checklist:

## Sample Checklist

### DIFF
--- a/framework/core/command_buffer.cpp
+++ b/framework/core/command_buffer.cpp
@@ -712,11 +712,23 @@ void CommandBuffer::flush_descriptor_state(VkPipelineBindPoint pipeline_bind_poi
 				}
 			}
 
-			// Request a descriptor set from the render frame, and write the buffer infos and image infos of all the specified bindings
-			auto &descriptor_set = command_pool.get_render_frame()->request_descriptor_set(descriptor_set_layout, buffer_infos, image_infos, command_pool.get_thread_index());
-			descriptor_set.update(bindings_to_update);
+			VkDescriptorSet descriptor_set_handle;
+			if (command_pool.get_render_frame()->get_descriptor_management_strategy() == DescriptorManagementStrategy::StoreInCache)
+			{
+				// Request a descriptor set from the render frame, and write the buffer infos and image infos of all the specified bindings
+				auto &descriptor_set = command_pool.get_render_frame()->request_descriptor_set(descriptor_set_layout, buffer_infos, image_infos, command_pool.get_thread_index());
+				descriptor_set.update(bindings_to_update);
+				descriptor_set_handle = descriptor_set.get_handle();
+			}
+			else
+			{
+				// Request a descriptor pool, allocate a descriptor set, write buffer and image data to it
+				auto &descriptor_pool = command_pool.get_render_frame()->request_descriptor_pool(descriptor_set_layout, command_pool.get_thread_index());
+				DescriptorSet descriptor_set{command_pool.get_render_frame()->get_device(), descriptor_set_layout, descriptor_pool, buffer_infos, image_infos};
+				descriptor_set.apply_writes();
+				descriptor_set_handle = descriptor_set.get_handle();
+			}
 
-			VkDescriptorSet descriptor_set_handle = descriptor_set.get_handle();
 
 			// Bind descriptor set
 			vkCmdBindDescriptorSets(get_handle(),
--- a/framework/core/command_buffer.cpp
+++ b/framework/core/command_buffer.cpp
@@ -623,9 +623,6 @@ void CommandBuffer::flush_descriptor_state(VkPipelineBindPoint pipeline_bind_poi
 
 			std::vector<uint32_t> dynamic_offsets;
 
-			// The bindings we want to update before binding, if empty we update all bindings
-			std::vector<uint32_t> bindings_to_update;
-
 			// Iterate over all resource bindings
 			for (auto &binding_it : resource_set.get_resource_bindings())
 			{
@@ -635,12 +632,6 @@ void CommandBuffer::flush_descriptor_state(VkPipelineBindPoint pipeline_bind_poi
 				// Check if binding exists in the pipeline layout
 				if (auto binding_info = descriptor_set_layout.get_layout_binding(binding_index))
 				{
-					// If update after bind is enabled, we store the binding index of each binding that need to be updated before being bound
-					if (update_after_bind && !(descriptor_set_layout.get_layout_binding_flag(binding_index) & VK_DESCRIPTOR_BINDING_UPDATE_AFTER_BIND_BIT_EXT))
-					{
-						bindings_to_update.push_back(binding_index);
-					}
-
 					// Iterate over all binding resources
 					for (auto &element_it : binding_resources)
 					{
@@ -716,7 +707,7 @@ void CommandBuffer::flush_descriptor_state(VkPipelineBindPoint pipeline_bind_poi
 			    command_pool.get_render_frame()->request_descriptor_set(descriptor_set_layout,
 			                                                            buffer_infos,
 			                                                            image_infos,
-			                                                            bindings_to_update,
+			                                                            update_after_bind,
 			                                                            command_pool.get_thread_index());
 
 			// Bind descriptor set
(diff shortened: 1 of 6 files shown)


framework/core/command_buffer.cpp
```diff
@@ -712,11 +712,22 @@ void CommandBuffer::flush_descriptor_state(VkPipelineBindPoint pipeline_bind_poi
...
+				descriptor_set_handle = descriptor_set.get_handle();
+			}
+			else
+			{
+				// Request a descriptor pool, allocate a descriptor set, write buffer and image data to it
+				auto &        descriptor_pool = command_pool.get_render_frame()->request_descriptor_pool(descriptor_set_layout, command_pool.get_thread_index());
+				DescriptorSet descriptor_set{command_pool.get_render_frame()->get_device(), descriptor_set_layout, descriptor_pool, buffer_infos, image_infos};
+				descriptor_set.apply_writes();
+				descriptor_set_handle = descriptor_set.get_handle();
+			}
```
asuessenbach: All operations here are done via `command_pool.get_render_frame()->`.
Looks like this functionality should be moved into `vkb::RenderFrame`, such that the `vkb::CommandBuffer` doesn't even need to know about the `DescriptorManagementStrategy`?

framework/core/descriptor_set.h
```diff
@@ -76,6 +76,11 @@ class DescriptorSet
 	 */
 	void update(const std::vector<uint32_t> &bindings_to_update = {});
 
+	/**
+	 * @brief Appies pending write operations without updating the state
```
asuessenbach: Appies -> Applies

framework/rendering/render_frame.cpp
```diff
@@ -180,10 +185,17 @@ DescriptorSet &RenderFrame::request_descriptor_set(DescriptorSetLayout &descript
 {
 	assert(thread_index < thread_count && "Thread index is out of bounds");
 
-	auto &descriptor_pool = request_resource(device, nullptr, *descriptor_pools.at(thread_index), descriptor_set_layout);
+	auto &descriptor_pool = request_descriptor_pool(descriptor_set_layout, thread_index);
 	return request_resource(device, nullptr, *descriptor_sets.at(thread_index), descriptor_set_layout, descriptor_pool, buffer_infos, image_infos);
 }
 
+DescriptorPool &RenderFrame::request_descriptor_pool(DescriptorSetLayout &descriptor_set_layout, size_t thread_index)
```
asuessenbach: `descriptor_set_layout` isn't modified here - should be a `const &`

framework/rendering/render_frame.h
```diff
@@ -40,6 +40,12 @@ enum BufferAllocationStrategy
 	MultipleAllocationsPerBuffer
 };
 
+enum DescriptorManagementStrategy
+{
+	StoreInCache,
+	CreateDirectly
```
asuessenbach, line 46: Maybe `Cached` and `Uncached`, instead of `StoreInCache` and `CreateDirectly`?

asuessenbach: ... and maybe this `enum` can even be replaced by a simple `bool`?
Or do you expect to potentially have more than just those two strategies?

framework/core/command_buffer.cpp
DmitryTsyganov: Good point, moved it

framework/core/descriptor_set.h
DmitryTsyganov: Fixed

framework/rendering/render_frame.cpp
DmitryTsyganov: Ended up deleting the method due to another comment, but noticed that `request_descriptor_set` and `descriptor_set` itself have the same issue, so added `const` there as well.

framework/rendering/render_frame.h
DmitryTsyganov: I chose `enum` over `bool` for 2 reasons:

1. There can potentially be more strategies, for example, cheap frame scope arena allocator managing descriptor sets preallocated from Vulkan (basically don't call `vkResetDescriptorPool` and `vkAllocateDescriptorSets` every frame). 
2. For consistency with `BufferAllocationStrategy` used in the same sample.

For this submission, I can change it to whatever you think is the best.

framework/core/command_buffer.cpp
asuessenbach: Now that this functionality has been moved, it reveals a potential other issue: those `bindings_to_update` are used in the `StoreInCache`-case only, but determined in `CommandBuffer::flush_descriptor_state` independently of that state.
Can that vector be determined from the data passed into `RenderFrame::request_descriptor_set`, if required only?

DmitryTsyganov: Moved collecting `bindings_to_update` to `RenderFrame::request_descriptor_set`. Unfortunately, I had to add a new function `collect_bindings_to_update` because  `bindings_to_update` have to be aggregated from `image_infos` and `buffer_infos`.
The complexity of `collect_bindings_to_update` doesn't affect performance, in `constant_data` sample there are only 3 `binding_index`s to check overall, and this number would not likely ever change significantly.

asuessenbach: Looks good now.
Just one question: Previously, each `binding_index` (given some conditions) from the `resource_set` was added to `bindings_to_update`. Now each `binding_index` (given the same conditions) from the `buffer_infos` and `image_infos` maps are used. Obviously, in this sample that is the same, but does that hold in general? Could that somehow be documented in code (by some assertions or such)?

DmitryTsyganov: Yes, it can. It's easy to check if a `binding_index` from the `resource_set` ended up as a key in either `buffer_infos` or `image_infos`. If it did, it would be evaluated in `collect_bindings_to_update`.

As before, I tested the code by running the `constant_data` sample (the only sample that sets `update_after_bind` to `true`) and the assertion doesn't fail.

In addition, I removed `std::move`, since both `VkDescriptorImageInfo` and `VkDescriptorBufferInfo` are trivial structs that can't benefit from move semantics.

asuessenbach: Just one very minor thing that caught my attention now: in
`else if (image_view != nullptr || sampler != VK_NULL_HANDLE)`
you should compare sampler with nullptr as well, as it's just a pointer and not a handle.

DmitryTsyganov: fixed

asuessenbach (approved):

gary-sweet (approved):

gpx1000 (approved):

PR #530, opened 2022-10-10
Add new sample: Vertex Input Dynamic State

Mateusz-Kowalewski-Mobica:
## Description

This PR adds a sample for presenting capabilities of VK_EXT_vertex_input_dynamic_state extension.
The sample is using one pipeline to draw two models that have different vertex input data layouts. One model was loaded by framework function "load_model", but second was generated directly in code.
Framework vk_initializers.h was updated by adding 2 new inline initialization code (for new structures from extension)

sample

Sample was tested on next 2 configurations:
Linux Ubuntu 20.04 
Nvidia Quadro P600
Intel i7-8750H
32 GB RAM
and
Windows 10
Nvidia Quadro P620
Intel i7-11800H
32 GB RAM

Note: Sample wasn't verified on MacOS and Android yet.

## General Checklist:

## Sample Checklist


TomAtkinsonArm:
Hey @Mateusz-Kowalewski-Mobica, I've started these pipelines for you. Please both contributors sign the CLA. Then we can move onto code reviews :)

gary-sweet (commented):
Compiles ok for me, and correctly reports that vertex input dynamic state is not supported, but I can't approve it since I can't run it.

Mateusz-Kowalewski-Mobica:
Hey guys,

I want to ask why some checks were cancelled? Like building projects under Windows or Ubuntu were just interrupted before any compile error appear.

SaschaWillems:
That's a good question. I can't see why they were cancelled. I'll try to rerun the onces.

SaschaWillems:
It looks like the checks are automatically cancelled if one of the other checks fails. Maybe @TomAtkinsonArm can help with this. 

If you need a hand with the failing checks, feel free to drop me a line and I'll try to help.

Mateusz-Kowalewski-Mobica:
As I look up on fail timestamps its look like all checks were cancelled after second failed test. So probably this is some kind of mechanism to not do every test when other fails.

@SaschaWillems Thanks for proposing help, for now I see all failures are related to the same root cause in code. I will fix it today, but when I will face a wall I of course will ask you for a help hand :)

Mateusz-Kowalewski-Mobica:
Sorry for little twists in commits history, but my colleague (Borys Szefler) was pushed one commit without proper github email link in past. Because of that we have trouble with signing CLA for him. Now all works fine.

If someone can, please re-run checks for the latest updated code.

samples/README.md
```diff
@@ -255,6 +255,10 @@ Uses the graphics pipeline library extensions to improve run-time pipeline creat
 **Extension**: [```VK_EXT_conditional_rendering```](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_EXT_conditional_rendering.html) <br/>
 Demonstrate how to do conditional rendering, dynamically discarding rendering commands without having to update command buffers. This is done by sourcing conditional rendering blocks from a dedicated buffer that can be updated without having to touch command buffers.
 
+### [Vertex input dynamic state](./extensions/vertex_dynamic_state)
+**Extension**: [```VK_EXT_vertex_input_dynamic_state```](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_vertex_input_dynamic_state.html) <br/>
+Demonstrate how to use vertex input bindings and attribute descriptions dynamically, which can reduce numbers of pipelines objects that are need to create.
```
SaschaWillems: I'd change spelling here slightly to: ", which can reduce the number of pipeline objects."

samples/extensions/vertex_dynamic_state/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+#add_definitions(-DDEBUG_SAMPLE)
+
+
+
+
+add_sample(
+    ID ${FOLDER_NAME}
+    CATEGORY ${CATEGORY_NAME}
+    AUTHOR "Mobica"
+    NAME "vertex_dynamic_state"
```
SaschaWillems: This name is displayed on the mobile launcher, so it should be the human readable name "Vertex Dynamic State" instead.

samples/extensions/vertex_dynamic_state/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+    draw_model(model1, pipeline1);
+    draw_model(model2, pipeline2);
+  };
+```
+
+However, with dynamic vertex input state multiple pipelines can be reduce in amount because of possibility to change vertex
+bindings and attribute descriptions on runtime by calling `vkCmdSetVertexInputEXT` before `draw_model`.
+Extension extends 2 structures: 
+"VkVertexInputBindingDescription" to "VkVertexInputBindingDescription2EXT"
```
SaschaWillems: I think those should be single quotes like above, so the the functions are properly highlighted by markdown.

SaschaWillems:
Did a review and the sample looks great. I added a few comments regarding the documentation, but only very minor things.

I also tested on two platforms:
- On Windows 10 with an NVIDIA RTX 2060 it works as expected.
- On Android 13 with a Google Pixel 4a it starts and closes without any errors

The Pixel 4a doesn't support the extension, so it's expected that the sample won't work there. But oddly instead of showing a message about lacking support for the extension, it just crashes for me. I guess this is not the sample's fault, but rather a bug in the framework.

Mateusz-Kowalewski-Mobica:
Thanks you @SaschaWillems for review. I implemented documentation fixes as you were suggested to me.

samples/README.md
```diff
@@ -255,6 +255,10 @@ Uses the graphics pipeline library extensions to improve run-time pipeline creat
 **Extension**: [```VK_EXT_conditional_rendering```](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_EXT_conditional_rendering.html) <br/>
 Demonstrate how to do conditional rendering, dynamically discarding rendering commands without having to update command buffers. This is done by sourcing conditional rendering blocks from a dedicated buffer that can be updated without having to touch command buffers.
 
+### [Vertex input dynamic state](./extensions/vertex_dynamic_state)
+**Extension**: [```VK_EXT_vertex_input_dynamic_state```](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_vertex_input_dynamic_state.html) <br/>
+Demonstrate how to use vertex input bindings and attribute descriptions dynamically, which can reduce the number of pipeline objects. that are need to create.
```
asuessenbach: Punctuation is a bit unexpected.

samples/extensions/vertex_dynamic_state/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+ 
+get_filename_component(FOLDER_NAME ${CMAKE_CURRENT_LIST_DIR} NAME)
+get_filename_component(PARENT_DIR ${CMAKE_CURRENT_LIST_DIR} PATH)
+get_filename_component(CATEGORY_NAME ${PARENT_DIR} NAME)
+add_definitions(-DVK_ENABLE_BETA_EXTENSIONS)
```
asuessenbach: Why are the BETA_EXTENSIONS needed here?

samples/extensions/vertex_dynamic_state/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+# Vertex input dynamic state
+
+![Sample](./images/sample.png)
+
+## Overview
+
+This sample demonstrates how to use the `VK_EXT_vertex_input_dynamic_state` extension, which eliminates the need to create multiple pipelines in case of different binding and/or attribute descriptions.
+
+This extension changes how attributes and bindings are managed. Instead of static description during pipeline creation, this extension allows developer to change attributes and bindings using a function before every draw.
```
asuessenbach: developer -> developers
or
developer -> the developer

samples/extensions/vertex_dynamic_state/README.md
```diff
@@ -0,0 +1,209 @@
...
+      vkb::initializers::vertex_input_binding_description(
+          0, sizeof(Vertex), VK_VERTEX_INPUT_RATE_VERTEX),
+  };
+
+  // Attribute descriptions
+  std::vector<VkVertexInputAttributeDescription> vertex_input_attributes_2 = {
+    vkb::initializers::vertex_input_attribute_description(
+        0, 0, VK_FORMAT_R32G32B32_SFLOAT, 0),                  // Position
+    vkb::initializers::vertex_input_attribute_description(
+        0, 1, VK_FORMAT_R32G32B32_SFLOAT, sizeof(float) * 6),  // Normal
```
asuessenbach: Maybe a hint here, that the normal is at a different offset than in vertex_input_attributes_1.

samples/extensions/vertex_dynamic_state/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+    vkCreateGraphicsPipelines(get_device().get_handle(), VK_NULL_HANDLE, 1,
+                              &graphics_create_info_2, VK_NULL_HANDLE, &pipeline2); 
+
+    draw_model(model1, pipeline1);
+    draw_model(model2, pipeline2);
+  };
+```
+
+However, with dynamic vertex input state multiple pipelines can be reduce in amount because of possibility to change vertex
```
asuessenbach: multiple pipelines can be reduce in amount -> the number of pipelines can be reduced ?

samples/extensions/vertex_dynamic_state/README.md
```diff
@@ -0,0 +1,209 @@
...
+bindings and attribute descriptions on runtime by calling `vkCmdSetVertexInputEXT` before `draw_model`.
+Extension extends 2 structures:<br>
```
asuessenbach: Extension -> This extension ?

samples/extensions/vertex_dynamic_state/README.md
```diff
@@ -0,0 +1,209 @@
...
+    draw_model(model2, pipeline2);
+  };
+```
+
+However, with dynamic vertex input state multiple pipelines can be reduce in amount because of possibility to change vertex
+bindings and attribute descriptions on runtime by calling `vkCmdSetVertexInputEXT` before `draw_model`.
+Extension extends 2 structures:<br> 
+`VkVertexInputBindingDescription` to `VkVertexInputBindingDescription2EXT`<br>
+`VkVertexInputAttributeDescription` to `VkVertexInputAttributeDescription2EXT`<br>
+For example of 2 sets of different vertex input data architectures "change_vertex_input_data" function was presented below. First model is loaded from assets (load_model function from framework), second model was created directly in code (have different vertex input data architecture).
```
asuessenbach: For example -> An example
was presented -> is presented

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -0,0 +1,634 @@
...
+
+#if VK_NO_PROTOTYPES
+	VkInstance instance = get_device().get_gpu().get_instance().get_handle();
+	assert(!!instance);
+	vkCmdSetVertexInputEXT = (PFN_vkCmdSetVertexInputEXT) vkGetInstanceProcAddr(instance, "vkCmdSetVertexInputEXT");
+	if (!vkCmdSetVertexInputEXT)
+	{
+		throw std::runtime_error("Unable to dynamically load vkCmdSetVertexInputEXT");
+	}
+#endif
```
asuessenbach: Why do you need to get that function pointer? Doesn't volk resolve that function?

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -0,0 +1,634 @@
...
+ * 	@brief Creating command buffers and drawing particular elements on window.
+ * 	@details Drawing object list:
+ * 			 - Skybox - cube that have background texture attached (easy way to generate background to scene).
+ * 			 - Object - cube that was placed in the middle with some reflection shader effect.
+ * 			 - Created model - cube that was created on runtime.
+ * 			 - UI - some statistic tab
+ * 
+ * 	@note In case of Vertex Input Dynamic State feature sample need to create model in runtime because of requirement to have different data structure.
+ * 		  By default function "load_model" from framework is parsing data from .gltf files and build it every time in declared structure (see Vertex structure in framework files).
+ * 		  Before drawing different models (in case of vertex input data structure) "change_vertex_input_data" fuction is call for dynamically change Vertex Input data.
```
asuessenbach: is call -> is called

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -0,0 +1,634 @@
...
+	clear_values[1].depthStencil = {0.0f, 0};
+
+	int i = -1;
+	for (auto &draw_cmd_buffer : draw_cmd_buffers)
+	{
+		i++;
+		auto command_begin = vkb::initializers::command_buffer_begin_info();
+		VK_CHECK(vkBeginCommandBuffer(draw_cmd_buffer, &command_begin));
+
+		auto draw_scene = [&] {
```
asuessenbach: `draw_scene` is used just once. That is, there's no need to make it a lambda function, you can just have this sequence of calls were `draw_scene` is called now.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -0,0 +1,634 @@
...
+	{
+		throw std::runtime_error("Unable to dynamically load vkCmdSetVertexInputEXT");
+	}
+#endif
+
+	vertex_input_features.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VERTEX_INPUT_DYNAMIC_STATE_FEATURES_EXT;
+	VkPhysicalDeviceFeatures2 device_features{};
+	device_features.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
+	device_features.pNext = &vertex_input_features;
+	vkGetPhysicalDeviceFeatures2(get_device().get_gpu().get_handle(), &device_features);
```
asuessenbach: You should check here that `vertex_input_features.vertexInputDynamicState` is `true`.
Moreover, `vertex_input_feature` does not need to be a member variable, as it is only used locally.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -0,0 +1,634 @@
...
+
+	camera.type = vkb::CameraType::LookAt;
+	camera.set_position({0.f, 0.f, -4.0f});
+	camera.set_rotation({0.f, 0.f, 0.f});
+	camera.set_perspective(60.f, (float) width / (float) height, 256.f, 0.1f);
```
asuessenbach, line 60: Maybe you could initialize the camera such that both cubes are clearly visible on startup?

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -0,0 +1,634 @@
...
+
+		VkImageSubresourceRange depth_range{range};
+		depth_range.aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT;
+
+		VkRenderPassBeginInfo render_pass_begin_info    = vkb::initializers::render_pass_begin_info();
+		render_pass_begin_info.renderPass               = render_pass;
+		render_pass_begin_info.framebuffer              = framebuffers[i];
+		render_pass_begin_info.renderArea.extent.width  = width;
+		render_pass_begin_info.renderArea.extent.height = height;
+		render_pass_begin_info.clearValueCount          = 3;
```
asuessenbach: There are just 2 clearValues! Use clearValues.size() instead.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -0,0 +1,634 @@
...
+}
+
+/**
+ *  @fn void vertex_dynamic_state::model_data_creation()
+ *  @brief Generate vertex input data for simple cube (position and normal vectors)
+ */
+void vertex_dynamic_state::model_data_creation()
+{
+	constexpr uint32_t vertex_count = 8;
+	SampleVertex *     vertices     = new SampleVertex[vertex_count];
```
asuessenbach: Why aren't you using a simple C-Array here (`SampleVertex vertices[vertex_count];`)? Or maybe a std::vector or a std::array?

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -0,0 +1,634 @@
...
+	vkFreeMemory(get_device().get_handle(), index_staging.memory, nullptr);
+
+	delete[] vertices;
+}
+
+/**
+ * 	@fn void vertex_dynamic_state::change_vertex_input_data(vertexDynamicStateVertexStruct_t variant)
+ * 	@brief Change Vertex Input data structure dynamically in runtime.
+ */
+void vertex_dynamic_state::change_vertex_input_data(vertexDynamicStateVertexStruct_t variant)
```
asuessenbach: If you reduce this function to just set the values that change (`vertex_bindings_description_ext.stride` and `vertex_attribute_description_ext[1].offset`), it would be more obvious what happens here.
Moreover, as there are just those two values that change, you could possibly completely remove this function (and that `vertexDynamicStateVertexStruct_t`) and set those values directly in `build_command_buffers()`.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.h
```diff
@@ -0,0 +1,101 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "api_vulkan_sample.h"
+
+class vertex_dynamic_state : public ApiVulkanSample
```
asuessenbach: I think, the naming conventions here would make this class be named `VertexDynamicState`, instead of `vertex_dynamic_state`.

asuessenbach (commented):
Great example!
I just have a few questions/suggestions...

asuessenbach:
And even though, this is an extension sample (not a performance sample), you might add some words on perf differences (if there are any) in the readme.

Mateusz-Kowalewski-Mobica:
I very appreciate for your valuable feedback @asuessenbach. I will take care of resolving each suggest/typo ASAP.

samples/README.md
Mateusz-Kowalewski-Mobica: I have implemented this change.

samples/extensions/vertex_dynamic_state/CMakeLists.txt
Mateusz-Kowalewski-Mobica: Thanks for explanation of this NAME variable. I have implemented this change.

samples/extensions/vertex_dynamic_state/README.md
Mateusz-Kowalewski-Mobica: Of course I want to use other braces. Thanks for spotting that. I have implemented this change.

samples/README.md
Mateusz-Kowalewski-Mobica: Yes, you have right. I was probably created that typo when I implementing documentation changes from Sascha review.

samples/extensions/vertex_dynamic_state/CMakeLists.txt
Mateusz-Kowalewski-Mobica: This is just some junk code that I forget to delete. Thanks for spotting that

samples/extensions/vertex_dynamic_state/README.md
Mateusz-Kowalewski-Mobica: I agree, this will be valuable for users to have this hint right there. I will add proper description.

samples/extensions/vertex_dynamic_state/README.md
Mateusz-Kowalewski-Mobica: I will correct that.

samples/extensions/vertex_dynamic_state/README.md
Mateusz-Kowalewski-Mobica: Your sentence sounds better. I will correct that.

samples/extensions/vertex_dynamic_state/README.md
Mateusz-Kowalewski-Mobica: I will correct that.

samples/extensions/vertex_dynamic_state/README.md
Mateusz-Kowalewski-Mobica: I will correct that.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: To be honest I didn't know that for sure. I was inherit this part of code when I try to utilize vkCmdSetVertexInputEXT command. I will try to dig deeper and maybe I will propose some other solution.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: I will correct that. In future I will take more care about descriptions in English.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: Yes, I agree with you. I was checked and everything work fine without lambda function. In next commit I will implement that change

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: Here I probably misunderstand documentation. 

> If the VkPhysicalDeviceVertexInputDynamicStateFeaturesEXT structure is included in the pNext chain of the [VkPhysicalDeviceFeatures2](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceFeatures2.html) structure passed to [vkGetPhysicalDeviceFeatures2](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPhysicalDeviceFeatures2.html), it is filled in to indicate whether each corresponding feature is supported. VkPhysicalDeviceVertexInputDynamicStateFeaturesEXT can also be used in the pNext chain of [VkDeviceCreateInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDeviceCreateInfo.html) to selectively enable these features.

And I want to enable this feature by passing `vertex_input_feature` to `device_features.pNext`. Additionally I implement: `void vertex_dynamic_state::request_gpu_features(vkb::PhysicalDevice &gpu)` where I do:
`requested_vertex_input_features.vertexInputDynamicState = VK_TRUE;`

So as I understand it correctly I can only use `request_gpu_features` to properly enable vertexInputDynamicState. I try to comment code from 67 to 71 line and everything works normally.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: That's great idea. I hate that I didn't do it before :)

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: Yes, you have right. Using size method is much more secure in case of that kind of mistakes.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: It's because of my habits (I works before mostly in C). I will change it to something more c++

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.h
Mateusz-Kowalewski-Mobica: Of course, you have a right. I will change it.

samples/README.md
```diff
@@ -257,7 +257,7 @@ Demonstrate how to do conditional rendering, dynamically discarding rendering co
 
 ### [Vertex input dynamic state](./extensions/vertex_dynamic_state)
 **Extension**: [```VK_EXT_vertex_input_dynamic_state```](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_vertex_input_dynamic_state.html) <br/>
-Demonstrate how to use vertex input bindings and attribute descriptions dynamically, which can reduce the number of pipeline objects. that are need to create.
+Demonstrate how to use vertex input bindings and attribute descriptions dynamically, which can reduce the number of pipeline objects that are need to create.
```
asuessenbach, line 261: that are need to create -> that need to be created?

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -371,12 +335,38 @@ void vertex_dynamic_state::build_command_buffers()
 		render_pass_begin_info.framebuffer              = framebuffers[i];
 		render_pass_begin_info.renderArea.extent.width  = width;
 		render_pass_begin_info.renderArea.extent.height = height;
-		render_pass_begin_info.clearValueCount          = 3;
+		render_pass_begin_info.clearValueCount          = clear_values.size();
```
asuessenbach: You probably should `static_cast<uint32_t>(clear_values.size())`

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -474,13 +464,14 @@ void vertex_dynamic_state::draw_created_model(VkCommandBuffer commandBuffer)
...
  *  @brief Generate vertex input data for simple cube (position and normal vectors)
  */
-void vertex_dynamic_state::model_data_creation()
+void VertexDynamicState::model_data_creation()
 {
-	constexpr uint32_t vertex_count = 8;
-	SampleVertex *     vertices     = new SampleVertex[vertex_count];
+	constexpr uint32_t                     vertex_count = 8;
+	std::array<SampleVertex, vertex_count> vertices;
+	//SampleVertex *     vertices     = new SampleVertex[vertex_count];
```
asuessenbach: Remove this comment?

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -549,7 +540,7 @@ void vertex_dynamic_state::model_data_creation()
 	    VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT,
 	    vertex_buffer_size,
 	    &vertex_staging.memory,
-	    vertices);
+	    &vertices);
```
asuessenbach: When using `std::array`, you should get the pointer to the vertices via `vertices.data()`.
And likewise for the indices.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: Yes, you have a right.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: I agree with reducing elements that are inside the function. About removing whole function and hardcode these lines into `build_command_buffers` I have mixed feelings. The function name is clearly telling users what's going on before drawing models. I can implement your suggestion and for visibility, I will add more comments near hardcoded lines with a simple explanation.

asuessenbach: You're right, function name and argument are telling the right story.
But it essentially "hides" important work on two data structures behind a function call. And the uninformed reader has to realize, that the data that is used in `vkCmdSetVertexInputEXT` is modified right there. He has to jump back and forth in code to see the connection.
When the function calls would be replaced by just two lines of code each (maybe with some nice comment), the important parts here are in one place.

Mateusz-Kowalewski-Mobica:
> And even though, this is an extension sample (not a performance sample), you might add some words on perf differences (if there are any) in the readme.

Regarding performance benefits, I will try asses them after I resolve all other issues/suggestions :)

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: I catch it and I agree with you. I will change it in the next commit.

Mateusz-Kowalewski-Mobica:
@TomAtkinsonArm I want to use the Re-request review functionality, but unfortunately behavior was strange and 2 people were removed.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -350,19 +350,37 @@ void VertexDynamicState::build_command_buffers()
 		vkCmdBindDescriptorSets(draw_cmd_buffer, VK_PIPELINE_BIND_POINT_GRAPHICS, pipeline_layout, 0, 1, &descriptor_set, 0, nullptr);
 
 		/* skybox */
-		change_vertex_input_data(VERTEX_DYNAMIC_STATE_USE_FRAMEWORK_VERTEX_STRUCT);
-		vkCmdSetVertexInputEXT(draw_cmd_buffer, 1, &vertex_bindings_description_ext, 2, vertex_attribute_description_ext);
+		/* First set of vertex input dynamic data (Vertex structure) */
+		vertex_bindings_description_ext[0].stride  = sizeof(Vertex);
+		vertex_attribute_description_ext[1].offset = sizeof(float) * 3;
```
asuessenbach: Maybe even better: `offsetof(Vertex, normal);` ?
And `offsetof(SampleVertex, normal)` for the other case?

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
```diff
@@ -350,19 +350,37 @@ void VertexDynamicState::build_command_buffers()
...
+		                       vertex_bindings_description_ext.data(),
+		                       static_cast<uint32_t>(vertex_attribute_description_ext.size()),
+		                       vertex_attribute_description_ext.data());
+
 		vkCmdBindPipeline(draw_cmd_buffer, VK_PIPELINE_BIND_POINT_GRAPHICS, skybox_pipeline);
 		draw_model(skybox, draw_cmd_buffer);
 
 		/* object */
-		vkCmdSetVertexInputEXT(draw_cmd_buffer, 1, &vertex_bindings_description_ext, 2, vertex_attribute_description_ext);
+		vkCmdSetVertexInputEXT(draw_cmd_buffer,
```
asuessenbach: Is this call to `vkCmdSetVertexInputEXT` needed?

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: I tried using `offsetof` before (see commit: https://github.com/KhronosGroup/Vulkan-Samples/commit/73330043bcc11d52fc012f0dae096601ea600336).
Unfortunately, I have CI failures under android and MacOS because of using `offsetof`, and I just back to `sizeof` method.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
Mateusz-Kowalewski-Mobica: Good point. I checked it and `vkCmdSetVertexInputEXT` needs to be called when we want to change vertex input data only. Not before each `vkCmdBindPipeline`. Thanks for spotting this. I will remove unnecessary `vkCmdSetVertexInputEXT`.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
asuessenbach: That's pretty strange! What errors did you get there?
Did you `#include <cstddef>`?

Mateusz-Kowalewski-Mobica: Unfortunately, as I see logs from previous commit CI are overwritten by a new CI check.
I didn't include `<cstddef>`, but I do mini research and this is included by going: <br> `api_vulkan_sample.h` -> `gui.h` -> `platform/input_events.h` and in last header we have `#include <cstddef>`.
As far as I know, because of this include daisy-chain it is not necessary to include `<cstddef>` directly in my sample header.

asuessenbach: Indeed, you should not need to include `cstddef` then.
But `offsetof` is used in other places in this repository, obviously without any issues. So maybe you could give it another try?

Mateusz-Kowalewski-Mobica: Sure, probably I know what I have done not quite right. I will commit changes within 30 minutes (please run CI check after that) :)

Mateusz-Kowalewski-Mobica: Okay, previously I use maybe strange expressions and Android and MacOS hate it, but now it works fine.

asuessenbach: Please don't forget to update the Readme.md accordingly :)

Mateusz-Kowalewski-Mobica: Of course, I just want to focus on refactoring code in the first place. From your side, do you have plans to give me more suggestions? Or can I assume that's close to the end? :)

asuessenbach: Nope, I don't see any additional issues.
Just the two yet to be resolved on getting the function pointer, and maybe some words on performance.

Mateusz-Kowalewski-Mobica: Okay, Thanks for your work and time. I will try to resolve them.

samples/extensions/vertex_dynamic_state/vertex_dynamic_state.cpp
borys-szefler-mobica: > Why do you need to get that function pointer? Doesn't volk resolve that function?

Our understanding of that concept was, that this is needed if vulkan is linked dynamically. Since in vulkan_core.h you can see:
```cpp
#ifndef VK_NO_PROTOTYPES
VKAPI_ATTR void VKAPI_CALL vkCmdSetVertexInputEXT(
    VkCommandBuffer                             commandBuffer,
    uint32_t                                    vertexBindingDescriptionCount,
    const VkVertexInputBindingDescription2EXT*  pVertexBindingDescriptions,
    uint32_t                                    vertexAttributeDescriptionCount,
    const VkVertexInputAttributeDescription2EXT* pVertexAttributeDescriptions);
#endif
```
We based this solution on dynamic_rendering sample where `vkCmdBeginRenderingKHR` and `vkCmdEndRenderingKHR` are resolved the same way.

Without this our code did not compile. Let us know if this is correct approach.

asuessenbach: Ah, I see... the problem is the volk version used here.
If you update it to a more recent one, you don't need to explicitly get the function pointer on your own.

Mateusz-Kowalewski-Mobica: Okay, the solution is easy in the case of implementation, but updating third party library will impact all created samples in this project. This kind of change needs to be discussed with other reviewers/contributors connected to this project. For now, we can't just update the code, but on Monday as I know will be Khronos Samples meeting. We can introduce the possibility of that kind of upgrade.

Mateusz-Kowalewski-Mobica: @TomAtkinsonArm @SaschaWillems @gary-sweet Can we discuss the possibility to implement that kind of change in Khronos Samples project? I didn't know if we have some regulations to stay with current versions of third-party libraries. Any feedback regarding this topic will be helpful to me.

In nutshell: the idea is to update `volk` library to a newer version so that we don't need to add this kind of code declaration in our Samples (see code pointed by @asuessenbach).

SaschaWillems: Volk is included as a submodule. So usually if a sample is added that needs an update to Volk, the PR for that sample also updates that submodule.

asuessenbach (approved):

SaschaWillems (approved):
Looks good and works fine for me (RTX 2060, Win10) with the latest changes. I'm happy to merge.

gpx1000 (approved):

gpx1000:
Works here

PR #558, opened 2022-11-29
Reduce memory footprint during gltf loading

### LINKED ISSUE #557: Some samples require excessive memory during gltf loading
The gltf loader is loading all the image data into system memory and then attempting to upload that via staging buffers to images all at once. That requires device allocations for all the staging buffers and the destination images to exist at the same time. On small memory systems that easily blows the budget when running the descriptor_management (Bonza4x) sample for example.

I have a PR that addresses this which I will add shortly.

gary-sweet:
## Description

The GLTF loader was loading all the image data into system memory and then attempting to upload that via staging buffers to device images all at once. That required device allocations for all the staging buffers and the destination images to exist at the same time. On small memory systems that easily blew the budget when running descriptor_management (Bonza4x) for example.

This change does the uploads in batches of ~64MB of image data. The staging buffers only need to cover that 64MB at any one time in that case.

Also made a small change to allow the GPU buffer->image conversions to overlap the CPU image loading. Rather than get()ing all the image load futures up-front, it now does them in the staging loop. Once 64KB of data is ready it will start the GPU conversion whilst continuing to load the remaining image on the CPU. This might also reduce the memory footprint slightly too since the image data and staging data should be finished with sooner.

Fixes #557

## General Checklist:

### DIFF
--- a/framework/gltf_loader.cpp
+++ b/framework/gltf_loader.cpp
@@ -506,41 +506,55 @@ sg::Scene GLTFLoader::load_scene(int scene_index)
 		image_components.push_back(fut.get());
 	}
 
-	// Upload images to GPU
-	std::vector<core::Buffer> transient_buffers;
-
-	auto &command_buffer = device.request_command_buffer();
-
-	command_buffer.begin(VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT, 0);
-
-	for (size_t image_index = 0; image_index < image_count; image_index++)
+	// Upload images to GPU. We do this in batches of 64MB of data to avoid needing
+	// double the amount of memory (all the images and all the corresponding buffers).
+	// This helps keep memory footprint lower which is helpful on smaller devices.
+	size_t image_index = 0;
+	while (image_index < image_count)
 	{
-		auto &image = image_components.at(image_index);
+		std::vector<core::Buffer> transient_buffers;
 
-		core::Buffer stage_buffer{device,
-		                          image->get_data().size(),
-		                          VK_BUFFER_USAGE_TRANSFER_SRC_BIT,
-		                          VMA_MEMORY_USAGE_CPU_ONLY};
+		auto &command_buffer = device.request_command_buffer();
 
-		stage_buffer.update(image->get_data());
+		command_buffer.begin(VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT, 0);
 
-		upload_image_to_gpu(command_buffer, stage_buffer, *image);
+		size_t batch_size = 0;
 
-		transient_buffers.push_back(std::move(stage_buffer));
+		// Deal with 64MB of image data at a time to keep memory footprint low
+		while (image_index < image_count && batch_size < 64 * 1024 * 1024)
+		{
+			auto &image = image_components.at(image_index);
+
+			core::Buffer stage_buffer{device,
+			                          image->get_data().size(),
+			                          VK_BUFFER_USAGE_TRANSFER_SRC_BIT,
+			                          VMA_MEMORY_USAGE_CPU_ONLY};
+
+			batch_size += image->get_data().size();
+
+			stage_buffer.update(image->get_data());
+
+			upload_image_to_gpu(command_buffer, stage_buffer, *image);
+
+			transient_buffers.push_back(std::move(stage_buffer));
+
+			image_index++;
+		}
+
+		command_buffer.end();
+
+		auto &queue = device.get_queue_by_flags(VK_QUEUE_GRAPHICS_BIT, 0);
+
+		queue.submit(command_buffer, device.request_fence());
+
+		device.get_fence_pool().wait();
+		device.get_fence_pool().reset();
+		device.get_command_pool().reset_pool();
+		device.wait_idle();
+
+		// Remove the staging buffers for the batch we just processed
+		transient_buffers.clear();
 	}
-
-	command_buffer.end();
-
-	auto &queue = device.get_queue_by_flags(VK_QUEUE_GRAPHICS_BIT, 0);
-
-	queue.submit(command_buffer, device.request_fence());
-
-	device.get_fence_pool().wait();
-	device.get_fence_pool().reset();
-	device.get_command_pool().reset_pool();
-	device.wait_idle();
-
-	transient_buffers.clear();
 
 	scene.set_components(std::move(image_components));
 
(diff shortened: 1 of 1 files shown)


framework/gltf_loader.cpp
```diff
@@ -501,46 +501,59 @@ sg::Scene GLTFLoader::load_scene(int scene_index)
...
+		size_t batch_size = 0;
 
-	for (size_t image_index = 0; image_index < image_count; image_index++)
-	{
-		auto &image = image_components.at(image_index);
+		// Deal with 64MB of image data at a time to keep memory footprint low
+		while (image_index < image_count && batch_size < 64 * 1024 * 1024)
+		{
+			// Wait for this image to complete loading, then stage for upload
+			image_components.push_back(image_component_futures.at(image_index).get());
```
asuessenbach, line 523: Why are you using at(), instead of [] here and in the next line?

framework/gltf_loader.cpp
```diff
@@ -501,46 +501,59 @@ sg::Scene GLTFLoader::load_scene(int scene_index)
...
+			image_index++;
+		}
 
-	device.get_fence_pool().wait();
-	device.get_fence_pool().reset();
-	device.get_command_pool().reset_pool();
-	device.wait_idle();
+		command_buffer.end();
+
+		auto &queue = device.get_queue_by_flags(VK_QUEUE_GRAPHICS_BIT, 0);
```
asuessenbach, line 545: Wouldn't VK_QUEUE_TRANSFER_BIT be the better choice here?

framework/gltf_loader.cpp
```diff
@@ -501,46 +501,59 @@ sg::Scene GLTFLoader::load_scene(int scene_index)
...
+
+		queue.submit(command_buffer, device.request_fence());
 
-	transient_buffers.clear();
+		device.get_fence_pool().wait();
+		device.get_fence_pool().reset();
+		device.get_command_pool().reset_pool();
+		device.wait_idle();
```
asuessenbach, line 552: There's probably no need to wait_idle here.

framework/gltf_loader.cpp
```diff
@@ -501,46 +501,59 @@ sg::Scene GLTFLoader::load_scene(int scene_index)
...
-	device.get_command_pool().reset_pool();
-	device.wait_idle();
+		command_buffer.end();
+
+		auto &queue = device.get_queue_by_flags(VK_QUEUE_GRAPHICS_BIT, 0);
+
+		queue.submit(command_buffer, device.request_fence());
 
-	transient_buffers.clear();
+		device.get_fence_pool().wait();
```
asuessenbach, line 549: If you would gather the transient_buffers per batch, together with the fence used for that batch, you could just get rid of the buffers that are done here, instead of waiting for the submit to be done.
The waiting could then be done after the surrounding while-loop, like this:
```
	std::deque<std::pair<std::vector<core::Buffer>, VkFence>> transient_buffers;
	while (image_index < image_count)
	{
		transient_buffers.push_back({});

		auto &command_buffer = device.request_command_buffer();

		command_buffer.begin(VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT, 0);

		size_t batch_size = 0;

		// Deal with 64MB of image data at a time to keep memory footprint low
		while (image_index < image_count && batch_size < 64 * 1024 * 1024)
		{
			// Wait for this image to complete loading, then stage for upload
			image_components.push_back(image_component_futures[image_index].get());

			auto &image = image_components[image_index];

			core::Buffer stage_buffer{device,
			                          image->get_data().size(),
			                          VK_BUFFER_USAGE_TRANSFER_SRC_BIT,
			                          VMA_MEMORY_USAGE_CPU_ONLY};

			batch_size += image->get_data().size();

			stage_buffer.update(image->get_data());

			upload_image_to_gpu(command_buffer, stage_buffer, *image);

			transient_buffers.back().first.push_back(std::move(stage_buffer));

			image_index++;
		}

		command_buffer.end();

		auto &queue = device.get_queue_by_flags(VK_QUEUE_TRANSFER_BIT, 0);

		transient_buffers.back().second = device.request_fence();
		queue.submit(command_buffer, transient_buffers.back().second);

		while (!transient_buffers.empty() && (vkGetFenceStatus(device.get_handle(), transient_buffers.back().second) == VK_SUCCESS))
		{
			transient_buffers.pop_front();
		}
	}

	device.get_fence_pool().wait();
	device.get_fence_pool().reset();
	device.get_command_pool().reset_pool();

	// Remove the staging buffers for the batch we just processed
	transient_buffers.clear();

```

framework/gltf_loader.cpp
gary-sweet: Because that's what the original code did.

framework/gltf_loader.cpp
gary-sweet: Probably. I didn't write this, or change it, it just moved.

framework/gltf_loader.cpp
gary-sweet: Again, not my code. Just rearranged slightly.

framework/gltf_loader.cpp
gary-sweet: Just to reiterate, these changes were intended to be the minimum needed to actually get the sample to run on my hardware.  There a lots of things I would have done differently if I had written this code originally. My intent here was to fit as closely with the original code as possible, to minimize risk, and keep the reviewing to a minimum.

It seems to me that further changes like you suggest, which is essentially refactoring the code to be more efficient, are better done in a separate PR, since they are unrelated to fixing the problem described in the issue.

JoseEmilio-ARM: Agree with @gary-sweet , this belongs in a different PR
@asuessenbach to avoid this confusion in the future, I usually use the 'hide whitespace' option: https://github.blog/2018-05-01-ignore-white-space-in-code-review/

JoseEmilio-ARM (approved):

framework/gltf_loader.cpp
asuessenbach: While my other three comments above might be attributed with "not my code", the comment here is a proposal to make you change potentially better. And it's not about whitespace at all!
I'm sorry if I went too far suggesting such changes... feel free to ignore it.

framework/gltf_loader.cpp
asuessenbach: Opened #577.

framework/gltf_loader.cpp
asuessenbach: Added #578.

marty-johnson59:
Hi @asuessenbach, per discuss on this on the last Samples call, we're thinking we'd like to proceed with merging this, but we don't want to miss some of the related issues you mentioned in comments above.  Can you create new issues to capture these so we don't lose track once this PR is merged?  Thanks!

SaschaWillems (approved):
LGTM and works fine on my desktop system.

PR #560, opened 2022-11-29
Reduce memory usage for OneAllocationPerBuffer

### LINKED ISSUE #559: OneAllocationPerBuffer option causes massive over-allocation of device memory
OneAllocationPerBuffer is used by the decriptor_management sample. It's intended to show the difference between having one buffer with one allocation, and having multiple allocations per buffer.

The way it is written however, each buffer is a minimum of 256KB regardless of the number of allocations. This sample is making thousands of small allocations, each of which is then trying to get its own 256KB buffer block. This quickly leads to memory exhaustion on small memory systems.

A have a PR that fixes this which I'll push.

gary-sweet:
## Description

OneAllocationPerBuffer is used by the decriptor_management sample. It's intended to show the difference between having one buffer with one allocation, and having multiple allocations per buffer.

The way it was written however, each buffer was a minimum of 256KB regardless of the number of allocations. This sample was making thousands of small allocations, each of which was then trying to get its own 256KB buffer block. This quickly led to memory exhaustion on small memory systems.

This change ensures that buffer blocks are allocated only at the size requested when OneAllocationPerBuffer is active.

Fixes #559

## General Checklist:

### DIFF
--- a/framework/buffer_pool.cpp
+++ b/framework/buffer_pool.cpp
@@ -85,7 +85,7 @@ BufferPool::BufferPool(Device &device, VkDeviceSize block_size, VkBufferUsageFla
 {
 }
 
-BufferBlock &BufferPool::request_buffer_block(const VkDeviceSize minimum_size)
+BufferBlock &BufferPool::request_buffer_block(const VkDeviceSize minimum_size, bool minimal)
 {
 	// Find the first block in the range of the inactive blocks
 	// which can fit the minimum size
@@ -101,8 +101,10 @@ BufferBlock &BufferPool::request_buffer_block(const VkDeviceSize minimum_size)
 
 	LOGD("Building #{} buffer block ({})", buffer_blocks.size(), usage);
 
+	VkDeviceSize new_block_size = minimal ? minimum_size : std::max(block_size, minimum_size);
+
 	// Create a new block, store and return it
-	buffer_blocks.emplace_back(std::make_unique<BufferBlock>(device, std::max(block_size, minimum_size), usage, memory_usage));
+	buffer_blocks.emplace_back(std::make_unique<BufferBlock>(device, new_block_size, usage, memory_usage));
 
 	auto &block = buffer_blocks[active_buffer_block_count++];
 
--- a/framework/buffer_pool.h
+++ b/framework/buffer_pool.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2019-2020, Arm Limited and Contributors
+/* Copyright (c) 2019-2022, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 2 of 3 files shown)


asuessenbach (commented):
Wouldn't it be simpler to instantiate the `BufferPool`s with `blockSize` of 0 in case of `OneAllocationPerBuffer`?
Would require an additional (optional) argument to the `RenderFrame` constructor, which seems to be the only one using the `BufferPool`.
And while looking at the `RenderFrame`
- why is `RenderFrame::allocate_buffer` throwing when the size is larger than some limit?
- with some new function `BufferPool::can_allocate(uint32_t size)`, or so, the second part of `RenderFrame::allocate_buffer` could be made a bit simpler.

And... just to be sure... that `BufferPool` really is that simple? It can just allocate, steadily growing, and be reset, no free?

gary-sweet:
> Wouldn't it be simpler to instantiate the `BufferPool`s with `blockSize` of 0 in case of `OneAllocationPerBuffer`? Would require an additional (optional) argument to the `RenderFrame` constructor, which seems to be the only one using the `BufferPool`.

Not really. `OneAllocationPerBuffer` isn't known at the point that `RenderFrame` is constructed; it's set later. Even if it wasn't, adding extra arguments there and changing the behaviour of the pool doesn't sound simpler to me.

> And while looking at the `RenderFrame`
> * why is `RenderFrame::allocate_buffer` throwing when the size is larger than some limit?

No idea. It's not my original code I'm afraid. Since it's unrelated to this fix, perhaps raise a new issue about it?

> * with some new function `BufferPool::can_allocate(uint32_t size)`, or so, the second part of `RenderFrame::allocate_buffer` could be made a bit simpler.

Again, not my code. I'm just trying to actually make the sample run on my hardware with the smallest change possible.

> And... just to be sure... that `BufferPool` really is that simple? It can just allocate, steadily growing, and be reset, no free?

You know what I'm going to say now - not my code ;-) But, my reading of the code is the same as yours. It never seems to get smaller. That does seem to be rather unfortunate, especially where staging buffers are concerned. I'll maybe take a look to see if that's easy enough to do something about in a different PR.

gary-sweet:
> ... especially where staging buffers are concerned. I'll maybe take a look to see if that's easy enough to do something about in a different PR.

So, `BufferPool`s are only used inside `RenderFrame`, and as such aren't used for staging buffers anyway. They only contain per-frame data, so I think it's probably OK that they don't shrink. I don't plan to change this behaviour.

asuessenbach:
You're right, that's not your code.
But when you modify it, you kind of adopt it, a little.
That is, I think, none of the things I've mentioned are worth to be an issue on their own. But they could easily be resolved (or explicitly stay how they are, with some reason) as part of this PR. But that might just be my personal view.

gary-sweet:
> ... they could easily be resolved (or explicitly stay how they are, with some reason) as part of this PR. But that might just be my personal view.

I'm pretty firmly of the opinion, personally, that commits should be as small as possible and shouldn't be changing anything even vaguely unrelated to the main purpose. There's nothing wrong with tiny commits. Keeping the responsibilities separated makes reviewing easier, blaming easier and bisecting easier down the road.

JoseEmilio-ARM (approved):
Similar to #558 , I agree with @gary-sweet that any unrelated changes belong in a different PR

asuessenbach:
Filed #579, #580,  and #581.

SaschaWillems (approved):
LGTM and works fin on my windows desktop system.

PR #568, opened 2022-12-05
add ninja-build to the Dockerfile.

gpx1000:
add ninja to the docker build so Android can build with it.

### DIFF
--- a/.github/docker/Dockerfile
+++ b/.github/docker/Dockerfile
@@ -40,6 +40,7 @@ RUN apt-get update && apt-get install -y \
         libglu1-mesa-dev \
         xorg-dev \
         adb \
+        ninja-build \
         software-properties-common
 
 RUN pip3 --no-cache-dir install \


ggfan:
This would work and needed, thank you!

The CMake installed by sdkmanager includes its own ninja binary, in the future, after this CL is in, might be better to use that one (it is tested to work with its including CMake version for sure)? To do this, I think we need to install tools in the front:
- Download the repo
- install one sdk so we have sdkmanager available on the system
- pull out the ndk/cmake version from app/build.gradle
- use sdkmanager to install the needed ndk / cmake versions
- set up the environment variable for NDK to point to the above NDK version (not any other versions) if intending to use NDK to do other things (No need to do this for Gradle build)
- put the installed cmake into the $PATH, use it to generate the android build scripts (this is different from what CI is doing now).
- `gradlew clean assembleDebug`. (or gradlew clean build) to build android apk.

tomadamatkinson:
I don't seem to need ninja installed directly in the CI Fixes branch? I am using the one from the NDK

ggfan:
If you are using the CMake port for Android Gradle Plugin (AGP), ninja is included and CMake will use that version of ninja (cmake knows where the path is, no need to put that ninja into $PATH); BUT if you want to use the CMake downloaded from kitware, ninja needed to be available (in the $PATH).

tomadamatkinson:
No longer required due to #585

PR #575, opened 2022-12-11
Vulkan Generators

tomadamatkinson:
Based on #574 

The commit history here is longer than it needs to be as it includes some of the changes from the upcoming Vulkan Builders. This will hopefully be squashed and rebased before that PR is merged.

This PR adds the ability to generate Vulkan code fairly easily. A small framework on top of the Vulkan Headers generator allows for small fixed-purpose generators to be created with a minimal amount of code. In the future, I aim to add a Vulkan test framework using these generators.

For now, a single generator is used to create the `get_structure_type_name` helper. This helper is later used in the `pNextChain` builder which allows for an easy way to collect a pNext chain without dealing with tracking the chain itself. It will be used in the Instance and Device builder to allow samples to append extensions to the create structs dynamically.

See `codegen/vulkan/readme.md` for basic documentation on the generator itself.


codegen/vulkan/generate_structure_type_names.py
```diff
@@ -0,0 +1,147 @@
...
+        is_new_feature = is_first_feature
+        nextFeatureName = self.featureName
+
+        # ... Or the first change between feature sets
+        if self.featureName is not self.lastUsedFeature:
+            is_new_feature = True
+
+        for member in members:
+            if member.name == "sType":
+                structureName = member.elem.get("values")
```
asuessenbach: Maybe `sTypeValue` would be a better name than `structureName`?
After all, it's not the structure name, but the value of the sType member.

codegen/vulkan/generate_structure_type_names.py
```diff
@@ -0,0 +1,147 @@
...
+                        section += f'#endif\n\n'
+
+                    section += f'#ifdef {nextFeatureName}\n'
+
+                    # Crucial: do not remove
+                    self.lastUsedFeature = self.featureName
+
+                if structureName:
+                    section += 'template <>\n' + \
+                        f'VkStructureType get_structure_type_name<{name}>()\n' + \
```
asuessenbach: With `get_structure_type_name`, I would expect to get some string, or so.
Why not just name it `get_structure_type`?

codegen/vulkan/generate_structure_type_names.py
tomadamatkinson: Good point I'll make this adjustment

codegen/vulkan/generate_structure_type_names.py
tomadamatkinson: I'll make this change as soon as I can :)

tomadamatkinson:
@asuessenbach resolved your comments and also renamed the generator and header along the same lines

components/vulkan/include/components/vulkan/common/structure_type_helpers.hpp
```diff
@@ -0,0 +1,3882 @@
...
+VkStructureType get_structure_type<VkCommandBufferBeginInfo>()
+{
+	return VK_STRUCTURE_TYPE_COMMAND_BUFFER_BEGIN_INFO;
+}
+template <>
+VkStructureType get_structure_type<VkRenderPassBeginInfo>()
+{
+	return VK_STRUCTURE_TYPE_RENDER_PASS_BEGIN_INFO;
+}
+
```
asuessenbach, line 270: Where do the empty line before all the `#endif`s come from?

asuessenbach (approved):

components/vulkan/include/components/vulkan/common/structure_type_helpers.hpp
tomadamatkinson: Not sure. It's not obvious looking at the generator.

I can run clang format on the generated file to try and clean it up a bit

tomadamatkinson: Fixed, was in the generator helper when processing sections

tomadamatkinson:
Fixed phantom line issue flagged by @asuessenbach. Merging on the original approval as no functional changes were made

PR #596, opened 2023-01-10
Apply Clang Tidy FIxes

tomadamatkinson:
## Description

Applies Clang Tidy to project owned source files

Removes the graphing library (experimenting with build times)


framework/stats/hwcpipe_stats_provider.cpp
```diff
@@ -90,31 +94,37 @@ HWCPipeStatsProvider::HWCPipeStatsProvider(std::set<StatIndex> &requested_stats)
 	{
 		switch (iter->second.type)
 		{
-			case StatType::Cpu:
-			{
+			case StatType::Cpu: {
```
asuessenbach: Why has the `{` moved off of its own line?

framework/stats/hwcpipe_stats_provider.cpp
```diff
@@ -90,31 +94,37 @@ HWCPipeStatsProvider::HWCPipeStatsProvider(std::set<StatIndex> &requested_stats)
...
 				}
 				else
 				{
 					iter = stat_data.erase(iter);
 				}
 				break;
 			}
-			case StatType::Gpu:
-			{
+			case StatType::Gpu: {
```
asuessenbach: As above.

asuessenbach (commented):
Looks really good, so far.
Just one question: why did you remove the precompilation of vulkan.hpp for all the vulkan.hpp-based samples?

tomadamatkinson:
> Looks really good, so far.
> 
> Just one question: why did you remove the precompilation of vulkan.hpp for all the vulkan.hpp-based samples?

This was causing a clang old style PCH error. I don't have the exact message to hand but clang tidy was warning against it

framework/stats/hwcpipe_stats_provider.cpp
tomadamatkinson: I used clang format 15 to format the code as this is the version we now use in the CI (latest version). It's likely that we need to tweak our clang format style file. Is this a blocker for the PR or something we can fix later?

tomadamatkinson:
It looks like the build times are the same when removing the cmake PCH usage

framework/stats/hwcpipe_stats_provider.cpp
asuessenbach: You probably just need to add
`     AfterCaseLabel : true`
in the `BraceWrapping` section in the .clang_format file.
Besides that: no, I wouldn't consider that as a blocker.

asuessenbach:
> It looks like the build times are the same when removing the cmake PCH usage

That's a bit surprising, when I introduced it, it gave me a few seconds. Not too much, compared to the overall build times here, but at least a bit.
If possible, I would like to keep the precompilation, as it also shows the interested reader how to manage that with CMake.

SaschaWillems:
I second that. While build times are still slow when doing full rebuilds, the precompiled headers actually matter at least on my setup ;)

SaschaWillems:
A probably stupid question: But don't we also need to update the .clang-format file?

I just checked out this PR locally in VS 2019, and if open one of files changed in this PR and then apply the code formatting in that file, some of the changes are reverted.

tomadamatkinson:
> A probably stupid question: But don't we also need to update the .clang-format file?
> 
> 
> 
> I just checked out this PR locally in VS 2019, and if open one of files changed in this PR and then apply the code formatting in that file, some of the changes are reverted.

I updated the CI to use clang format 15 and clang tidy 15. You may need to update your local versions. Or we downgrade the CI back to a more broadly used clang version. I think we where using 11 before?

SaschaWillems:
Visual Studio includes the Clang-Toolset, so the only way to update is when Microsoft updates Visual Studio to a newer toolset. I just updated my VS2019 to the most recent version, and clang-format reports version 12. So unless we want people to update to VS2022 (which seems to have v15), we probably can't update the requireents.

tomadamatkinson:
> Visual Studio includes the Clang-Toolset, so the only way to update is when Microsoft updates Visual Studio to a newer toolset. I just updated my VS2019 to the most recent version, and clang-format reports version 12. So unless we want people to update to VS2022 (which seems to have v15), we probably can't update the requireents.


I'm not against downgrading the clang toolset to match MSVC. The default shipped on MacOS seems to be 8.0. I guess we can't make everyone happy.

One option is to use the clang toolset version that best fits us and to ask users to add a custom version for their environment. On mac you can use Homebrew to update the clang toolset to 15. With MSVC it looks like you can define a custom toolset https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170#custom_llvm_location

gpx1000:
> > Visual Studio includes the Clang-Toolset, so the only way to update is when Microsoft updates Visual Studio to a newer toolset. I just updated my VS2019 to the most recent version, and clang-format reports version 12. So unless we want people to update to VS2022 (which seems to have v15), we probably can't update the requireents.
> 
> I'm not against downgrading the clang toolset to match MSVC. The default shipped on MacOS seems to be 8.0. I guess we can't make everyone happy.
> 
> One option is to use the clang toolset version that best fits us and to ask users to add a custom version for their environment. On mac you can use Homebrew to update the clang toolset to 15. With MSVC it looks like you can define a custom toolset https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170#custom_llvm_location

I really don't like downgrading for any vender.  We should pick the right version for us and document how to get that version everywhere.

tomadamatkinson:
> I really don't like downgrading for any vender.  We should pick the right version for us and document how to get that version everywhere.

I prefer using the latest as it tends to have the most false positives fixed. Especially in clang tidy.

framework/gltf_loader.cpp
```diff
@@ -627,7 +631,9 @@ sg::Scene GLTFLoader::load_scene(int scene_index)
 				vkb::sg::Texture *tex = textures.at(gltf_value.second.TextureIndex());
 
 				if (texture_needs_srgb_colorspace(gltf_value.first))
+				{
```
ggfan, line 642: Applying the brace to the single line code can be optional?

tomadamatkinson: Warned against by clang tidy

gpx1000: I thought this was a setting in clang-tidy to accept style guides (i.e. can accept Google style guide as an example, there's obviously several others).  Don't have to use KnR for brace placement.

.github/workflows/build.yml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
   push:
-    branches: [master]
+    branches: [main]
```
gpx1000, line 7: this is obviously correct for after we've switched to main from master.  However, I just wanted to confirm that it hasn't happened yet.  Thus do we want this in a different PR?

CMakeLists.txt
```diff
@@ -27,18 +27,6 @@ list(APPEND CMAKE_MODULE_PATH
     ${CMAKE_CURRENT_SOURCE_DIR}/bldsys/cmake
     ${CMAKE_CURRENT_SOURCE_DIR}/bldsys/cmake/module)
 
-# Precompiled header macro. Parameters are source file list and filename for pch cpp file.
```
gpx1000, line 30: We should probably clean up the precompiled header generation commands anyway.  However, I believe clang-tidy will work with PCH.

gpx1000 (dismissed):
overall I'm very happy with this.  We're finally getting some of the warnings removals ;-)

SaschaWillems:
> > > Visual Studio includes the Clang-Toolset, so the only way to update is when Microsoft updates Visual Studio to a newer toolset. I just updated my VS2019 to the most recent version, and clang-format reports version 12. So unless we want people to update to VS2022 (which seems to have v15), we probably can't update the requireents.
> > 
> > 
> > I'm not against downgrading the clang toolset to match MSVC. The default shipped on MacOS seems to be 8.0. I guess we can't make everyone happy.
> > One option is to use the clang toolset version that best fits us and to ask users to add a custom version for their environment. On mac you can use Homebrew to update the clang toolset to 15. With MSVC it looks like you can define a custom toolset https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170#custom_llvm_location
> 
> I really don't like downgrading for any vender. We should pick the right version for us and document how to get that version everywhere.

I'm totally fine with that :)

As Tom mentions, it looks like you can manually update clang-tidy in VS2019, so I'll try that.

Will do a proper review once I've updated.

.github/workflows/build.yml
tomadamatkinson: Oops it's because I'm based from my test main branch. I will remove this later

framework/gltf_loader.cpp
tomadamatkinson: Is this a biggy? I prefer using braces as it makes the code branch more visible at a glance. If I see it I then get an opportunity to try and remove it by simplifying the code

framework/stats/hwcpipe_stats_provider.cpp
tomadamatkinson: Thanks! that worked

tomadamatkinson:
I'm having trouble with the PCH stuff. Experimenting with a few things and so I've moved the PR back into WIP.

Removed graphing as this tool was never 100% introduced. All in all just playing with the build.

We need some way of breaking the circular dependency between apps (samples), plugins and the framework. This is something that framework v2.0 does and is why the build times are not effected when introducing samples.

Lets see where this goes. Im open to ideas

tomadamatkinson:
@gpx1000 @SaschaWillems @asuessenbach @ggfan

Vulkan Samples Call 23/01/22 we discussed unblocking the CI by merging this without PCH support and later focus on fixing the build times.

Please approve this PR if you agree with this approach or provide suggestion of a different approach :)

gpx1000 (approved):

ggfan (approved):

asuessenbach (approved):

tomadamatkinson:
@marty-johnson59 I believe this has 3 approvals. Can this go in?

Karina-Murawko-Wisniewska-Mobica:
@tomadamatkinson @SaschaWillems The README file was not updated. https://github.com/KhronosGroup/Vulkan-Samples/blob/main/docs/build.md#clang-format-and-visual-studio
Does the PR means devs should switch to latest clang-format or to some specific version? 
Formatting is failing with clang-format-8 now (due to usage of AfterCaseLabel which is valid from version 9).

tomadamatkinson:
Hi @Karina-Murawko-Wisniewska-Mobica,

Here is the link to the docker file repository for clang tools https://github.com/KhronosGroupActions/clang-tools

In Alpine 3.17 clang 15.0.0 is in use. I can update the readme later to reflect this change

PR #613, opened 2023-02-03
Add Swapchain Recreation sample

ShabbyX:
## Description

The focus of this sample is handling swapchain recreation in an ideal way.  In this change, swapchain recreation is dealt with without the use of VK_EXT_swapchain_maintenance1 extension.  A follow up will add a path that uses various features of this extension; most importantly the present fence.

This sample has been tested on Nvidia/Linux/X11, as well as SwiftShader/Linux/X11.

## General Checklist:

## Sample Checklist

### DIFF
--- a/samples/api/swapchain_recreation/swapchain_recreation.cpp
+++ b/samples/api/swapchain_recreation/swapchain_recreation.cpp
@@ -1,0 +1,965 @@
+/* Copyright (c) 2023, Google
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 the "License";
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "swapchain_recreation.h"
+
+#include "common/logging.h"
+#include "common/vk_common.h"
+#include "glsl_compiler.h"
+#include "platform/filesystem.h"
+#include "platform/platform.h"
+
+static constexpr uint32_t INVALID_IMAGE_INDEX = std::numeric_limits<uint32_t>::max();
+
+void SwapchainRecreation::get_queue()
+{
+	queue = &get_device().get_suitable_graphics_queue();
+
+	// Make sure presentation is supported on this queue.  This is practically always the case;
+	// if a platform/driver is found where this is not true, all queues supporting
+	// VK_QUEUE_GRAPHICS_BIT need to be queried and one that supports presentation picked.
+	VkBool32 supports_present = VK_FALSE;
+	vkGetPhysicalDeviceSurfaceSupportKHR(get_gpu_handle(), queue->get_family_index(), get_surface(), &supports_present);
+
+	if (!supports_present)
+	{
+		throw std::runtime_error("Default graphics queue does not support preesnt.");
+	}
+}
+
+void SwapchainRecreation::query_surface_format()
+{
+	uint32_t surface_format_count;
+	vkGetPhysicalDeviceSurfaceFormatsKHR(get_gpu_handle(), get_surface(), &surface_format_count, nullptr);
+	std::vector<VkSurfaceFormatKHR> supported_surface_formats(surface_format_count);
+	vkGetPhysicalDeviceSurfaceFormatsKHR(get_gpu_handle(), get_surface(), &surface_format_count, supported_surface_formats.data());
+
+	// We want to get an SRGB image format that matches our list of preferred format candiates
+	// We initialize to the first supported format, which will be the fallback in case none of the preferred formats is available
+	surface_format             = supported_surface_formats[0];
+	auto preferred_format_list = std::vector<VkFormat>{VK_FORMAT_R8G8B8A8_SRGB, VK_FORMAT_B8G8R8A8_SRGB, VK_FORMAT_A8B8G8R8_SRGB_PACK32};
+
+	for (auto &candidate : supported_surface_formats)
+	{
+		if (std::find(preferred_format_list.begin(), preferred_format_list.end(), candidate.format) != preferred_format_list.end())
+		{
+			surface_format = candidate;
+			break;
+		}
+	}
+}
+
+void SwapchainRecreation::query_present_modes()
+{
+	uint32_t present_mode_count = 0;
+	VK_CHECK(vkGetPhysicalDeviceSurfacePresentModesKHR(get_gpu_handle(), get_surface(), &present_mode_count, nullptr));
+
+	present_modes.resize(present_mode_count);
+	VK_CHECK(vkGetPhysicalDeviceSurfacePresentModesKHR(get_gpu_handle(), get_surface(), &present_mode_count, present_modes.data()));
+
+	adjust_desired_present_mode();
+}
+
+void SwapchainRecreation::adjust_desired_present_mode()
+{
+	// The FIFO present mode is guaranteed to be present.
+	if (desired_present_mode == VK_PRESENT_MODE_FIFO_KHR)
+	{
+		return;
+	}
+
+	// When switching to MAILBOX, fallback to IMMEDIATE if not available and back to FIFO if
+	// neither are available.
+	if (desired_present_mode == VK_PRESENT_MODE_MAILBOX_KHR && std::find(present_modes.begin(), present_modes.end(), desired_present_mode) != present_modes.end())
+	{
+		return;
+	}
+
+	desired_present_mode = VK_PRESENT_MODE_IMMEDIATE_KHR;
+	if (std::find(present_modes.begin(), present_modes.end(), desired_present_mode) == present_modes.end())
+	{
+		LOGW("Neither MAILBOX nor IMMEDIATE are supported, falling back to FIFO");
+		desired_present_mode = VK_PRESENT_MODE_FIFO_KHR;
+	}
+}
+
+void SwapchainRecreation::create_render_pass()
+{
+	VkAttachmentDescription attachment = {0};
+	attachment.format                  = surface_format.format;
+	attachment.samples                 = VK_SAMPLE_COUNT_1_BIT;
+	attachment.loadOp                  = VK_ATTACHMENT_LOAD_OP_CLEAR;
+	attachment.storeOp                 = VK_ATTACHMENT_STORE_OP_STORE;
+	attachment.stencilLoadOp           = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
+	attachment.stencilStoreOp          = VK_ATTACHMENT_STORE_OP_DONT_CARE;
+	attachment.initialLayout           = VK_IMAGE_LAYOUT_UNDEFINED;
+	attachment.finalLayout             = VK_IMAGE_LAYOUT_PRESENT_SRC_KHR;
+
+	VkAttachmentReference color_ref = {0, VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL};
+
+	VkSubpassDescription subpass = {0};
+	subpass.pipelineBindPoint    = VK_PIPELINE_BIND_POINT_GRAPHICS;
+	subpass.colorAttachmentCount = 1;
+	subpass.pColorAttachments    = &color_ref;
+
+	// Create a dependency from external such that srcStageMask matches WSI semaphore wait stage
+	// (pWaitDstStageMask)
+	VkSubpassDependency dependency = {0};
+	dependency.srcSubpass          = VK_SUBPASS_EXTERNAL;
+	dependency.dstSubpass          = 0;
+	dependency.srcStageMask        = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
+	dependency.dstStageMask        = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
+	dependency.srcAccessMask       = 0;
+	dependency.dstAccessMask       = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
+	dependency.dependencyFlags     = 0;
+
+	VkRenderPassCreateInfo rp_info = {VK_STRUCTURE_TYPE_RENDER_PASS_CREATE_INFO};
+	rp_info.attachmentCount        = 1;
+	rp_info.pAttachments           = &attachment;
+	rp_info.subpassCount           = 1;
+	rp_info.pSubpasses             = &subpass;
+	rp_info.dependencyCount        = 1;
+	rp_info.pDependencies          = &dependency;
+
+	VK_CHECK(vkCreateRenderPass(get_device_handle(), &rp_info, nullptr, &render_pass));
+}
+
+/**
+ * @brief Initializes the Vulkan swapchain.
+ */
+void SwapchainRecreation::init_swapchain()
+{
+	VkSurfaceCapabilitiesKHR surface_properties;
+	VK_CHECK(vkGetPhysicalDeviceSurfaceCapabilitiesKHR(get_gpu_handle(), get_surface(), &surface_properties));
+
+	if (surface_properties.currentExtent.width == 0xFFFFFFFF)
+	{
+		swapchain_extents = VkExtent2D{400, 300};
+	}
+	else
+	{
+		swapchain_extents = surface_properties.currentExtent;
+	}
+
+	// Do tripple-buffering when possible.  This is clamped to the min and max image count limits.
+	uint32_t desired_swapchain_images = std::max(surface_properties.minImageCount, 3u);
+	if (surface_properties.maxImageCount > 0)
+	{
+		desired_swapchain_images = std::min(desired_swapchain_images, surface_properties.maxImageCount);
+	}
+
+	// Find a supported composite type.
+	VkCompositeAlphaFlagBitsKHR composite = VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR;
+	if (surface_properties.supportedCompositeAlpha & VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR)
+	{
+		composite = VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR;
+	}
+	else if (surface_properties.supportedCompositeAlpha & VK_COMPOSITE_ALPHA_INHERIT_BIT_KHR)
+	{
+		composite = VK_COMPOSITE_ALPHA_INHERIT_BIT_KHR;
+	}
+	else if (surface_properties.supportedCompositeAlpha & VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR)
+	{
+		composite = VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR;
+	}
+	else if (surface_properties.supportedCompositeAlpha & VK_COMPOSITE_ALPHA_POST_MULTIPLIED_BIT_KHR)
+	{
+		composite = VK_COMPOSITE_ALPHA_POST_MULTIPLIED_BIT_KHR;
+	}
+
+	VkSwapchainKHR old_swapchain = swapchain;
+
+	VkSwapchainCreateInfoKHR info{VK_STRUCTURE_TYPE_SWAPCHAIN_CREATE_INFO_KHR};
+	info.surface          = get_surface();
+	info.minImageCount    = desired_swapchain_images;
+	info.imageFormat      = surface_format.format;
+	info.imageColorSpace  = surface_format.colorSpace;
+	info.imageExtent      = swapchain_extents;
+	info.imageArrayLayers = 1;
+	info.imageUsage       = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
+	info.imageSharingMode = VK_SHARING_MODE_EXCLUSIVE;
+	info.preTransform     = VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR;
+	info.compositeAlpha   = composite;
+	info.presentMode      = desired_present_mode;
+	info.clipped          = true;
+	info.oldSwapchain     = old_swapchain;
+
+	// Note: the above info sets preTransform to `VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR`.  This
+	// is **not optimal** on devices that support rotation and will lead to measurable
+	// performance loss.  It is strongly recommended that `surface_properties.currentTransform`
+	// be used instead.  However, the application is required to handle preTransform elsewhere
+	// accordingly.
+
+	VK_CHECK(vkCreateSwapchainKHR(get_device_handle(), &info, nullptr, &swapchain));
+
+	current_present_mode = desired_present_mode;
+
+	// Schedule destruction of the old swapchain resources once this frame's submission is finished.
+	submit_history[submit_history_index].swapchain_garbage.push_back(std::move(swapchain_objects));
+
+	// Schedule destruction of the old swapchain itself once its last presentation is finished.
+	if (old_swapchain != VK_NULL_HANDLE)
+	{
+		schedule_old_swapchain_for_destruction(old_swapchain);
+	}
+
+	// Get the swapchain images.
+	uint32_t image_count;
+	VK_CHECK(vkGetSwapchainImagesKHR(get_device_handle(), swapchain, &image_count, nullptr));
+
+	swapchain_objects.images.resize(image_count, VK_NULL_HANDLE);
+	swapchain_objects.views.resize(image_count, VK_NULL_HANDLE);
+	swapchain_objects.framebuffers.resize(image_count, VK_NULL_HANDLE);
+	VK_CHECK(vkGetSwapchainImagesKHR(get_device_handle(), swapchain, &image_count, swapchain_objects.images.data()));
+
+	for (uint32_t index = 0; index < image_count; ++index)
+	{
+		init_swapchain_image(index);
+	}
+}
+
+/**
+ * @brief Called to initialize resources for a swapchain image.
+ */
+void SwapchainRecreation::init_swapchain_image(uint32_t index)
+{
+	assert(swapchain_objects.views[index] == VK_NULL_HANDLE);
+
+	VkImageViewCreateInfo view_info{VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO};
+	view_info.viewType                    = VK_IMAGE_VIEW_TYPE_2D;
+	view_info.format                      = surface_format.format;
+	view_info.image                       = swapchain_objects.images[index];
+	view_info.subresourceRange.levelCount = 1;
+	view_info.subresourceRange.layerCount = 1;
+	view_info.subresourceRange.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
+	view_info.components.r                = VK_COMPONENT_SWIZZLE_R;
+	view_info.components.g                = VK_COMPONENT_SWIZZLE_G;
+	view_info.components.b                = VK_COMPONENT_SWIZZLE_B;
+	view_info.components.a                = VK_COMPONENT_SWIZZLE_A;
+
+	VK_CHECK(vkCreateImageView(get_device_handle(), &view_info, nullptr, &swapchain_objects.views[index]));
+
+	VkFramebufferCreateInfo fb_info{VK_STRUCTURE_TYPE_FRAMEBUFFER_CREATE_INFO};
+	fb_info.renderPass      = render_pass;
+	fb_info.attachmentCount = 1;
+	fb_info.pAttachments    = &swapchain_objects.views[index];
+	fb_info.width           = swapchain_extents.width;
+	fb_info.height          = swapchain_extents.height;
+	fb_info.layers          = 1;
+
+	VK_CHECK(vkCreateFramebuffer(get_device_handle(), &fb_info, nullptr, &swapchain_objects.framebuffers[index]));
+}
+
+/**
+ * @brief When a swapchain is retired, the resources associated with its images are scheduled to be
+ * cleaned up as soon as the last submission using those images is complete.  This function is
+ * called at such a moment.
+ *
+ * The swapchain itself is not destroyed until known safe.
+ */
+void SwapchainRecreation::cleanup_swapchain_objects(SwapchainObjects &garbage)
+{
+	for (VkImageView view : garbage.views)
+	{
+		vkDestroyImageView(get_device_handle(), view, nullptr);
+	}
+	for (VkFramebuffer framebuffer : garbage.framebuffers)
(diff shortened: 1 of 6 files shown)


samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,1120 @@
...
+VkResult SwapchainRecreation::acquire_next_image(uint32_t *index)
+{
+	PerFrame &frame = submit_history[submit_history_index];
+
+	// Use a fence to know when acquire is done.  Without VK_EXT_swapchain_maintenance1, this
+	// fence is used to infer when the _previous_ present to this image index has finished.
+	// There is no need for this with VK_EXT_swapchain_maintenance1.
+	VkFence acquire_fence = has_maintenance1 ? VK_NULL_HANDLE : get_fence();
+
+	VkResult result = vkAcquireNextImageKHR(get_device_handle(), swapchain, UINT64_MAX, frame.acquire_semaphore, acquire_fence, index);
```
charles-lunarg: How does this prevent frame.acquire_semaphore from being 'in-use' at the time of the call to `vkAcquireNextImageKHR`?

ShabbyX: I'm not sure I understood the question. Are you concerned that acquire_semaphore may be used elsewhere at the same time this call is being made? That can't happen due to CPU throttling. See `setup_frame`. The semaphores are being recycled (instead of newed/deleted), but they are always recycled when they could have otherwise been deleted.

charles-lunarg: The spec language I was thinking of is this:

> If semaphore is not VK_NULL_HANDLE it must not have any uncompleted signal or wait operations pending
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkAcquireNextImageKHR.html

Just means the semaphore has to be completely free, unlike vkQueueSubmit which allows things to be in use as long as it will be unsignalled when its next used.

Ahhh thats where the fence is. I was wondering where the 'wait for previous frame to complete is' and its there. 
Thanks for clarifying!

framework/rendering/hpp_render_target.h
```diff
@@ -19,6 +19,7 @@
 
 #include "core/hpp_image.h"
 #include <functional>
+#include <memory>
```
SaschaWillems: Any reason for adding this? If this is required (by the sample) you'll also have to update the copyright in this header to 2023 ;)

ShabbyX: Yes, `unique_ptr` needs it. The other samples seem to have gotten lucky with it being included through some other header.

>  you'll also have to update the copyright in this header to 2023

Done (weird, but ok)

marty-johnson59:
Hi @ShabbyX, any chance you can join the Vulkan Samples call on Monday (8am PT) to brief the team on this one?  Subscribe to Vulkan DevRel to see the meetings on Causeway.

ShabbyX:
Limited this PR to the base sample without using the VK_EXT_swapchain_maintenance1 extension.

I couldn't reproduce the clang-format diff with clang-format 8.0.0 or 8.0.1 on my Linux machine. Unfortunately the diff output from the bot has whitespace collapsed, so it's not useful for me to apply the change by hand. I worked around it (hopefully) by breaking the expression that was confusing clang-format.

SaschaWillems (dismissed):
LGTM and works fine on my system (Windows 10 and NV RTX 2060).

Thank you very much for this sample :)

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,936 @@
...
+	black.color = {{0, 0, 0, 1.0f}};
+
+	VkClearValue gray;
+	gray.color = {{0.5f, 0.5f, 0.5f, 1.0f}};
+
+	VkClearValue colorful;
+	colorful.color = {{
+	    (frame_number % 256) / 255.0f,
+	    ((frame_number + 63) % 256) / 255.0f,
+	    ((frame_number + 128) % 256) / 255.0f,
```
gpx1000: above 3 lines narrow convert from uint32_t to float.  Static_cast should help.

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,936 @@
...
+		}
+		else
+		{
+			assert(present_operation.present_semaphore != VK_NULL_HANDLE);
+
+			// Otherwise accumulate it in cleanup data.
+			cleanup.semaphores.push_back(present_operation.present_semaphore);
+
+			// Accumulate any previous swapchains that are pending destruction too.
+			for (SwapchainCleanupData &old_swapchain : present_operation.old_swapchains)
```
gpx1000: old_swapchain shadows the argument to the function.

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,936 @@
...
+		return;
+	}
+
+	const auto &key_button = static_cast<const vkb::KeyInputEvent &>(input_event);
+	if (key_button.get_action() != vkb::KeyAction::Up)
+	{
+		return;
+	}
+
+	switch (key_button.get_code())
```
gpx1000, line 937: use default for the switch statement to avoid adding warnings.

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,936 @@
...
+}
+
+void SwapchainRecreation::input_event(const vkb::InputEvent &input_event)
+{
+	if (input_event.get_source() != vkb::EventSource::Keyboard)
+	{
+		return;
+	}
+
+	const auto &key_button = static_cast<const vkb::KeyInputEvent &>(input_event);
```
gpx1000, line 931: static_cast shouldn't be used to convert to a base class.

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,936 @@
...
+
+		// Move clean up data to the next (now first) present operation, if any.  Note that
+		// there cannot be any clean up data on the rest of the present operations, because
+		// the first present already gathers every old swapchain to clean up.
+		assert(std::all_of(present_history.begin(), present_history.end(), [](const PresentOperation &op) {
+			return op.old_swapchains.empty();
+		}));
+		present_history.front().old_swapchains = std::move(present_operation.old_swapchains);
+
+		// Put the present operation at the end of the queue so it's revisited after the
```
gpx1000: ',' before so

samples/api/swapchain_recreation/swapchain_recreation.h
```diff
@@ -0,0 +1,195 @@
...
+		 * the image.  Only relevant when VK_EXT_swapchain_maintenance1 is not supported;
+		 * otherwise a fence is always associated with the present operation.
+		 */
+		uint32_t image_index = std::numeric_limits<uint32_t>::max();
+	};
+
+  public:
+	SwapchainRecreation();
+
+	virtual ~SwapchainRecreation();
```
gpx1000: use override instead of virtual.

gpx1000 (commented):
think there's some copy in the last PR and this one, duplicating comments here as I'm not sure which one is current.

gpx1000 (commented):
Oh, sadly resizing the window with this sample will cause validation warnings if smaller than the full size of the box.

ShabbyX:
> Oh, sadly resizing the window with this sample will cause validation warnings if smaller than the full size of the box.

Ah probably the clear area is outside the render area; a clamp on the clear area should do. I'll put up a PR by Monday.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: The conversion is from an integer in `[0, 256)` though, there's no actual narrowing happening.

Added `static_cast` nevertheless.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Done. Please consider adding `-Wshadow` in the future.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Done.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Nack for this change. This is how input events are handled everywhere else.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Done

samples/api/swapchain_recreation/swapchain_recreation.h
ShabbyX: Done

ShabbyX:
> > Oh, sadly resizing the window with this sample will cause validation warnings if smaller than the full size of the box.
> 
> Ah probably the clear area is outside the render area; a clamp on the clear area should do. I'll put up a PR by Monday.

There was that, which is fixed. Also fixed vkCmdClearAttachments complaining about 0-sized clears when the window size is extremely small.

gpx1000 (dismissed):

SaschaWillems (dismissed):

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,950 @@
...
+#include "platform/filesystem.h"
+#include "platform/platform.h"
+
+static constexpr uint32_t INVALID_IMAGE_INDEX = std::numeric_limits<uint32_t>::max();
+
+void SwapchainRecreation::get_queue()
+{
+	queue = &get_device().get_suitable_graphics_queue();
+
+	// Make sure presentation is supported on this queue.  This is practically always the case;
```
asuessenbach, line 32: Isn't this check for present support already done in `get_suitable_graphics_queue()`?

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,950 @@
...
+}
+
+void SwapchainRecreation::query_surface_format()
+{
+	uint32_t format_count;
+	vkGetPhysicalDeviceSurfaceFormatsKHR(get_gpu_handle(), get_surface(), &format_count, nullptr);
+	std::vector<VkSurfaceFormatKHR> formats(format_count);
+	vkGetPhysicalDeviceSurfaceFormatsKHR(get_gpu_handle(), get_surface(), &format_count, formats.data());
+
+	if (format_count == 1 && formats[0].format == VK_FORMAT_UNDEFINED)
```
asuessenbach: Maybe the format determination should be done like it's done in hello_triangle (since #622, at least).

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,950 @@
...
+			{
+				LOGI("Disabling V-Sync");
+				desired_present_mode = VK_PRESENT_MODE_MAILBOX_KHR;
+			}
+			break;
+		default:
+			break;
+	}
+
+	query_present_modes();
```
asuessenbach, line 959: Are the present modes supposed to change over time?
Wouldn't calling `adjust_desired_present_mode()` be enough, here?
(Assuming `query_present_modes()` is called once after surface creation)

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,950 @@
...
+		VK_CHECK(result);
+
+		cleanup_present_operation(present_operation);
+		present_history.pop_front();
+	}
+
+	// The present history can grow indefinitely if a present operation is done on an index
+	// that's never acquired in the future.  In that case, there's no fence associated with that
+	// present operation.  Move the offending entry to last, so the resources associated with
+	// the rest of the present operations can be duly freed.
```
asuessenbach, line 563: How would I provoke such a situation?

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,950 @@
...
+VkDevice SwapchainRecreation::get_device_handle()
+{
+	return get_device().get_handle();
+}
+
+SwapchainRecreation::SwapchainRecreation()
+{
+	LOGI("USAGE:");
+	LOGI(" - Press v to enable v-sync");
+	LOGI(" - Press n to disable v-sync");
```
asuessenbach: Maybe move this output at some place such that it's more visible to the user. Like in `SwapchainRecreation::prepare` (after calling `VulkanSample::prepare`)

samples/api/swapchain_recreation/swapchain_recreation.h
```diff
@@ -0,0 +1,195 @@
...
+#include "platform/application.h"
+#include "vulkan_sample.h"
+
+/**
+ * @brief A sample that implements best practices in handling present resources and swapchain
+ * recreation, for example due to window resizing or present mode changes.
+ */
+class SwapchainRecreation : public vkb::VulkanSample
+{
+	struct SwapchainObjects
```
asuessenbach, line 30: To have consistent naming on those structs, maybe rename them to `SwapchainData`, `PerFrameData`, `SwapchainCleanupData`, `PresentData` ?

asuessenbach (commented):
Overall, this sample looks great!
I have just a couple of minor issues/questions.
Besides that, we should think about integrating your way of swapchain handling into the framework!

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: No? That seems to only check for `VK_QUEUE_GRAPHICS_BIT`.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Ok yeah, this was a copy-paste of that, but hello_triangle changed.

Done.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: You'd be surprised (as I was), but apparently dragging a window through monitors or something like that can affect supported present modes.

It came up during VK_EXT_swapchain_maintenance1 that there's (an existing) race condition about that (swapchain created successfully with one present mode, then fails present right after because present modes changed)

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Sounds good, done.

samples/api/swapchain_recreation/swapchain_recreation.h
ShabbyX: At that point `Data` is just redundant; all structs are data!

If anything, it's probably more meaningful to `s/SwapchainCleanupData/SwapchainCleanupObjects`

ShabbyX:
> we should think about integrating your way of swapchain handling into the framework!

On a previous call an issue with Nvidia came up that was written off as a driver bug, but I mentioned it's likely due to mishandling of swapchain present semaphores (happened previously to us in ANGLE with Nvidia). Someone signed up to do exactly that (integrate stuff from this sample into the framework), but I don't recall who.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Simple :)

Step 1: Write a driver that does this
Step 2: Run this sample

But seriously, it's out of our control. If you really want to provoke this, you could probably modify the swapchain code in mesa to create 1 additional image, then only return that image on like the 5th acquire and never again.

As to why I'm guarding against this, it's because this might actually happen in practice. Like if you create a MAILBOX swapchain with 3 images, then have a burst of frames that makes the swapchain return all images, then really slow down and cycle through images 0 and 1, then image 3 may very well never be returned again by the swapchain implementation. You could imagine the same with FIFO for that matter.

gpx1000:
> > we should think about integrating your way of swapchain handling into the framework!
> 
> On a previous call an issue with Nvidia came up that was written off as a driver bug, but I mentioned it's likely due to mishandling of swapchain present semaphores (happened previously to us in ANGLE with Nvidia). Someone signed up to do exactly that (integrate stuff from this sample into the framework), but I don't recall who.

that would be me :) I'm waiting for this PR to land before I put it into the framework.  I'm always happy to help.

samples/api/swapchain_recreation/swapchain_recreation.h
asuessenbach: > If anything, it's probably more meaningful to s/SwapchainCleanupData/SwapchainCleanupObjects

Sounds reasonable as well, yes.
What made me open up this comment is `PresentOperation`, which sounds a bit strange for a struct.

ShabbyX: Ok I hear it (it sounds like it's for doing something, rather than holding data). The struct represents a "present operation" in the history of presents, I don't have a good name for it though really; `PresentOperationData` or `PresentOperationInfo` might do, but again `Data` and `Info` feels redundant :shrug: 

I have no strong opinion here though, I can always `sed` if you feel strongly about this.

gary-sweet (commented):
I'm seeing some odd behaviour with this sample.
The reported fps just continuously gets lower and lower over time.

I assume that's not intentional? Does anyone else see that? Obviously much easier to see in non-vsync mode.

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,930 @@
...
+void SwapchainRecreation::render(uint32_t index)
+{
+	PerFrame &frame = submit_history[submit_history_index];
+
+	VkCommandBuffer             cmd;
+	VkCommandBufferAllocateInfo cmd_buf_info{VK_STRUCTURE_TYPE_COMMAND_BUFFER_ALLOCATE_INFO};
+	cmd_buf_info.commandPool        = frame.primary_command_pool;
+	cmd_buf_info.level              = VK_COMMAND_BUFFER_LEVEL_PRIMARY;
+	cmd_buf_info.commandBufferCount = 1;
+	VK_CHECK(vkAllocateCommandBuffers(get_device_handle(), &cmd_buf_info, &cmd));
```
gary-sweet: I believe the cause of the slowing down over time is because you are allocating a new command buffer every frame but never freeing it.  You reset the command pool in setup_frame() but that just puts any command buffers in the pool back into 'initial' state, it doesn't free them. The list of command buffers in the pool just keeps growing, which causes vkResetCommandPool to take longer and longer for me.

gary-sweet (changes requested):
Wondering if it would make more sense to allocate a few command buffers up-front and then recycle them.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Oops, fixed. Let me know if it's ok now

ShabbyX:
> Wondering if it would make more sense to allocate a few command buffers up-front and then recycle them.

Probably, but that's not really the focus of this sample.

samples/api/swapchain_recreation/swapchain_recreation.cpp
gary-sweet: > Oops, fixed. Let me know if it's ok now

It no longer slows over time, thanks.

gary-sweet:
> Probably, but that's not really the focus of this sample.

Sure, but you are already storing the command_buffer in the 'frame' structure. Every frame you free it and then reallocate it which just feels rather wasteful when you already have it and just need to reset it.

ShabbyX:
> > Probably, but that's not really the focus of this sample.
> 
> Sure, but you are already storing the command_buffer in the 'frame' structure. Every frame you free it and then reallocate it which just feels rather wasteful when you already have it and just need to reset it.

Ok done, hopefully mistake-free

gary-sweet (dismissed):
Thanks, looks much better now.

samples/api/swapchain_recreation/swapchain_recreation.h
asuessenbach: You're right, `Data` or `Info` would be kind of redundant. But as we are used from vulkan, structures are not named `*Create`, but `*CreateInfo`, at least for easier separation from the actual operation.
That is, ending on `Info` here would be a good fit, I think. Even though it's redundant.

ShabbyX: Done

gpx1000:
Looks like you need to rebase or at least merge main into this branch to get past the CI problem.

ShabbyX:
> Looks like you need to rebase or at least merge main into this branch to get past the CI problem.

Done, but a couple of points:

- The CI check here should really only check the files that are modified by the PR, not any other file
- The CI step should merge the PR for the sake of this check, rather than checkout, so it wouldn't matter where the PR was branched off.

gpx1000:
> > Looks like you need to rebase or at least merge main into this branch to get past the CI problem.
> 
> Done, but a couple of points:
> 
> * The CI check here should really only check the files that are modified by the PR, not any other file
> * The CI step should merge the PR for the sake of this check, rather than checkout, so it wouldn't matter where the PR was branched off.

The CI check does only check the files that are modified.  The check itself I think is now also checking CMakeLists.txt when it wasn't checking them originally.  The intention was originally to always check CMakeLists.txt so the problem is more of an effect of fixing a bug in the CI.
Completely agreed on the second point.  This is something we should open an issue for and make sure we address.

gary-sweet (dismissed):

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,929 @@
...
+	if (desired_present_mode != current_present_mode)
+	{
+		recreate_swapchain();
+	}
+
+	uint32_t index;
+	auto     res = acquire_next_image(&index);
+
+	// Handle outdated error in acquire.
+	if (res == VK_SUBOPTIMAL_KHR || res == VK_ERROR_OUT_OF_DATE_KHR)
```
cforfang, line 890: Hi @ShabbyX, I'm wondering if the logic here might be problematic on Android. Specifically, if pre-rotation is not honoured[1] the WSI calls will tend to return `VK_SUBOPTIMAL_KHR` since Android 10 IIRC. If we imagine that happening here, it seems like the below branch would be entered, `recreate_swapchain()` is called, however then it would early-out because surface dimensions are unchanged, and then `acquire_next_image()` is called on the line below.

On the face of it, that seems like we'd have `acquire_next_image()` called twice in a row without any swap-chain recreation in between -- which seems incorrect.

It looks like the code as it is now effectively assumes VK_SUBOPTIMAL_KHR is an error code which should trigger re-creation (thus also a re-acquire), but (a) the current code seems like it won't actually do re-creation in this scenario (leading to a potential bug), and (b) many apps still do not honour pre-rotation, so treating VK_SUBOPTIMAL_KHR this way in the sample might be a bad idea (would lead to unnecessary swapchain re-creation _every frame_ in such apps if they try to implement similar code as in the sample, as `VK_SUBOPTIMAL_KHR` will keep being returned given the app never actually start pre-rotating)

I should say, I say this only having glancing over the code here so I might've missed something, but wanted to check if this could be a concern. :)

[1] Note that the sample as-is _does_ seem to honour it, as you have `info.preTransform     = surface_properties.currentTransform`. A way to confirm the above issue might be to change that and see if you now get SUBOPTIMAL returned every frame.

samples/api/swapchain_recreation/swapchain_recreation.cpp
```diff
@@ -0,0 +1,929 @@
...
+void SwapchainRecreation::create_render_pass()
+{
+	VkAttachmentDescription attachment = {0};
+	attachment.format                  = surface_format.format;
+	attachment.samples                 = VK_SAMPLE_COUNT_1_BIT;
+	attachment.loadOp                  = VK_ATTACHMENT_LOAD_OP_CLEAR;
+	attachment.storeOp                 = VK_ATTACHMENT_STORE_OP_STORE;
+	attachment.stencilLoadOp           = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
+	attachment.stencilStoreOp          = VK_ATTACHMENT_STORE_OP_DONT_CARE;
+	attachment.initialLayout           = VK_IMAGE_LAYOUT_UNDEFINED;
```
cforfang, line 109: It might look like this hits my favorite obscure sync issue :) You have:
- A render-pass with transition from UNDEFINED
- An acquire semaphore wait in COLOR stage
- No custom external subpass dependencies

The problem is, the UNDEFINED transition can trigger a memory write, and when this happens is defined by the subpass dependencies. The default implicit subpass dependency doesn't 'hook' onto any stage except TOP stage (srcStageMask = TOP), and so there's nothing preventing the UNDEFINED transition from happening ASAP after the TOP stage. The problem then is, the semaphore wait doesn't happen before the COLOR stage -- after TOP -- meaning the transition can happen before the semaphore signals. So we have a race where a write to the image (to transition it) might happen before it's legal to touch the image (before the semaphore signals).

Some ways to fix:
- Wait on the acquire semaphore in TOP -- this is suboptimal sync wise, as you have e.g. vertex-shading blocked on the semaphore unnecessarily.
- Set initialLayout to VK_IMAGE_LAYOUT_PRESENT_SRC_KHR so there's no transition generated.
- Add an external subpass-dependency which waits on the COLOR stage, so the semaphore is included in the render-pass's dependencies.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Hi. You are right, the code _should_ be modified to account for rotation (like swap the dimensions after the query if 90/270 degree rotations). Probably the offset/extents of the clear command should also be rotated.

May I ask for help here though? I'm not really well setup to run the samples on Android, so I'd appreciate it if someone can follow up with the necessary changes to make it correct in the presence of rotation. I agree it's important to get that right in the sample, but it's rather beyond the scope of what I signed up to demonstrate in this sample.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Ugh right :roll_eyes: the semaphore wait stage doesn't interact well with the implicit layout transition. I think the best solution is to craft the external subpass-dependency, do you mind drafting a change and I'll incorporate it?

SaschaWillems (dismissed):

marty-johnson59:
Hi @cforfang, we discussed this briefly on the Samples call and are wondering if you recommend holding off merging this sample to resolve the 2 unresolved threads above, or if they are something we can defer for now?  Let us know.  Thanks!

samples/api/swapchain_recreation/swapchain_recreation.cpp
cforfang: This should do the trick I believe:

```
		std::array<VkSubpassDependency, 1> dependencies{};
		dependencies[0].srcSubpass      = VK_SUBPASS_EXTERNAL;
		dependencies[0].dstSubpass      = 0;
		// srcStageMask must match WSI semaphore wait stage (pWaitDstStageMask)
		dependencies[0].srcStageMask    = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
		dependencies[0].dstStageMask    = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
		dependencies[0].srcAccessMask   = 0;
		dependencies[0].dstAccessMask   = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
		dependencies[0].dependencyFlags = 0;

		rp_info.dependencyCount        = static_cast<uint32_t>(dependencies.size());
		rp_info.pDependencies          = dependencies.data();
```

cforfang:
> Hi @cforfang, we discussed this briefly on the Samples call and are wondering if you recommend holding off merging this sample to resolve the 2 unresolved threads above, or if they are something we can defer for now? Let us know. Thanks!

Hey Marty, the sync issue is quite obscure and shouldn't cause issues in practice, but it's also very easy to fix -- I finally got to posting some code for it now, so can hopefully just apply it and it's all good :)

For the SUBOPTIMAL issue -- in practice it would probably only cause issues on Android, and since Android might not support this for a bit there's probably no rush (and as mentioned it doesn't seem to affect this sample anyway; would only be an issue for somebody 'porting' this code elsewhere).

But for me the thing is that if somebody is affected it's not always easy to spot this -- things will just run slower than expected, I suspect. So it would be good if the code here is 'good' in the sense of not coming with this risk. It seems like one could just remove `VK_SUBOPTIMAL_KHR` from the two `if (res == VK_SUBOPTIMAL_KHR || res == VK_ERROR_OUT_OF_DATE_KHR)` branches here and then it's fine -- i.e. resolving this is an easy fix anyway -- and probably this would also be the 'proper' fix anyway for apps which cannot support pre-rotation in their apps? 

So if it's possibly to try this, see if it works fine on current platforms, that might be preferable overall, I think, but it's not a critical issue. :)

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: Thanks, I put that in.

samples/api/swapchain_recreation/swapchain_recreation.cpp
ShabbyX: > It looks like the code as it is now effectively assumes VK_SUBOPTIMAL_KHR is an error code which should trigger re-creation

The thing is it's different per platform. Like on non-Android if you get SUBOPTIMAL because the user dragged the window corner and changed the size, you _do_ want to recreate the swapchain.

An alternative here is to just set the preTransform to 0, and comment that "rotation is not handled, please don't do it / you need to handle it properly in your application --> see this other sample (if we have any)".

cforfang: Aah, I see! That's a bit annoying in that case... Just a comment sounds pretty good to me, either keeping the code as-is but pointing out that if you modify / don't handle pre-rotation it might be a good idea to _also_ modify the re-creation-logic as well -- or like you say stop handling pre-rotation and point out the consequences.

I think my preference would be the former because then the code is good and would work fine on Android, but it still helpfully points out this is something to keep in mind if porting it elsewhere (specifically to somewhere where pre-rotation isn't handled).

ShabbyX: I opted for using identity for preTransform and ignore SUBOPTIMAL on Android + added warnings. Dealing with rotation is very important, but I feel like it deserves its own sample where it can be thoroughly explained.

SaschaWillems (approved):
LGTM and works fine for me on Win11 with a RTX 2060

gary-sweet (approved):
Looks fine to me

damienleone (approved):
LGTM. Tested on NVIDIA Linux and verified usage of VK_EXT_swapchain_maintenance1.

marty-johnson59:
Great, thanks!  Merging with 3 approvals..

AkumaRRR:
```
Program: ...ild\windows\app\bin\debug\AMD64\vulkan_samples.exe
File: C:\Users\Alex\Desktop\Vulkan-Sample...\vulkan_...ple.cpp
Line: 568

Expression: render_context && "Render context is not valid"
```

PR #619, opened 2023-02-16
Scene Graph and Gltf Loader

tomadamatkinson:
Framework v1.0 defines components with classes that had complex lifecycles. This caused lots of bloated code and didn't allow for much flexibility on a per-sample basis due to the boilerplate required. Component storage was also managed by our codebase and didn't allow for the efficient traversal of multiple component types at the same time. This left us with lots of coupling between components through pointers and a lot of iterative loops to find different components for a given node.

Most code in v1.0 acted like a system but was held inside other managing classes. Combine managers with coupled components and we where left with a scene graph that was not as flexible as we originally wanted.

Framework v2.0 takes a more data-driven approach. Most of the v1.0 components can be transformed into structs very easily. This dramatically reduces the code needed to represent that data in our engine. Using Entt as the backing store for node data we can add any structs to the scene graph. Samples can create information for nodes specific to their samples and iterate over nodes using Entt views.

In terms of CPU culling and rendering, we can create a system that iterates over the AABB and world matrix of each mesh and marks meshes as Culled using a struct. We can then only render meshes that contain a VulkanMesh, Transform, and are not marked Culled. It looks something like this:

```cpp
// find all geometries to render
auto view = registry->view<SceneNode, VulkanMesh>(entt::exclude<Culled>);
for (auto entity : view)
{
	SceneNode  &node        = view.get<SceneNode>(entity);
	VulkanMesh &vulkan_mesh = view.get<VulkanMesh>(entity);

	auto model = node.ptr->world_matrix();
}
```

# How it works

The scene graph is entirely represented by a `Node`. Nodes can be related together to create a hierarchy. Nodes can be queried for their world matrix to find their position relative to all other nodes in the scene. All Node components are managed through an ECS.

Components are now PoD structs. Components should not contain too complex logic but may contain functions it reduces repeated behaviour. Currently, I have most of the Mesh and Transform structs defined. Animations and Cameras are in development.

A gltf loader transforms the gltf spec into a series of PoD structs and attaches this to nodes in the scene hierarchy. The gltf loader no longer needed to care about vulkan as the PoD representing the logic scene do not have any vulkan objects.

In the future, we can create systems that allocate Vulkan objects for each Mesh in the scene. These meshes can then be queried in a geometry pass. Decoupling the Vulkan from the logical representation allows us to implement more best-practice techniques in the framework such as safe Vulkan garbage collection and streaming techniques.

# TODO
[ ] more detail in the PR
[ ] complete the simple gltf test and add more scenarios or models to test against
[ ] load images
[ ] create a framework samples folder with a simple rendering sample
[ ] finish camera, animation and light loading from the scene
[ ] test case for loading a scene from the assets folder


gpx1000 (approved):
A lot to review but it looks good and compiles and works for what I can test

PR #600, opened 2023-01-20
Add basic descriptor buffer sample

SaschaWillems:
## Description

This PR adds a basic sample for the VK_EXT_descriptor_buffer extension. The sample renders multiple cubes and replaces descriptor sets with descriptor buffers introduced with this extension.

I have tested the sample on Windows 10 with an nvidia RTX 2060.

The sample also includes an extensive readme/tutorial.

**Note:** The sample does not display a UI. Mixing descriptor buffers and descriptor sets (which the UI is using) shouldn't be done (and triggers validation).

Fixes #587

## General Checklist:

Please ensure the following points are checked:

- [x] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [x] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
- [x] I have commented any added functions (in line with Doxygen)
- [x] I have commented any code that could be hard to understand
- [x] My changes do not add any new compiler warnings
- [x] My changes do not add any new validation layer errors or warnings
- [x] I have used existing framework/helper functions where possible
- [x] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [x] My changes do not add any regressions
- [x] I have tested every sample to ensure everything runs correctly
- [x] This PR describes the scope and expected impact of the changes I am making

## Sample Checklist

If your PR contains a new or modified sample, these further checks must be carried out *in addition* to the General Checklist:
- [x] I have tested the sample on at least one compliant Vulkan implementation
- [x] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [x] I have stated on what implementation the sample has been tested so that others can test on different implementations and platforms
- [x] Any dependent assets have been merged and published in downstream modules
- [x] For new samples, I have added a paragraph with a summary to the appropriate chapter in the [samples readme](./../samples/README.md)

PR #541, opened 2022-10-27
Mesh shader

gpx1000:
## Description

This is the most basic of mesh shading examples.  Another PR is incoming with the LOD sample.

## General Checklist:

Please ensure the following points are checked:

- [x] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [x] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
- [x] I have commented any added functions (in line with Doxygen)
- [x] I have commented any code that could be hard to understand
- [x] My changes do not add any new compiler warnings
- [ ] My changes do not add any new validation layer errors or warnings
- [x] I have used existing framework/helper functions where possible
- [x] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [x] My changes do not add any regressions
- [x] I have tested every sample to ensure everything runs correctly
- [x] This PR describes the scope and expected impact of the changes I am making

## Sample Checklist

If your PR contains a new or modified sample, these further checks must be carried out *in addition* to the General Checklist:
- [x] I have tested the sample on at least one compliant Vulkan implementation
- [ ] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [ ] I have stated on what implementation the sample has been tested so that others can test on different implementations and platforms
- [x] Any dependent assets have been merged and published in downstream modules
- [ ] For new samples, I have added a paragraph with a summary to the appropriate chapter in the [samples readme](./../samples/README.md)

PR #624, opened 2023-02-19
mesh shader culling

gpx1000:
## Description

Adding a mesh sample that demonstrates basic culling.  Include a tutorial to describe basic ideas and link to best practices.

## General Checklist:

## Sample Checklist


gpx1000:
This PR depends upon #541 being merged first.  It will not build correctly until changes in there are merged into here.

gpx1000:
glslang was updated with a breaking change for us.  Given that #541 has one approval, I figured I'd do the required changes here and get glslang to the latest version. Thus we need to merge #541 and then this to capture the latest glslang required updates.

This PR is ready to review on its' own now.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,221 @@
...
+- See the License for the specific language governing permissions and
+- limitations under the License.
+-
+-->
+
+# Mesh Shader Culling
+
+## Overview
+
+This sample demonstrates how to incorporate the Vulkan extension ```VK_EXT_mesh_shader_exclusive ```, and
```
SaschaWillems: Should read `VK_EXT_mesh_shader`.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Enabling mesh shading
+
+To enable the mesh shading feature, the following 
+extensions are required:
+
+1) ```VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2```
+2) ```VK_KHR_SPIRV_1_4```
+3) ```VK_EXT_MESH_SHADER```
+4) ```VK_API_VERSION_1_1```
```
SaschaWillems: Strictly spoken this is not an extension, maybe move it out of this list and simply list "Vulkan 1.1" as a base requirement.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+* ```cull_center_x``` and ```cull_center_y``` determines 
+the translation of the cull mask
+* ```cull_radius```defines the size of the cull mask. 
+* ```meshlet_density``` defines the total number of meshlets used for the sample.
+
+## Task Shader
+
+A simple task shader processes the data from the uniform 
+buffer, and then send it to its associated mesh
```
SaschaWillems: sends instead of send

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,221 @@
...
+emits a simple mesh task.
+
+In general, the Mesh pipeline refers to a new pipeline 
+which replaces everything before the fragment shader 
+with a task shading stage that can call other mesh 
+shading stages.
+
+* A task shader is used to emit mesh shader(s) (or not)
+* A mesh shader has the responsibility to generate 
+  primatives.
```
SaschaWillems: primitives instead of primatives

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,221 @@
...
+whether a meshlet may be generated or not. 
+
+In this sample, a circular visual zone is centered at the origin, with an adjustable radius, controlled by the gui.
+Whenever a meshlet moves out of the visual zone, its generation process will be skipped.
+
+```glsl
+// the actual position of each meshlet:
+vec4 position = displacement + sharedData.positionTransformation;
+float squareRadius = position.x * position.x + position.y * position.y;
+// Cull Logic: only is the meshlet center position is within the view circle defined by the cull radius,
```
SaschaWillems: Only if instead of only is

shaders/mesh_shader_culling/mesh_shader_culling.mesh
```diff
@@ -0,0 +1,115 @@
...
+        // this is the displacement with respect to the origin, displacement is used in cull logic:
+
+        vec4 displacement = vec4(deltaPosition_xy, 0.0f, 1.0f);
+
+        // the actual position of each meshlet:
+
+        vec4 position = displacement + sharedData.positionTransformation;
+        float squareRadius = position.x * position.x + position.y * position.y;
+
+        // Cull Logic: only is the meshlet center position is within the view circle defined by the cull radius,
```
SaschaWillems: Only if instead of only is

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,221 @@
...
+[Culling with NVIDIA Mesh Shaders](https://www.youtube.com/watch?v=n3cnUHYGbpw)
+
+## UI overlay
+
+![Sample](./images/mesh_shader_culling_ui_overlay.png)
+
+A simple UI overlay is created to help showcase the mesh shader cull feature. Where, the radius of the circular cull
+mask can be adjusted using the "Cull Radius" slider, which ranges from ```0.5f``` to ```2.0f```. And the meshlet density
+level can be controlled using the combo-box. Where, a selection of 4-by-4, 6-by-6, and 8-by-8 of density opinions is
+provided. Where, users may use keyboard input ```W``` ```A``` ```S``` ```D``` to move the square mesh around the scene.
```
SaschaWillems: Maybe it's because I'm not a native speaker, but those "Where," make this cumbersome to read.

samples/extensions/mesh_shader_culling/mesh_shader_culling.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+		vkCmdEndRenderPass(draw_cmd_buffers[i]);
+
+		VK_CHECK(vkEndCommandBuffer(draw_cmd_buffers[i]));
+	}
+}
+
+void MeshShaderCulling::setup_descriptor_pool()
+{
+	// I still build the pool using a vector, since practically it cannot be just one descriptor pool size.
```
SaschaWillems: This comment (starting with "I still build..." is used multiple times. It sounds like a left-over.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+The mesh pipeline can create its own vertices as is 
+done in this sample.  Or it can receive them from the 
+application the same way one would for compute shaders 
+when working with models.
+
+Thus, we disable the ```pVertexInputState``` and 
+```pInputAssemblyState```.
+
+It is expected the normal use case would be to send
```
SaschaWillems: What exactly does this mean?

SaschaWillems (changes requested):
The sample itself works fine on my system. I added a few comments though, mostly for the readme and code comments. In general the readme feels a bit cumbersome to read, at least for me as a non-native speaker.

samples/extensions/mesh_shader_culling/mesh_shader_culling.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+	// Configure application version
+	set_api_version(VK_API_VERSION_1_1);
+
+	// Adding instance extension
+	add_instance_extension(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+	// Adding device extensions
+	add_device_extension(VK_KHR_SPIRV_1_4_EXTENSION_NAME);
+	add_device_extension(VK_EXT_MESH_SHADER_EXTENSION_NAME);
+	add_device_extension(VK_KHR_SHADER_FLOAT_CONTROLS_EXTENSION_NAME);
```
SaschaWillems, line 33: Is this really required for this sample?

SaschaWillems:
The sample also need to be added to the samples listing readme.

gpx1000:
@SaschaWillems If you get a chance, I just reworked the tutorial.  Please give it a read and let me know if that is easier to understand.  All comments should be addressed from above as well.

SaschaWillems:
TVM. This reads a lot better now! 👍

SaschaWillems (dismissed):

asuessenbach:
I would like to review this, but when I hit the "Start a review" button or the "Add single comment" button in the "Files changed" tab nothing happens.
Any ideas?

gpx1000:
I'm not sure, it seems to work okay when I try to do it here.  If you have clion or Android studio you can do reviews through the GitHub plugin.

marty-johnson59:
Hi @aurolou, can you test this one out on AMD?  This is a DT sample - would be good to make sure it's OK on your HW.  Also, please look at Sascha's descriptor buffer sample again (PR 600) - hopefully the TDR issues is fixe now.  Thanks for your help!

asuessenbach:
I tried to configure a solution for this PR (Win10, VS2022, using CMake GUI), but it fails. It seems to miss pthread.h and pthreads.lib.
Any advice how to resolve that?

SaschaWillems:
@asuessenbach: Did you do a clean checkout and cmake cache delete? I've been having similar troubles with other PRs in the past cause I'm usually just switching PR branches.

asuessenbach:
@SaschaWillems Indeed, deleting the caches and updating some submodules did help. Thanks for that hint.

samples/extensions/mesh_shader_culling/mesh_shader_culling.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+MeshShaderCulling::MeshShaderCulling()
+{
+	title = "Mesh shader culling";
+
+	// Configure application version
+	set_api_version(VK_API_VERSION_1_1);
+
+	// Adding instance extension
+	add_instance_extension(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
```
asuessenbach: No need to add `VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME` with `VK_API_VERSION_1_1`

samples/extensions/mesh_shader_culling/mesh_shader_culling.cpp
```diff
@@ -0,0 +1,310 @@
...
+	// Adding device extensions
+	add_device_extension(VK_KHR_SPIRV_1_4_EXTENSION_NAME);
```
asuessenbach, line 31: using `VK_KHR_SPIRV_1_4_EXTENSION_NAME` also requires `VK_KHR_SHADER_FLOAT_CONTROLS_EXTENSION_NAME`

framework/glslang_default_resource_limits.h
asuessenbach, line 1: This file should be added to the framework/CMakeLists.txt.
And: why do you introduce it at all and don't use `GetDefaultResources()` from `third_party\glslang\glslang\Public\ResourceLimits.h` ?

samples/CMakeLists.txt
```diff
@@ -48,15 +48,15 @@ endforeach ()
...
-    "hlsl"
+        #API Samples
+        "hello_triangle"
+        "dynamic_uniform_buffers"
+        "texture_loading"
+        "hdr"
+        "instancing"
+        "compute_nbody"
+        "terrain_tessellation"
+        "hlsl"
```
asuessenbach: No need to indent those samples.

samples/CMakeLists.txt
```diff
@@ -77,30 +78,30 @@ set(ORDER_LIST
...
+        "16bit_storage_input_output"
+        "16bit_arithmetic"
+        "async_compute"
+        "multi_draw_indirect"
+        "texture_compression_comparison"
 
-    #Tooling samples
-    "profiles"
+        #Tooling samples
+        "profiles"
```
asuessenbach: Indenting again.

samples/api/hpp_hlsl_shaders/hpp_hlsl_shaders.h
```diff
@@ -1,4 +1,4 @@
-/* Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
+/* Copyright (c) 2022-2023, NVIDIA CORPORATION. All rights reserved.
```
asuessenbach: No need to just adjust the copyright, when nothing else changed in this file.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+pipeline](#creating-pipeline) 
+3) [generate a simple mesh using meshlets](#mesh-shader) 
+4) [establish a basic cull logic for the meshlets.](#mesh-shader-culling)
+
+## Enabling mesh shading
+
+To enable the mesh shading feature, the following 
+extensions are required: (NB: ```VK_API_VERSION_1_1``` is required as a base requirement)
+
+1) ```VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2```
```
asuessenbach: With `VK_API_VERSION_1_1`, you don't need `VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME` any more, as it's promoted into that version

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+2) ```VK_KHR_SPIRV_1_4```
```
asuessenbach: `VK_KHR_SPIRV_1_4_EXTENSION_NAME` also requires `VK_KHR_SHADER_FLOAT_CONTROLS_EXTENSION_NAME`

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+3) ```VK_EXT_MESH_SHADER```
```
asuessenbach: To list the extension here, you should either use their actual name (`VK_KHR_spirv_1_4`) or their named define (`VK_KHR_SPIRV_1_4_EXTENSION_NAME`)

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+// 1) define some structure if more than one variable data sharing is desired:
+struct SharedData
+{
+    vec4  positionTransformation;
+    int   N;
+    int   meshletsNumber;
+    float subDimension;
+    float cullRadius;
+};
+// 2) use the following command to "establish the connection" to its associated mesh shader:
```
asuessenbach: It's not exactly a command. It's more like a variable with a storage class specifier

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+with an (optional) task shading stage that can call other mesh 
+shading stages.
+
+* A task shader (optional) is used to emit mesh shader(s)
+* A mesh shader has the responsibility to generate 
+  primitives.
+
+More details about emitting a mesh task can be found in the attached article:
+
+[Introduction to Turing Mesh Shaders](https://developer.nvidia.com/blog/introduction-turing-mesh-shaders/)
```
asuessenbach, line 131: Maybe also mention this article: [Mesh Shading for Vulkan](https://www.khronos.org/blog/mesh-shading-for-vulkan)

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+GPU manufactures have recommended best practices for 
+their hardware in setting the work group and mesh size 
+number. Further reading can be found here:
+
+* [Meshlet Size tradeoffs](https://zeux.io/2023/01/16/meshlet-size-tradeoffs/)
+* NVIDIA - [Mesh Shaders in Turing](https://on-demand.gputechconf.com/gtc-eu/2018/pdf/e8515-mesh-shaders-in-turing.pdf)
+* AMD - [Sampler feedback ultimate in Mesh shaders](https://gpuopen.com/wp-content/uploads/slides/AMD_RDNA2_DirectX12_Ultimate_SamplerFeedbackMeshShaders.pdf)
+* [Timur's blog](https://timur.hu/blog/2022/mesh-and-task-shaders)
```
asuessenbach, line 140: We're probably allowed to link to AMD and NVIDIA pages. But what about the other two?

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+* ```cull_center_x``` and ```cull_center_y``` determines 
+the translation of the cull mask
+* ```cull_radius```defines the size of the cull mask. 
+* ```meshlet_density``` defines the total number of meshlets used for the sample.
+
+## Task Shader
+
+A task shader is an optional stage responsible for launching mesh shaders.  It has two purposes:
```
asuessenbach: As you mention, the task shader is optional. Maybe you should also mention what's done when you don't provide one.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+- Unless required by applicable law or agreed to in writing, software
+- distributed under the License is distributed on an "AS IS" BASIS,
+- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+- See the License for the specific language governing permissions and
+- limitations under the License.
+-
+-->
+
+# Mesh Shader Culling
+
```
asuessenbach, line 22: Maybe you could add an image of this sample here, to show the expected results.

samples/extensions/mesh_shader_culling/mesh_shader_culling.cpp
```diff
@@ -0,0 +1,310 @@
...
+	{
+		return;
+	}
+
+	draw();
+
+	if (camera.keys.left || camera.keys.right || camera.keys.up || camera.keys.down)
+	{
+		ubo_cull.cull_center_x = (-1.0f) * camera.position.x;
+		ubo_cull.cull_center_y = (-1.0f) * camera.position.z;
```
asuessenbach: Maybe a little comment: why is `cull_center` the negative `camera.position`?
And: why do you explicitly multiply with `(-1.0f)`, instead of just `-camera.position.x`

samples/extensions/mesh_shader_culling/mesh_shader_culling.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+#pragma once
+
+#include "api_vulkan_sample.h"
+#include "glsl_compiler.h"
+
+class MeshShaderCulling : public ApiVulkanSample
+{
+  private:
```
asuessenbach, line 28: Why is just the `density_level` private and all the other data public?
I would have expected every data to be private.

shaders/mesh_shader_culling/mesh_shader_culling.mesh
```diff
@@ -0,0 +1,115 @@
...
+// 1) the same structure must be defined (name can vary however):
+
+struct SharedData
+{
+    vec4  positionTransformation;
+    int   N;
+    int   meshletsNumber;
+    float subDimension;
+    float cullRadius;
+};
```
asuessenbach: Maybe move the `struct SharedData` into a header and include that in both, the task and the fragment shader?

shaders/mesh_shader_culling/mesh_shader_culling.mesh
```diff
@@ -0,0 +1,115 @@
...
+
+const vec2 unitVertex_0 = vec2(-0.5f,  0.5f);
+const vec2 unitVertex_1 = vec2( 0.5f,  0.5f);
+const vec2 unitVertex_2 = vec2( 0.5f, -0.5f);
+const vec2 unitVertex_3 = vec2(-0.5f, -0.5f);
+
+// 2) index data associated with the vertex data, triangle defined using right-hand rule
+
+const uvec3 unitPrimitive_0 = uvec3(0, 1, 2);
+const uvec3 unitPrimitive_1 = uvec3(2, 0, 3);
```
asuessenbach: The two triangles have different winding. Doesn't matter here, but it would be better to use `(2, 3 0)` for the second triangle.

shaders/mesh_shader_culling/mesh_shader_culling.task
```diff
@@ -0,0 +1,96 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#extension GL_EXT_mesh_shader: require
+
+layout(local_size_x = 64, local_size_y = 1, local_size_z = 1) in;
```
asuessenbach: Why are you using `local_size_x = 64`, while `vkCmdDrawMeshTasksEXT` is called with `num_workgroups_x = 1`?

shaders/mesh_shader_culling/mesh_shader_culling.task
```diff
@@ -0,0 +1,96 @@
...
+void main()
+{
+    // Example of processing the shared data:
+
+    sharedData.positionTransformation = vec4(ubo.cull_center_x, ubo.cull_center_y, 0.0f, 0.0f);
+    sharedData.cullRadius = ubo.cull_radius;
+
+    if (ubo.meshlet_density == 0)
+    {
+        // Attempt to create a 6 by 6 meshlets array
```
asuessenbach: 4 by 4 ?

shaders/mesh_shader_culling/mesh_shader_culling.task
```diff
@@ -0,0 +1,96 @@
...
+
+        sharedData.N = 6;
+        sharedData.meshletsNumber = 6 * 6;
+        sharedData.subDimension = squareDimension / 6.0f;
+    }
+    else if (ubo.meshlet_density == 2)
+    {
+        // Attempt to create a 6 by 6 meshlets array
```
asuessenbach: 8 by 8?

shaders/mesh_shader_culling/mesh_shader_culling.mesh
```diff
@@ -0,0 +1,115 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#extension GL_EXT_mesh_shader: require
+
+layout(local_size_x = 64, local_size_y = 1, local_size_z = 1) in;
```
asuessenbach: local_size_x = 1 ?

shaders/mesh_shader_culling/mesh_shader_culling.mesh
```diff
@@ -0,0 +1,115 @@
...
+layout(triangles, max_vertices = 64 * 4, max_primitives = 124 * 2) out;
```
asuessenbach: max_primitives = 64 * 2 ?

asuessenbach (commented):
Overall, this is a good example for mesh culling.
Currently, you spawn one mesh shader per task shader (with just one task shader). Maybe the value of mesh shaders would be better represented by spawning N*N mesh shaders from your task shader?

marty-johnson59:
Hi @aurolou, could you check this sample on AMD HW before we merge?  We think it's ready.   Also, if you could re-run the descriptor buffer sample ( [PR 600: ](https://github.com/KhronosGroup/Vulkan-Samples/pull/600) ), that one is hopefuly working now too.  Thanks!

shaders/mesh_shader_culling/mesh_shader_culling.mesh
```diff
@@ -0,0 +1,115 @@
...
+
+        // the actual position of each meshlet:
+
+        vec4 position = displacement + sharedData.positionTransformation;
+        float squareRadius = position.x * position.x + position.y * position.y;
+
+        // Cull Logic: only is the meshlet center position is within the view circle defined by the cull radius,
+        // then this very meshlet will be generated. Otherwise, meshlet will NOT be generated.
+
+        if (squareRadius < sharedData.cullRadius * sharedData.cullRadius)
```
cforfang: If I'm reading this right then some elements of `gl_MeshVerticesEXT[]` and `gl_PrimitiveTriangleIndicesEXT[]` are not written to depending on this culling logic, is that right? However, `SetMeshOutputsEXT()` is always the same value regardless of the culling logic?

Is the implication then that the consumer of `gl_MeshVerticesEXT` and `gl_PrimitiveTriangleIndicesEXT` may be reading either (a) some undefined values (if the culling logic caused those elements to _not_ be written to), or (b) defined-to-be-something (zero?) values? 

Basically, I'm wondering if this is well-defined (specifically, if `gl_MeshVerticesEXT` and `gl_PrimitiveTriangleIndicesEXT` have well-defined default values and if this is relied on here) -- and also if this could really be called _culling_ when it seems more like the same number of primitives are always output, it's just than when 'culled' they are set up as degenerate primitives (thus _processed_ but not actually visible)?

samples/extensions/mesh_shader_culling/mesh_shader_culling.cpp
```diff
@@ -0,0 +1,310 @@
...
+	{
+		if (drawer.slider_float("Cull Radius: ", &ubo_cull.cull_radius, 0.5f, 2.0f))
+		{
+			update_uniform_buffers();
+		}
+		if (drawer.combo_box("Meshlet Density Level: ", &density_level, {"4 x 4", "6 x 6", "8 x 8"}))
+		{
+			ubo_cull.meshlet_density = static_cast<float>(density_level);
+			update_uniform_buffers();
+		}
```
asuessenbach, line 359: Maybe you could add some pipeline statistics here (FS, TS, MS invocations, for example)

asuessenbach (commented):
The current approach with one mesh shader invocation performing a loop does not really reflect the power of task/mesh shaders. Maybe you could change it to have N*N mesh shaders.
Besides that, some pipeline statistics like number of Task Shader, Mesh Shader, and Fragment Shader invocations could be of interest.
If you like, I could send you a patch with a few modifications to showcase that.
Finally, I think it might be worth to substantially increase the fragment shader complexity, to actually get some improvement with the mesh shader approach.

samples/extensions/mesh_shader_culling/mesh_shader_culling.cpp
```diff
@@ -0,0 +1,310 @@
...
+	if (device)
+	{
+		vkDestroyPipeline(get_device().get_handle(), pipeline, nullptr);
+		vkDestroyPipelineLayout(get_device().get_handle(), pipeline_layout, nullptr);
+		vkDestroyDescriptorSetLayout(get_device().get_handle(), descriptor_set_layout, nullptr);
+	}
+}
+
+void MeshShaderCulling::request_gpu_features(vkb::PhysicalDevice &gpu)
+{
```
asuessenbach, line 56: You should check for taskShader and meshShader support here.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+defining/creating the vertex information that the 
+standard fragment pipeline utilizes.
+
+The mesh pipeline can create its own vertices as is 
+done in this sample.  Or it can receive them from the 
+application the same way one would for compute shaders 
+when working with models.
+
+Thus, we disable the ```pVertexInputState``` and 
+```pInputAssemblyState```.
```
Venemo: Suggestion:

Thus, we disable the ```pVertexInputState``` and 
```pInputAssemblyState``` by setting them to NULL.

File: samples/extensions/mesh_shader_culling/README.md (hunk 22 above)
Comment (Venemo) [reply samples/extensions/mesh_shader_culling/README.md]: Additionally, please mention that task shaders are highly recommended in order to get the best performance out of a mesh shading pipeline.

File: samples/extensions/mesh_shader_culling/README.md [hunk 37]
```diff
@@ -0,0 +1,188 @@
...
+the translation of the cull mask
+* ```cull_radius```defines the size of the cull mask. 
+* ```meshlet_density``` defines the total number of meshlets used for the sample.
+
+## Task Shader
+
+A task shader is an optional stage responsible for launching mesh shaders.  It has two purposes:
+
+* Decide how many mesh shaders to launch in the workgroup.
+* Create task payload to share with the mesh shader.
```
Venemo: I think "share" does not describe well what is happening. Suggestion:

"Create a task payload that mesh shaders will have read-only access to."

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+  * taskPayloadSharedEXT type variable in GLSL can exist at most once it gets implicitly used by EmitMeshTasksEXT 
+    and behaves like shared memory.
```
Venemo: Suggestion:

... behaves like shared memory in the task shader and like a read-only SSBO in a mesh shader.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+Once a variable is defined with taskPayloadSharedEXT in the task shader, it will be shared with the mesh shader when 
+EmitMeshTasksEXT is called.
+
+In general, the Mesh pipeline refers to a new pipeline 
+which replaces everything before the fragment shader 
+with an (optional) task shading stage that can call other mesh 
+shading stages.
+
+* A task shader (optional) is used to emit mesh shader(s)
```
Venemo: Suggestion to clarify this sentence:

A task shader (optional) is used to launch mesh shader workgroup(s)

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+* A mesh shader has the responsibility to generate 
+  primitives.
```
Venemo: and vertices

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+// Example of the data shared with its associated mesh shader:
+// 1) define some structure if more than one variable data sharing is desired:
+struct SharedData
+{
+    vec4  positionTransformation;
+    int   N;
+    int   meshletsNumber;
+    float subDimension;
+    float cullRadius;
+};
```
Venemo, line 111: It should be mentioned that all GPU vendors recommend to use as little task payload as possible, eg. by packing the data to fewer bits etc.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+number. Further reading can be found here:
+
+* [Meshlet Size tradeoffs](https://zeux.io/2023/01/16/meshlet-size-tradeoffs/)
+* NVIDIA - [Mesh Shaders in Turing](https://on-demand.gputechconf.com/gtc-eu/2018/pdf/e8515-mesh-shaders-in-turing.pdf)
+* AMD - [Sampler feedback ultimate in Mesh shaders](https://gpuopen.com/wp-content/uploads/slides/AMD_RDNA2_DirectX12_Ultimate_SamplerFeedbackMeshShaders.pdf)
+* [Timur's blog](https://timur.hu/blog/2022/mesh-and-task-shaders)
+
+## Mesh Shader
+
+A mesh shader executes on the same workgroup as the task shader that emitted it if a task shader exists. A
```
Venemo: This is not true. The mesh shader is NOT part of the same workgroup as the task shader. One task shader workgroup can launch many workgroups of mesh shaders (which can read the payload from the task workgroup that launched them).

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+* [Meshlet Size tradeoffs](https://zeux.io/2023/01/16/meshlet-size-tradeoffs/)
+* NVIDIA - [Mesh Shaders in Turing](https://on-demand.gputechconf.com/gtc-eu/2018/pdf/e8515-mesh-shaders-in-turing.pdf)
+* AMD - [Sampler feedback ultimate in Mesh shaders](https://gpuopen.com/wp-content/uploads/slides/AMD_RDNA2_DirectX12_Ultimate_SamplerFeedbackMeshShaders.pdf)
+* [Timur's blog](https://timur.hu/blog/2022/mesh-and-task-shaders)
+
+## Mesh Shader
+
+A mesh shader executes on the same workgroup as the task shader that emitted it if a task shader exists. A 
+mesh shader is responsible for generating 
+vertices and indices
```
Venemo: vertices and primitives (primitives are represented by their vertex indices)

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,188 @@
...
+outColor[k * 4 + 1] = color;
+outColor[k * 4 + 2] = color;
+outColor[k * 4 + 3] = color;
+```
+
+More details of meshlets generation can be found in the attached article:
+
+[Using Mesh Shaders for Professional Graphics](https://developer.nvidia.com/blog/using-mesh-shaders-for-professional-graphics/)
+
+## Meshlets culling
```
Venemo: What is described below is not meshlet culling but per-primitive culling.

To clarify:

* Per-meshlet culling is done in the task shader. One example of how this can be done is having a bounding box of the entire meshlet and test whether the entire meshlet would be invisible. This is highly recommended because it reduces the number of mesh shader workgroups launched by that shader, so it saves compute (and bandwidth) resources by not launching any mesh shader workgroups for meshlets that would be invisible.
* Per-primitive culling is done in the mesh shader and its purpose is to have the mesh shader emit fewer vertices and primitives. This is typicaly less beneficial and is only recommended for scenes with very high geometric complexity or on GPUs whose fixed-function culling is slower than shaders. This is no longer recommended by neither AMD nor NVidia for their latest generation GPUs.

Venemo (changes requested):

framework/glslang_default_resource_limits.h
gpx1000: technically headerfiles shouldn't need to be added to targets in CMake.  CMake is designed to find those header files as needed.  However, thanks for the suggestion to use GetDefaultResources.  I've removed the new header file and used that throughout.

samples/extensions/mesh_shader_culling/README.md
gpx1000: That's actually a good question.  I know Timur was commenting on this so @Venemo do you mind if we refernce your blog article?  I'll remove the other reference.

Venemo: I think it's fine to mention my article. Let's ask @zeux about the other one.

zeux: Feel free to link to my article.

gpx1000: Thanks guys, much appreciated.  I'll leave both of your articles in.

gpx1000:
> 

Sure,  I always accept help when it's offered :)

aurolou (dismissed):

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,198 @@
...
+-
+-->
+
+# Mesh Shader Culling
+![Mesh Shader Culling](./images/mesh_shader_culling.png)
+
+## Overview
+
+This sample demonstrates how to incorporate the Vulkan extension [```VK_EXT_mesh_shader```](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_EXT_mesh_shader), and
+introduces mesh shader culling using meshlets.
```
Venemo: The term "mesh shader culling" is vague, it should either be:

* per-meshlet culling in a task shader, or
* per-primitive culling in a mesh shader

samples/extensions/mesh_shader_culling/README.md
Venemo: I like the additional paragraph. I also recommend to modify the above sentence to: "A task shader is an optional but recommended stage...." to emphasize this even better.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+* [Meshlet Size tradeoffs](https://zeux.io/2023/01/16/meshlet-size-tradeoffs/)
+* NVIDIA - [Mesh Shaders in Turing](https://on-demand.gputechconf.com/gtc-eu/2018/pdf/e8515-mesh-shaders-in-turing.pdf)
+* AMD - [Sampler feedback ultimate in Mesh shaders](https://gpuopen.com/wp-content/uploads/slides/AMD_RDNA2_DirectX12_Ultimate_SamplerFeedbackMeshShaders.pdf)
+* [Timur's blog](https://timur.hu/blog/2022/mesh-and-task-shaders)
+
+## Mesh Shader
+
+A mesh shader does not execute on the same workgroup as the task shader that emitted it if a task shader exists.  
+Many workgroups can be  One task shader workgroup can launch many mesh shader workgroups.
```
Venemo: I think this paragraph still needs editing. Suggestion:

Task and mesh shaders are executed in workgroups similar to compute shaders. Each task shader workgroup can launch many mesh shader workgroups. Each mesh shader workgroup is responsible for generating vertices and primitives. The API allows for any logic, but a typical application it is recommended to organize this around meshlets, which are a small group of vertices and primitives. Typically, each task shader invocation processes a group of meshlets and each mesh shader workgroup processes one meshlet.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,198 @@
...
+[Using Mesh Shaders for Professional Graphics](https://developer.nvidia.com/blog/using-mesh-shaders-for-professional-graphics/)
+
+## Per-primitive culling
+
+This sample uses a simple cull functionality from the mesh 
+shader. The intention in mesh shading is to only generate geometry that is relevant to the scene.  Thus, the correct 
+method of culling is to decide via a condition if a meshlet should be generated or not.
+
+In this sample, a circular visual zone is centered at the origin, with an adjustable radius, controlled by the gui.
+When a meshlet moves out of the visual zone, its generation process will be skipped.
```
Venemo: Sorry but this is still unclear to me from this text. Does the sample implement per-primitive or per-meshlet culling? Or maybe both?

Venemo (changes requested):

samples/extensions/mesh_shader_culling/README.md
Venemo: (or both)

samples/extensions/mesh_shader_culling/README.md
asuessenbach: You need to also add the image to the PR.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -0,0 +1,198 @@
...
+3) [generate a simple mesh using meshlets](#mesh-shader) 
+4) [establish a basic cull logic for the meshlets.](#mesh-shader-culling)
+
+## Enabling mesh shading
+
+To enable the mesh shading feature, the following 
+extensions are required: (NB: ```VK_API_VERSION_1_1``` is required as a base requirement)
+
+1) ```VK_KHR_SPIRV_1_4_EXTENSION_NAME```
+2) ```VK_EXT_MESH_SHADER```
```
asuessenbach: `VK_EXT_MESH_SHAEDER` -> `VK_EXT_MESH_SHADER_EXTENSION_NAME`

samples/extensions/mesh_shader_culling/mesh_shader_culling.cpp
```diff
@@ -0,0 +1,429 @@
...
+			vkDestroyQueryPool(get_device().get_handle(), query_pool, nullptr);
+			vkDestroyBuffer(get_device().get_handle(), query_result.buffer, nullptr);
+			vkFreeMemory(get_device().get_handle(), query_result.memory, nullptr);
+		}
+	}
+}
+
+void MeshShaderCulling::request_gpu_features(vkb::PhysicalDevice &gpu)
+{
+	// Query whether the device supports buffer device addresses
```
asuessenbach: "Query" -> "Check"
"buffer device addresses" -> "task and mesh shaders"

Venemo (commented):
The readme looks good to me now.

shaders/mesh_shader_culling/mesh_shader_culling.mesh
```diff
@@ -0,0 +1,94 @@
...
+
+#extension GL_EXT_mesh_shader: require
+#extension GL_GOOGLE_include_directive: require
+
+layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
+layout(triangles, max_vertices = 4, max_primitives = 2) out;
+
+#include "mesh_shader_culling/mesh_shader_shared.h"
+
+// 2) using the following command to "establish the connection" (variable name can vary):
```
asuessenbach: `2)` ?

shaders/mesh_shader_culling/mesh_shader_culling.task
```diff
@@ -0,0 +1,84 @@
...
+layout (binding = 0) uniform UBO
+{
+    float cull_center_x;
+    float cull_center_y;
+    float cull_radius;
+    float meshlet_density;
+} ubo;
+
+
+// 2) use the following command to "establish the connection" to its associated mesh shader:
```
asuessenbach: `2)` ?

shaders/mesh_shader_culling/mesh_shader_shared.h
```diff
@@ -0,0 +1,28 @@
...
+// 1) define some structure if more than one variable data sharing is desired:
+
+struct SharedData
+{
+	vec4  positionTransformation;
+	int   N;
+	int   meshletsNumber;
+	float subDimension;
+	float cullRadius;
+};
```
asuessenbach, line 34: In order to reduce payload size: `N` and `meshletsNumber` are not used in the mesh shader, `subDimension` could be easily calculated out of a constant and `gl_WorkGroupSize`. And from `positionTransformation`, only `x` and `y` need to be carried. Thus the size of `SharedData` could be reduced from 8 * 4 byte to 3 * 4 byte.
And one could even argue, that those three values are part of an UBO, that could be passed into the mesh shader as well, instead of the payload. But in order demonstrate how a payload is managed, it should probably be kept, maybe with a comment on how to remove it.

samples/extensions/mesh_shader_culling/README.md
```diff
@@ -23,7 +23,7 @@
 ## Overview
 
 This sample demonstrates how to incorporate the Vulkan extension [```VK_EXT_mesh_shader```](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_EXT_mesh_shader), and
-introduces mesh shader culling using meshlets.
+introduces per primative culling in a mesh shader.
```
asuessenbach: primative -> primitive

shaders/mesh_shader_culling/mesh_shader_shared.h
```diff
@@ -16,13 +16,12 @@
...
-// 1) define some structure if more than one variable data sharing is desired:
+// Note that all of these could be removed.  One can pass positionTransformation, subDimension, and cullRadius from a UBO.
+// however, in the interests of demonstrating how to use shared data between task and mesh shaders.
 
 struct SharedData
 {
-	vec4  positionTransformation;
-	int   N;
-	int   meshletsNumber;
+	vec2  positionTransformation;
```
asuessenbach: When you change from vec4 to vec2, you should adjust the usage in the task and mesh shader!

asuessenbach (commented):
After adjusting the task and mesh shader to the positionTransformation being a vec2 now, I get a different image at startup: the blocks are much larger and shifted to the lower right, color varies per column only, and I have to reduce the cull radius to see any culling at all.
Seems, you have to fine-adjust the settings here.

shaders/mesh_shader_culling/mesh_shader_culling.mesh
```diff
@@ -51,7 +51,8 @@ void main()
     vec4 displacement = vec4(deltaPosition_xy, 0.0f, 1.0f);
 
     // the actual position of each meshlet:
-    vec4 position = displacement + sharedData.positionTransformation;
+    vec4 posTrans = vec4(sharedData.positionTransformation.x, sharedData.positionTransformation.y, 0.0, 0.0);
+    vec4 position = displacement + posTrans;
     float squareRadius = position.x * position.x + position.y * position.y;
```
asuessenbach: It's a bit surprising, that you extend `deltaPosition_xy` and `sharedData.positionTransform` from `vec2` to `vec4`, and then just use `x` and `y`.
Why not just
`vec2 position = deltaPosition_xy + sharedData.positionTransformation;`

shaders/mesh_shader_culling/mesh_shader_culling.mesh
```diff
@@ -68,10 +69,10 @@ void main()
...
-        gl_MeshVerticesEXT[0].gl_Position = vec4(2.0 * sharedData.subDimension * unitVertex_0, 0.0f, 1.0f) + sharedData.positionTransformation + displacement;
-        gl_MeshVerticesEXT[1].gl_Position = vec4(2.0 * sharedData.subDimension * unitVertex_1, 0.0f, 1.0f) + sharedData.positionTransformation + displacement;
-        gl_MeshVerticesEXT[2].gl_Position = vec4(2.0 * sharedData.subDimension * unitVertex_2, 0.0f, 1.0f) + sharedData.positionTransformation + displacement;
-        gl_MeshVerticesEXT[3].gl_Position = vec4(2.0 * sharedData.subDimension * unitVertex_3, 0.0f, 1.0f) + sharedData.positionTransformation + displacement;
+        gl_MeshVerticesEXT[0].gl_Position = vec4(2.0 * sharedData.subDimension * unitVertex_0, 0.0f, 1.0f) + posTrans + displacement;
+        gl_MeshVerticesEXT[1].gl_Position = vec4(2.0 * sharedData.subDimension * unitVertex_1, 0.0f, 1.0f) + posTrans + displacement;
+        gl_MeshVerticesEXT[2].gl_Position = vec4(2.0 * sharedData.subDimension * unitVertex_2, 0.0f, 1.0f) + posTrans + displacement;
+        gl_MeshVerticesEXT[3].gl_Position = vec4(2.0 * sharedData.subDimension * unitVertex_3, 0.0f, 1.0f) + posTrans + displacement;
```
asuessenbach: Here, `posTrans + displacement` is the above calculated `position`.
Moreover, due to the w-component, `displacement` is a point while `posTrans` is a vector. That is, you're adding two points and a vector which leads to, well, something strange, with a w-component of 2.0f. You should not do that.
If you would just use a `vec2 position`, you could calculate
`vec4(2.0f * sharedData.subDimension * unitVertex_N + position, 0.0f, 1.0f);`

asuessenbach (commented):
The blocks are nicely centered on startup now, but with the first keystroke, it jumps to a completely different position. Only from there on, movement looks reasonable.

samples/extensions/mesh_shader_culling/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+add_sample_with_tags(
+        ID ${FOLDER_NAME}
+        CATEGORY ${CATEGORY_NAME}
+        AUTHOR "Holochip Corporation"
+        NAME "Mesh Shader Culling"
+        DESCRIPTION "Demonstrate and showcase a sample application using mesh shader culling functionality"
+        SHADER_FILES_GLSL
+        "mesh_shader_culling/mesh_shader_culling.task"
+        "mesh_shader_culling/mesh_shader_culling.mesh"
+        "mesh_shader_culling/mesh_shader_culling.frag")
```
asuessenbach: Add `mesh_shader_shared.h` here?

shaders/mesh_shader_culling/mesh_shader_culling.mesh
gpx1000: :) Did it without thinking.  Good catch

samples/extensions/mesh_shader_culling/CMakeLists.txt
gpx1000: I went ahead and did this; however, we should have a talk in the group about best practices for header files used only in shaders being included in CMake.
CMake's normal rules are that you don't include header files in your targets, they are discovered by the build tools and CMake does the right thing from there.  However, CMake can't do the right thing in shader includes because no tooling exists that will discover and inform CMake about that header file.  We might even take it so far as to have a talk about if we want to introduce a best practice of providing the shader includes with a different extension so tooling can potentially do syntax highlighting and it will be easier to write other tools like static analysis.

SaschaWillems (dismissed):

SaschaWillems (dismissed):

asuessenbach:
@gpx1000 I've sent you another version of a task/mesh shader that might show the "power of task shaders" a bit more...
After dispatching N*N task shader workgroups with vkCmdDrawMeshTasksEXT, there are 2*2 task shader invocations per workgroup. Some culling is performed in that shader, and a variable number of mesh shader workgroups (0....4) is emitted from one invocation per task shader workgroup. There are 2*2 mesh shader invocations per workgroup, each emitting 2 triangular primitives with 4 vertices.
That is, the final drawing looks a bit different than before.
Again, feel free to use that approach to your liking.

gpx1000:
> asuessenbach

Thanks for the contribution; it is now part of the sample.

asuessenbach (dismissed):

SaschaWillems (dismissed):

PR #653, opened 2023-03-28
Fix missing ktx lib

tomadamatkinson:
## Description

After updating the build system to build and compile much quicker we also set KTX to be a shared library by default. Due to how the build system propagates libraries across folders the runtime linker can not find KTX. There are two ways to fix this.

1. Use a static library instead (easy - compiled into the executable)
2. Package the ktx library next to the executable (hard - must support all platforms)

I chose the easy option :)

## General Checklist:

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -163,7 +163,7 @@ set(KTX_INCLUDE_DIRS
     ${KTX_DIR}/other_include
 )
 
-add_library(ktx SHARED ${KTX_SOURCES})
+add_library(ktx STATIC ${KTX_SOURCES})
 
 target_compile_definitions(ktx PUBLIC LIBKTX)
 if (WIN32)


gpx1000 (approved):
Technically there's a third option.  Use the RUNTIME_LIB definition for executables as this is a windows only problem (ldconfig handles this correctly else where).  However, yes, I agree easy option and it doesn't cost THAT much to have one more static lib.

gpx1000:
Or another option is to build the ktx lib to the binary output directory of vulkan_samples.  Then Windows will find it too.

tomadamatkinson:
> Technically there's a third option. Use the RUNTIME_LIB definition for executables as this is a windows only problem (ldconfig handles this correctly else where). However, yes, I agree easy option and it doesn't cost THAT much to have one more static lib.

Forgot about that one! Pesky windows

Might be worth giving it a go at some point. But as we used a static lib before and main is currently broken for windows this might be a reasonable stop gap :)

> Or another option is to build the ktx lib to the binary output directory of vulkan_samples. Then Windows will find it too.

This is option 2

asuessenbach (approved):

SaschaWillems:
Odd. When I reviewed the other PR I didn't run into any problems with ktx. I guess because it was already there from an older build.

But nonetheless this looks good to me.

SaschaWillems (approved):

gary-sweet (approved):

marty-johnson59:
Merging w/ 3 approvals

PR #655, opened 2023-03-29
WIP: Redesign functionality of GLSLcompiler class

Szymon-Rajczakowski:
## Description

Redesign functionality of GLSLcompiler class. Change name of this class to ShaderCompiler. Add support for HLSL, GLSL and SPV files. 

## General Checklist:

## Sample Checklist


SaschaWillems:
Not sure if this should be a separate sample. Imo it would be better if this was an addition to the existing basic raytracing sample, e.g. maybe a compile time define so people can toggle between GLSL and HLSL.

samples/extensions/raytracing_basic/raytracing_basic.cpp
```diff
@@ -62,6 +62,31 @@ RaytracingBasic::~RaytracingBasic()
 	}
 }
 
+VkPipelineShaderStageCreateInfo RaytracingBasic::load_spirv_shader(const std::string &filename, VkShaderStageFlagBits stage)
```
asuessenbach: Maybe, instead of introducing a sample-level function `load_spirv_shader`, you could extent the `load_shader` function in vk_common.cpp to also handle files with `.spv` exension by just not compiling them?
You could also easily make it a run-time switch instead of compile-time.

shaders/khr_ray_tracing_basic/hlsl_closesthit.rchit
```diff
@@ -0,0 +1,33 @@
...
+
+struct Attribute
+{
+    float2 bary;
+};
+
+[shader("closesthit")]
+void main(inout Payload payload, in Attribute attribs)
+{
+    const float3 barycentrics = float3(1.0 - attribs.bary.x - attribs.bary.y, attribs.bary.x, attribs.bary.y);
```
asuessenbach: I don't know much about HLSL, but wouldn't it be better to have `1.0f` here, instead of `1.0`?

samples/extensions/raytracing_basic/raytracing_basic.cpp
Szymon-Rajczakowski: The run-time switch sounds good, nevertheless @SaschaWillems suggested defining compile time in a comment above. What should be the preferred solution for switching shaders in load_shader, compile-time switch or runtime switch?

SaschaWillems: I'm happy with both. Feel free to implement it the way you prefer.

framework/common/vk_common.cpp
```diff
@@ -356,25 +356,36 @@ int32_t get_bits_per_pixel(VkFormat format)
...
-	// Compile the GLSL source
-	if (!glsl_compiler.compile_to_spirv(vkb::find_shader_stage(file_ext), buffer, "main", {}, spirv, info_log))
+		// Compile the GLSL source
+		if (!shader_compiler.compile_to_spirv(vkb::find_shader_stage(file_ext), buffer, "main", {}, spirv, info_log))
+		{
+			LOGE("Failed to compile shader, Error: {}", info_log.c_str());
+			return VK_NULL_HANDLE;
+		}
+	}
+	if (src_language == vkb::ShaderSourceLanguage::VK_HLSL)
```
asuessenbach: Should probably be `else if` here

framework/common/vk_common.cpp
```diff
@@ -356,25 +356,36 @@ int32_t get_bits_per_pixel(VkFormat format)
...
-		LOGE("Failed to compile shader, Error: {}", info_log.c_str());
-		return VK_NULL_HANDLE;
+		vkb::ShaderCompiler shader_compiler;
+		// Extract extension name from the glsl shader file
+		file_ext = file_ext.substr(file_ext.find_last_of(".") + 1);
+		if (!shader_compiler.compile_to_spirv(vkb::find_shader_stage(file_ext), buffer, "main", {}, spirv, info_log))
+		{
+			LOGE("Failed to compile shader, Error: {}", info_log.c_str());
+			return VK_NULL_HANDLE;
+		}
```
asuessenbach: Where's the difference between the two cases here, GLSL vs. HLSL?

framework/common/vk_common.cpp
```diff
@@ -383,6 +394,11 @@ VkShaderModule load_shader(const std::string &filename, VkDevice device, VkShade
 	module_create_info.codeSize = spirv.size() * sizeof(uint32_t);
 	module_create_info.pCode    = spirv.data();
```
asuessenbach: These two assignments are only needed when not VK_SPV.

framework/shader_compiler.cpp
```diff
@@ -81,48 +81,58 @@ inline EShLanguage FindShaderLanguage(VkShaderStageFlagBits stage)
...
+                                      vkb::ShaderSourceLanguage   src_language)
 {
 	// Initialize glslang library.
 	glslang::InitializeProcess();
 
 	EShMessages messages = static_cast<EShMessages>(EShMsgDefault | EShMsgVulkanRules | EShMsgSpvRules);
 
+	if (src_language == vkb::ShaderSourceLanguage::VK_HLSL)
+	{
+		messages = static_cast<EShMessages>(EShMsgReadHlsl | EShMsgDefault | EShMsgVulkanRules | EShMsgSpvRules);
```
asuessenbach: Maybe a bit more clear, that you just add some hsls support:
`messages = static_cast<EShMessages>(messages | EShMsgReadHlsl);`

framework/shader_compiler.h
```diff
@@ -56,17 +56,18 @@ class GLSLCompiler
...
-	                      const std::vector<uint8_t> &glsl_source,
-	                      const std::string &         entry_point,
-	                      const ShaderVariant &       shader_variant,
+	                      const std::vector<uint8_t> &shader_source,
+	                      const std::string          &entry_point,
+	                      const ShaderVariant        &shader_variant,
 	                      std::vector<std::uint32_t> &spirv,
-	                      std::string &               info_log);
+	                      std::string                &info_log,
+	                      vkb::ShaderSourceLanguage   src_language = vkb::ShaderSourceLanguage::VK_GLSL);
```
asuessenbach: I think, the logical position of the `src_language` argument would be between the `stage` and the `shader_source`. Even though, it would need adjustments all over the samples and the framework.
Or maybe even as a template argument to `compile_to_spirv`?
Any opinion on that?

samples/extensions/raytracing_basic/raytracing_basic.cpp
```diff
@@ -603,21 +578,16 @@ void RaytracingBasic::create_ray_tracing_pipeline()
...
 	    Each shader group points at the corresponding shader in the pipeline
 	*/
 	std::vector<VkPipelineShaderStageCreateInfo> shader_stages;
-
 	// Ray generation group
 	{
-#ifdef HLSL_SHADER
-		shader_stages.push_back(load_spirv_shader("khr_ray_tracing_basic/hlsl_raygen.rgen.spv", VK_SHADER_STAGE_RAYGEN_BIT_KHR));
-#else
 		shader_stages.push_back(load_shader("khr_ray_tracing_basic/raygen.rgen", VK_SHADER_STAGE_RAYGEN_BIT_KHR));
```
asuessenbach: No differentiation between GLSL and HLSL anymore?

samples/extensions/raytracing_basic/raytracing_basic.cpp
```diff
@@ -630,11 +600,7 @@ void RaytracingBasic::create_ray_tracing_pipeline()
 
 	// Ray miss group
 	{
-#ifdef HLSL_SHADER
-		shader_stages.push_back(load_spirv_shader("khr_ray_tracing_basic/hlsl_miss.rmiss.spv", VK_SHADER_STAGE_MISS_BIT_KHR));
-#else
 		shader_stages.push_back(load_shader("khr_ray_tracing_basic/miss.rmiss", VK_SHADER_STAGE_MISS_BIT_KHR));
```
asuessenbach: As above.

samples/extensions/raytracing_basic/raytracing_basic.cpp
```diff
@@ -647,11 +613,7 @@ void RaytracingBasic::create_ray_tracing_pipeline()
 
 	// Ray closest hit group
 	{
-#ifdef HLSL_SHADER
-		shader_stages.push_back(load_spirv_shader("khr_ray_tracing_basic/hlsl_closesthit.rchit.spv", VK_SHADER_STAGE_CLOSEST_HIT_BIT_KHR));
-#else
-		shader_stages.push_back(load_shader("khr_ray_tracing_basic/closesthit.rchit", VK_SHADER_STAGE_CLOSEST_HIT_BIT_KHR));
-#endif
+		shader_stages.push_back(load_shader("khr_ray_tracing_basic/hlsl_closesthit.rchit.spv", VK_SHADER_STAGE_CLOSEST_HIT_BIT_KHR, vkb::VK_SPV));
```
asuessenbach: As above.

asuessenbach (commented):
I tried to run hlsl_shaders, but get the following error:
```
[error] [framework\common\vk_common.cpp:386] Failed to compile shader, Error: ERROR: #version: ES shaders for SPIR-V require version 310 or higher
ERROR: :20: '' :  syntax error, unexpected LEFT_BRACKET
ERROR: 2 compilation errors.  No code generated.
```
Would I need to update anything on my end?

Besides that: is there any means to switch between HSLS and GLSL now?

framework/shader_compiler.h
Szymon-Rajczakowski: I've overloaded the compile_to_spirv method. Do you think this is a good solution?

framework/common/vk_common.cpp
Szymon-Rajczakowski: I have refactored this method and removed unnecessary logic. I will resolve all conversation from old vk_common.cpp. I would appreciate it if you can find the time to review the latest changes

samples/extensions/raytracing_basic/raytracing_basic.cpp
Szymon-Rajczakowski: I have made a change that will allow you to switch between GLSL shaders and SPV shaders using the UI. Please take a look at the latest changes

samples/extensions/raytracing_basic/raytracing_basic.cpp
```diff
@@ -577,18 +577,20 @@ void RaytracingBasic::create_ray_tracing_pipeline()
 
 	VK_CHECK(vkCreatePipelineLayout(get_device().get_handle(), &pipeline_layout_create_info, nullptr, &pipeline_layout));
 
-	// Ray tracing shaders require SPIR-V 1.4, so we need to set the appropriate target environment for the glslang compiler
-	vkb::GLSLCompiler::set_target_environment(glslang::EShTargetSpv, glslang::EShTargetSpv_1_4);
+	// Ray tracing shaders require SPIR-V 1.4, so we need to set the appropriate target envinroment for the glslang compiler
```
asuessenbach: introduced typo: envinroment

framework/api_vulkan_sample.cpp
```diff
@@ -427,12 +427,12 @@ void ApiVulkanSample::create_pipeline_cache()
 	VK_CHECK(vkCreatePipelineCache(device->get_handle(), &pipeline_cache_create_info, nullptr, &pipeline_cache));
 }
 
-VkPipelineShaderStageCreateInfo ApiVulkanSample::load_shader(const std::string &file, VkShaderStageFlagBits stage)
+VkPipelineShaderStageCreateInfo ApiVulkanSample::load_shader(const std::string &file, VkShaderStageFlagBits stage, vkb::ShaderSourceLanguage src_language)
```
asuessenbach: Would you mind to also extend HPPApiVulkanSample::load_shader accordingly?

framework/api_vulkan_sample.cpp
```diff
@@ -828,17 +828,17 @@ void ApiVulkanSample::update_render_pass_flags(uint32_t flags)
 	dependencies[0].srcSubpass      = VK_SUBPASS_EXTERNAL;
 	dependencies[0].dstSubpass      = 0;
 	dependencies[0].srcStageMask    = VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT;
-	dependencies[0].dstStageMask    = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
+	dependencies[0].dstStageMask    = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT | VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT;
```
asuessenbach: Could you please adjust the corresponding values in HPPApiVulkanSample::setup_render_pass?

framework/common/vk_common.cpp
```diff
@@ -356,25 +356,28 @@ int32_t get_bits_per_pixel(VkFormat format)
 	}
 }
 
-VkShaderModule load_shader(const std::string &filename, VkDevice device, VkShaderStageFlagBits stage)
+VkShaderModule load_shader(const std::string &filename, VkDevice device, VkShaderStageFlagBits stage, vkb::ShaderSourceLanguage src_language)
```
asuessenbach, line 378: Would you mind to also extend `load_shader` in hpp_vk_common.h?

framework/shader_compiler.h
```diff
@@ -54,19 +54,28 @@ class GLSLCompiler
...
 	                      std::vector<std::uint32_t> &spirv,
-	                      std::string &               info_log);
+	                      std::string                &info_log);
+
+	bool compile_to_spirv(VkShaderStageFlagBits       stage,
+	                      const std::vector<uint8_t> &shader_source,
+	                      const std::string          &entry_point,
+	                      const ShaderVariant        &shader_variant,
+	                      std::vector<std::uint32_t> &spirv,
+	                      std::string                &info_log);
```
asuessenbach: Instead of having two flavours of `compile_to_spirv` I would prefer to have just one. Maybe with some default value for the `src_language`.
And please extend the HPPShaderCompiler accordingly.

framework/api_vulkan_sample.cpp
lorna-mcneill-mobica: Looks like the HPPApiVulkanSample::setup_render_pass flags have diverged from the ApiVulkanSample::setup_render_pass flags already
for example HPP dependencies[1].srcStageMask has
eColorAttachmentOutput | eEarlyFragmentTests | eLateFragmentTests
whereas non-HPP dependencies[1].srcStageMask only has VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT 
I guess we should align those to whichever is correct first...?

lorna-mcneill-mobica: @asuessenbach 

You mean HPPApiVulkanSample::update_render_pass_flags?

I see that on KhronosGroup:main, the HPP and non-HPP versions of update_render_pass_flags have diverged, eg 

dependencies[1].srcStageMask    = vk::PipelineStageFlagBits::eColorAttachmentOutput | vk::PipelineStageFlagBits::eEarlyFragmentTests | vk::PipelineStageFlagBits::eLateFragmentTests;

versus 

dependencies[1].srcStageMask    = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;

(not sure whether there are other differences)

I guess those shoule be aligned to whichever is correct too?

asuessenbach: > I guess those should be aligned to whichever is correct too?

That would be perfect, yes!

lorna-mcneill-mobica: @asuessenbach 
I had a look at the change that caused that diversion. It's quite a significant change made to fix various synchronisation errors reported by validation.
If it's Ok with you I'd prefer to keep that "re-alignment" separate from Szymon's change so as not to muddy the waters?

(In that case I suppose I would do that as a separate PR)

asuessenbach: Sure, I'm fine with that.
We should just make sure to not miss that re-alignment.

.copyrightignore
asuessenbach, line 1: Does this file really belong to your PR?

scripts/copyright.py
asuessenbach, line 1: Does this file really belong to your PR?

lorna-mcneill-mobica: Ack, good catch - this was a temporary local change to track down the files causing the copyright checker to fail, I will back out the change!

.copyrightignore
lorna-mcneill-mobica: @asuessenbach 
Yeah, I updated this to tell the copyright checker to ignore .spv files (since Szymon added a number of binary .spv files that we do not want to be copyright checked and these were causing a failure in the CI checks)

asuessenbach (dismissed):

samples/extensions/raytracing_basic/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2021, Sascha Willems
```
SaschaWillems: I don't think it's okay to replace the existing copyright. If you add to a file, please append your copyright instead.

framework/shader_compiler.h
```diff
@@ -33,7 +33,7 @@ namespace vkb
 {
 /// Helper class to generate SPIRV code from GLSL source
```
SaschaWillems: With the compiler now being able to support other languages, the comments need to be changed too. They still refer to GLSL only.

framework/shader_compiler.h
```diff
@@ -54,19 +54,21 @@ class GLSLCompiler
 	static void reset_target_environment();
 
 	/**
-	 * @brief Compiles GLSL to SPIRV code
+	 * @brief Compiles GLSL/HLSL to SPIRV code
```
SaschaWillems, line 58: Is this correct? If so, what tool is used to compile HLSL tp SPIR-V (dxc)? Or does it simply load offline compiled SPV? If it's the latter, the comment needs to be changed.

samples/api/hlsl_shaders/hlsl_shaders.cpp
```diff
@@ -410,7 +410,7 @@ void HlslShaders::prepare_pipelines()
 
 	// Load shaders
 	std::array<VkPipelineShaderStageCreateInfo, 2> shader_stages{};
-	shader_stages[0] = load_hlsl_shader("hlsl_shaders/hlsl_shader.vert", VK_SHADER_STAGE_VERTEX_BIT);
+	shader_stages[0] = load_shader("hlsl_shaders/hlsl_shader.vert", VK_SHADER_STAGE_VERTEX_BIT, vkb::ShaderSourceLanguage::VK_HLSL);
```
SaschaWillems: This change doesn't seem correct. This is a standalone sample that is supposed to use it's own hlsl loading function, but this line would use the one from the framework instead. The change should be reverted.

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021-2023, NVIDIA CORPORATION. All rights reserved.
```
SaschaWillems: I don't think it's okay to replace the existing copyright. If you add to a file, please append your copyright instead.

SaschaWillems (changes requested):

SaschaWillems:
I did a first quick review and added a few comments.

One thing I'm not sure about is the direction of this PR.

On one hand it seems to add HLSL support to the framework's compiler, but only through glslang, which is very very limited. And on the other hand it also adds direct loading of SPIR-V for e.g. the basic ray tracing sample.

Tbh. I think we should discuss a different approach that better fits our framework strategy.

We've been talking about HLSL support for quite some time and one option would be adding direct SPIR-V loading to the shader compiler (as a derived class maybe)  so we can add HLSL to all samples with offline compiled SPIR-V files via DXC. The goal is also to not have to change every sample but to support shader language selection at framework level instead. And imo we should move into that direction.

I appreciate the work put in this PR, but it feels a bit like a stop gap towards our actual goal and we might revert a large part of that PR in the future.

I'll wait for other opinions, and I'd happily discuss this in the next samples call :)

Patryk-Jastrzebski-Mobica:
Hi @SaschaWillems,
We have discussed this problem and think that the best solution will be to extend HLSL support (add raytracing support) in the glslang repository. At the moment @Dawid-Lorenz-Mobica is investigating how to add raytracing support for the HLSL language in the glslang repository.

SaschaWillems:
Tbh I'm not sure if that's the right approach. HLSL in glslang is very limited and the actual reference compiler is DXC. So if we want to add HLSL support we should either integrate DXC or offline compile HLSL to SPIR-V and just use that.

SaschaWillems:
@Szymon-Rajczakowski-Mobica: Can you solve the merge conflict? We talked about this on our last call and would like to merge this as early as  possible, so we can start working on HLSL shaders :)

framework/shader_compiler.h
SaschaWillems: That comment still needs to be updated.

samples/api/hlsl_shaders/hlsl_shaders.cpp
SaschaWillems: This still needs to be reverted.

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
SaschaWillems: This is still open. This sample was done by Andreas (Nvidia).

SaschaWillems:
I gave this a try and added/updated a few comments (see above) that needs to be resolved.

Several build checks also failed, e.g. Ubuntu and Android builds. These need to be fixed too.

One thing that should be changed is that right now instead of the actual human readable HLSL files, the SPV files are added to the project:

image

That doesn't make a lot of sense, I'd expect the hlsl source files in here.

We should also discuss folder structure. Right now (for the one sample were HLSL was added), GLSL and HLSL files are in the same folder, with the HLSL ones being prefixed. I'd prefer to separate them via folders (similar to what I do in my samples). I propose we do this with sub folders like this:

* \shaders
  * \sample
    * \glsl <-- GLSL shaders here
    * \hlsl <-- HLSL shaders and compiled SPV in here

SaschaWillems:
> Besides that: is there any means to switch between HSLS and GLSL now?

I was wondering the same. I don't see a way to select HLSL or GLSL. We should do that as a command line switch.

pawel-jastrzebski-mobica:
> > Besides that: is there any means to switch between HSLS and GLSL now?
> 
> I was wondering the same. I don't see a way to select HLSL or GLSL. We should do that as a command line switch.

@SaschaWillems: I really don't know how to command-line switch should be used in context of this change. Since the PR adds the ability compile HLSL to SPV for both off-line and on-line use-case. Please explain the functionality that could be switched using command-line.

SaschaWillems:
Selecting between running a sample with the GLSL based shader or the HLSL based shader.

docs/build.adoc
```diff
@@ -208,13 +212,20 @@ build\windows\app\bin\Release\AMD64\vulkan_samples.exe
 * C{pp}14 Compiler
 * <<cmake-options,CMake Options>>
 * <<3d-models,3D models>>
+* Vulkan SDK
```
SaschaWillems: Can this be reverted? We don't require the SDK to build our samples.

docs/build.adoc
```diff
@@ -243,11 +254,18 @@ cmake --build build/linux --config Release --target vulkan_samples -j$(nproc)
 * https://vulkan.lunarg.com/doc/sdk/latest/mac/getting_started.html[Vulkan SDK] `./install_vulkan.py`
 * <<cmake-options,CMake Options>>
 * <<3d-models,3D models>>
+* Vulkan SDK
```
SaschaWillems: Same as above.

SaschaWillems:
Also odd, I get three out-of-place hlsl related entries in the MSVC project browser now:

image

Probably left overs. Would be good if this PR did not include any HLSL shaders at all, but just the code required to load SPIR-V. We'll add the HLSL in separate PRs.

pawel-jastrzebski-mobica:
@SaschaWillems : do you want to remove the whole off-line shader compilation functionality from the PR?

SaschaWillems:
Nope, having a way to compile the HLSL shaders vie the build system is fine. Problem is where those projects are put in the project structure. They appear at the top level, but should instead be part of the sample itself.

framework/shader_compiler.h
Seweryn-Zielas-Mobica: Updated

framework/shader_compiler.h
Seweryn-Zielas-Mobica: Glslang is used for compilation at runtime and for offline compilation DirectX Shader Compiler is being used.

samples/api/hlsl_shaders/hlsl_shaders.cpp
Seweryn-Zielas-Mobica: Reverted

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
Seweryn-Zielas-Mobica: Fixed

docs/build.adoc
Seweryn-Zielas-Mobica: Reverted

docs/build.adoc
Seweryn-Zielas-Mobica: Reverted

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
Seweryn-Zielas-Mobica: I had to revert fix, it caused CI checks for copyrights to fail

Seweryn-Zielas-Mobica:
I changed way of using dxc for offline compilation. I had problem with **add_custom_target/add_custom_command** that created targets in top level view of Visual Studio solution, I was only found way how to move it into other folders, but couldn't move it under other targets like specific samples. From what I red and understood, this is most probably not possible. 
Instead of  **add_custom_target/add_custom_command**  is used **execute_process** which seems to work as good or maybe even better than previous solution, as now code for compiling shaders seems more straightforward and less bulky.

The difference in how it works now is rather not significant, before HLSL shaders were compiled at the start of executing build command, e.g. **"cmake --build build/linux --config Release --target vulkan_samples -j$(nproc)"**
Now they are compiled after running command for generating project, e.g. "cmake -G "Unix Makefiles" -Bbuild/linux -DCMAKE_BUILD_TYPE=Release"

I also updated cmake to only compile shaders if dxc executable is found, by searching for it under **VULKAN_SDK** path.

There is still glslang used for compilation HLSL shaders at sample runtime, and rather big modification of **hpp_hello_triangle.cpp** which I'm not sure if is needed.

There is still some work left to do, mostly stylistic like naming scheme and directories/folders paths, but core functionality of this PR that is compiling HLSL shaders seems to be done.

SaschaWillems (changes requested):
Can you undo the addition of HLSL shaders for the ray tracing basic sample. I'm already working on adding actual HLSL shaders, and to clearly divide work this PR should only add the offline compilation part.

Seweryn-Zielas-Mobica:
@SaschaWillems As we discussed that we should only add offline compilation for HLSL, I've created created new PR that only do this, as it seemed easier and less misguiding when looking at this PR title.
New PR: #990

marty-johnson59:
Closing this as we're going a different direction

PR #657, opened 2023-04-02
WIP: Offline Shader Compiler

tomadamatkinson:
> NOTE: This is an early implementation of a shader compiler. The code is not cleaned up and not all edge cases are covered. This PR aims to get early feedback from the WG on offline shader compilation in V2.0 and act as a demonstration of what might be possible and beneficial for current and future sample development

# Description

We have discussed offline shader compilation several times and so I thought I would attempt it in FV2. The benefit of this is that I can make more drastic redesigns to our shader pipeline than in FV1 without affecting the repository. Offline shader compilation gives us lots of positive improvements to our shader pipeline. I decided to take this a step forward and allow for the SPIRV binaries to be embedded in the sample itself. This removes the need to load the shader at runtime.

# Goals

- GLSL Shader Compilation
- SPIRV Shader Reflection (Previously we could only reflect GLSL)
- Per shader compile targets `shader__<name>`
- All shaders compile target `vkb__shaders`
- Header Generation (compile safe structs which match those in shaders)
- Descriptor Set Layout generation (not yet implemented)
- Compile time safety / developer feedback time (if the shader fails or changes, there should be repercussions at compile time)

# GLSL Compiler

Uses the previous compiler but reduces the function definition to as few parameters as possible. We throw errors (as is the suggested approach in v2.0).

# SPIRV Reflection

I have added Spirv-Reflect to the repository to provide SPIRV reflection. This has the added bonus of working with any language we choose to implement. Before GLSL reflection was the only option making HLSL support cumbersome. Resources are reflected in a resource hierarchy: Bound Resources -> Types. This hierarchy can be used to generate headers or create Vulkan descriptor sets in the future.

# CMake

`vkb__register_shader`, as the name implies, was added to simplify shader registration. Under the hood, this generates a Json descriptor of the shader which is then passed to the `shader_compiler`. The entire process is automatic.

```cmake
vkb__register_shader(
    NAME "base_fragment"
    LANGUAGE GLSL
    SHADER_TYPE Fragment
    SHADER_PATH ${CMAKE_CURRENT_SOURCE_DIR}/base.frag
    INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}
    DEFINES # these act as constants
      "MAX_LIGHT_COUNT=${MAX_LIGHT_COUNT}"
    VARIANT_DEFINES # these trigger variants
      "HAS_BASE_COLOR_TEXTURE"
)
```

The above will:
- Generate the `shader__base_fragment` target
- Create a `build/shaders/base_fragment.json` shader descriptor
- At build time call the shader compiler which outputs `build/shaders/base_fragment.generated.hpp`

### Compilation Triggers

Compiling shaders currently takes around 50 seconds when done in parallel (`cmake --build build --target vkb__shaders --parallel`). This includes the shader compiler and its dependencies.

A recompilation is triggered when:
- The Json descriptors contents changes (a check is performed to only update this if `vkb__register_shaders` inputs have changed)
- The shader file is changed
- The `shader_compiler` is altered

In most cases shaders will compile only once

### Compile Target Graph

framework

# Output

`shaders/constant_data/buffer_array.vert` outputs the following. We can see the structs have been reflected and the two variants have been generated. Shader hashes are easily generated from `shaders::CompileConfig::hash()`

```cpp
#pragma once

/* DO NOT EDIT */
/* This file was generated by the shader compiler */

#include <cstdint>
#include <vector>
#include <components/shaders/shader_resources.hpp>

namespace shaders
{

// clang-format off
	constexpr uint32_t BufferArrayVertex__15193367717478976081[] = []; // removed for demo

	constexpr size_t BufferArrayVertex__15193367717478976081_size = 524;

	constexpr uint32_t BufferArrayVertex__15330343479558377201[] = []; // removed for demo

	constexpr size_t BufferArrayVertex__15330343479558377201_size = 588;

// clang-format on

class BufferArrayVertex
{
	public:
		using UnknownType = uint32_t; // used to represent poorly reflected types

		struct MVPUniform
		{
                         // std::array is inline: sizeof(MVPUniform) == 128 == sizeof(float) * 16 * 2
			std::array<std::array<float, 4>, 4> model;
			std::array<std::array<float, 4>, 4> view_proj;
		};

		struct MVPUniformArray
		{
			std::array<MVPUniform, 250> uniform_data;
		};

		std::vector<uint32_t> get_variant(const shaders::CompilerConfig &variant) const
		{
			switch (variant.hash())
			{
				case 15193367717478976081:
					return std::vector<uint32_t>(BufferArrayVertex__15193367717478976081, BufferArrayVertex__15193367717478976081 + BufferArrayVertex__15193367717478976081_size);
					break;
				case 15330343479558377201:
					return std::vector<uint32_t>(BufferArrayVertex__15330343479558377201, BufferArrayVertex__15330343479558377201 + BufferArrayVertex__15330343479558377201_size);
					break;
				default:
					throw std::runtime_error("Invalid shader variant");
			}
};

}        // namespace shaders

```

# Sample

I envision a sample using a shader like so

```cpp
#include <shaders/constant_data/buffer_array_fragment.generated.hpp>


EXPORT_CLIB int sample_main(PlatformContext *platform_context)
{
  shaders::CompilerConfig config;
  config.stage = VK_SHADER_STAGE_FRAGMENT_BIT;
  config.entry_point = "main";
  config.defines = {};

  auto SPIRV = shaders::BufferArrayFragment::get_variant(config);
  auto descriptor_set_layout = shaders::BufferArrayFragment::get_layout(config);
  // ... create pipeline, etc

 shaders::BufferArrayFragment::MVPUniform mvp_uniform;
 mvp_uniform.model = {}; //...
 mvp_uniform.view_proj = {}; //...
}
```


SaschaWillems:
Does this also affect the api samples or is this mostly aimed at samples that require reflection? If it'll still be possible to simply load a GLSL or offline compiled SPIR-V I'm more than fine with this, esp. as this seems like an important base for being able to add in HLSL (offline) compiled shaders.

cmd/shader_compiler/include/shader_descriptor.hpp
```diff
@@ -0,0 +1,63 @@
...
+	TessellationEvaluation,
+	RayGeneration,
+	Intersection,
+	AnyHit,
+	ClosestHit,
+	Miss,
+	Callable,
+	Task,
+	Mesh,
+	Count
```
SaschaWillems, line 27: What shader type does `Count` represent?

components/shaders/src/glslang/default_resource_limits.cpp
SaschaWillems, line 1: Do we really need that file? glslang already has ResourceLimits.h/.cpp

SaschaWillems:
Currently we add shaders to the project file in a sample's CMake. Is that still the case with this change? It looks like we then would have to add a separate CMake file inside each sample's shader folder.

cmd/shader_compiler/include/shader_descriptor.hpp
tomadamatkinson: Count is just the end of the types... A better name would be "Unknown"

components/shaders/src/glslang/default_resource_limits.cpp
tomadamatkinson: ResourceLimits are a private part of that library and isnt exposed without hacky compile target stuff. I was having some compile issues as well so I believe I just chose to copy it

tomadamatkinson:
> Does this also affect the api samples or is this mostly aimed at samples that require reflection? If it'll still be possible to simply load a GLSL or offline compiled SPIR-V I'm more than fine with this, esp. as this seems like an important base for being able to add in HLSL (offline) compiled shaders.

The reflection part of this was more me playing around and seeing what was possible with the reflection and compilation interfaces. Doing this exercise was a good way to add more support to the compiler and reflection as there was a tangible output. Unit tests didn't quite cut it.

I quite like the idea of having compile time validation that a shader will work with a sample or not, but I believe that would need a little more work.

That said this shouldn't effect any samples. The shader components can be used on or offline. This is just an example of an automated compile time process to compile shaders and package them into an executable.

> Currently we add shaders to the project file in a sample's CMake. Is that still the case with this change? It looks like we then would have to add a separate CMake file inside each sample's shader folder.

I chose not to do this for now as it made more sense to keep the CMake close to its target. But each shader has a custom compile target generated for it so it can be linked to a sample that way. If you where then to compile hello_triangle by itself, this would only compile the shaders referenced against that sample. 


Thank you for taking the time to review, its very appreciated :)

Happy to make any changes or push this in a direction which best fits the use-case, this was just me doing the leg work to get something functional working

bldsys/cmake/component_helpers.cmake
```diff
@@ -41,48 +41,59 @@ function(vkb__register_component)
...
-
-        if(TARGET_INCLUDE_DIRS)
-            target_include_directories("vkb__${TARGET_NAME}" INTERFACE ${TARGET_INCLUDE_DIRS})
-        endif()
-
-        target_compile_features("vkb__${TARGET_NAME}" INTERFACE cxx_std_17)
+        target_link_libraries("vkb__${TARGET_NAME}" INTERFACE ${TARGET_HEADER_ONLY})
     endif()
 
     set_property(TARGET "vkb__${TARGET_NAME}" PROPERTY FOLDER "components/${TARGET_NAME}")
```
asuessenbach, line 96: `... FOLDER "components" )` would remove the additional hierarchy level.

bldsys/cmake/shader_helpers.cmake
```diff
@@ -0,0 +1,144 @@
...
+
+    if(NOT TARGET_NAME OR TARGET_NAME STREQUAL "")
+        message(FATAL_ERROR "Shader name not specified")
+    endif()
+
+    if(NOT TARGET_LANGUAGE)
+        message(FATAL_ERROR "Shader language not specified: GLSL, HLSL")
+    endif()
+
+    set(SHADER_LANGUAGE "")
```
asuessenbach, line 37: Why an additional variable `SHADER_LANGUAGE` ? Isn't `TARGET_LANGUAGE` sufficient?

bldsys/cmake/shader_helpers.cmake
```diff
@@ -0,0 +1,144 @@
...
+
+    if(NOT TARGET_SHADER_PATH)
+        message(FATAL_ERROR "Shader path not specified")
+    endif()
+
+    set(ENTRY_POINT "main")
+
+    if(TARGET_ENTRY_POINT)
+        set(ENTRY_POINT ${TARGET_ENTRY_POINT})
+    endif()
```
asuessenbach, line 64: Maybe, instead of introducing `ENTRY_POINT`:
```
	if (NOT TARGET_ENTRY_POINT)
		set(TARGET_ENTRY_POINT "main")
	endif()
```

bldsys/cmake/shader_helpers.cmake
```diff
@@ -0,0 +1,144 @@
...
+
+    if(TARGET_INCLUDE_DIRS)
+        set(ARRAY "[")
+
+        foreach(DIR ${TARGET_INCLUDE_DIRS})
+            string(APPEND ARRAY "\"${DIR}\",")
+        endforeach()
+
+        string(REGEX REPLACE ",$" "" ARRAY ${ARRAY})
+        string(APPEND ARRAY "]")
```
asuessenbach, line 82: Could just replace ",$" with "]", instead of replacing with "" and appending "]".

bldsys/cmake/shader_helpers.cmake
```diff
@@ -0,0 +1,144 @@
...
+
+    if(TARGET_DEFINES)
+        set(ARRAY "[")
+
+        foreach(DEFINE ${TARGET_DEFINES})
+            string(APPEND ARRAY "\"${DEFINE}\",")
+        endforeach()
+
+        string(REGEX REPLACE ",$" "" ARRAY ${ARRAY})
+        string(APPEND ARRAY "]")
```
asuessenbach, line 94: as above

bldsys/cmake/shader_helpers.cmake
```diff
@@ -0,0 +1,144 @@
...
+
+    if(TARGET_VARIANT_DEFINES)
+        set(ARRAY "[")
+
+        foreach(DEFINE ${TARGET_VARIANT_DEFINES})
+            string(APPEND ARRAY "\"${DEFINE}\",")
+        endforeach()
+
+        string(REGEX REPLACE ",$" "" ARRAY ${ARRAY})
+        string(APPEND ARRAY "]")
```
asuessenbach, line 106: as above

bldsys/cmake/shader_helpers.cmake
```diff
@@ -0,0 +1,144 @@
...
+        DEPENDS shader_compiler ${JSON_FILE} ${TARGET_SHADER_PATH}
+        COMMENT "Generating shader header for ${TARGET_NAME}"
+    )
+
+    add_custom_target(
+        "shader__${TARGET_NAME}"
+        DEPENDS
+        ${OUTPUT_FILE}
+    )
+
```
asuessenbach, line 141: Maybe add something like this here:
`set_property(TARGET "shader__${TARGET_NAME}" PROPERTY FOLDER "shaders")`

bldsys/cmake/test_helpers.cmake
```diff
@@ -56,29 +55,32 @@ function(vkb__register_tests)
...
     if(${VKB_WARNINGS_AS_ERRORS})
-        message(STATUS "Warnings as Errors Enabled")
-        if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+        if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
             target_compile_options(${TARGET_NAME} PRIVATE -Werror)
-        elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+        elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
             target_compile_options(${TARGET_NAME} PRIVATE /W3 /WX)
         endif()
     endif()
```
asuessenbach, line 77: The same code sequence on handling VKB_WARNINGS_AS_ERRORS as in component_helpers.cmake and multiple times in this file -> move it into a little helper function?

cmd/shader_compiler/src/header_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+	{
+		if (i == 0)
+		{
+			pascal += toupper(snake[0]);
+			continue;
+		}
+
+		if (snake[i] == '_')
+		{
+			pascal += toupper(snake[i + 1]);
```
asuessenbach, line 29: Fails on a `snake` with a trailing `_`

cmd/shader_compiler/src/header_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+		{
+			if (i % 50 == 0)
+			{
+				if (i != 0)
+				{
+					output_stream << std::endl;
+				}
+
+				output_stream << "\t\t";
+				output_stream << variant.spirv[i];
```
asuessenbach, line 62: The rest of the if-body is the same as the else-body.

cmd/shader_compiler/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
+project(shader_compiler)
+
+set(SOURCES
+    src/shader_descriptor.cpp
+    src/shader_compiler.cpp
+    src/header_generator.cpp
+)
+
+add_library(shader_compiler_lib STATIC ${SOURCES})
```
asuessenbach, line 9: I would prefer to also have the headers added here.
It's easier to navigate from the IDE, then.

cmd/shader_compiler/src/header_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+				if (i != variant.spirv.size() - 1)
+				{
+					output_stream << ",";
+				}
+			}
+			else
+			{
+				output_stream << variant.spirv[i];
```
asuessenbach, line 71: As this is binary data, maybe format the numbers as hex?
```
output_stream << std::setbase(16) << std::setfill('0');
output_stream << "\tconstexpr uint32_t ...
```
`output_stream << "0x" << std::setw(8) << variant.spirv[i];
`

cmd/shader_compiler/src/header_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+		for (const auto &member : type.members)
+		{
+			switch (member->type)
+			{
+				case shaders::ShaderRawDataType::Array: {
+					auto array_member = std::static_pointer_cast<shaders::ArrayShaderData>(member);
+
+					output_stream << "\t\t\t"
+					              << "std::array<" << array_member->type_name << ", " << array_member->element_count << "> " << array_member->struct_member_name << ";" << std::endl;
+					continue;
```
asuessenbach, line 109: `continue` in a switch case works, but is unusual. I would expect a `break` here.

cmd/shader_compiler/src/header_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+					output_stream << "\t\t\t" << detail::snake_to_pascal(struct_member->type_name) << " " << struct_member->struct_member_name << ";" << std::endl;
+					continue;
+				}
+
+				case shaders::ShaderRawDataType::Unknown: {
+					LOGW("Unknown shader data type encountered! - reflector not yet supported for this type!");
+					continue;
+				}
+
+				default: {
```
asuessenbach, line 148: No error or warning in the default case?

cmd/shader_compiler/src/header_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+		pascal += snake[i];
+	}
+
+	return pascal;
+}
+}        // namespace detail
+
+void generate_raw_variant_pointers(const ShaderDescriptor &descriptor, std::ofstream &output_stream)
+{
+	output_stream << "// clang-format off" << std::endl;
```
asuessenbach, line 43: Do you want to clang-format the generated files, at some point?
Currently they are not formatted, so why switch formatting off here?

cmd/shader_compiler/src/header_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+			}
+		}
+
+		output_stream << "\t\t};" << std::endl;
+		output_stream << std::endl;
+	}
+}
+
+void generate_interface_bindings(const ShaderDescriptor &descriptor, std::ofstream &output_stream)
+{
```
asuessenbach, line 161: What is this function supposed to generate?

cmd/shader_compiler/src/header_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+	output_stream << "\t\t\t{" << std::endl;
+
+	for (const auto &variant_pair : descriptor.variants)
+	{
+		auto &hash    = variant_pair.first;
+		auto &variant = variant_pair.second;
+
+		output_stream << "\t\t\t\tcase " << hash << ":" << std::endl;
+		output_stream << "\t\t\t\t\treturn std::vector<uint32_t>(" << detail::snake_to_pascal(descriptor.name) << "__" << hash << ", " << detail::snake_to_pascal(descriptor.name) << "__" << hash << " + " << detail::snake_to_pascal(descriptor.name) << "__" << hash << "_size);" << std::endl;
+		output_stream << "\t\t\t\t\tbreak;" << std::endl;
```
asuessenbach, line 190: No `break` needed after a `return` statement.

cmd/shader_compiler/src/header_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	output_stream << "\t\t\t\tdefault:" << std::endl;
+	output_stream << "\t\t\t\t\tthrow std::runtime_error(\"Invalid shader variant\");" << std::endl;
+	output_stream << "\t\t\t}" << std::endl;
```
asuessenbach, line 195: Missing `output_stream << "\t\t}" << std::endl;`

cmd/shader_compiler/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+set_target_properties(
+    shader_compiler
+    PROPERTIES
+        CXX_STANDARD 17
+        CXX_STANDARD_REQUIRED YES
+        CXX_EXTENSIONS NO
+        ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
+        LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
+        RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
+)
```
asuessenbach, line 43: Why are you separating the shader_compiler into an executable and a library?

cmd/shader_compiler/src/header_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+void generate_raw_variant_pointers(const ShaderDescriptor &descriptor, std::ofstream &output_stream)
+{
+	output_stream << "// clang-format off" << std::endl;
+
+	for (const auto &variant_pair : descriptor.variants)
+	{
+		auto &hash    = variant_pair.first;
+		auto &variant = variant_pair.second;
+
+		output_stream << "\tconstexpr uint32_t " << detail::snake_to_pascal(descriptor.name) << "__" << hash << "[] = [" << std::endl;
```
asuessenbach, line 50: As the function `get_variant` can't be `constexpr`, you could as well make this a `const std::vector<uint32_t>`, and let `get_variant` return a `const std::vector<uint32_t> &`, without any copy operations.

asuessenbach (commented):
Overall, this looks very promising!
Just a lot of comments with the first 13 files in this huge PR. Will probably have a look at the other files later on as well.

tomadamatkinson:
Thanks @asuessenbach for reviewing. Sorry for the size of PR :sweat_smile:

Luckily there are quite a few files in here with very small changes such as the per shader cmake target configurations.

If you are happy with the general direction that it is heading then I will do a second pass and try and clean up some of the more thrown together areas of code. Hopefully this will reduce the PR line count

cmd/shader_compiler/src/shader_compiler.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+  private:
+	std::vector<std::string> search_paths;
+};
+
+std::vector<std::vector<std::string>> generate_combinations(const std::vector<std::string> &input_set)
+{
+	std::vector<std::vector<std::string>> output_set;
+	const std::size_t                     input_size       = input_set.size();
+	const std::size_t                     max_combinations = 1 << input_size;        // equivalent to pow(2, input_size)
```
asuessenbach, line 55: You should at least assert here, that input_size <= 64.
The number of combinations can get huge here pretty easy! What is that supposed to be?

cmd/shader_compiler/src/shader_compiler.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+	for (std::size_t i = 0; i < max_combinations; ++i)
+	{
+		std::vector<std::string> current_combination;
+		current_combination.reserve(input_size);        // reserve memory for maximum possible size
```
asuessenbach, line 60: I think, always reserving the maximum size is a bit too generous.
std::popcount(i) would be a more precise size, with C++20 at least. That could be calculated in a little helper function as well.

cmd/shader_compiler/src/shader_compiler.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+		for (std::size_t j = 0; j < input_size; ++j)
```
asuessenbach, line 62: No need to run to `input_size` every time. `8 * sizeof(size_t) - std::countl_zero(i)` should be enough (or `(1<<j)<=i`).

cmd/shader_compiler/src/shader_compiler.cpp
```diff
@@ -0,0 +1,184 @@
...
+			}
+		}
+
+		output_set.push_back(current_combination);
+	}
+
+	return output_set;
+}
+
+std::string read_file(const std::string &file_name)
```
asuessenbach, line 76: Don't you already have some `read_file` function in filesystem.hpp?

cmd/shader_compiler/src/shader_compiler.cpp
```diff
@@ -0,0 +1,184 @@
...
+		auto   entire_include = include_directive + matched_include + include_trail;
+		size_t pos            = expanded_contents.find(entire_include);
+		size_t end            = pos + entire_include.size();        // keep the qoutes as they need to be removed
+
+		std::string include_contents = read_file(resolver.resolve_include(matched_include));
+		include_contents             = expand_includes(resolver, include_contents);
+
+		expanded_contents.replace(pos, end - pos + 1, include_contents);
+
+		pos += include_contents.size();
```
asuessenbach, line 111: `pos` isn't used in the `PartialMatch`, so no need to adjust it here.

cmd/shader_compiler/src/shader_compiler.cpp
```diff
@@ -0,0 +1,184 @@
...
+	auto contents            = expand_includes(resolver, read_file(descriptor.path));
+	auto contents_bytes      = std::vector<uint8_t>(contents.begin(), contents.end());
+
+	for (auto &define_combination : define_combinations)
+	{
+		// This is effectively a variant of the shader
+		shaders::CompilerConfig config;
+		config.stage = to_vk_shader_stage(descriptor.type);
+
+		for (auto &define : descriptor.defines)
```
asuessenbach, line 145: This loop could be moved out of the surrounding loop.

cmd/shader_compiler/src/shader_compiler.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+	return shader;
+}
+
+void compile_and_reflect_shader(ShaderDescriptor &descriptor)
+{
+	IncludeResolver resolver(descriptor.include_folders);
+
+	auto define_combinations = generate_combinations(descriptor.variant_defines);
+	auto contents            = expand_includes(resolver, read_file(descriptor.path));
```
asuessenbach, line 136: Is expanding the includes really needed? Wouldn't that be done by the actual compiler?

components/platform/include/components/platform/platform.hpp
```diff
@@ -37,7 +37,7 @@ class PlatformContext
 	PlatformContext()          = default;
 	virtual ~PlatformContext() = default;
 
-	virtual std::vector<std::string_view> arguments() const = 0;
+	virtual std::vector<std::string> arguments() const = 0;
```
asuessenbach, line 40: Why don't you return a `const std::vector<std::string> &` ?

components/platform/include/components/platform/platforms/windows.hpp
```diff
@@ -27,15 +30,63 @@ VKBP_ENABLE_WARNINGS()
...
+	WideCharToMultiByte(CP_UTF8, 0, &wstr[0], wstr_len, &str[0], str_len, NULL, NULL);
+
+	return str;
+}
+}        // namespace detail
+
 class WindowsContext final : public PlatformContext
 {
   public:
 	WindowsContext()          = default;
```
asuessenbach, line 58: Why does the WindowsContext not get the arguments as input, like all the other Contexts?
Or at least, move the argument getting that's currently done in the `arguments()` function into the constructor.

components/shaders/include/components/glslang/shader_compiler.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include <glslang/Public/ShaderLang.h>
+VKBP_ENABLE_WARNINGS()
+
+#include <components/shaders/compiler.hpp>
+
+namespace components
+{
+namespace glslang
+{
+using namespace ::glslang;
```
asuessenbach, line 15: You should not use a `using namespace` directive in a header .

asuessenbach (commented):
I have to admit, I don't understand your file hierarchies, especially the include paths.
Why do you have, for example `components\common\include\components\common`, instead of just `components\common\include`?
Moreover, but that seems to be a general issue, I really like to have headers added to a project, which would make them easier to access from an IDE.

tomadamatkinson:
> I have to admit, I don't understand your file hierarchies, especially the include paths.
Why do you have, for example components\common\include\components\common, instead of just components\common\include?
Moreover, but that seems to be a general issue, I really like to have headers added to a project, which would make them easier to access from an IDE.

The path can be split up into two parts: The component `components\common` and the include path `include\components\common`. The include path is structured in a way to avoid header include collisions if files are named the same. In source you include `components/common/some_header.hpp`.

I agree that it seems a little bloated but it enforces the idea that each component is in itself its own library. Headers are not mixed between libraries and it solidifies a barrier between libraries. A header specific to the `vulkan` component shoud not be shown in the `model_loaders` component.

We can address this issue when integrating into the main repo. For now it may be easier to view the v2.0 branch using the file explorer view

components/shaders/include/components/shaders/compiler.hpp
```diff
@@ -0,0 +1,54 @@
...
+	bool is_valid() const noexcept
+	{
+		return stage != VK_SHADER_STAGE_ALL && !entry_point.empty();
+	}
+
+	size_t hash() const noexcept
+	{
+		size_t hash = 0;
+
+		hash ^= std::hash<std::string>{}(entry_point);
```
asuessenbach, line 31: xor'ing is a pretty bad hash combiner. Maybe something like what is used in boost (or vulkan_hash.hpp as well) is a bit better?

components/shaders/include/components/shaders/shader.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+namespace components
+{
+namespace shaders
+{
+class ShaderStage
+{
+  public:
+	virtual VkShaderStageFlagBits stage() const     = 0;
+	virtual ShaderResources       resources() const = 0;
```
asuessenbach, line 36: A `ShaderResources` might be a pretty heavy object. Maybe return a `const &`, instead?

components/shaders/include/components/shaders/shader.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+	size_t hash() const
+	{
+		return _hash;
+	}
+
+  private:
+	std::string                 _name;
+	std::vector<ShaderStagePtr> _stages;
+	size_t                      _hash;
```
asuessenbach, line 75: `_name` and `_hash` are never set.

components/shaders/include/components/shaders/shader_resource.hpp
```diff
@@ -0,0 +1,322 @@
...
+	Array,
+	RuntimeArray,
+	Struct,
+	Unknown
+};
+
+struct RawShaderData
+{
+	ShaderRawDataType type{ShaderRawDataType::Unknown};
+	std::string       struct_member_name;
```
asuessenbach, line 63: The `struct_member_name` probably belongs into the `StructShaderData`. Maybe by changing its member to `std::vector<std::pair<std::string, std::shared_ptr<RawShaderData>>>`.

components/shaders/include/components/shaders/shader_resource.hpp
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	static ShaderResources merge(const ShaderResources &a, const ShaderResources &b)
+	{
+		ShaderResources copy = a;
+		copy.merge(b.resources);
+		return copy;
+	}
+
+	std::vector<StructShaderData> structs_in_topological_order() const
```
asuessenbach, line 179: I don't get, what this function is supposed to do. Could you provide some explanation?

components/shaders/include/components/shaders/shader_resource.hpp
```diff
@@ -0,0 +1,322 @@
...
+				{
+					found = true;
+					break;
+				}
+			}
+
+			if (!found)
+			{
+				resources.push_back(new_resource);
+			}
```
asuessenbach, line 241: Maybe:
```
auto it = std::find_if(resources.begin(), resources.end(), [&new_resource](ShaderResource const & resource){ return resource.name == new_resource.name; });
if (it == resources.end())
{
    resources.push_back(new_resource);
}
```

components/shaders/include/components/shaders/shader_resource.hpp
```diff
@@ -0,0 +1,322 @@
...
+
+			if (structs.find(struct_data->type_name) == structs.end())
+			{
+				structs[struct_data->type_name] = *struct_data;
+
+				for (auto &member : struct_data->members)
+				{
+					if (member->type == ShaderRawDataType::Struct)
+					{
+						if (auto struct_member = std::static_pointer_cast<StructShaderData>(member))
```
asuessenbach, line 282: Can this check fail?

components/shaders/include/components/shaders/shader_resource.hpp
```diff
@@ -0,0 +1,322 @@
...
+						}
+					}
+
+					if (member->type == ShaderRawDataType::Array)
+					{
+						if (auto array_member = std::static_pointer_cast<ArrayShaderData>(member))
+						{
+							if (array_member->element_type->type == ShaderRawDataType::Struct)
+							{
+								if (auto struct_member = std::static_pointer_cast<StructShaderData>(array_member->element_type))
```
asuessenbach, line 294: Can this check fail?

components/shaders/src/glslang/shader_compiler.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+#include "default_resource_limits.h"
+
+namespace components
+{
+namespace glslang
+{
+namespace detail
+{
+inline EShLanguage find_shader_language(VkShaderStageFlagBits stage)
```
asuessenbach, line 11: Maybe replace this function with a simple map?

components/shaders/include/components/glslang/shader_compiler.hpp
```diff
@@ -0,0 +1,35 @@
...
+class GlslangShaderCompiler : public shaders::ShaderCompiler
+{
+  public:
+	GlslangShaderCompiler(LogCallback &&log_callback = nullptr);
+	~GlslangShaderCompiler() = default;
+
+	// Compile shader source code to SPIR-V
+	std::vector<uint32_t> compile_spirv(const shaders::CompilerConfig &config, const std::vector<uint8_t> &shader_source) const override;
+
+	EShTargetLanguage        target_language         = EShTargetLanguage::EShTargetSpv;
```
asuessenbach, line 28: At least `target_language` could probably by a `const static` ?

components/shaders/src/glslang/shader_compiler.cpp
```diff
@@ -0,0 +1,146 @@
...
+	std::string source   = std::string(shader_source.begin(), shader_source.end());
+
+	const char *file_name_list[1]  = {""};
+	const char *shader_source_data = reinterpret_cast<const char *>(source.data());
+
+	glslang::TShader shader(language);
+	shader.setStringsWithLengthsAndNames(&shader_source_data, nullptr, file_name_list, 1);
+	shader.setEntryPoint(config.entry_point.c_str());
+	shader.setSourceEntryPoint(config.entry_point.c_str());
+	shader.setEnvTarget(target_language, target_language_version);
```
asuessenbach, line 97: Can target_language and version really vary?

components/shaders/include/components/shaders/compiler.hpp
SaschaWillems: Or what about using a standard, not implementation dependent hash algorithm like SHA256 instead?

PR #667, opened 2023-04-19
Add sample with ability to enable and disable barriers

Michal-Masiukiewicz-Mobica:
## Description

This PR adds new sample which shows barriers usage and allows to disable each used barrier from GUI, it should help to understand barriers (especially with enabled synchronization validation layers). 

The sample is variation of HDR sample implemented with VK_KHR_dynamic_rendering to show explicitly each required barrier.

I hoped that disabling barriers would show artifacts on screen but actually in most cases no change is seen, but I was able to get some artifacts after disabling barriers and switching from and to fullscreen mode (Linux, AMD, Wayland).

I am unsure if such sample is desired.
Waiting for any feedback :)

## Framework change
To display GUI with VK_KHR_dynamic_rendering small change in framework was required (added virtual method prepare_gui).

## Notes

Tested on:
Windows 10
Nvidia Quadro P600
Intel i7-8750H
32 GB RAM

## General Checklist:

## Sample Checklist


gary-sweet:
> I am unsure if such sample is desired ...

My personal opinion is that we shouldn't be deliberately causing invalid usage like this in samples.
On my driver for example, turning off all the barriers still results in a correct result. People could naively see this and just assume they don't need to use barriers, which would be very wrong. We should be showing best practice, not bad practice in my opinion.

SaschaWillems:
I'm with Gary on this. There are a lot of implementations where this will not show any errors without barriers, and such a sample could cause confusion or even lead people to think that stuff just works without barriers.

Michal-Masiukiewicz-Mobica:
Seems like sample with improper usage is not desired. Closing.

PR #678, opened 2023-04-25
Change external_files_dir logic and permission logic.

### LINKED ISSUE #646: Android 11 Permissions
## Issue:
Vulkan Samples currently require access to directories/files outside the app's directory. Android 11 changed how storage permissions work and at the moment the only way to bypass the storage permission requirements is to request the MANAGE_EXTERNAL_STORAGE permission with adb. 
The Vulkan Samples application actually requests this permission already but it doesn't appear to work reliably, at least on S23.
Similarly, we also have to manually create the output log directory to avoid exceptions being generated.

## Proposed Fix:
* Change the build system to pack the assets and shaders into the APK
* Add an AssetManager class which makes use of AAssetManager to load assets
* Change the filesystem and other asset loading code to support this
* Modify the Java code to change the location of the external directory so that Vulkan Samples can output to a directory inside the app (Android/data/com.khronos.vulkan_samples/...)

A similar solution to this was previously proposed, and rejected: https://github.com/KhronosGroup/Vulkan-Samples/issues/39, but we feel that this is worth revisiting.

stmazoni:
## Description
Current implementation has problem, it doesn't show sample list well, I tested on Pixel 6 with Android 13.

I removed changing external_files_dir path. It changed directory path from '/storage/emulated/0/Android/data/com.khronos.vulkan_samples/files' to '/storage/emulated/0/com.khronos.vulkan_samples/'. But on the build.md file guide to copy resource to '/storage/emulated/0/Android/data/com.khronos.vulkan_samples/files'. And also it is better to use '/storage/emulated/0/Android/data/com.khronos.vulkan_samples', because it is deleted automatically when application is deleted.

And about the permission, current logic makes not showing sample list on the Android 13 device. it is happen because WRITE_EXTERNAL_STORAGE and READ_EXTERNAL_STORAGE has no effect after API level 33, current logic think application doesn't have permissions. However it can be used without permission request after API level 19. https://developer.android.com/reference/android/Manifest.permission#READ_EXTERNAL_STORAGE https://developer.android.com/reference/android/Manifest.permission#WRITE_EXTERNAL_STORAGE
I add if API level is same or higher than 19(Kitkat), skip the checkPermissions() and assume it already have these permissions. It is also related with not working on S23 device on https://github.com/KhronosGroup/Vulkan-Samples/issues/646.

Fixes #646 

## General Checklist:

## Sample Checklist

### DIFF
--- a/app/android/java/com/khronos/vulkan_samples/SampleLauncherActivity.java
+++ b/app/android/java/com/khronos/vulkan_samples/SampleLauncherActivity.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2019-2021, Arm Limited and Contributors
+/* Copyright (c) 2019-2023, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
--- a/app/android/java/com/khronos/vulkan_samples/SampleLauncherActivity.java
+++ b/app/android/java/com/khronos/vulkan_samples/SampleLauncherActivity.java
@@ -74,10 +74,6 @@ public class SampleLauncherActivity extends AppCompatActivity {
             File external_files_dir = getExternalFilesDir("");
             File temp_files_dir = getCacheDir();
             if (external_files_dir != null && temp_files_dir != null) {
-                // User no longer has permissions to access applications' storage, save files in
-                // top level (shared) external storage directory
-                String shared_storage = external_files_dir.getPath().split(Pattern.quote("Android"))[0];
-                external_files_dir = new File(shared_storage, getPackageName());
                 initFilePath(external_files_dir.toString(), temp_files_dir.toString());
             }
 
@@ -90,7 +86,7 @@ public class SampleLauncherActivity extends AppCompatActivity {
         permissions.add(new Permission(Manifest.permission.WRITE_EXTERNAL_STORAGE, 1));
         permissions.add(new Permission(Manifest.permission.READ_EXTERNAL_STORAGE, 2));
 
-        if (checkPermissions()) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT || checkPermissions()) {
             // Permissions previously granted skip requesting them
             parseArguments();
             showSamples();


r-potter:
@r-kaczynski will check for Samsung devices.

r-kaczynski:
<del>Tested this branch on S23+ and S22 and after granting permissions I get this screen:
image
Pressing the "Grant Permissions" button does nothing. I ran into a similar issue when trying to change the java code.</del>

Same as @SaschaWillems. I checked again and pushing the assets solves the problem.

SaschaWillems (changes requested):
~~Doesn't work on my Pixel 4a either (latest firmware, Android 13). I do get a file permission dialog, but granting the permission doesn't return to the samples. If I manually go back, all samples simply crash at startup.~~

Just checked again, and it seems that the build step simply didn't push the assets. After manually pushing them, and setting the permissions the samples work fine.

stmazoni:
> Just checked again, and it seems that the build step simply didn't push the assets. After manually pushing them, and setting the permissions the samples work fine.

It is not related with this PR.
I also cannot push assets, even adb push assets folder is not working. I pushed with using Windows file explorer. It seems because of security, and happen when pushed folder has another folders internally. I am checking it is security  policy or not in internally.

JoseEmilio-ARM:
> ~Tested this branch on S23+ and S22 and after granting permissions I get this screen: image Pressing the "Grant Permissions" button does nothing. I ran into a similar issue when trying to change the java code.~
> 
> Same as @SaschaWillems. I checked again and pushing the assets solves the problem.

I've also noticed that when installing the samples on a new device, this screen might also appear if not granting notifications permissions (i.e. we do not have `<uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>` in the [manifest](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/app/android/AndroidManifest.xml)) which makes the error message about assets misleading. Manually granting notifications permissions fixed the issue for me.

JoseEmilio-ARM:
Hi @stmazoni  , since #713 has now been merged, can we close this PR now?

PR #690, opened 2023-05-05
Add Common Component

### LINKED ISSUE #689: Implement the initial component build system
## Description

To be able to use the new framework we need to implement the CMake functions required to construct components.

## Acceptence Criteria

- Add CMake helpers
- Add a single component (Common? Core?)
- Include tests
- Folder structure agreed upon - works in all IDEs

tomadamatkinson:
## Description

Implements component helpers and the common component as the default example

At this point, we need to agree on the following:
- Folder structure of components
- Documentation method (README.md per component)
- IDE support / how components show in the IDE

Fixes #689

## Structure of a Component

Root folder `components/<component_name>` represents a single component

Include directory `<root_dir>/include/components/<component_name>` contains all public headers for a given component. Nesting the components headers with `components` reduces header name collisions when linking against multiple components. Source files need only `#include <components/component_name/header.hpp>`

Source directory `<root_dir>/src` contains all private headers and source files. Private headers should be used to hide complexity from samples. In v1.0 everything is public which lead to a lot of accidental coupling

Test directory `<root_dir>/tests` contains all `<test_file>.test.cpp` files

### VSCode

image

### Visual Studio

image

### CI Test Output

`--output-on-failure` provides an error report for only the tests that fail. Tests are run in parallel and are extremely quick as they do not require the entire framework to initialise before running.

```
Run ctest -C Release --test-dir "build/ubuntu" --output-on-failure
Internal ctest changing into directory: /home/runner/work/Vulkan-Samples/Vulkan-Samples/build/ubuntu
Test project /home/runner/work/Vulkan-Samples/Vulkan-Samples/build/ubuntu
    Start 1: test__core
1/1 Test #1: test__core .......................   Passed    0.00 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.01 sec
```


third_party/CMakeLists.txt
```diff
@@ -363,3 +363,8 @@ add_library(opencl INTERFACE)
 set(OPENCL_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/opencl)
 target_sources(opencl INTERFACE ${OPENCL_INCLUDE_DIR}/CL/opencl.h)
 target_include_directories(opencl INTERFACE ${OPENCL_INCLUDE_DIR})
+
+# Catch2
+add_subdirectory(catch2)
+set_property(TARGET Catch2 PROPERTY FOLDER "third_party")
```
SaschaWillems: The other 3rd party modules are in the "ThirdParty" folder, would be good to have this consistent.

image

SaschaWillems (changes requested):
Folder structure looks like expected for me and I'm fine with how this is structured.

Also tested with Android Studio and looking good too:

image

One thing that feels a bit odd is the directory structure though. E.g. why `components\core\include\core\util` instead of just `components\core\include\util` or just `components\core\include`?

third_party/CMakeLists.txt
gpx1000: I would prefer keeping with the idea that header files are in the same folder as source files.  The reasoning behind this is the tooling assumes it and it takes less work to keep things working.  For example in CMake, you only need to include the source filename in the target description and it will find the header file automatically if it's in the same directory for purposes of determining if the file has changed etc.

tomadamatkinson: I'm not sure I quite understand where you are coming from. In V1.0 we still had to manually add headers to source groups for them to show up in IDEs. In the v2.0 branch I do not manually add any headers and recompilation has been fine with CMake on all platforms. I often recompile after small changes so I would like to think that I have tested this case thoroughly - any steps I can reproduce to hit this recompilation mishap?

The biggest issues I've had are with solution compatibility with Visual Studio. I haven't tried this on other IDEs. If this looks a bit off in your IDE of choice could you share a screenshot?

In terms of the header folder structure, it may seem a little overkill right now. However, when this folder structure grows the change is a little more natural. I am happy to drop the extra `core` / `components` subfolders. But these have helped reduce header collisions in my platform migration branch (that's a work in progress). It is a similar pattern to `#inlcude <spdlog/fmt/fmt.hpp>` the extra folders help "namespace" the headers.

Very stylistic choices, so let me know more about what sort of styles you think may work better and ill be happy to try and incorporate them.

gpx1000: Adding headers to sourcegroups in order for an IDE to work is a bug in the IDE.  The problem with the work around of including the header files in the source groups is that it means that cmake will erroniously treat those header files as if they were source files and try to compile them.  Most of the time; this isn't a problem because not a lot of code goes into header files which, when compiled, creates bugs.  However, it is incorrect and it causes slow down in compile time and configuration time.  To confirm this, take a look in the .git folder for your listing of dependencies for the project.
When CMake goes through the creation of the dependencies, you want it to realize that header files are dependencies and not source files.  This way, if you touch a header file without listing it; it will cause a recompile correctly.

Visual Studio has recently received support for loading cmake projects directly.  I know that VS has recently been trying to fix some of these IDE bugs.  You should find that both XCode and CLion or (Android Studio) should behave correctly with regards to header files.  I don't know if it would help to try loading the project in VS as a cmake project rather than using kitware's VS generator command step.  However, this should hopefully be the correct thing to do with CMake projects.
If we don't get a good VS style solution; we can continue the work-around; but someday in the future we should review the decision again.  The likelihood of there being a real compilation bug that is caused by this is remote.

Something that might be a happy middle ground for the folder structure is to use header only target libraries.  Each folder has it's own CMakeLists.txt and the include directory merely sets the headers.  Then to use them; make them dependencies of the consuming target in CMake.  You can then just say "#include <fmt.hpp>"  or similar without worry about the folder structure.

tomadamatkinson: With `vkb__register_component` we can tweak this to use both. Interface for headers and static or shared for headers + source. I do want to try and keep the public and private headers as separate as possible as it encourages only exposing the public stuff in the future. One issue we can see in samples and many other projects is how coupled these different areas get over time. So separation by default can be a good thing

bldsys/cmake/component_helper.cmake
```diff
@@ -0,0 +1,163 @@
...
+    set(multiValueArgs SRC HEADERS LINK_LIBS INCLUDE_DIRS)
+
+    cmake_parse_arguments(TARGET "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+    if(TARGET_NAME STREQUAL "")
+        message(FATAL_ERROR "NAME must be defined in vkb__register_tests")
+    endif()
+
+    set(TARGET "vkb__${TARGET_NAME}")
+    set(TARGET_FOLDER "components/${TARGET_NAME}")
```
asuessenbach: Is this additional folder level needed/useful?
In VS, it looks like this:
image
Instead, I would prefer this view:
image

bldsys/cmake/component_helper.cmake
```diff
@@ -0,0 +1,163 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## Bucket target for all components
+add_custom_target(vkb__components)
+
+set_property(TARGET vkb__components PROPERTY FOLDER "components")
```
asuessenbach: What is vkb__components supposed to be used for? Just for dependency management?
If so, does it really belong into the "components" folder? After all, it is not a component.

bldsys/cmake/component_helper.cmake
```diff
@@ -0,0 +1,163 @@
...
+            if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+                target_compile_options(${TARGET} PRIVATE -Werror)
+                # target_compile_options(${TARGET} PRIVATE -Wall -Wextra -Wpedantic -Werror)
+            elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+                target_compile_options(${TARGET} PRIVATE /W3 /WX)
+                # target_compile_options(${TARGET} PRIVATE /W4 /WX)
+            endif()
+        endif()
+    else()
+        message("INTERFACE: ${TARGET}")
```
asuessenbach: Should be `message(STATUS "...` ?

components/core/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+vkb__register_component(
+    NAME core
+    HEADERS
+        include/core/util/strings.hpp
+        include/core/util/error.hpp
+        include/core/util/hash.hpp
```
asuessenbach, line 22: Why is there an additional folder level named "util"?

components/core/include/core/util/strings.hpp
```diff
@@ -0,0 +1,38 @@
...
+{
+/**
+ * @brief Replaces all occurrences of a substring with another substring.
+ */
+std::string replace_all(std::string str, const std::string &from, const std::string &to);
+
+/**
+ * @brief Removes all occurrences of a set of characters from the end of a string.
+ */
+std::string trim_right(const std::string &str, const std::string &chars = "");
```
asuessenbach: Maybe `chars` should default to whitespace. This is probably the most used function with trimming.

components/core/include/core/util/strings.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @brief Removes all occurrences of a set of characters from the beginning of a string.
+ */
+std::string trim_left(const std::string &str, const std::string &chars = "");
```
asuessenbach: As above.

bldsys/cmake/component_helper.cmake
```diff
@@ -0,0 +1,163 @@
...
+
+    if(TARGET_NAME STREQUAL "")
+        message(FATAL_ERROR "NAME must be defined in vkb__register_tests")
+    endif()
+
+    if(NOT TARGET_SRC)
+        message(FATAL_ERROR "One or more source files must be added to vkb__register_tests")
+    endif()
+
+    set(TARGET_FOLDER "components/${TARGET_COMPONENT}")
```
asuessenbach: Maybe set the folder to something like "tests" ?

bldsys/cmake/component_helper.cmake
tomadamatkinson: It's a custom target that allows us to compile just components. Quite useful when you just want to rebuild components and not their dependants (samples, tests, executables)

third_party/CMakeLists.txt
tomadamatkinson: Any actions from this conversation? Other than Saschas `third_party` rename

bldsys/cmake/component_helper.cmake
tomadamatkinson: That makes a lot more sense! Will adjust this

components/core/CMakeLists.txt
tomadamatkinson: What I found with the v2.0 branch was the weird multiple nested components. This was mentioned on one of the PRs but I decided to address it at a different point.

The aim of core is to contain all the interfaces used throughout the rest of the components. 

```
core/window.hpp
core/model_loader.hpp
core/image_loader.hpp
```

and so on...

This stops the multiple nested components issue seen in the assets component on the v2.0 branch

The aim is to keep this header and abstract interface only... however, its likely it might evolve

tomadamatkinson: Util has landed there for now as it is common to have core utilities used throughout all the components which dont take up much space and don't warrant their own component

bldsys/cmake/component_helper.cmake
tomadamatkinson: Ah this was so that the test target showed up close to the component target. But I can remove the `${TARGET_COMPONENT}` part so that it matches your proposed changes in the screenshots

components/core/include/core/util/strings.hpp
tomadamatkinson: updated the tests to include the default case

components/core/include/core/util/strings.hpp
tomadamatkinson: updated the tests to include the default case

bldsys/cmake/component_helper.cmake
tomadamatkinson: this should now be seen in the latest commit

bldsys/cmake/component_helper.cmake
asuessenbach: Maybe it should be located in parallel to vkb__tests, that is, on the top level, not inside the components folder?

tomadamatkinson:
@asuessenbach @SaschaWillems I believe I have resolved all comments :)

asuessenbach (approved):

SaschaWillems (approved):

gpx1000 (approved):

tomadamatkinson:
Merging on three approvals

PR #701, opened 2023-05-11
Custom Main

### LINKED ISSUE #698: Replace entrypoint with CUSTOM_MAIN
A first step in removing the platform would be to replace our current entrypoint (`app/main.cpp`) with a `CUSTOM_MAIN` declaration.

```cpp
CUSTOM_MAIN(context)
{
  //... pass context to platform

  //... existing platform lifecycle
}
```

This provides the starting point of breaking the acyclic platform dependency

tomadamatkinson:
## Description

The main goal from this PR is to lay the foundation of a custom entry point which can be reused for multiple executables. Currently this is only used in the `vulkan_samples` CLI. Down the line this can be used in sample single entry point and in compile time executables such as offline shader compilation / reflection or asset pack preparation.

This PR starts small by moving only the entry point, arguments and file paths away from the frameworks platform. Following this we can remove the windows which will eventually allow us to remove the platform in its entirety. We can do this as samples only require the window interface and some boolean checks if we are running in a specific mode. All other platform functionality is to facilitate the vulkan samples cli lifecycle

Recompile times when tweaking a platform-specific implementation are extremely quick as we no longer need to recompile parts of the framework. 

The folder structure may seem a little overkill whilst this PR targets just the entry points. But these folders will fill out after moving the windows.

- [x] Unix
- [x] Unix D2D
- [x] Macos (via Unix)
- [x] Android
- [x] Windows

Fixes #698 
Based off #690 

## General Checklist:

## Sample Checklist


tomadamatkinson:
@gary-sweet can you help test the Unix D2D implementation, please? Nothing should have changed but always good to be sure.

I'm also interested if we can set up a local device to test with. Is there any way to emulate D2D or can you share any links to some getting started articles? I'm struggling to find anything of substance but it's likely that I don't fully understand what is going on with the Unix D2D stuff

tomadamatkinson:
Tested and works on all platforms that I have access too. I would greatly appreciate a second test on an android device. Currently I only have access to a Google Pixel 3

gary-sweet:
> @gary-sweet can you help test the Unix D2D implementation, please? Nothing should have changed but always good to be sure.

Sorry for the delay Tom - I was on holiday for a couple of weeks, then out of commission with covid for a week and then chasing my tail trying to catch up for a week.  I'll see what I can do today, but I have slightly hampered capability right now as I'm still having to work from home.

gary-sweet:
I'm not seeing any issues here other than an odd build warning that I don't seem to get without this change. It's not obvious what has changed to cause the warning though.

```
Vulkan-Samples/samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp: In member function ‘std::pair<vk::Result, unsigned int> HPPHelloTriangle::acquire_next_image()’:
Vulkan-Samples/samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp:503:23: warning: ignoring return value of ‘vk::Result vk::Device::waitForFences(const vk::ArrayProxy<const vk::Fence>&, vk::Bool32, uint64_t, const Dispatch&) const [with Dispatch = vk::DispatchLoaderDynamic; vk::Bool32 = unsigned int; uint64_t = long unsigned int]’, declared with attribute nodiscard [-Wunused-result]
  503 |   device.waitForFences(per_frame_data[image].queue_submit_fence, true, UINT64_MAX);
      |   ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from Vulkan-Samples/third_party/vulkan/include/vulkan/vulkan.hpp:6681,
                 from Vulkan-Samples/framework/common/hpp_error.h:22,
                 from Vulkan-Samples/framework/core/hpp_instance.h:20,
                 from Vulkan-Samples/framework/platform/hpp_application.h:22,
                 from Vulkan-Samples/samples/api/hpp_hello_triangle/hpp_hello_triangle.h:20,
                 from Vulkan-Samples/samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp:18:
Vulkan-Samples/third_party/vulkan/include/vulkan/vulkan_funcs.hpp:1499:42: note: declared here
 1499 |                                          Device::waitForFences( VULKAN_HPP_NAMESPACE::ArrayProxy<const VULKAN_HPP_NAMESPACE::Fence> const & fences,
      |                                          ^~~~~~
```

gary-sweet:
> I'm also interested if we can set up a local device to test with. Is there any way to emulate D2D or can you share any links to some getting started articles? I'm struggling to find anything of substance but it's likely that I don't fully understand what is going on with the Unix D2D stuff

Hmmm, not that I know of.  Our test devices only exist in our labs or at customers using them for early access. The overhead of maintaining the build and run environment for one would be large even if we were allowed to share one (which I highly doubt).

We didn't invent the D2D APIs in Vulkan, we just found them and used them as they fit our needs. I'm not sure which vendor initially created them tbh.

SaschaWillems:
This looks really good!

Compiles fine for me on:
- Windows 11
- Android 13 (built from Windows)

Runs fine on:
- Windows 11

Sadly can't test on Android due to the still persisting permission problem (which is out of the scope of this PR).

SaschaWillems (dismissed):

SaschaWillems:
One thing I noticed though is that there are now two build steps for each platform in the CI:

image

samples/extensions/open_cl_interop/open_cl_interop.cpp
```diff
@@ -613,9 +613,9 @@ void OpenCLInterop::prepare_open_cl_resources()
 
 	cl_data->command_queue = clCreateCommandQueue(cl_data->context, cl_data->device_id, 0, &result);
 
-	auto   kernel_source      = vkb::fs::read_shader("open_cl_interop/procedural_texture.cl");
-	auto   kernel_source_data = kernel_source.data();
-	size_t kernel_source_size = kernel_source.size();
+	auto        kernel_source      = vkb::fs::read_shader("open_cl_interop/procedural_texture.cl");
```
SaschaWillems, line 615: was this change intentional? This could possibly cause merge conflicts with my OpenCL PR, as I had to rename and move this arm specific sample.

tomadamatkinson: Will need to look into this further but this PR should not need to change samples. Likely a rebase issue

tomadamatkinson:
The new group builds and runs unit tests. These are generally quicker than seen in this PR but it is likely that GitHub's recent CI/CD outages have degraded their performance

Locally building and compiling the unit tests takes a matter of seconds and run in sub seconds

These tests do not require compiling the framework

samples/extensions/open_cl_interop/open_cl_interop.cpp
tomadamatkinson: Fixed

tomadamatkinson:
Just got a chance to come back to this, fixed PR comments @SaschaWillems @gary-sweet

SaschaWillems (dismissed):

gpx1000 (dismissed):
LGTM; I tested it on Linux and works well here.  Will need to work this back into the iOS work that's in progress.

app/main.cpp
```diff
@@ -19,32 +19,30 @@
...
+#elif defined(PLATFORM__LINUX_D2D)
 #	include "platform/unix/unix_d2d_platform.h"
-int main(int argc, char *argv[])
-{
-	vkb::UnixD2DPlatform platform{argc, argv};
-#else
+#elif defined(PLATFORM__LINUX) || defined(PLATFORM__MACOS)
 #	include "platform/unix/unix_platform.h"
-int main(int argc, char *argv[])
+#endif
```
asuessenbach, line 34: Wouldn't this sequence (lines 24-32) be a perfect fit for some `platform/platform.hpp`?
Besides that: some `#else` branch with an error message?

app/main.cpp
```diff
@@ -19,32 +19,30 @@
...
+#if defined(PLATFORM__ANDROID)
+	vkb::AndroidPlatform platform{context};
+#elif defined(PLATFORM__WINDOWS)
+	vkb::WindowsPlatform platform{context};
+#elif defined(PLATFORM__LINUX_D2D)
+	vkb::UnixD2DPlatform platform{context};
+#elif defined(PLATFORM__LINUX)
+	vkb::UnixPlatform platform{context, vkb::UnixType::Linux};
+#elif defined(PLATFORM__MACOS)
+	vkb::UnixPlatform platform{context, vkb::UnixType::Mac};
```
asuessenbach, line 47: And could this sequence (lines 36-46) be moved into some `platform/platform.inl` or hidden behind some `PLATFORM`-macro?
Besides that, again: some `#else` branch with some error message?

components/CMakeLists.txt
```diff
@@ -13,6 +13,15 @@
...
 
-add_subdirectory(core)
\ No newline at end of file
+add_subdirectory(core)
+
+if(ANDROID)
+    add_subdirectory(android)
+elseif(WIN32)
+    add_subdirectory(windows)
+elseif(APPLE)
```
asuessenbach: -> `elseif(APPLE OR UNIX)` ?

components/core/CMakeLists.txt
```diff
@@ -33,4 +33,42 @@ vkb__register_tests(
         tests/strings.test.cpp
     LINK_LIBS
         vkb__core
-)
\ No newline at end of file
+)
+
+if(ANDROID)
+    message(STATUS "android")
```
asuessenbach: Is this message useful?

components/CMakeLists.txt
```diff
@@ -13,6 +13,15 @@
...
+    add_subdirectory(unix) # reuses then unix platform
+elseif(UNIX)
+    add_subdirectory(unix)
+endif()
```
asuessenbach, line 27: Some `else()` branch with an error message?

components/unix/include/unix/context.hpp
```diff
@@ -0,0 +1,46 @@
...
+#pragma once
+
+#include <string>
+
+#include <core/platform/context.hpp>
+
+namespace vkb
+{
+/**
+ * @brief MS Windows platform context
```
asuessenbach: MS Windows -> Unix

components/unix/include/unix/context.hpp
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @warning Use in extreme circumstances with code guarded by the PLATFORM__WINDOWS define
```
asuessenbach: PLATFORM__WINDOWS ??

components/core/include/core/platform/context.hpp
```diff
@@ -0,0 +1,47 @@
...
+{
+  public:
+	PlatformContext()          = default;
+	virtual ~PlatformContext() = default;
+
+	virtual std::vector<std::string_view> arguments() const = 0;
+
+	virtual std::string_view external_storage_directory() const = 0;
+
+	virtual std::string_view temp_directory() const = 0;
```
asuessenbach: Why do you make those three functions pure virtual? Do you expect to have any platform that has that data in a different form?
And why do you return a `std::vector<std::string_view>`, instead of just a `const std::vector<std::string>&`? You're approach probably requires some preparation in every arguments() implementation (if the implementation stays in the derived classes).

asuessenbach (commented):
This PR mostly looks good, just a couple of minor issues and some questions.
And by the way, I am of the opinion that headers should be listed as part of the projects. Even if it's just to make them easily accessible from an IDE.

app/main.cpp
tomadamatkinson: I would say yes. But later I have completely removed the Platform class from existence  :)

app/main.cpp
tomadamatkinson: Same as above. Platform is not needed in the grand scheme of things. It is a mega class coupling many concepts of the framework together. I replace this by moving it to the Vulkan Samples CLI executable with plugins and the CLI

tomadamatkinson: I can add the else though :)

app/main.cpp
asuessenbach: Well, I'm not asking to put that stuff into some platform class. It's just some platform specific code that I would expect to be in a more central header so that every client/user needs to include just one header and doesn't need to think about handling every supported platform. Especially if one day we get another platform to be supported.

tomadamatkinson: Platform in this case represents the lifecycle of the Vulkan Samples CLI application.
- Initialize
- Main Loop
- Terminate

`CUSTOM_MAIN` facilitates hiding new platforms from the user. In later PRs we can remove the need for a platform class and rely solely on CUSTOM_MAIN which makes lines 24-32 redundant. This is why I am reluctant to move them elsewhere as they will soon become deprecated

tomadamatkinson: I have added the else statement though as a nicer message than a random set of compile errors

components/CMakeLists.txt
tomadamatkinson: It is not, removed :)

components/core/CMakeLists.txt
tomadamatkinson: It is not, removed :)

components/core/include/core/platform/context.hpp
tomadamatkinson: Agreed, replaced with std::string for simplicity

tomadamatkinson:
Will need to re-test this on the available platforms

components/core/include/core/platform/context.hpp
asuessenbach: Those functions are still pure virtual.
All three derived classes hold their data in the very same format and the three functions do the very same in each case. That is, you could move the data into the base class and make the functions non-virtual.
Besides that, I would return `const&` of the data, not a copy of the data.

tomadamatkinson: I think i made them pure virtual  due to android being the outlier with its `static std:;string` usage

I have kept them virtual just incase there are other platforms in the future which need to change a few things. Also moved to `const&` as suggested. Changes are inbound just building and testing now

asuessenbach (dismissed):

gary-sweet (dismissed):
LGTM

components/windows/include/windows/context.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <Windows.h>
+#include <core/platform/context.hpp>
+
+namespace vkb
+{
+/**
+ * @brief MS Windows platform context
+ *
+ * @warning Use in extreme circumstances with code guarded by the PLATFORM__WINDOWS define
```
SaschaWillems, line 30: Just curious, but what does "use in extreme circumstances..." mean?

components/windows/src/context.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+WindowsPlatformContext::WindowsPlatformContext(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR lpCmdLine, INT nCmdShow) :
+    PlatformContext{}
+{
+	_external_storage_directory = "";
+	_temp_directory             = get_temp_path_from_environment();
+	_arguments                  = get_args();
+
+	// allocate a console for this app
```
SaschaWillems, line 90: I guess this should todo can be removed?

SaschaWillems (dismissed):

components/android/include/android/context.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include <game-activity/native_app_glue/android_native_app_glue.h>
+
+namespace vkb
+{
+/**
+ * @brief MS Windows platform context
+ *
+ * @warning Use in extreme circumstances with code guarded by the PLATFORM__WINDOWS define
+ */
```
asuessenbach, line 33: MS Windows -> Android ?
PLATFORM__WINDOWS -> PLATFORM__ANDROID ?

framework/platform/platform.h
```diff
@@ -179,7 +175,7 @@ class Platform
 	std::vector<Plugin *> plugins;
 
 	/// Static so can be set via JNI code in android_platform.cpp
```
asuessenbach: It's not static anymore?

asuessenbach (commented):
Just two minor comment issues left.

components/android/include/android/context.hpp
tomadamatkinson: Fixed!

framework/platform/platform.h
tomadamatkinson: Fixed, but added a comment for the other static variables

components/windows/include/windows/context.hpp
tomadamatkinson: We don't want to pollute the code base with platform specific code, this is just a warning against that. I didn't spend too much time thinking about the comments as I am likely to do a more formal documentation pass later

PR #692, opened 2023-05-05
Pack assets into APK to fix Android 11 permission issues

r-kaczynski:
## Description

Vulkan Samples stores all it's assets and output files outside the app's directory. Android 11's permission update made this strategy no longer work.
My solution to this is to pack the assets with the APK and change where output files are located.
I've tested almost all API and performance samples on Windows, Android and Linux. I was not able to test the multi_draw_indirect sample on Linux as my device was not capable of running it.
The impact on build times is negligible (around 1% increase) and the APK's size goes up to 600MB and 550MB for debug and release respectively.
This PR is a proposal to change how assets are loaded on Android. This PR overlaps with https://github.com/KhronosGroup/Vulkan-Samples/pull/678, and we are happy for the team to take whichever solution you prefer
Fixes #646

## General Checklist:

Please ensure the following points are checked:

- [x] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [x] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
- [x] I have commented any added functions (in line with Doxygen)
- [x] I have commented any code that could be hard to understand
- [x] My changes do not add any new compiler warnings
- [x] My changes do not add any new validation layer errors or warnings
- [x] I have used existing framework/helper functions where possible
- [ ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [x] My changes do not add any regressions
- [x] I have tested every sample to ensure everything runs correctly
- [x] This PR describes the scope and expected impact of the changes I am making

## Sample Checklist

If your PR contains a new or modified sample, these further checks must be carried out *in addition* to the General Checklist:
- [x] I have tested the sample on at least one compliant Vulkan implementation
- [x] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [x] I have stated on what implementation the sample has been tested so that others can test on different implementations and platforms
- [x] Any dependent assets have been merged and published in downstream modules

PR #370, opened 2021-10-11
[Android] Check for external storage permissions

JoseEmilio-ARM:
## Description

Request additional storage permissions for Android 11, and legacy permissions for previous versions

Tested on Samsung Galaxy S20 (Android 10) and S21 (Android 11)

Fixes #369 

## General Checklist:

Please ensure the following points are checked:

- [x] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [x] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
- [n/a] I have commented any added functions (in line with Doxygen)
- [x] I have commented any code that could be hard to understand
- [x] My changes do not add any new compiler warnings
- [x] My changes do not add any new validation layer errors or warnings
- [x] I have used existing framework/helper functions where possible
- [x] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [x] My changes do not add any regressions
- [x] I have tested every sample to ensure everything runs correctly
- [x] This PR describes the scope and expected impact of the changes I am making

PR #713, opened 2023-06-07
Do not require access to external storage

### LINKED ISSUE #646: Android 11 Permissions

JoseEmilio-ARM:
## Description

Yet another option to fix the issue with permissions in the latest Android versions, see #678 and #692

Rather than directly running `adb push --sync` to the application's storage, which can cause permission issues, sync shaders and assets to temporary storage, then use `adb shell` to copy them to the application's location. This is all possible without requesting special external storage permissions, so the changes from #370 were reverted.

Fixes #646

## General Checklist:

- [n/a] I have used existing framework/helper functions where possible

### DIFF
--- a/app/android/AndroidManifest.xml
+++ b/app/android/AndroidManifest.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--
-- Copyright (c) 2019-2021, Arm Limited and Contributors
+- Copyright (c) 2019-2023, Arm Limited and Contributors
 -
 - SPDX-License-Identifier: Apache-2.0
 -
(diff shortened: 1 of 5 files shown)


SaschaWillems (approved):
This is the first Android permission fix that actually works for me. I can finally (again) run the samples on my Pixel 4a. So LGTM, and happy to merge as soon as possible :)

tomadamatkinson (approved):
Works locally for me!!

PR #245, opened 2021-01-27
enable performance warnings

samwalls-arm:
## Description

Allows debug callbacks for messages categorised as "perf" (performance warnings, Best Practices etc.)

This means that Best Practice warnings can be emitted from this vulkan application (without this change, they do not appear).

You can try to see performance warnings with the following:
```
export VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation
export VK_LAYER_SETTINGS_PATH=<path-to-my-txt>
```

vk_layer_settings.txt
```
khronos_validation.debug_action=VK_DBG_LAYER_ACTION_LOG_MSG
khronos_validation.report_flags=info,warn,perf,error
khronos_validation.log_filename=stdout
khronos_validation.enables=VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT,VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_ARM
```
(or the equivalent in Vulkan-Configurator)

Then run the desired sample.

I look forward to any feedback or advice about this.

Thanks,
Sam Walls

## General Checklist:

Please ensure the following points are checked:

- [x] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [x] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
- [ ] I have commented any added functions (in line with Doxygen)
- [ ] I have commented any code that could be hard to understand
- [ ] My changes do not add any new compiler warnings
- [ ] My changes do not add any new validation layer errors or warnings
- [ ] I have used existing framework/helper functions where possible
- [ ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [ ] My changes do not add any regressions
- [x] I have tested every sample to ensure everything runs correctly
- [x] This PR describes the scope and expected impact of the changes I am making

## Sample Checklist

If your PR contains a new or modified sample, these further checks must be carried out *in addition* to the General Checklist:
- [ ] I have tested the sample on at least one compliant Vulkan implementation
[...]

PR #737, opened 2023-07-01
Rework CMake validation layer options and add flags for best practices and sync validation

SaschaWillems:
## Description

This PR adds two new CMake options related to validation layers:

- `VKB_VALIDATION_LAYERS_BEST_PRACTICES`
Enables the best practices layer.
- `VKB_VALIDATION_LAYERS_SYNCHRONIZATION`
Enables the synchronization validation layer.

These flags let developers easily toggle what part of the validation layers are active without having to e.g. use vkconfig from SDK.

Enabling one of these flags also implicitly enables validation layers. This was also retrofitted for the already existing `VKB_VALIDATION_LAYERS_GPU_ASSISTED`. So as soon as one of these flags is set, validation layers are enabled. This makes them less confusing to use.

## General Checklist:


tomadamatkinson (dismissed):

JoseEmilio-ARM:
LGTM, but curious, how is this necessary now, #245 suggested these messages should already have been visible? Oh, it looks like it required the user enabling some settings externally, whereas this PR enables them programmatically 👍

JoseEmilio-ARM (dismissed):

framework/CMakeLists.txt
```diff
@@ -486,6 +486,10 @@ if(${VKB_VALIDATION_LAYERS_GPU_ASSISTED})
     endif()
 endif()
 
+if(${VKB_VALIDATION_LAYERS_BEST_PRACTICES})
```
gpx1000, line 490: Does it make sense to enforce validation layers to on a they are needed so seeing this also sets validation layers?

framework/core/hpp_instance.cpp
```diff
@@ -204,7 +204,7 @@ HPPInstance::HPPInstance(const std::string                            &applicati
 	enable_extension(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME, available_instance_extensions, enabled_extensions);
 #endif
 
-#if (defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)) && defined(VKB_VALIDATION_LAYERS_GPU_ASSISTED)
+#if (defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)) && (defined(VKB_VALIDATION_LAYERS_GPU_ASSISTED) || defined(VKB_VALIDATION_LAYERS_BEST_PRACTICES))
```
gpx1000, line 207: Think we might want to rethink the truth table on this one.  If you set validation best practices and it's release this would be true.

gpx1000 (changes requested):
Small things to consider on this one.  It's a good change and we need it.

bldsys/cmake/global_options.cmake
```diff
@@ -47,6 +47,7 @@ endif()
 set(VKB_WARNINGS_AS_ERRORS ON CACHE BOOL "Enable Warnings as Errors")
 set(VKB_VALIDATION_LAYERS OFF CACHE BOOL "Enable validation layers for every application.")
 set(VKB_VALIDATION_LAYERS_GPU_ASSISTED OFF CACHE BOOL "Enable GPU assisted validation layers for every application.")
+set(VKB_VALIDATION_LAYERS_BEST_PRACTICES OFF CACHE BOOL "Enable best practices validation layers for every application.")
```
asuessenbach: I think, enabling GPU_ASSISTED or BEST_PRACTICES should implicitly enable VALIDATION_LAYER.

docs/build.md
```diff
@@ -94,6 +94,16 @@ Enable GPU assisted Validation Layers, used primarily for VK_EXT_descriptor_inde
...
 
+**Note:** Only works with `VKB_VALIDATION_LAYERS` enabled
+
+### VKB_VALIDATION_LAYERS_BEST_PRACTICES
+
+Enable [best practices Validation Layers](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/best_practices.md).
+
+**Default:** `OFF`
+
+**Note:** Only works with `VKB_VALIDATION_LAYERS` enabled
```
asuessenbach: Instead of depending on VKB_VALIDATION_LAYERS, I would implicitly enable it.

framework/core/hpp_instance.cpp
asuessenbach: > If you set validation best practices and it's release this would be true.

Only, if VKB_VALIDATION_LAYERS is true. And in that case, you really want it, even with the release build.

If GPU_ASSISTED and BEST_PRACTICES would imply VALIDATION_LAYERS, we would just need to check
`#if (defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS))`

asuessenbach (commented):
While introducing an option for the BEST_PRACTICES layer, we could also introduce an option for SYNCHRONIZATION_VALIDATION ?
Which would make implicit enabling of VALIDATION_LAYER even more important.

SaschaWillems:
I deliberately didn't change how and when the general layer flag is enabled and just expanded the way that the GPU assisted layer works.

But sure, I could integrate all of your feedback and rescope the PR. Will take a look at that.

framework/core/hpp_instance.cpp
SaschaWillems: Yeah, I'd prefer to be able to enable validation even in release builds.

framework/CMakeLists.txt
SaschaWillems: Yes, that makes sense. I'll enable layers if one of those features will be enabled.

asuessenbach (approved):

gary-sweet (approved):

gpx1000 (approved):

PR #740, opened 2023-07-04
Add a color picker UI element

Krzysztof-Dmitruk-Mobica:
## Description

The color picker is available in ImGui, but not in the Drawer class, so I added it there. The color picker functionality is useful for the dynamic blending sample I'm working on now. I used the color write enable sample to show the color picker in action.

## General Checklist:

### DIFF
--- a/framework/gui.cpp
+++ b/framework/gui.cpp
@@ -1245,25 +1245,45 @@ void Drawer::text(const char *formatstr, ...)
 	va_end(args);
 }
 
-bool Drawer::color_picker(const char *caption, const char channel_count, float *color, uint16_t width, ImGuiColorEditFlags flags)
+bool Drawer::color_picker(const char *caption, std::array<float, 3> &color, float width, ImGuiColorEditFlags flags)
 {
-	assert((channel_count == 3 || channel_count == 4) && "Channel count value must be 3 or 4");
 	bool res;
-	if (width)
-		ImGui::PushItemWidth(width);
-	switch (channel_count)
-	{
-		case 3:
-			res = ImGui::ColorPicker3(caption, color, flags);
-			break;
-		case 4:
-			res = ImGui::ColorPicker4(caption, color, flags);
-			break;
-	}
+	ImGui::PushItemWidth(width);
+	res = ImGui::ColorPicker3(caption, color.data(), flags);
+	ImGui::PopItemWidth();
 	if (res)
-	{
 		dirty = true;
-	};
+	return res;
+}
+
+bool Drawer::color_picker(const char *caption, std::array<float, 4> &color, float width, ImGuiColorEditFlags flags)
+{
+	bool res;
+	ImGui::PushItemWidth(width);
+	res = ImGui::ColorPicker4(caption, color.data(), flags);
+	ImGui::PopItemWidth();
+	if (res)
+		dirty = true;
+	return res;
+}
+
+bool Drawer::color_edit(const char *caption, std::array<float, 3> &color, float width, ImGuiColorEditFlags flags)
+{
+	bool res;
+	ImGui::PushItemWidth(width);
+	res = ImGui::ColorEdit3(caption, color.data(), flags);
+	ImGui::PopItemWidth();
+	if (res)
+		dirty = true;
+	return res;
+}
+
+bool Drawer::color_edit(const char *caption, std::array<float, 4> &color, float width, ImGuiColorEditFlags flags)
+{
+	bool res;
+	res = ImGui::ColorEdit4(caption, color.data(), flags);
+	if (res)
+		dirty = true;
 	return res;
 }
 
--- a/framework/gui.cpp
+++ b/framework/gui.cpp
@@ -1245,4 +1245,26 @@ void Drawer::text(const char *formatstr, ...)
 	va_end(args);
 }
 
+bool Drawer::color_picker(const char *caption, const char channel_count, float *color, uint16_t width, ImGuiColorEditFlags flags)
+{
+	assert((channel_count == 3 || channel_count == 4) && "Channel count value must be 3 or 4");
+	bool res;
+	if (width)
+		ImGui::PushItemWidth(width);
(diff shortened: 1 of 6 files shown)


tomadamatkinson (dismissed):
LGTM! I think we should push to use more complex UI items like this. We should look at using Implot and possibly imnode in the future

SaschaWillems (dismissed):

framework/gui.cpp
```diff
@@ -1245,4 +1245,26 @@ void Drawer::text(const char *formatstr, ...)
 	va_end(args);
 }
 
+bool Drawer::color_picker(const char *caption, const char channel_count, float *color, uint16_t width, ImGuiColorEditFlags flags)
+{
+	assert((channel_count == 3 || channel_count == 4) && "Channel count value must be 3 or 4");
```
asuessenbach: If channel_count needs to be 3 or 4, you could use some bool parameter. Or maybe an enum with just two possible values.

To stop the separation of channel_count and color, and still let the compiler tell you when you're wrong, you could also change the interface to a template on channel_count, like this:
```
	template <size_t N>
	bool color_picker(std::string const &caption, std::array<float, N> &colors, uint16_t width = 0, ImGuiColorEditFlags flags = 0)
	{
		static_assert((N == 3) || (N == 4), "booh");
```

samples/extensions/color_write_enable/color_write_enable.cpp
```diff
@@ -539,15 +539,11 @@ void ColorWriteEnable::on_update_ui_overlay(vkb::Drawer &drawer)
...
-		if (drawer.slider_float("Red", &background_r_value, 0.0f, 1.0f))
-		{
-			build_command_buffers();
-		}
-		if (drawer.slider_float("Green", &background_g_value, 0.0f, 1.0f))
-		{
-			build_command_buffers();
-		}
-		if (drawer.slider_float("Blue", &background_b_value, 0.0f, 1.0f))
+		if (drawer.color_picker("", 3, background_color.data(), 200,
```
asuessenbach: With a width of 200 the color picker is a bit too narrow to display the selcted color value. Maybe use a width of 300, instead.

asuessenbach (commented):
Nice idea.
Would you mind to port the `color_picker` function to hpp_gui as well?

framework/gui.cpp
Krzysztof-Dmitruk-Mobica: I felt bad about making my method the first template in the Drawer class if it wasn't necessary. Instead, I created two overloaded functions for different sizes of the array. I think this solution is fine.

samples/extensions/color_write_enable/color_write_enable.cpp
Krzysztof-Dmitruk-Mobica: I changed it to 300. On my screen, 200 pixels looks fine. You can see it in the screenshot I added to the initial message. Maybe the problem is DPI.

framework/gui.cpp
asuessenbach: Well, instead of four functions, each one doing more or less the same, I would prefer a templatized approach, like this:
```
	enum class ColorOp
	{
		Edit,
		Pick
	};

	template <ColorOp OP, size_t N>
	bool color_op(const std::string & caption, std::array<float, N> &color, float width = 0.0f, ImGuiColorEditFlags flags = 0)
	{
		static_assert((N == 3) || (N == 4), "booh");

		ImGui::PushItemWidth(width);
		bool res = color_op_impl<OP, N>(caption.c_str(), color.data(), flags);
		ImGui::PopItemWidth();
		if (res)
			dirty = true;
		return res;
	}
```
With some private implementation functions (unfortunately, we're at C++14, not C++17, where we could use some `constexpr if` instead) like that:
```
	template <ColorOp OP, size_t N>
	bool color_op_impl(const char *caption, float *colors, ImGuiColorEditFlags flags)
	{
		assert(false);
	}

	template <>
	bool color_op_impl<ColorOp::Edit, 3>(const char *caption, float *colors, ImGuiColorEditFlags flags)
	{
		return ImGui::ColorEdit3(caption, colors, flags);
	}

	template <>
	bool color_op_impl<ColorOp::Edit, 4>(const char *caption, float *colors, ImGuiColorEditFlags flags)
	{
		return ImGui::ColorEdit4(caption, colors, flags);
	}

	template <>
	bool color_op_impl<ColorOp::Pick, 3>(const char *caption, float *colors, ImGuiColorEditFlags flags)
	{
		return ImGui::ColorPicker3(caption, colors, flags);
	}

	template <>
	bool color_op_impl<ColorOp::Pick, 4>(const char *caption, float *colors, ImGuiColorEditFlags flags)
	{
		return ImGui::ColorPicker4(caption, colors, flags);
	}
```
But, of course, your solution works as well, and if nobody else objects, I'd be ok as well.
(As there's no vulkan stuff with this set of functions, we could even move them out to some place, where they could be called by both, vkb::Drawer and vkb::HPPDrawer. But that might be a different story.)

Krzysztof-Dmitruk-Mobica: After giving it a second thought, I really liked your idea. I changed the code according to it.

gary-sweet:
> LGTM! I think we should push to use more complex UI items like this. We should look at using Implot and possibly imnode in the future

Just please be aware that not all platforms have very usable inputs for the samples. I have to drive the samples from the Linux console using the keyboard inputs. I'm not sure how well that will work for complex input UI items.

gary-sweet:
> > LGTM! I think we should push to use more complex UI items like this. We should look at using Implot and possibly imnode in the future
> 
> Just please be aware that not all platforms have very usable inputs for the samples. I have to drive the samples from the Linux console using the keyboard inputs. I'm not sure how well that will work for complex input UI items.

Turns out that the color picker can be driven from the keyboard, so no issue with this one at least.

asuessenbach (approved):

tomadamatkinson (approved):

PR #757, opened 2023-07-17
Add Tracy Profiler Support

tomadamatkinson:
## Description

Adds Tracy Profiler support. This will allow samples to show the stats and usage that they want to show to users without needing to clutter the Sample GUI

This can be extended in the future to instrument up core parts of the framework and samples that we want to draw the users attention too. It can also be used to show how multi-threading can be used with Vulkan

For now I have hooked this into our stats system so that we can show more stat information to users without worrying about cluttering the main GUI. We can still use the main sample GUI to show stats aswell!! I also added support for VMA Memory Budgets to show heap usage

We can use this to track other useful information such as the amount of vertices we expect to render for a given frame or the amount of geometry in the scene graph

Tracy also has Vulkan support however I have not played around with this yet

image

String names for VkMemoryTypeFlags are taking from vulkan.hpp instead of adding to the existing vkb::to_string() set

## General Checklist:

## Sample Checklist


framework/stats/stats.cpp
```diff
@@ -279,6 +287,154 @@ void Stats::push_sample(const StatsProvider::Counters &sample)
...
+		{
+			continue;
+		}
+
+		float average = 0.0f;
+		for (auto &v : c.second)
+		{
+			average += v;
+		}
+		average /= c.second.size();
```
asuessenbach, line 395: Maybe: `float average = std::accumulate( c.second.begin(), c.second.end(), 0.0f ) / c.second.size();`

asuessenbach (commented):
Is the concurrency stuff in the *.yml files related to adding tracy?

Besides that, I'm not familiar with tracy. How would I get that nice little tracy window?

tomadamatkinson:
> Is the concurrency stuff in the *.yml files related to adding tracy?
> 
> 
> 
> Besides that, I'm not familiar with tracy. How would I get that nice little tracy window?

The concurrency stuff isn't related, I can make a small PR for these changes if needed. Just threw it in here as I had lots of jobs running and not automatically cancelling outdated ones.

To use Tracy you should be able to run './scripts/tracy.py' and it will build and run the window. I have only tested this on Linux and you do need to install system libraries.

I can give it a go on windows at some point this week and make sure it works. It is likely that it won't work first time

asuessenbach:
> It is likely that it won't work first time

And you're right, tracy on windows is not supported.

> The concurrency stuff isn't related,

You can keep it in, if you like. I just wanted to make that I didn't miss something here.

SaschaWillems:
> > It is likely that it won't work first time
> 
> And you're right, tracy on windows is not supported.
> 

I'm pretty sure tracy works on Windows. I'm using it in one of my projects and I mainly develop on windows ;)

tomadamatkinson:
Tracy is supported. It is likely my script doesn't work on windows - Yet!

The script was meant as a build helper as you need the same Tracy version as TracyClient version used in the app.

I've only had chance to get this helper working on Linux but i should have some free time tonight to test it out on windows.

The latest Tracy download from GitHub should also work (again not tested this as I was focused on the script)

gary-sweet:
The build script for Tracy doesn't work for me in Linux sadly:

```
../../../server/TracySourceView.cpp: In member function ‘bool tracy::SourceView::Disassemble(uint64_t, const tracy::Worker&)’:
../../../server/TracySourceView.cpp:736:46: error: ‘CS_GRP_BRANCH_RELATIVE’ was not declared in this scope
  736 |                 else if( detail.groups[j] == CS_GRP_BRANCH_RELATIVE && opType < OpType::Branch ) opType = OpType::Branch;
      |                                              ^~~~~~~~~~~~~~~~~~~~~~
../../../server/TracySourceView.cpp:739:46: error: ‘CS_GRP_PRIVILEGE’ was not declared in this scope
  739 |                 else if( detail.groups[j] == CS_GRP_PRIVILEGE && opType < OpType::Privileged )
      |                                              ^~~~~~~~~~~~~~~~
```

asuessenbach:
> The latest Tracy download from GitHub should also work (again not tested this as I was focused on the script)

Not exactly. I get this error:
image

tomadamatkinson:
@asuessenbach sorry for the delay on this. Each time I have thought I would have time I've had a distraction

I'm going to pin the submodule version to v9.1.0 https://github.com/wolfpld/tracy/tree/v0.9.1

This would make the versions compatible. I think I am currently on v.9.2.0 (using main)

Will also try the Windows build now with the script

@gary-sweet what is the version of your capstone library? I believe that build issue is coming from a missing dependency. The first printout at the top of the script is some install dependencies. It may work after building those

tomadamatkinson:
Wow, the Windows build is much more involved. Pinning the Tracy version to v0.9.1 makes our client compatible with the profiler

Added some download instructions to the script 

image

I would download automatically but as the release page packages Tracy in a 7zip archive I feel like it is best to let the user do this

After rebuilding samples I get this

image

framework/stats/stats.cpp
tomadamatkinson: Nice, yeah a bit more concise

tomadamatkinson:
Moved concurrency changes to another PR

gary-sweet:
> @gary-sweet what is the version of your capstone library? I believe that build issue is coming from a missing dependency. The first printout at the top of the script is some install dependencies. It may work after building those

```
libcapstone-dev:
  Installed: 4.0.1+really+3.0.5-1build1
```
 I literally just used the "sudo apt install libglfw3-dev libcapstone-dev libdbus-glib-1-dev libfreetype-dev" instruction from the script.

gary-sweet:
It's a bit weird though. The capstone change log would indicate that CS_GRP_BRANCH_RELATIVE was introduced in version 4.0. Grepping for that in /usr/include/capstone (which is where the build is looking) yields nothing however, despite Ubuntu claiming to have version 4.0.1.

tomadamatkinson:
Locally I have

```
libcapstone-dev/jammy,jammy,now 4.0.2-5 amd64 [installed]
libcapstone4/jammy,jammy,now 4.0.2-5 amd64 [installed,automatic]
```

Interesting how your lib says `4.0.1+really+3.0.5-1build1` as in its 4.0.1 but actually 3.0.5? What OS are you using?

gary-sweet:
> Interesting how your lib says `4.0.1+really+3.0.5-1build1` as in its 4.0.1 but actually 3.0.5? What OS are you using?

It's Ubuntu 20.04

tomadamatkinson:
Looking at making an Ubuntu docker build to export a usable executable. Here are the system dependencies for each platform as stated in the Tracy docs. I think the latest version of Tracy has revisions to the build system to make this work better but in general the build process is very much system specific.

image

Will post an update later

framework/stats/stats.cpp
```diff
@@ -279,6 +288,149 @@ void Stats::push_sample(const StatsProvider::Counters &sample)
...
+
+	// We know that we will only ever have one device in the system, so we can cache the labels
+	if (labels.size() == 0)
+	{
+		VkPhysicalDeviceMemoryProperties memory_properties;
+		vkGetPhysicalDeviceMemoryProperties(device.get_gpu().get_handle(), &memory_properties);
+
+		labels.reserve(memory_properties.memoryHeapCount);
+
+		for (size_t heap = 0; heap < memory_properties.memoryHeapCount; heap++)
```
asuessenbach, line 419: Why don't you just use the flags from the memoryHeaps?

framework/stats/stats.cpp
```diff
@@ -279,6 +288,149 @@ void Stats::push_sample(const StatsProvider::Counters &sample)
...
+			for (size_t type = 0; type < memory_properties.memoryTypeCount; type++)
+			{
+				if (memory_properties.memoryTypes[type].heapIndex == heap)
+				{
+					flags |= memory_properties.memoryTypes[type].propertyFlags;
+				}
+			}
+
+			// TODO: Using VulkanHPP to_string here instead of vkb::to_string. After deciding on the framework direction we should adjust this
+			std::stringstream ss;
```
asuessenbach: You don't have to use a `std::stringstream`:
`labels.push_back("Heap " + std::to_string(heap) + " " + vk::to_string(vk::MemoryPropertyFlags{flags}));
`

asuessenbach (commented):
Works nice now with Tracy 0.9.1 on Win10!

SaschaWillems:
Support for tracy would be great addition 👍 

But on windows you'll get a firewall notification when running the samples (at least for the first time), which may irritate users.

image

So I'm not sure if we should always enable it by default, or if it's better to hide it behind a build time or runtime flag.

SaschaWillems:
> I would download automatically but as the release page packages Tracy in a 7zip archive I feel like it is best to let the user do this

Can we do that in the python script? There should be [7z libraries](https://github.com/miurahr/py7zr) for python, so maybe we can make use of them to unpack the Windows version.

tomadamatkinson:
I'd be happy to add that if you believe it would make the process easier! It's a little frustrating that Tracy doesn't have a cmake build for their profiler

Happy to hide this behind a VKB_ENABLE_TRACY=ON flag

SaschaWillems (changes requested):
The android build fails for me with hundreds of errors from VMA:
```
V:/Vulkan-KHR-Samples/Vulkan-Samples/third_party/vma/include/vk_mem_alloc.h:2767:12: error: pointer is missing a nullability type specifier (_Nonnull, _Nullable, or _Null_unspecified) [-Werror,-Wnullability-completeness]
  static void* vma_aligned_alloc(size_t alignment, size_t size)
             ^
  V:/Vulkan-KHR-Samples/Vulkan-Samples/third_party/vma/include/vk_mem_alloc.h:2767:12: note: insert '_Nullable' if the pointer may be null
  static void* vma_aligned_alloc(size_t alignment, size_t size)
             ^
               _Nullable
  V:/Vulkan-KHR-Samples/Vulkan-Samples/third_party/vma/include/vk_mem_alloc.h:2767:12: note: insert '_Nonnull' if the pointer should never be null
  static void* vma_aligned_alloc(size_t alignment, size_t size)
             ^
               _Nonnull
```

Btw. I think we also need to adjust our documentation so people know that we have tracy support.

SaschaWillems:
Last time I checked on Windows this worked fine with Tracy 0.9.1, but if I try to connect now I get an error a protocol mismatch error? Did anything change recently?
image

asuessenbach (dismissed):

components/core/include/core/util/profiling.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+	static Plot *get_instance()
+	{
+		static_assert((std::is_same<T, int64_t>::value || std::is_same<T, double>::value || std::is_same<T, float>::value), "PlotStore only supports int64_t, double and float");
+		static Plot instance;
+		return &instance;
+	}
+
+	std::unordered_map<const char *, T> plots;
+};
```
asuessenbach, line 130: With `TRACY_ENABLE` defined, `get_instance` and `plots` are private, otherwise they are public?

asuessenbach (commented):
It compiles fine, both with and without VKB_ENABLE_TRACY defined.
But I get the same protocol mismatch error as Sascha got three weeks ago.

framework/stats/stats.cpp
```diff
@@ -17,10 +17,19 @@
...
 
 #include "stats/stats.h"
-#include "core/device.h"
 
+#include <common/error.h>
+#include <core/util/profiling.hpp>
+
+VKBP_DISABLE_WARNINGS()
+#include <vk_mem_alloc.h>
+#include <vulkan/vulkan.hpp>
```
asuessenbach, line 23: What warnings from vulkan.hpp do you want to hide here?

tomadamatkinson:
Picked this branch back up as the Tracy support does work. This branch got stalled on that fact that Tracy is hard to build for different platforms - this is a known limitation of the project.

I have created [this pr](https://github.com/wolfpld/tracy/pull/707) in an attempt to help improve Tracy in this area. Unfortunately you will not be able to use this branch to build Tracy due to the strict protocol constraints. If you would like to test this PR you will need to install Tracy from the release page or by following the Tracy.pdf guide.

If you are using version of ubuntu that does not include capstone 4+ you will need to manually build and install capstone 4+ in your distro for Tracy to build.

`VKB_PROFILING` was added to make Tracy's usage opt in. By default this option is OFF

For now I have only instrumented up the GLTF Loader. We do not need to instrument all functions with Tracy. It is useful to track the performance of high level systems such as scene iterations or multi threaded scenarios.

I have tested on Ubuntu 23.04, Windows 11 and Android. Tracy is disabled for Android for now as there are some networking considerations

(I will be addressing any CI issues over the next few days, There are likely going to be some nuances between actual machines and github runners)

gpx1000 (dismissed):

asuessenbach:
Unfortunately, I get an "Incompatible protocol" error, again:
image

I just downloaded the release v0.10 from https://github.com/wolfpld/tracy/tags, started the Tracy.exe and got this.

framework/common/error.h
```diff
@@ -21,39 +21,12 @@
 #include <stdexcept>
 #include <string>
 
+#include "core/util/error.hpp"
```
asuessenbach, line 24: Why do you separate error.hpp for the other headers?

framework/gltf_loader.cpp
```diff
@@ -29,6 +29,8 @@ VKBP_DISABLE_WARNINGS()
 #include <glm/gtc/type_ptr.hpp>
 VKBP_ENABLE_WARNINGS()
 
+#include <core/util/profiling.hpp>
```
asuessenbach, line 30: Why do you separate profiling.hpp from the other headers below?

tomadamatkinson:
Hey @asuessenbach, this branch is pinned to v0.10 can you confirm that you updated your submodules before building?

Before this branch was v0.9.1 so it may be that you still have that version checked out

SaschaWillems:
Works fine for me 👍🏻. Will try to do a proper review this weekend.

asuessenbach:
@tomadamatkinson You're right, I somehow missed to get the check out the right version. Works now!

SaschaWillems (dismissed):

SaschaWillems:
@tomadamatkinson: Could you fix the the merge conflicts? Would' love to see this merged, so we can add additional functionality based on your work. If you can't find the time, I'll happily try to help :)

tomadamatkinson:
@SaschaWillems ill take a look!

tomadamatkinson:
image

Rebased and tested on linux. To work with android devices i believe we need to port forward with adb and also give the application local network permissions. I have not tested this in this PR. On desktop this should work fine out of the box.

Works with Tracy version 0.10. When Tracy has a new official release with the CMake build system i helped with we can bump the submodule and point to that allowing users to seamlessly build the profiler.

I cannot test on Windows or Mac today but may be able too in the near future

@SaschaWillems let me know if there are any tweaks you want adding to this

SaschaWillems (approved):
Thank you very much. Worked out of the box for me on Windows with the Tracy 0.10 binary release 👍🏻

gpx1000 (approved):
Awesome, works here.

gary-sweet (approved):
Doesn't break anything for me

marty-johnson59:
Merging - 3 approvals

PR #774, opened 2023-08-08
Validation Layer Fixes

tomadamatkinson:
## Description

Fixes Validation Layer issues where samples will crash from missing validation layers

Tested on windows

## General Checklist:

## Sample Checklist

- [x] For new samples, I have added a tutorial README.md file to guide users through what they need to know to implement code using this feature. For example, see [conditional_rendering](./../samples/extensions/conditional_rendering)

### DIFF
--- a/framework/core/hpp_instance.cpp
+++ b/framework/core/hpp_instance.cpp
@@ -21,11 +21,23 @@
 #include <core/hpp_physical_device.h>
 #include <volk.h>
 
+#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS) || (defined(VKB_VALIDATION_LAYERS_GPU_ASSISTED) || defined(VKB_VALIDATION_LAYERS_BEST_PRACTICES))
+#	define USE_VALIDATION_LAYERS 1
+#endif
+
+#if defined(USE_VALIDATION_LAYERS) && (defined(VKB_VALIDATION_LAYERS_GPU_ASSISTED) || defined(VKB_VALIDATION_LAYERS_BEST_PRACTICES) || defined(VKB_VALIDATION_LAYERS_SYNCHRONIZATION))
+#	define USE_VALIDATION_LAYER_FEATURES 1
+#endif
+
+#ifdef USE_VALIDATION_LAYERS
+#	define USE_VULKAN_LOGGER
+#endif
+
 namespace vkb
 {
 namespace
 {
-#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
+#ifdef USE_VULKAN_LOGGER
 VKAPI_ATTR VkBool32 VKAPI_CALL debug_utils_messenger_callback(VkDebugUtilsMessageSeverityFlagBitsEXT message_severity, VkDebugUtilsMessageTypeFlagsEXT message_type,
                                                               const VkDebugUtilsMessengerCallbackDataEXT *callback_data,
                                                               void                                       *user_data)
@@ -181,7 +193,7 @@ HPPInstance::HPPInstance(const std::string                            &applicati
 {
 	std::vector<vk::ExtensionProperties> available_instance_extensions = vk::enumerateInstanceExtensionProperties();
 
-#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
+#ifdef USE_VULKAN_LOGGER
 	// Check if VK_EXT_debug_utils is supported, which supersedes VK_EXT_Debug_Report
 	const bool has_debug_utils  = enable_extension(VK_EXT_DEBUG_UTILS_EXTENSION_NAME,
 	                                               available_instance_extensions, enabled_extensions);
@@ -204,7 +216,7 @@ HPPInstance::HPPInstance(const std::string                            &applicati
 	enable_extension(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME, available_instance_extensions, enabled_extensions);
 #endif
 
-#if (defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)) && (defined(VKB_VALIDATION_LAYERS_GPU_ASSISTED) || defined(VKB_VALIDATION_LAYERS_BEST_PRACTICES))
+#ifdef USE_VALIDATION_LAYER_FEATURES
 	bool validation_features = false;
 	{
 		std::vector<vk::ExtensionProperties> available_layer_instance_extensions = vk::enumerateInstanceExtensionProperties(std::string("VK_LAYER_KHRONOS_validation"));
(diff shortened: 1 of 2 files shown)


asuessenbach:
I think, for a "hotfix", you would just need to replace
`#ifdef VKB_VALIDATION_LAYERS`
with
`#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)`
on line 273 in hpp_instance.cpp and on line 285 in instance.cpp.
For the "correct" handling (having VKB_VALIDATION_LAYERS not implicitly enabled with VKB_DEBUG) @SaschaWillems volunteered to work that out.

tomadamatkinson:
This PR does that. The only extra change is moving the logic to a single place at the top of the source file so that it is easier to tweak and reason with.

When you have the more complex OR and AND logic in multiple places throughout the file it is easier to introduce bugs by accident

framework/core/hpp_instance.cpp
```diff
@@ -21,11 +21,23 @@
 #include <core/hpp_physical_device.h>
 #include <volk.h>
 
+#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS) || (defined(VKB_VALIDATION_LAYERS_GPU_ASSISTED) || defined(VKB_VALIDATION_LAYERS_BEST_PRACTICES))
```
asuessenbach: With #737, the VKB_VALIDATION_LAYERS_* implicitly enable VKB_VALIDATION_LAYERS.
So this condition (here and in instance.cpp) could be simplified. (Otherwise, it would be missing the check on VKB_VALIDATION_LAYERS_SYNCHRONIZATION)

SaschaWillems: Exactly. There is no need to check for the additional validation flags here.

SaschaWillems (changes requested):
See Andreas' comments. This PR should be simplified.

framework/core/hpp_instance.cpp
tomadamatkinson: Sorry just got back onto samples. Fixed this. Thank you for the suggestion, forgot that the layers are now implicitly enabled

framework/core/hpp_instance.cpp
```diff
@@ -21,11 +21,23 @@
...
+#ifdef USE_VALIDATION_LAYERS
+#	define USE_VULKAN_LOGGER
+#endif
+
 namespace vkb
 {
 namespace
 {
-#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
+#ifdef USE_VULKAN_LOGGER
```
gpx1000: Do we ever envision this being enabled independently of USE_VALIDATION_LAYERS?  If not, then maybe let's just use USe_VALIDATION_LAYERS as the test here.  Makes it slightly less complex and doesn't create a possibility of someone accidentally enabling code that might not work independent of other macros.

framework/core/instance.cpp
```diff
@@ -22,11 +22,23 @@
...
+#	define USE_VALIDATION_LAYERS 1
+#endif
+
+#if defined(USE_VALIDATION_LAYERS) && (defined(VKB_VALIDATION_LAYERS_GPU_ASSISTED) || defined(VKB_VALIDATION_LAYERS_BEST_PRACTICES) || defined(VKB_VALIDATION_LAYERS_SYNCHRONIZATION))
+#	define USE_VALIDATION_LAYER_FEATURES 1
+#endif
+
+#ifdef USE_VALIDATION_LAYERS
+#	define USE_VULKAN_LOGGER
+#endif
```
gpx1000, line 31: Do we want to centralize this logic instead of copying it to two places?

gpx1000 (changes requested):

framework/core/hpp_instance.cpp
tomadamatkinson: I doubt we ever will. However, having `USE_VULKAN_LOGGER` highlights where code is more specifically logger related and not validation layer related. After all, the logger can be used by other layers to communicate with developers. The logger macro should work independently to `USE_VALIDATION_LAYERS`, right now we just didn't need that functionality in order to fix the bug on main

framework/core/instance.cpp
tomadamatkinson: I am reluctant to add more headers that would couple the CPP and C frameworks together. Later I imagine we will chose one framework over the other. These macros are less likely to change than the duplicated code in both frameworks and so there is relatively less risk keeping them duplicated as there is making a framework change.

For now I would like to keep them copied

gpx1000: fair enough on both accounts.

gpx1000 (dismissed):

framework/core/hpp_instance.cpp
```diff
@@ -21,11 +21,23 @@
...
+#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
+#	define USE_VALIDATION_LAYERS 1
+#endif
+
+#if defined(USE_VALIDATION_LAYERS) && (defined(VKB_VALIDATION_LAYERS_GPU_ASSISTED) || defined(VKB_VALIDATION_LAYERS_BEST_PRACTICES) || defined(VKB_VALIDATION_LAYERS_SYNCHRONIZATION))
+#	define USE_VALIDATION_LAYER_FEATURES 1
+#endif
+
+#ifdef USE_VALIDATION_LAYERS
+#	define USE_VULKAN_LOGGER
```
SaschaWillems: Is there a situation where you can enable `USE_VULKAN_LOGGER` without `USE_VALIDATION_LAYERS`? If not, do we need `USE_VULKAN_LOGGER` at all?

tomadamatkinson: See my above reply to @gpx1000 😄 

Although our use-case is tied to the validation layers, i didn't think it was a requirement that the loggers can only be used with the validation layers?

I was under the impression that any layer can hook into the logger and use it to communicate messages from a layer to an application? If that is the case, then there is value keeping this separate to demonstrate that the logger code is not exclusive to the layers

SaschaWillems: Sorry, didn't saw that comment "thanks" to github's horrible UI.

In that case it makes sense for future use cases.

But I kinda fear that it may confuse people. A define that simply defines another define adds another layer, making code a bit less clear IMO. 

I'm fine having a dedicated logger define, but I think everything that's related to validation, which IMO also includes the debug callback, should be toggled by a define that tells people looking at the code that this is somehow related to validation.

asuessenbach (dismissed):
Two of your new defines are defined to be a specific value, while the third one is just defined.
Is that for any reason?
Defining to a specific value seems to imply, that that define could under different circumstances also be defined to a different value, which is not the case here.

framework/core/hpp_instance.cpp
tomadamatkinson: @SaschaWillems is there a specific change that you would like to see here? I can address this at the same time as @asuessenbach's latest comment

SaschaWillems: Since there is no explicit use-case for `USE_VULKAN_LOGGER` we may as well remove that and keep it in mind for further discussions on that topic.

But tbh. the most important part is getting this merged with so many related issues coming up.

SaschaWillems (approved):

gpx1000 (approved):

tomadamatkinson:
Merging on two as discussed in the Working Group call

PR #780, opened 2023-08-17
Vk ext shader object optional layer

gpx1000:
## Description

This fixes the remaining issues in the newly accepted/merged VK_EXT_shader_object PR.


SaschaWillems (dismissed):
LGTM

.github/workflows/check.yml
```diff
@@ -75,7 +75,15 @@ jobs:
...
-      run: echo "line-count=$(echo "${{ steps.clang-diff.outputs.changes }}" | grep -c +++)" >> $GITHUB_OUTPUT
+      run: |
+        echo "line-count=$(echo "${{ steps.clang-diff.outputs.changes }}" | grep -c +++)" >> $GITHUB_OUTPUT
+        echo "${{ steps.clang-diff.outputs.changes }}" >> patch.txt
+    - if: ${{ steps.count-diff.outputs.line-count > 0 }}
+      uses: actions/upload-artifact@v3
+      with:
+        name: diff-output
+        path: patch.txt
+
```
asuessenbach: Are these changes related to this PR?

samples/extensions/shader_object/shader_object.cpp
```diff
@@ -230,6 +230,11 @@ void ShaderObject::setup_framebuffer()
 // Create render pass for UI drawing
 void ShaderObject::setup_render_pass()
 {
+	// delete existing render pass
+	if (render_pass != VK_NULL_HANDLE)
```
asuessenbach, line 235: Is `setup_render_pass` called multiple times?
If so, you would need to initialize `render_pass` with `VK_NULL_HANDLE` to make this check work in debug mode. If not, you don't need to destroy the render pass here.

asuessenbach (commented):
This PR works.
But instead of improving the somewhat hacky usage of `VulkanSample::get_validation_layers`, I would completely remove that function, and instead introduce some `VulkanSample::add_instance_layer` (or maybe just `VulkanSample::add_layer`, as there's nothing like a device layer), similar to `VulkanSample::add_instance_extension`, and handle it analogously.

samples/extensions/shader_object/shader_object.cpp
marty-johnson59: Hi @asuessenbach , per discussion today, can  you open an issue so we don't lose the context?  We're going to merge this one soon w/o capturing this - but want to circle back later

asuessenbach: Sorry, I missed the meeting where this was discussed...
Why should this issue not be resolved right here?

SaschaWillems: The idea is to fix this after this PR has been merged. The PR is kinda urgent, and the line you highlighted is not new to the PR but already present in the base sample.

asuessenbach: @SaschaWillems so it seems, you know about the issue to be resolved later. Would you be so kind and file that issue, instead of me guessing what's meant to be done here?

SaschaWillems: Sorry, I was under the impression that this change was already present beforehand. But as this code is new to this PR you're right to mention this and it indeed needs fixing in this PR.

gpx1000: I did this change because renderpass was being created by the vulkansample in the framework automatically.  I can take this change out, but then validation layers correctly identifies a render_pass object not getting released.  The rest of the changes are updated to conform with your requests.

SaschaWillems: That makes sense. We should keep that change then :)

SaschaWillems (dismissed):

framework/core/hpp_instance.cpp
```diff
@@ -74,6 +74,41 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL debug_callback(VkDebugReportFlagsEXT flags
 }
 #endif
 
+bool validate_layers(std::unordered_map<const char *, bool> &required,
```
asuessenbach: The `required` layers here actually are not required (they're allowed to be optional), but requested. Maybe rename the argument?

framework/core/hpp_instance.cpp
```diff
@@ -74,6 +74,41 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL debug_callback(VkDebugReportFlagsEXT flags
...
+		for (auto &available_layer : available)
+		{
+			if (strcmp(available_layer.layerName, layer.first) == 0)
+			{
+				found = true;
+				break;
+			}
+		}
+
+		if (!found)
```
asuessenbach: Maybe we can replace the loop over the `available`s with this check:
`		if (std::none_of(available.begin(), available.end(), [&layer](vk::LayerProperties const &lp) { return lp.layerName == layer.first; }))
`

framework/core/hpp_instance.cpp
```diff
@@ -74,6 +74,41 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL debug_callback(VkDebugReportFlagsEXT flags
...
+	for (auto &rem : remove_vec)
+	{
+		required.erase(rem);
+	}
+	return true;
+}
+
 bool validate_layers(const std::vector<const char *>        &required,
                      const std::vector<vk::LayerProperties> &available)
 {
```
asuessenbach, line 105: In case, you want to change a little more, this version of `validate_layers` could look like this:
```
	auto requiredButNotAvailableIt =
	    std::find_if(required.begin(),
	                 required.end(),
	                 [&available](char const *layer) {
		                 return std::none_of(available.begin(), available.end(), [&layer](vk::LayerProperties const &lp) { return lp.layerName == layer; });
	                 });
	if (requiredButNotAvailableIt != required.end())
	{
		LOGE("Validation Layer {} not found", *requiredButNotAvailableIt);
		return false;
	}
	return true;

```

framework/core/hpp_instance.cpp
```diff
@@ -299,6 +335,20 @@ HPPInstance::HPPInstance(const std::string                            &applicati
 		throw std::runtime_error("Required validation layers are missing.");
 	}
 
+	std::unordered_map<const char *, bool> layers = (std::unordered_map<const char *, bool>) (requested_layers);
```
asuessenbach: No.
Never cast the `const` away.
If you would handle the `requested_layers` the same way as the `required_extensions` (which should be named `requested_extensions`!), you would not even need the new `validate_layers` version.
And, by the way, the `layers` are never used here!

framework/core/instance.cpp
```diff
@@ -76,6 +76,41 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL debug_callback(VkDebugReportFlagsEXT flags
 }
 #endif
 
+bool validate_layers(std::unordered_map<const char *, bool> &required,
+                     const std::vector<VkLayerProperties>   &available)
```
asuessenbach, line 80: The same comments as for hpp_instance.cpp hold here.

samples/extensions/shader_object/README.adoc
```diff
@@ -436,13 +436,15 @@ The layer can be shipped with your application, and it will disable itself if a
 The emulation layer can be enabled by adding `VK_LAYER_KHRONOS_shader_object` to `ppEnabledLayerNames` in `VkDeviceCreateInfo`.
 
 The sample framework already has an existing abstraction normally used for enabling the validation layer.
-This sample repurposes this mechanism to instead load the emulation layer.
+This
```
asuessenbach: Newline here?

Besides that, is that description still valid? We now have some `add_layer` and `get_layers`. None of them explicitly mentions validation layers.

samples/extensions/shader_object/README.adoc
```diff
@@ -436,13 +436,15 @@ The layer can be shipped with your application, and it will disable itself if a
...
+sample repurposes this mechanism to instead load the emulation layer. NB: the layer is set to be optional as that's
+what the false does.
 
 [,CPP]
 ----
-const std::vector<const char *> ShaderObject::get_validation_layers()
+const std::unordered_map<const char *, bool> ShaderObject::get_validation_layers()
```
asuessenbach, line 442: There's no `ShaderObject::get_validation_layers` anymore.

samples/extensions/shader_object/shader_object.cpp
```diff
@@ -415,6 +416,10 @@ void ShaderObject::load_assets()
 
 	VkSamplerCreateInfo sampler_create_info = vkb::initializers::sampler_create_info();
 
+	// destroy created sampler before re-creating
+	vkDestroySampler(get_device().get_handle(), heightmap_texture.sampler, nullptr);
+	vkDestroySampler(get_device().get_handle(), terrain_array_textures.sampler, nullptr);
+
 	// Setup a mirroring sampler for the height map
```
gpx1000, line 425: looks like the calling load_assets multiple times bug has been fixed.

PR #787, opened 2023-08-26
Antora

gpx1000:
## Description

This PR enables gathering all images and ascidoc files in the correct place for antora to find it.


antora/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+    endforeach ()
+
+    foreach(image ${IMAGES_R})
+        file(RELATIVE_PATH relpage ${CMAKE_SOURCE_DIR} ${image})
+        get_filename_component(directory ${relpage} DIRECTORY)
+        file(COPY ${image} DESTINATION ${CMAKE_CURRENT_LIST_DIR}/modules/ROOT/images/${directory})
+    endforeach ()
+endfunction()
+
+if(CMAKE_BUILD_TYPE MATCHES "Debug")
```
tomadamatkinson: We should use a flag closer to `VKB_GENERATE_ANTORA_SITE=ON`. Gathering resources doesn't need to be specific to a Debug build and can be run in a separate CI task specific for Antora

oddhack:
@gpx1000 should I go ahead and merge this now, per our call this afternoon? It's got a pending-review flag on it and I don't know what the Samples process is.

BTW the Antora nav.adoc is a kind of general nuisance when trying to maintain both Antora and regular asciidoctor builds. For the Vulkan spec I may refactor the top-level vkspec.adoc a bit to make it possible to extract nav.adoc from it with a script. Otherwise MRs will add / rearrange chapters in vkspec.adoc and it will be very easy to overlook the corresponding update to nav.adoc.

gpx1000:
Let's hold off on merging this.  We'll merge it in our Samples call on Monday (@marty-johnson59 I think it's on the Monday call list correct?)

antora/CMakeLists.txt
gpx1000: This has been addressed.

SaschaWillems (approved):

tomadamatkinson (approved):

PR #794, opened 2023-09-05
Add an overwrite prompt to the generate sample script

Krzysztof-Dmitruk-Mobica:
If the directory with the sample exists, generate.py will prompt whether to overwrite it. The story behind this PR is a painful one.

### DIFF
--- a/scripts/generate.py
+++ b/scripts/generate.py
@@ -117,6 +117,24 @@ def generate_sample_help(subparsers, template):
         default=None,
     )
 
+def prompt_if_path_exists(output_dir, name):
+    path = os.path.join(output_dir, name)
+
+    if os.path.exists(path):
+        yes = {'yes', 'y'}
+        no = {'no', 'n', ''}
+        print(terminal_colors.WARNING
+              + f"The output directory {path} is not empty. Would you like to overwrite its content with the sample template? [y/N]"
+              + terminal_colors.END)
+        while True:
+            choice = input().lower()
+            if choice in yes:
+                return True
+            elif choice in no:
+                return False
+            else:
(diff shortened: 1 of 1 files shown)


scripts/generate.py
```diff
@@ -117,6 +117,24 @@ def generate_sample_help(subparsers, template):
...
     )
 
+def prompt_if_path_exists(output_dir, name):
+    path = os.path.join(output_dir, name)
+
+    if os.path.exists(path):
+        yes = {'yes', 'y'}
+        no = {'no', 'n', ''}
+        print(terminal_colors.WARNING
+              + f"The output directory {path} is not empty. Would you like to overwrite its content with the sample template? [y/N]"
```
gary-sweet, line 127: I assume the capitalized 'N' is meant to indicate that this is the default response if you just hit return, but that doesn't appear to be the behaviour of the script; it will print a message and wait for another input.

Krzysztof-Dmitruk-Mobica: Unfortunately, I was unable to reproduce the error. I tested it on:
Ubuntu 23.04,
Python 3.11.4,
zsh 5.9 (not sure if this one is important).
My colleague, whom I asked to test it on Windows, also couldn't reproduce this error. What system are you testing it on?

SaschaWillems (approved):
Works fine for me. Tested with the Windows Terminal (Win11) and Ubuntu 22.04 (via WSL).

scripts/generate.py
gary-sweet: Sorry - my bad. I completely missed the '' in the 'no' list which makes my review incorrect.

gary-sweet (approved):

PR #827, opened 2023-10-14
[sample] Order-independent transparency using per-pixel linked lists

mdagois:
* Added a new sample that implements an order-independent transparency algorithm using per-pixel linked lists
* Added a new parameter, _instance_count_, to ApiVulkanSample::draw_model (defaults to 1)
* Fixed an issue forcing samples to enable anisotropic sampling when using ApiVulkanSample::load_texture

- [ ] ~I have tested every sample to ensure everything runs correctly~
- [X] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [ ] ~If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)~
- [X] For new samples, I have added a paragraph with a summary to the appropriate chapter in the [samples readme](./../samples/README.md)
- [X] For new samples, I have added a tutorial README.md file to guide users through what they need to know to implement code using this feature. For example, see [conditional_rendering](./../samples/extensions/conditional_rendering)
- [X] For new samples, I have added a link to the [Antora navigation](./../antora/modules/ROOT/nav.adoc) so that the sample will be listed at the Vulkan documentation site


tomadamatkinson:
Apologies, I may have marked this ready for review from draft... please move back to draft if that was intended

I have enabled CI for this PR 😄

mdagois:
> Apologies, I may have marked this ready for review from draft... please move back to draft if that was intended
> 
> I have enabled CI for this PR 😄

Ah that explains the mysterious behavior. :)
I was about to mark it ready for review, and got confused when I saw it was already in that state!

shaders/oit_linked_lists/combine.frag
```diff
@@ -0,0 +1,106 @@
...
+                    const uvec2 tmp = sortedFragments[i];
+                    sortedFragments[i] = sortedFragments[j];
+                    sortedFragments[j] = tmp;
+                }
+            }
+        }
+    }
+
+    // Merge the fragments to get the final color
+    vec4 color = vec4(0.0f, 0.0f, 0.0f, 1.0f);
```
gary-sweet: Would this be more accurate as the clear color of the frame rather than black (maybe as a push constant, or in your SceneConstants)?

samples/api/oit_linked_lists/oit_linked_lists.cpp
```diff
@@ -0,0 +1,482 @@
...
+	}
+	else
+	{
+		throw std::runtime_error("This sample requires support for buffers and images stores and atomic operations in the fragment shader stage");
+	}
+}
+
+void OITLinkedLists::on_update_ui_overlay(vkb::Drawer &drawer)
+{
+	drawer.checkbox("Sort fragments", &sort_fragments);
```
gary-sweet, line 119: Another checkbox that would slowly rotate the camera would be really helpful for those of us with limited input capabilities.

`drawer.checkbox("Rotate", &rotate);`

samples/api/oit_linked_lists/oit_linked_lists.cpp
```diff
@@ -0,0 +1,482 @@
...
+	return true;
+}
+
+void OITLinkedLists::render(float delta_time)
+{
+	if (!prepared)
+	{
+		return;
+	}
+
```
gary-sweet, line 91: See below for my "Rotate" checkbox comment.

You could do something as simple as:
```
static float rot = 0.0f;
camera.type = vkb::CameraType::LookAt;
camera.set_rotation({rot, rot, 0.0f});
if (rotate)
	rot += delta_time * 5.0f;
```


Probably better with `rot` as a member variable though.

shaders/oit_linked_lists/combine.frag
```diff
@@ -0,0 +1,106 @@
...
+    for(int i = 0; i < fragmentCount; ++i)
+    {
+		const vec4 fragmentColor = unpackUnorm4x8(sortedFragments[i].x);
+		color.rgb = mix(color.rgb, fragmentColor.rgb, fragmentColor.a);
```
gary-sweet: I've personally always found standard alpha blending looks very unrealistic. Since you're actually performing all the blend calculations here you could actually make use of properly filtered colour blending (i.e. the way that glass or gel films actually behave).

Here's something I hacked in to do that (using another checkbox to control filtered vs standard).
Filtering the colours is just essentially multiplying src colour and dst colour. The opacity (alpha) then just controls how much you mix between fully transparent (just src*dst) and fully opaque (just src).
You have the alpha from 0.2 to 1.0 in the sample, but this kind of filtering actually looks nicer biased towards lower alpha values.

Try it, see if you like it. I can attach a video of it running also.

```
// Merge the fragments to get the final color
vec4 color = vec4(0.3f, 0.3f, 0.3f, 1.0f);
for(int i = 0; i < fragmentCount; ++i)
{
    const vec4 fragmentColor = unpackUnorm4x8(sortedFragments[i].x);
     vec3 first = color.rgb;
     if (sceneConstants.filtered == 1U)
         first *= fragmentColor.rgb;
    color.rgb = mix(first, fragmentColor.rgb, fragmentColor.a);
    color.a *= 1.0f - fragmentColor.a;
    }
if (sceneConstants.filtered == 1U)
    color.a = 1.0f;
else
    color.a = 1.0f - color.a;
outFragColor = color;
```

samples/api/oit_linked_lists/README.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+:pp: {plus}{plus}
+
+image::./images/sample.png[Sample]
+
+== Overview
+
+This sample implements an order-independent transparency (OIT) algorithm using per-pixel ordered linked lists.
+It renders 64 spheres with random color and opacity (from 0.2 to 1.0).
+It produces pixel-perfert results.
```
gary-sweet: typo: perfect, not perfert

samples/api/oit_linked_lists/README.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+== Algorithm
+
+The OIT algorithm consists of two passes: the _gather_ pass and the _combine_ pass.
+
+During the gather pass, the transparent geometry is rendered per-pixel order linked lists.
```
gary-sweet: rendered **into**

samples/api/oit_linked_lists/README.adoc
mdagois: Fixed!

samples/api/oit_linked_lists/README.adoc
mdagois: Fixed!

samples/api/oit_linked_lists/oit_linked_lists.cpp
mdagois: I added the option.

samples/api/oit_linked_lists/oit_linked_lists.cpp
mdagois: I added a call to camera.rotate when the option is enabled.

shaders/oit_linked_lists/combine.frag
mdagois: I use this default color so that pixels with no transparent geometry aren't visually affected by the combine pass.
The computed transparent color is blended into the color attachment (see around line 414 in *oit_linked_lists.cpp*), so it should produce correct results, even with a complex background.

shaders/oit_linked_lists/combine.frag
mdagois: Oh, excellent point!
I completely forgot to point out that the algorithm gives more freedom over the blending equation.

I added support for your equation and a corresponding checkbox.
I also updated the README to point out that the blending equation can be customized.
Could you please have a look to make sure the rendering is correct with the new option?

gary-sweet: > Could you please have a look to make sure the rendering is correct with the new option?

It doesn't look quite right as it stands. I believe you'll need to disable the native blending when using the filtered option. Anything that just blocks the background looks wrong otherwise. You can do this by just setting color.a to 1.0 in the filtered case.

```
if (sceneConstants.filtered_blending == 1U)
    color.a = 1.0f;
else
    color.a = 1.0f - color.a;
```

gary-sweet:
Looks like you're also missing the checklists that should appear in the initial description.
That's particularly important for new samples.

It looks pretty close now though, to me at least. Thanks

shaders/oit_linked_lists/combine.frag
gary-sweet: > > Could you please have a look to make sure the rendering is correct with the new option?
> 
> It doesn't look quite right as it stands. I believe you'll need to disable the native blending when using the filtered option. Anything that just blocks the background looks wrong otherwise. You can do this by just setting color.a to 1.0 in the filtered case.
> 
> ```
> if (sceneConstants.filtered_blending == 1U)
>     color.a = 1.0f;
> else
>     color.a = 1.0f - color.a;
> ```

Actually, this isn't right either. Let me have a play.

gary-sweet: So, the filtered blend isn't going to work unless we have the background color available in the shader. The filtering is multiplicative, which means just using 0,0,0,1 as the initial color forces the first color to multiply by zero, resulting in zero (obviously). If you're happy to accept a fixed color background, it could work, otherwise perhaps it's just a step too far for this sample. It would require a multi-pass approach to work against any background sadly.

.gitignore
```diff
@@ -8,6 +8,7 @@ tests/system_test/artifacts
 tests/system_test/tmp
 output/*
 .mypy_cache
+tags
```
asuessenbach, line 11: What `tags` do you want to ignore here?

samples/api/oit_linked_lists/oit_linked_lists.cpp
```diff
@@ -0,0 +1,489 @@
...
+	for (uint32_t l = 0; l < kInstanceLayerCount; ++l)
+	{
+		for (uint32_t c = 0; c < kInstanceColumnCount; ++c)
+		{
+			for (uint32_t r = 0; r < kInstanceRowCount; ++r)
+			{
+				const uint32_t instance_index =
+				    (l * kInstanceColumnCount * kInstanceRowCount) +
+				    (c * kInstanceRowCount) + r;
```
asuessenbach: You could as well define the instance_index in the outer-most loop, and increment it in the inner-most loop:
```
	for (uint32_t l = 0, instance_index = 0; l < kInstanceLayerCount; ++l)
	{
		for (uint32_t c = 0; c < kInstanceColumnCount; ++c)
		{
			for (uint32_t r = 0; r < kInstanceRowCount; ++r, ++instance_index)

```

mdagois:
> Looks like you're also missing the checklists that should appear in the initial description. That's particularly important for new samples.
> 
> It looks pretty close now though, to me at least. Thanks

I added the checklist to the PR description.

samples/api/oit_linked_lists/oit_linked_lists.h
```diff
@@ -0,0 +1,110 @@
...
+	{
+		kInstanceRowCount    = 4,
+		kInstanceColumnCount = 4,
+		kInstanceLayerCount  = 4,
+		kInstanceCount       = kInstanceRowCount * kInstanceColumnCount * kInstanceLayerCount,
+
+		kFragmentsPerPixelAverage = 8,
+
+		kLinkedListEndSentinel = 0xFFFFFFFFU,
+	};
```
asuessenbach: Those constants are more or less unrelated. Maybe consider to use `static constexpr`, instead,

shaders/oit_linked_lists/combine.frag
mdagois: Ah too bad!
I removed the filter blending option, because the necessary changes would probably stray too far from the intent of the sample, which is mostly to use the per-pixel linked lists for OIT.

shaders/oit_linked_lists/combine.frag
```diff
@@ -0,0 +1,112 @@
...
+
+	// Get the first fragment index in the linked list
+	uint fragmentIndex = imageLoad(linkedListHeadTex, ivec2(gl_FragCoord.xy)).r;
+	// Reset the list head for the next frame
+	imageStore(linkedListHeadTex, ivec2(gl_FragCoord.xy), uvec4(LINKED_LIST_END_SENTINEL, 0, 0, 0));
+
+	// Copy the fragments into local memory for sorting
+	uvec2 sortedFragments[SORTED_FRAGMENT_MAX_COUNT];
+	uint fragmentCount = 0U;
+	while(fragmentIndex != LINKED_LIST_END_SENTINEL && fragmentCount < SORTED_FRAGMENT_MAX_COUNT)
```
asuessenbach: You just consider SORTED_FRAGMENT_MAX_COUNT layers on a fragment and ignore any further ones. That way, depending on rendering order and camera position, you might even miss the nearest one.
Maybe you could, if there are more than SORTED_FRAGMENT_MAX_COUNT layers, mix-in the farthest layers in arbitrary order and only have the nearest SORTED_FRAGMENT_MAX_COUNT layers sorted?
That way, it might be even worth to make SORTED_FRAGMENT_MAX_COUNT a UI-controlled parameter.

asuessenbach (commented):
That's a great sample!
Just a few minor issues and a potential extension...

.gitignore
mdagois: It is for ctags. I use those with gvim, but I believe emacs' users might also use them.

samples/api/oit_linked_lists/oit_linked_lists.cpp
mdagois: Fixed.

samples/api/oit_linked_lists/oit_linked_lists.h
mdagois: Fixed.

shaders/oit_linked_lists/combine.frag
mdagois: Ah, very good point.

Technically, it should be possible to do a coarse sort first to limit the issue, but it would stray away from the purpose of the sample.
I plan to write a multipass sample, with a completely different technique, that removes the issue.

I added an explanation about the performance/correctness trade-off in the README.
I also added an option to control the maximum number of sorted fragments per pixel, so that users can visualize the artifacts when that number is low.

mdagois:
Following the discussion with Gary, I decided to add a background image to show that the transparent color is properly composed over the background one.
There is a new option to control the grayscale applied to the background image.
I believe the sample is complete now. Please review again. :)

gary-sweet (dismissed):
LGTM

shaders/oit_linked_lists/combine.frag
asuessenbach: You don't need to do a coarse sort first. Instead of sorting the first encountered fragments, you could as well keep them sorted:
```
	while ( fragmentIndex != LINKED_LIST_END_SENTINEL )
	{
		const uvec3 fragment = fragmentBuffer.data[fragmentIndex];
		fragmentIndex = fragment.x;
		if ( sceneConstants.sortFragments == 1U)
		{
			if ( fragmentCount < sortedFragmentMaxCount )
			{
				uint i=fragmentCount;
				for ( ; (0<i) && (fragment.z < sortedFragments[i-1].y) ; --i)
				{
					sortedFragments[i] = sortedFragments[i-1];
				}
				sortedFragments[i] = fragment.yz;
				++fragmentCount;
			}
			else if ( sortedFragments[0].y <= fragment.z )
			{
				color = blend( color, sortedFragments[0].x);
				uint i=0;
				for ( ; (i<sortedFragmentMaxCount-1) && (sortedFragments[i+1].y <= fragment.z) ; ++i)
				{
					sortedFragments[i] = sortedFragments[i+1];
				}
				sortedFragments[i] = fragment.yz;
			}
			else
			{
				color = blend( color, fragment.y);
			}
		}
		else
		{
			sortedFragments[fragmentCount] = fragment.yz;
			++fragmentCount;
		}
	}
```
Granted, a bit more complex, but I think it's more accurate. And probably not that much more expensive.
When you skip the blending operation in this sequence, you would nicely see the increasing depth when you increase the sortedFragmentMaxCount.

asuessenbach:
I get a validation layer warning about the `gather_render_pass`:
`Validation Warning: [ UNASSIGNED-BestPractices-vkCmdBeginRenderPass-ClearValueWithoutLoadOpClear ] | MessageID = 0x6beca064 | This render pass does not have VkRenderPassCreateInfo.pAttachments->loadOp == VK_ATTACHMENT_LOAD_OP_CLEAR but VkRenderPassBeginInfo.clearValueCount > 0. VkRenderPassBeginInfo.pClearValues will be ignored and no attachments will be cleared.
`

mdagois:
> I get a validation layer warning about the `gather_render_pass`: `Validation Warning: [ UNASSIGNED-BestPractices-vkCmdBeginRenderPass-ClearValueWithoutLoadOpClear ] | MessageID = 0x6beca064 | This render pass does not have VkRenderPassCreateInfo.pAttachments->loadOp == VK_ATTACHMENT_LOAD_OP_CLEAR but VkRenderPassBeginInfo.clearValueCount > 0. VkRenderPassBeginInfo.pClearValues will be ignored and no attachments will be cleared. `

Oops, this should be fixed now.

shaders/oit_linked_lists/combine.frag
mdagois: Ah, I see!
I misunderstood you original comment.
Let me try it.

mdagois: I used your algorithm for the combine pass. It works very well!
I can barely see any artifacts even with 4-5 sorted fragments.

I assume it might get worse with lower transparency, but it is way better than just ignoring the fragments.
Please have a look and thanks for the trick!

asuessenbach: Looks great!
Glad, I could help here.

asuessenbach:
Unfortunately, resolving the validation layer warning above resulted in an error, this time about the `ApiVulkanSample::render_pass`:
`Validation Error: [ VUID-VkRenderPassBeginInfo-clearValueCount-00902 ] Object 0: handle = 0x1d986e73d10, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xab64de0000000020, type = VK_OBJECT_TYPE_RENDER_PASS; | MessageID = 0x4de051e3 | In vkCmdBeginRenderPass the VkRenderPassBeginInfo struct has a clearValueCount of 0 but there must be at least 2 entries in pClearValues array to account for the highest index attachment in VkRenderPass 0xab64de0000000020[] that uses VK_ATTACHMENT_LOAD_OP_CLEAR is 2. Note that the pClearValues array is indexed by attachment number so even if some pClearValues entries between 0 and 1 correspond to attachments that aren't cleared they will be ignored. The Vulkan spec states: clearValueCount must be greater than the largest attachment index in renderPass specifying a loadOp (or stencilLoadOp, if the attachment has a depth/stencil format) of VK_ATTACHMENT_LOAD_OP_CLEAR (https://vulkan.lunarg.com/doc/view/1.3.261.1/windows/1.3-extensions/vkspec.html#VUID-VkRenderPassBeginInfo-clearValueCount-00902)
`

mdagois:
> Unfortunately, resolving the validation layer warning above resulted in an error, this time about the `ApiVulkanSample::render_pass`: `Validation Error: [ VUID-VkRenderPassBeginInfo-clearValueCount-00902 ] Object 0: handle = 0x1d986e73d10, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xab64de0000000020, type = VK_OBJECT_TYPE_RENDER_PASS; | MessageID = 0x4de051e3 | In vkCmdBeginRenderPass the VkRenderPassBeginInfo struct has a clearValueCount of 0 but there must be at least 2 entries in pClearValues array to account for the highest index attachment in VkRenderPass 0xab64de0000000020[] that uses VK_ATTACHMENT_LOAD_OP_CLEAR is 2. Note that the pClearValues array is indexed by attachment number so even if some pClearValues entries between 0 and 1 correspond to attachments that aren't cleared they will be ignored. The Vulkan spec states: clearValueCount must be greater than the largest attachment index in renderPass specifying a loadOp (or stencilLoadOp, if the attachment has a depth/stencil format) of VK_ATTACHMENT_LOAD_OP_CLEAR (https://vulkan.lunarg.com/doc/view/1.3.261.1/windows/1.3-extensions/vkspec.html#VUID-VkRenderPassBeginInfo-clearValueCount-00902) `

My bad, this time it should be good for both render passes.

samples/api/oit_linked_lists/oit_linked_lists.cpp
```diff
@@ -0,0 +1,530 @@
...
+		throw std::runtime_error("This sample requires support for buffers and images stores and atomic operations in the fragment shader stage");
+	}
+
+	if (gpu.get_features().samplerAnisotropy)
+	{
+		gpu.get_mutable_requested_features().samplerAnisotropy = VK_TRUE;
+	}
+	else
+	{
+		throw std::runtime_error("This sample requires support for anisotropic sampling");
```
asuessenbach: Why does this sample require support for anisotropic sampling?
Could this just be optional?

mdagois: Ah! I had completely forgotten about this.
I enabled aniso because it was necessary to remove a validation error when using `ApiVulkanSample::load_texture`.
Other samples are enabling aniso as well (at least the 2-3 ones I have checked).

I think I fixed the issue in `ApiVulkanSample::load_texture`, but I am not too familiar with `ApiVulkanSample` code, so please have a look.
I removed aniso from the sample and it does not trigger any validation error anymore after the fix.

asuessenbach (approved):

gary-sweet (approved):

samples/api/oit_linked_lists/README.adoc
```diff
@@ -0,0 +1,93 @@
...
+| Background grayscale
+| Specify the grayscale value by which the background color is multiplied (0.0 to 1.0)
+| 
+|===
+
+== Requirements
+
+The sample requires _fragmentStoresAndAtomics_ from VkPhysicalDeviceFeatures.
+
+== Tests
```
SaschaWillems, line 87: I don't think this paragraph is necessary. I'd remove it.

SaschaWillems (approved):
Great sample. I only had a minor nitpick with the readme, but that shouldn't stop us from merging this.

marty-johnson59:
3 approvals - merging

PR #835, opened 2023-10-20
Allow backspace to exit modal UI dialogs

gary-sweet:
## Description

For those of us navigating the UI with just a keyboard there was no way to close a color-picker dialog for example.

The escape key is currently used to exit a sample, so I've chosen the backspace key as a way to exit the current modal UI navigation.

## General Checklist:

### DIFF
--- a/framework/gui.cpp
+++ b/framework/gui.cpp
@@ -148,6 +148,7 @@ Gui::Gui(VulkanSample &sample_, const Window &window, const Stats *stats,
 	io.KeyMap[ImGuiKey_UpArrow]    = static_cast<int>(KeyCode::Up);
 	io.KeyMap[ImGuiKey_DownArrow]  = static_cast<int>(KeyCode::Down);
 	io.KeyMap[ImGuiKey_Tab]        = static_cast<int>(KeyCode::Tab);
+	io.KeyMap[ImGuiKey_Escape]     = static_cast<int>(KeyCode::Backspace);
 
 	// Default font
 	fonts.emplace_back(default_font, font_size * dpi_factor);
--- a/framework/hpp_gui.cpp
+++ b/framework/hpp_gui.cpp
@@ -114,6 +114,7 @@ HPPGui::HPPGui(HPPVulkanSample &sample_, const vkb::Window &window, const vkb::s
 	io.KeyMap[ImGuiKey_UpArrow]    = static_cast<int>(KeyCode::Up);
 	io.KeyMap[ImGuiKey_DownArrow]  = static_cast<int>(KeyCode::Down);
 	io.KeyMap[ImGuiKey_Tab]        = static_cast<int>(KeyCode::Tab);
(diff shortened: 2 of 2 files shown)


asuessenbach:
I can't tell, if backspace would be an appropriate key to exit a modal UI dialog.
But if this is considered to be acceptable, would you mind to adjust hpp_gui.cpp as well?

gary-sweet:
> Would you mind to adjust hpp_gui.cpp as well?

Done. Though I have to admit I find having all of this duplicated code in the framework fairly objectionable. Is there no way we can share this stuff instead of duplicating it all?

asuessenbach:
> Is there no way we can share this stuff instead of duplicating it all?

Well, I could imagine to introduce some base class for the framework classes, and derive a vulkan-based and a vulkan.hpp-based class from it, moving all the non-vulkan-related stuff into that base class.
Or both classes could call some helper functions to do the common stuff.

asuessenbach (approved):

SaschaWillems:
> > Is there no way we can share this stuff instead of duplicating it all?
> 
> Well, I could imagine to introduce some base class for the framework classes, and derive a vulkan-based and a vulkan.hpp-based class from it, moving all the non-vulkan-related stuff into that base class. Or both classes could call some helper functions to do the common stuff.

That would be great. Working on the framework is already kinda frustrating, and having to duplicate every change for C/C++ based samples makes it even more tedious. I'd really appreciate any effort to simplify the framework, esp in regards to the C/C++ separation.

asuessenbach:
> > > Is there no way we can share this stuff instead of duplicating it all?
> > 
> > 
> > Well, I could imagine to introduce some base class for the framework classes, and derive a vulkan-based and a vulkan.hpp-based class from it, moving all the non-vulkan-related stuff into that base class. Or both classes could call some helper functions to do the common stuff.
> 
> That would be great. Working on the framework is already kinda frustrating, and having to duplicate every change for C/C++ based samples makes it even more tedious. I'd really appreciate any effort to simplify the framework, esp in regards to the C/C++ separation.

What way should we go then? The base-class or the helper-functions approach? Or is there even some other option?
I think, the helper-function approach would be less intrusive (especially with some class hierarchies we already have), while the base-class approach would probably become clearer.

SaschaWillems:
Probably best to discuss the framework direction on our next call. I don't want to pollute this PR ;)

tomadamatkinson (approved):
LGTM

tomadamatkinson:
+1 for the framework discussion. I find that I have to bin most changes as there is too much overhead with so much duplication

gpx1000:
I think we did discuss it and decided to do just the CPP version in the framework?  We should validate that again in the next meeting and then I'll put together a PR to address it.

gpx1000 (approved):

SaschaWillems (approved):

marty-johnson59:
4 approvals -merging

PR #846, opened 2023-11-06
WIP: Offline Shader Compilation

tomadamatkinson:
Disclaimer: PoC does not yet work. Creating a PR for working group visibility

.
.
.
.
.
.
.
.
.
.
.
.
.
.
.

## Description

Please read the [contribution guidelines](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md)

Fixes #<issue number>

## General Checklist:

- [ ] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#Code-Style)
- [ ] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#Copyright-Notice-and-License-Template)

 - [ ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#General-Requirements)

## Sample Checklist

- [ ] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#General-Requirements)
- [ ] For new samples, I have added a link to the [Antora navigation](./../antora/modules/ROOT/nav.adoc) so that the sample will be listed at the Vulkan documentation site


shaders/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+file(GLOB_RECURSE GLSL_SHADER_FILES ${CMAKE_CURRENT_SOURCE_DIR}/*.glsl)
+file(GLOB_RECURSE HLSL_SHADER_FILES ${CMAKE_CURRENT_SOURCE_DIR}/*.hlsl)
+
+set(SHADERS ${GLSL_SHADER_FILES} ${HLSL_SHADER_FILES})
+
+foreach(SHADER_FILE ${SHADERS})
+    string(REPLACE ${CMAKE_CURRENT_SOURCE_DIR}/ "" SHADER_NAME ${SHADER_FILE})
+    string(REPLACE "/" "_" SHADER_NAME ${SHADER_NAME})
+    string(REPLACE "." "_" SHADER_NAME ${SHADER_NAME})
+    string(REPLACE "\\" "_" SHADER_NAME ${SHADER_NAME})
```
jherico:
```suggestion
    file(TO_CMAKE_PATH ${SHADER_FILE} SHADER_FILE)
    file(RELATIVE_PATH SHADER_RELATIVE_PATH ${CMAKE_CURRENT_SOURCE_DIR} ${SHADER_FILE})
    string(REGEX REPLACE "[\\./]" "_" SHADER_NAME ${SHADER_RELATIVE_PATH})
```

components/filesystem/include/filesystem/filesystem.hpp
```diff
@@ -0,0 +1,59 @@
...
+	virtual void                 write_file(const std::string &path, const std::vector<uint8_t> &data) = 0;
+
+	void write_file(const std::string &path, const std::string &data);
+
+	// Read the entire file into a string
+	std::string read_file(const std::string &path);
+
+	// Read the entire file into a vector of bytes
+	template <typename T = uint8_t>
+	std::vector<T> read_binary_file(const std::string &path)
```
jherico, line 42: You might consider a function along the lines of

```c++
template<typename T = uint8_t>
void with_file_contents(const std::string& filename, const std::function<void(size_t, const T*)>& handler);
```

In my experience, most applications only need the asset data for the scope of the call where it's read.  While `read_binary_file` returns a `std::vector` and thus _always_ allocates heap space, the implementation of something like `with_file_contents` is free to give a pointer to a memory mapped file, or something like the result of an `AAsset_getBuffer(...)` call on Android, which can lower the amount of actual heap space used in the application.  Obviously anything going into a staging buffer will consume host memory anyway, but at least this would potentially cut that in half.

components/core/include/core/util/hash.hpp
```diff
@@ -38,4 +43,11 @@ inline void hash_combine(size_t &seed, const T &v)
 
 	hash_combine(seed, hasher(v));
 }
+
+inline std::string sha256(const std::string &input)
```
jherico, line 47: Might want to consider using a `std::array<uint8_t, 32>`  (or `std::array<char, 64>` if you want to keep it hex encoded) for hashes since `std::array` can be `constexpr` and doesn't ever dynamically allocate.

jherico:
I'm curious why you took the path of switching all the existing GLSL headers to have the extension `glsl`.  You _could_ either assume that everything not ending in `hlsl` was instead GLSL, or you could inspect the file contents for a version directive. That would remove from the changed files most of the files that are just being renamed and have their copyright touched.

Also, you might want to consider abseil as a replacement for separate hash/string manipulation/logging libraries.  You could also use it to replace `fmt` as well, but I personally like `fmt` more.

components/filesystem/include/filesystem/filesystem.hpp
tomadamatkinson: I spun of a filesystem PR to simplify this one. I will move this comment there tomorrow

tomadamatkinson:
Thanks @jherico for taking a look.

In terms of hashing and abseil. I would prefer to keep the single header approach for now. No need to import an entirely new library for the sake of a sha256 algorithm.

My goal with this hash approach was to identify binary equivalent files to reduce the amount of compiled shaders that would need to be shipped to a device or stored in headers. The json indexes help to keep track of this in a more human readable way - end of the day we are trying to show people what is happening under the hood

In terms of glsl. I added the type of the shader in the file path as in the future we are wanting to create hlsl alternatives to your glsl shaders. Having the type would allow us to store hlsl/glsl in the same folder with less confusion

I am open to improvements here though. Maybe my renaming was preemptive and that specific change could come in a later PR if needed

.gitmodules
```diff
@@ -56,6 +56,15 @@
 [submodule "third_party/catch2"]
 	path = third_party/catch2
 	url = https://github.com/catchorg/Catch2.git
+[submodule "third_party/spirv-reflect"]
+	path = third_party/spirv-reflect
+	url = https://github.com/KhronosGroup/SPIRV-Reflect.git
+[submodule "third_party/json"]
+	path = third_party/json
+	url = https://github.com/nlohmann/json.git
```
jherico, line 64: tinygltf already includes a version of this JSON parser.  I understand a possible reluctance to have whatever is depending on this version depend on tinygltf instead, but I'd also be concerned about possible issues if any project depending on both gets the version it wasn't expecting when it tries to include the header.

jherico: ok, I was mistaken... there are already 2 versions of this parser in the codebase (hwcpipe has one embedded) and this will make 3.

jherico:
I'm confused as to why this particular build is generating some errors on the CI runs when the files themselves haven't changed.

SaschaWillems:
CI errors are due to some files having been renamed but not all files referring to them have been updated yet. But it's a draft anyway.

As the description suggest, Tom made this as base for discussing possible ways of doing this.

jherico:
~Sorry, I should habe been more specific.  I saw an error where the compiler.complained about RAND_MAX cast to float losing precision.~

Never mind, I think I get it.

jherico:
Playing around with locally building this and diving into the shader preprocessing a bit more this morning.  One thing I'm noticing is the heavy use of `add_custom_target`.  Anything made by `add_custom_target` is considered always out of date.  

I previously had a similar solution in my project where every single shader was a dependency of a single massive shaders target, but that had the annoying by-product of causing a rebuild of every single shader if I wanted to tweak one and re-run a given sample.  

Instead I would recommend replacing the `add_custom_target` items with `add_custom_command` where you're able to specify the both the inputs and the outputs of the command, so the output are only regenerated if they're actually out of date compared to the inputs.  This means you don't have a top-level target for each shader file, so what you end up doing is when you get to a target that actually requires a shader (like the say the `add_sample` lists the required shaders) you
* Add the shader source to the specific target
* Also add the generated files to the target, which forces them to be built if they're out of date, when you build the project.

You might still need a separate step to build something like an atlas of all the shaders, but it would make it much more convenient for someone working on a given sample to edit the shader used for that sample without triggering the burden of rebuilding every single shader.

If you're interested I can create a branch off of your code and tweak it to do this.  This wouldn't prevent the framework or components from declaring shader dependencies, and specific shaders could still easily be referenced by more than one target in the project.

tomadamatkinson:
This PR is no longer needed. The complexity of this PR is to support our implementation of variants. The working group has opted to reduce our variant usage allowing this PR to be closed

PR #857, opened 2023-11-30
WIP: Refactor Resource Cache

tomadamatkinson:
## Description

The resource cache is an easy to use area of the framework to lazily create resources. The resource cache is also one of the hardest to understand when debugging or digging through its code. The issue with the cache is that it holds the complexity needed for the `hpp_pipeline_cache` and `pipeline_cache` samples.

This PR moves this functionality from the framework to the sample. The cache code is easier to understand and the specialisation for pipeline caching is now local to the sample and so easier for users of the sample!

I found that the resource cache itself has been a hurdle when trying to make larger changes. This is likely due to its complexity and that the current implementation is not super readable. In order to fix some of this, i implemented some nice to have container types in `core/containers/*` (with tests). Removing the record/replay made the cache code easier to understand. This PR is a useful precursor to #657 but is not a blocker

## General Checklist:

- [ ] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#Code-Style)
- [ ] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#Copyright-Notice-and-License-Template)

 - [ ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#General-Requirements)

## Sample Checklist

- [ ] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#General-Requirements)
- [ ] For new samples, I have added a link to the [Antora navigation](./../antora/modules/ROOT/nav.adoc) so that the sample will be listed at the Vulkan documentation site


samples/extensions/open_cl_interop/open_cl_interop.cpp
```diff
@@ -572,7 +572,7 @@ void OpenCLInterop::prepare_shared_image()
 	sampler_create_info.magFilter   = VK_FILTER_LINEAR;
 	sampler_create_info.minFilter   = VK_FILTER_LINEAR;
 	sampler_create_info.mipmapMode  = VK_SAMPLER_MIPMAP_MODE_LINEAR;
-	sampler_create_info.maxLod      = (float) 1;
```
SaschaWillems, line 575: I'm not a C++ pro, so I was wondering why the static cast here is better? Does the old code generate warnings?

samples/extensions/open_cl_interop/open_cl_interop.cpp
```diff
@@ -620,12 +620,12 @@ void OpenCLInterop::prepare_shared_image()
 	mem_properties.push_back((cl_mem_properties) handle);
 #else
 	int fd = get_vulkan_memory_handle(shared_image.memory);
-	mem_properties.push_back((cl_mem_properties) CL_EXTERNAL_MEMORY_HANDLE_OPAQUE_FD_KHR);
```
SaschaWillems, line 623: Same as above. Since I wrote that code I'm curious to know what type of cast is better and which one we should use in the future.

SaschaWillems (changes requested):
Gave this a first spin and it's looking good and working fine so far :)

But this PR also contains a lot of changes not directly related to the resource cache. Many of these seem to be code cleanup related like changing type casts to explicit casts. Maybe move them out of this PR to make reviewing easier?

framework/core/hpp_device.h
```diff
@@ -24,9 +24,10 @@
 #include <core/hpp_queue.h>
 #include <core/hpp_vulkan_resource.h>
 #include <hpp_fence_pool.h>
-#include <hpp_resource_cache.h>
 #include <vulkan/vulkan.hpp>
 
+#include <hpp_resource_cache.h>
```
asuessenbach, line 29: Why have you moved this include a little?

framework/core/hpp_shader_module.h
```diff
@@ -77,33 +77,22 @@ struct HPPShaderResource
 	std::string           name;
 };
 
-class HPPShaderSource : private vkb::ShaderSource
```
asuessenbach, line 80: Why have you removed HPPShaderSource and HPPShaderVariant?

framework/hpp_resource_cache.cpp
```diff
@@ -58,76 +65,48 @@ void HPPResourceCache::clear_framebuffers()
...
-
-const HPPResourceCacheState &HPPResourceCache::get_internal_state() const
-{
-	return state;
-}
-
 vkb::core::HPPComputePipeline &HPPResourceCache::request_compute_pipeline(vkb::rendering::HPPPipelineState &pipeline_state)
 {
-	return request_resource(device, recorder, compute_pipeline_mutex, state.compute_pipelines, pipeline_cache, pipeline_state);
+	vk::PipelineCache pipeline_cache{VK_NULL_HANDLE};
```
asuessenbach, line 73: Maybe use the same comment on the pipeline_cache as in resource_cache.cpp

framework/hpp_resource_cache.cpp
```diff
@@ -58,76 +65,48 @@ void HPPResourceCache::clear_framebuffers()
...
+vkb::core::HPPFramebuffer &HPPResourceCache::request_framebuffer(const vkb::rendering::HPPRenderTarget &render_target, const vkb::core::HPPRenderPass &render_pass)
 {
-	return request_resource(device, recorder, framebuffer_mutex, state.framebuffers, render_target, render_pass);
+	return request_resource(device, state.framebuffers, render_target, render_pass);
 }
 
 vkb::core::HPPGraphicsPipeline &HPPResourceCache::request_graphics_pipeline(vkb::rendering::HPPPipelineState &pipeline_state)
 {
-	return request_resource(device, recorder, graphics_pipeline_mutex, state.graphics_pipelines, pipeline_cache, pipeline_state);
+	vk::PipelineCache pipeline_cache{VK_NULL_HANDLE};
```
asuessenbach, line 98: Maybe use the same comment on the pipeline_cache as in resource_cache.cpp

samples/extensions/open_cl_interop/open_cl_interop.cpp
asuessenbach: Especially, if you simply write
`sampler_create_info.maxLod = 1.0f;`
you don't need any cast at all.

bldsys/cmake/global_options.cmake
```diff
@@ -60,7 +60,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "bin/${CMAKE_BUILD_TYPE}/${TARGET_ARCH}")
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "lib/${CMAKE_BUILD_TYPE}/${TARGET_ARCH}")
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "lib/${CMAKE_BUILD_TYPE}/${TARGET_ARCH}")
 
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
asuessenbach: Why do you need to increase the CXX_STANDARD?

components/core/include/core/containers/cache_map.hpp
asuessenbach, line 1: Would be good to have this file listed in some CMakeLists.txt as well.

components/core/include/core/containers/type_map.hpp
asuessenbach, line 1: Would be good to have this file listed in some CMakeLists.txt as well.

asuessenbach (commented):
Some first comments...
Besides that, I don't see the benefit of those new classes `ContainerWrapper` and `CacheMap`. Wouldn't just using `std::unordered_map`, plus maybe one or two helper functions be a less complex approach?

framework/rendering/hpp_render_frame.h
```diff
@@ -157,10 +157,10 @@ class HPPRenderFrame
 	std::map<uint32_t, std::vector<std::unique_ptr<vkb::core::HPPCommandPool>>> command_pools;
 
 	/// Descriptor pools for the frame
-	std::vector<std::unique_ptr<std::unordered_map<std::size_t, vkb::core::HPPDescriptorPool>>> descriptor_pools;
+	CacheMap<size_t, std::unique_ptr<CacheMap<size_t, vkb::core::HPPDescriptorPool>>> descriptor_pools;
```
asuessenbach, line 160: Is it worth to have a `CacheMap` here, instead of a `std::vector`? What do you gain from that change?

framework/resource_cache.h
```diff
@@ -80,82 +68,32 @@ class ResourceCache
...
-
-	RenderPass &request_render_pass(const std::vector<Attachment> &   attachments,
-	                                const std::vector<LoadStoreInfo> &load_store_infos,
-	                                const std::vector<SubpassInfo> &  subpasses);
-
-	Framebuffer &request_framebuffer(const RenderTarget &render_target,
-	                                 const RenderPass &  render_pass);
-
-	void clear_pipelines();
+	virtual ShaderModule        &request_shader_module(VkShaderStageFlagBits stage, const ShaderSource &glsl_source, const ShaderVariant &shader_variant = {});
```
asuessenbach, line 76: When you make all those functions virtual, you need to make all the corresponding functions in HPPResourceCache virtual as well and have them in the same order to keep them congruent.
-> A loud call for some unifying approach between HPP and non-HPP again!

framework/resource_cache.h
```diff
@@ -80,82 +68,32 @@ class ResourceCache
...
+  protected:
 	Device &device;
 
-	ResourceRecord recorder;
-
-	ResourceReplay replayer;
-
-	VkPipelineCache pipeline_cache{VK_NULL_HANDLE};
-
+	mutable std::mutex mutex;
```
asuessenbach: This mutex is unused?

samples/extensions/open_cl_interop/open_cl_interop.cpp
tomadamatkinson: Fixed with suggested

samples/extensions/open_cl_interop/open_cl_interop.cpp
tomadamatkinson: clang tidy recommends using C++ style casts over C style casts. I believe a lot of these changes is me attempting to build with clang tidy warnings as errors on

components/core/include/core/containers/cache_map.hpp
tomadamatkinson: Added to components/core/CMakeLists.txt

components/core/include/core/containers/type_map.hpp
tomadamatkinson: Added to component/core/CMakeLists.txt

framework/hpp_resource_cache.cpp
tomadamatkinson: Added

framework/hpp_resource_cache.cpp
tomadamatkinson: Added

framework/resource_cache.h
tomadamatkinson: Added virtual to missing functions and order is now the same as recommended. Would you be able to point me to some docs about this technique. I'd like to understand it more but couldn't find anything which explained it well

> A loud call for some unifying approach between HPP and non-HPP again!

We'll get there. I'm all for simplifying the framework and choosing a single version of the API to use for the majority. It very hard to make any useful change maintaining two versions of the same framework

framework/resource_cache.h
tomadamatkinson: Yup removed. I think it was something i was playing with

framework/core/hpp_shader_module.h
tomadamatkinson: The classes do not need a fascade as they are identical and do not include Vulkan code which needed to be be duplicated

framework/core/hpp_device.h
tomadamatkinson: Moved as i was getting some cyclical dependencies issues

bldsys/cmake/global_options.cmake
tomadamatkinson: Fixed, not needed. I was likely playing with something at the time that needed 17 but then chose not to use it

framework/core/hpp_shader_module.h
asuessenbach: Maybe you can add some TODO comment to resolve that cyclicyal dependencies?

framework/resource_cache.h
asuessenbach: > Would you be able to point me to some docs about this technique.

I don't get what "technique" you mean here...
Currently, all the HPP- and corresponding non-HPP-classes are need to be binary identical, as they tend to be used in both HPP- and non-HPP calls and classes.
That is, all class/struct pairs need to have binary identical members in identical order, and apparently also need to have identical sets of virtual functions.

framework/core/hpp_shader_module.h
tomadamatkinson: This might be resolved by combining the duplicate headers we have between the two versions of the framework

framework/resource_cache.h
tomadamatkinson: This explains the issue i was having with the HPP sample. I should be able to test this at some point today

framework/core/hpp_device.h
asuessenbach: That's strange. I don't get any issues when it's sorted in as before.

framework/core/hpp_shader_module.h
asuessenbach: Sure... but that might take a while 'til it's all done.

samples/performance/hpp_pipeline_cache/hpp_resource_record.cpp
```diff
@@ -0,0 +1,264 @@
...
+	      load_store_infos);
+
+	write_subpass_info(stream, subpasses);
+
+	return render_pass_indices.back();
+}
+
+size_t HPPResourceRecord::register_graphics_pipeline(vk::PipelineCache /*pipeline_cache*/, rendering::HPPPipelineState &pipeline_state)
+{
+	LOGI("HPPResourceRecord::register_graphics_pipeline");
```
asuessenbach, line 203: Maybe you could remove all those logging calls to make the hpp-version as fast as the non-hpp-version.

samples/extensions/open_cl_interop/open_cl_interop.cpp
SaschaWillems: Thank you very much. Good to know :)

tomadamatkinson:
Closing as this will be resolved by reducing the complexity of the framework. That is the priority

PR #858, opened 2023-11-30
Fix: Wrong pragma under GCC

### LINKED ISSUE #826: Wrong pragma under GCC
Hi,
I saw that the macro VKBP_DISABLE_WARNINGS  has wrong behavior when compiling using GCC https://github.com/KhronosGroup/Vulkan-Samples/blob/main/framework/common/error.h#L43.

Shouldn't be GCC instead of clang?

Thanks

tomadamatkinson:
Fixes #826


gpx1000 (approved):

SaschaWillems (approved):

tomadamatkinson:
Merging on two as small change

PR #869, opened 2023-12-28
[sample] Order-independent transparency using depth peeling

mdagois:
## Description

Added a new sample that implements an order-independent transparency algorithm using depth peeling.

## General Checklist:

- [x] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#Code-Style)
- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [ ] ~~I have tested every sample to ensure everything runs correctly~~

 - [x] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#General-Requirements)

## Sample Checklist

- [ ] ~~If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.md#General-Requirements)~~
- [x] For new samples, I have added a tutorial README.md file to guide users through what they need to know to implement code using this feature. For example, see [conditional_rendering](./../samples/extensions/conditional_rendering)
- [x] For new samples, I have added a link to the [Antora navigation](./../antora/modules/ROOT/nav.adoc) so that the sample will be listed at the Vulkan documentation site


samples/api/oit_depth_peeling/README.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+Each _gather_ pass renders one layer of transparent geometry.
+The first pass renders the first layer, the second pass the second layer, etc.
+The N^th^ layer consists of all the N^th^ fragments of each pixel when the fragments are ordered from front to back.
+
+The _combine_ pass is a screen-space operation.
+It merges the layer images from back to front to produce the final result.
+
+The algorithm can produce pixel-perfect results, even with intersecting geometry.
+When there are more geometry layers than gather passes, the backmost layers get skipped, but the visual results stay stable (i.e. no flickering pixels).
```
asuessenbach, line 49: Maybe you could blend-in those remaining fragments similarly to how they are handled in the `oit_linked_lists` sample?

samples/api/oit_depth_peeling/images/sample.png
asuessenbach, line 1: Maybe you could use an image with opacity less than 1.0 ?

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
```diff
@@ -0,0 +1,558 @@
...
+}
+
+void OITDepthPeeling::render(float delta_time)
+{
+	if (!prepared)
+	{
+		return;
+	}
+
+	back_layer_index = std::max(back_layer_index, front_layer_index);
```
asuessenbach: Maybe instead of taking the max of `front_` and `back_layer_index`, prevent overlapping in `on_update_ui_overlay`, like
```
	drawer.slider_int("Front layer index", &front_layer_index, 0, back_layer_index);
	drawer.slider_int("Back layer index", &back_layer_index, front_layer_index, kLayerMaxCount - 1);

```

samples/api/oit_depth_peeling/oit_depth_peeling.h
```diff
@@ -0,0 +1,112 @@
...
+	VkPipeline       gather_pipeline;
+	VkPipelineLayout combine_pipeline_layout;
+	VkPipeline       combine_pipeline;
+	VkPipeline       background_pipeline;
+
+	int32_t camera_auto_rotation = false;
+	float   background_grayscale = 0.3f;
+	float   object_alpha         = 1.0f;
+	int32_t front_layer_index    = 0;
+	int32_t back_layer_index     = kLayerMaxCount;
```
asuessenbach: `back_layer_index` should probably be initialized with `kLayerMaxCount - 1`

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
```diff
@@ -0,0 +1,558 @@
...
+	render_pass_begin_info.renderArea.extent.height = height;
+	render_pass_begin_info.clearValueCount          = 2;
+	render_pass_begin_info.pClearValues             = clear_values;
+
+	for (int32_t i = 0; i < draw_cmd_buffers.size(); ++i)
+	{
+		VK_CHECK(vkBeginCommandBuffer(draw_cmd_buffers[i], &command_buffer_begin_info));
+		{
+			// Gather passes
+			for (uint32_t l = 0; l < kLayerMaxCount; ++l)
```
asuessenbach: You always run over `kLayerMaxCount` passes. Maybe it's worth to rebuild the command buffers on `[back|front]_layer_index` changes, and only run as many passes as needed?

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
```diff
@@ -0,0 +1,558 @@
...
+
+	VkFramebufferCreateInfo framebuffer_create_info = vkb::initializers::framebuffer_create_info();
+	framebuffer_create_info.width                   = width;
+	framebuffer_create_info.height                  = height;
+	framebuffer_create_info.layers                  = 1;
+	framebuffer_create_info.attachmentCount         = 2;
+	for (uint32_t i = 0; i < kLayerMaxCount; ++i)
+	{
+		VK_CHECK(vkCreateRenderPass(get_device().get_handle(), &render_pass_create_info, nullptr, gather_render_pass + i));
+		framebuffer_create_info.renderPass = gather_render_pass[i];
```
asuessenbach: I think, you don't need `kLayerMaxCount` identical `gather_render_pass`es. One should suffice.

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
```diff
@@ -0,0 +1,558 @@
...
+		VkDescriptorSetLayoutCreateInfo descriptor_layout_create_info = vkb::initializers::descriptor_set_layout_create_info(set_layout_bindings.data(), static_cast<uint32_t>(set_layout_bindings.size()));
+		VK_CHECK(vkCreateDescriptorSetLayout(get_device().get_handle(), &descriptor_layout_create_info, nullptr, &combine_descriptor_set_layout));
+	}
+
+	{
+		std::vector<VkDescriptorPoolSize> pool_sizes = {
+		    vkb::initializers::descriptor_pool_size(VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 4),
+		    vkb::initializers::descriptor_pool_size(VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER, 11),
+		};
+		const uint32_t             num_descriptor_sets         = 3;
```
asuessenbach: Maybe some comment where those magic numbers (4, 11, and 3) come from?

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
```diff
@@ -0,0 +1,558 @@
...
+		std::vector<VkWriteDescriptorSet> write_descriptor_sets = {
+		    vkb::initializers::write_descriptor_set(gather_descriptor_set[i], VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 0, &scene_constants_descriptor),
+		    vkb::initializers::write_descriptor_set(gather_descriptor_set[i], VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER, 1, &depth_texture_descriptor),
+		};
+		vkUpdateDescriptorSets(get_device().get_handle(), static_cast<uint32_t>(write_descriptor_sets.size()), write_descriptor_sets.data(), 0, NULL);
+	}
+
+	{
+		std::vector<VkWriteDescriptorSet> write_descriptor_sets;
+		write_descriptor_sets.resize(3);
```
asuessenbach: You can just write
`		std::vector<VkWriteDescriptorSet> write_descriptor_sets( 3 );
`
to get a vector of size 3.

shaders/oit_depth_peeling/gather.vert
```diff
@@ -0,0 +1,44 @@
...
+	float object_alpha;
+	int   front_layer_index;
+	int   back_layer_index;
+} sceneConstants;
+
+////////////////////////////////////////
+
+void main()
+{
+	gl_Position = sceneConstants.projection * sceneConstants.view * sceneConstants.model * vec4(inPos, 1.0f);
```
asuessenbach: Maybe it's worth to combine those three matrices into one?

asuessenbach (commented):
When starting this sample, I get lots of warnings from the Vulkan Configurator:
`Validation Warning: [ UNASSIGNED-BestPractices-TransitionUndefinedToReadOnly ] | MessageID = 0x9f63e654 | vkCmdPipelineBarrier(): pImageMemoryBarriers[0] VkImageMemoryBarrier is being submitted with oldLayout VK_IMAGE_LAYOUT_UNDEFINED and the contents may be discarded, but the newLayout is VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, which is read only.
`
on the call to `vkb::image_layout_transition` in oit_depth_peeling.cpp, line 163.

samples/api/oit_depth_peeling/images/sample.png
mdagois: Yes, I updated the screenshot to use 0.3.

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
mdagois: I tried that before, and I felt the UI behavior was a bit unintuitive.
I changed to what you suggested, and if that looks good to you, then it is fine by me.

samples/api/oit_depth_peeling/oit_depth_peeling.h
mdagois: Oops, fixed.

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
mdagois: I used constants to convey the meaning of those numbers better.

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
mdagois: Fixed.

shaders/oit_depth_peeling/gather.vert
mdagois: Fixed.

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
mdagois: You are right! I changed the code to use a single render pass.

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
mdagois: I cannot skip the front layers completely, as I need their depth information to extract the subsequent layers. Technically, I could skip rendering any color attachment for layers before **front_layer_index**, but that would require additional changes for what is more of a "debug" visualization feature.

I did the simple change of skipping any layer after **back_layer_index** though, since nor their depth nor color are not used in any way.

Is this OK with you?

mdagois:
@asuessenbach I had to update a few copyright year that are not from my company (to please the copyright quality check). Is it OK?

mdagois:
> When starting this sample, I get lots of warnings from the Vulkan Configurator: `Validation Warning: [ UNASSIGNED-BestPractices-TransitionUndefinedToReadOnly ] | MessageID = 0x9f63e654 | vkCmdPipelineBarrier(): pImageMemoryBarriers[0] VkImageMemoryBarrier is being submitted with oldLayout VK_IMAGE_LAYOUT_UNDEFINED and the contents may be discarded, but the newLayout is VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, which is read only. ` on the call to `vkb::image_layout_transition` in oit_depth_peeling.cpp, line 163.

I think I fixed this issue. I was transitioning a yet unused depth texture in the first gather pass.
Could you please check again?

samples/api/oit_depth_peeling/README.adoc
mdagois: I could, but I think it is less important to use the blend-in when using depth peeling.
With the linked lists, when the fragment buffer was running out of space, the algorithm could produce artifacts dependent on the camera position.
With depth peeling, we always get the closest fragments, so there are no obvious visible artifacts.
I could use an OIT approximation, like Meshkin, for the last layer, but that would stray away from the traditional depth peeling.
What do you think?

asuessenbach: Sure thing, blending-in the remaining fragments would go add some complexity that the "standard" depth-peeling approach has not. But the visible artifacts are essentially the same as with the linked-list approach, which also always renders the frontmost fragments in correct order.
One idea to make this sample even better might be to make the number of layers dynamic by counting the number of fragments drawn per layer and stop when that number is below some (controllable) threshold?

samples/api/oit_depth_peeling/README.adoc
```diff
@@ -0,0 +1,84 @@
...
+endif::[]
+
+:pp: {plus}{plus}
+
+image::./images/sample.png[Sample]
+
+== Overview
+
+This sample implements an order-independent transparency (OIT) algorithm using depth peeling.
+It renders a single torus whose opacity can be controller via the UI.
```
asuessenbach: typo: controller -> controlled

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
asuessenbach: I think, it's better this way.

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
```diff
@@ -0,0 +1,566 @@
...
+		    vkb::initializers::descriptor_set_layout_binding(VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER, VK_SHADER_STAGE_FRAGMENT_BIT, 1),
+		    vkb::initializers::descriptor_set_layout_binding(VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER, VK_SHADER_STAGE_FRAGMENT_BIT, 2, kLayerMaxCount),
+		};
+		VkDescriptorSetLayoutCreateInfo descriptor_layout_create_info = vkb::initializers::descriptor_set_layout_create_info(set_layout_bindings.data(), static_cast<uint32_t>(set_layout_bindings.size()));
+		VK_CHECK(vkCreateDescriptorSetLayout(get_device().get_handle(), &descriptor_layout_create_info, nullptr, &combine_descriptor_set_layout));
+	}
+
+	{
+		const uint32_t num_gather_pass_combined_image_sampler = 1 * kDepthCount;
+		const uint32_t num_gather_pass_uniform_buffer         = 1 * kDepthCount;
```
asuessenbach: I would skip those `1 *`.

samples/api/oit_depth_peeling/README.adoc
mdagois: Fixed.

samples/api/oit_depth_peeling/oit_depth_peeling.cpp
mdagois: Removed.

samples/api/oit_depth_peeling/README.adoc
mdagois: Do you mean using something like conditional rendering?

asuessenbach: > Do you mean using something like conditional rendering?

You're asking for the dynamic layer count? I thought about using occlusion queries to get the number of fragments drawn in a pass, and stop the gather passes when it's below some threshold.
Note, though, you should not wait for idle after a gather-pass to get the number of fragments, before you trigger the next gather-pass or the final combine-pass. Instead, you might take the number of fragments of the second last gather-pass...

asuessenbach (dismissed):

samples/api/oit_depth_peeling/README.adoc
mdagois: Ah yeah, that makes sense. Maybe I could even discard fragments when the previous layers' cumulative opacity is 1.0. That would draw only a single layer when the mesh is fully opaque.

gary-sweet:
I'm seeing quite a lot of flickering triangles when I reduce the object opacity.
I'll see if I can attach a small video.

gary-sweet:
https://github.com/KhronosGroup/Vulkan-Samples/assets/20819159/1d50129f-f175-43bc-8b42-5e1a309851d2

mdagois:
> I'm seeing quite a lot of flickering triangles when I reduce the object opacity. I'll see if I can attach a small video.

Oh, interesting. Thanks for the video!

Could you tell me which GPU you tested on?
The algorithm should never produce those flickers, but I have been quite aggressive with the barriers and it may be causing sync issues.

I only ever tested on AMD GPU.

gary-sweet:
> Could you tell me which GPU you tested on? The algorithm should never produce those flickers, but I have been quite aggressive with the barriers and it may be causing sync issues.

It's our own Broadcom Videocore VII GPU

gary-sweet:
So, the flickering looks like it might be an issue on our side, which I'm investigating.

shaders/oit_depth_peeling/gather.frag
```diff
@@ -0,0 +1,38 @@
...
+
+////////////////////////////////////////
+
+layout(set = 0, binding = 1) uniform sampler2D depthTex;
+
+////////////////////////////////////////
+
+void main()
+{
+	const float layerDepth = texture(depthTex, gl_FragCoord.xy / textureSize(depthTex, 0).xy).r;
```
gary-sweet: Could you change this to:
`const float layerDepth = texelFetch(depthTex, ivec2(gl_FragCoord.xy), 0).r;`

It removes the divide and is therefore more efficient.

samples/api/oit_depth_peeling/README.adoc
gary-sweet: I did a similar thing many, many years ago when I added depth-peeling to a visualizer I was working on at a previous company. I used occlusion queries there.

I also peeled the other way around, so farthest layers first. That has a couple of advantages (IIRC - it was nearly 20 years ago).
1. The first layer includes the backdrop, so requires no special handling. Any kind of blending works against the backdrop too. You can do multiplicative blending (i.e. real filtered colours) and it just works.
2. You don't need to keep all of the depth layers and combine them at the end. You just need the last depth buffer that was written. 

You do need to know when to stop, which means you do need occlusion counts.

shaders/oit_depth_peeling/gather.frag
mdagois: Fixed.

samples/api/oit_depth_peeling/README.adoc
mdagois: It sounds like your algorithm has a lot of advantages.

I implemented depth peeling years ago in the context of a game (it was not used when the game launched though, as we used a different technique instead).
At the time, we wanted to control the number of layers for performance reasons, even if that meant the result was not exact. In that case, starting from the front was the only way we could come up with.

asuessenbach (dismissed):

samples/api/oit_depth_peeling/README.adoc
asuessenbach: You could as well do a two-layer front-to-back peeling with a slightly adjusted blending operation, as described for example in https://developer.download.nvidia.com/SDK/10/opengl/src/dual_depth_peeling/doc/DualDepthPeeling.pdf

mdagois: Yes, I actually tried that one back in the day.
But, our use case was lots of transparents layers very likely to converge to full opacity in 4-5 layers, and a hard limit on the number of layers (IIRC, 4).
So, a run-of-the-mill depth peeling was actually the most applicable one.

We did use the single pass approximations presented in this paper though!

mdagois:
> So, the flickering looks like it might be an issue on our side, which I'm investigating.

Hey @gary-sweet,
Did you find out if the flickering issue was on your driver side?
Or might it be in the sample?

gary-sweet:
> Did you find out if the flickering issue was on your driver side?

@mdagois  The flickering is an issue on our side related to a tiny mismatch between the fixed-function Z value and the value of gl_FragCoord.z in certain circumstances on older hardware.

While I was pondering why we hadn't seen such an issue previously, I started to think about how depth-peeling is usually implemented, and I would suggest that it normally just uses the fixed-function depth comparison operations rather than having to modify the shaders to do the depth comparison.  Having to modify the fragment shaders to be able to use them with depth-peeling isn't an optimal solution, so we may want to think about modifying this sample (which is presumably supposed to be showing good-practice) to use fixed-function depth instead.

mdagois:
> > Did you find out if the flickering issue was on your driver side?
> 
> @mdagois The flickering is an issue on our side related to a tiny mismatch between the fixed-function Z value and the value of gl_FragCoord.z in certain circumstances on older hardware.
> 
> While I was pondering why we hadn't seen such an issue previously, I started to think about how depth-peeling is usually implemented, and I would suggest that it normally just uses the fixed-function depth comparison operations rather than having to modify the shaders to do the depth comparison. Having to modify the fragment shaders to be able to use them with depth-peeling isn't an optimal solution, so we may want to think about modifying this sample (which is presumably supposed to be showing good-practice) to use fixed-function depth instead.
> 
> Doing that also has other advantages, like only needing a single depth-buffer - not the eight that you have now for example.

Ah, I see. That sounds more like a hardware issue than an algorithm one though.
I doubt any modern hardware would have that problem.

The original version of depth peeling (link below) relies on both fixed-function Z and a depth texture read to "fake" having a second depth buffer.
The paper is from the early 2000's, so the algorithm was working on pretty old hardware.
I believe the sample is doing exactly what the paper presents.
It uses the fixed-function Z to get the closest layer, but use the texture read (from the depth buffer of the previous gather pass) to reject fragments that were already rendered.
There are two assumptions that makes this work afaik:
1. the same shader is run over and over so we get the same gl_FragCoord.z on each fragment
2. 32-bit float depth buffer are used, so we can exactly compare gl_FragCoord.z to the value read from the depth texture

From what I understand, the old hardware you are running the sample on has an issue with assumption (2).
The problem is that other depth peeling algorithms do rely on the same trick.
At least, dual depth peeling does.
I believe back-first depth peeling would too, otherwise, I am not sure how it could discard back layers that have already been rendered.

Even then, the advantage of the original depth peeling is that you can stop early.
Dual depth peeling or back first depth peeling cannot do that.
They of course have other advantages, like less gather passes or lower resource usage.

Ultimately, I think the sample should probably implement all three algorithms, and let the user choose which algorithm to use.
I can add a TODO if necessary and list pros/cons of each technique to the README.
I just cannot commit a given timeline for implementing those though, as I mostly do that on my free time.

Btw, the algorithm does not use 8 depth buffers, but just 2.
It does however use 8 color attachments.
I assume that's what you meant.

Original depth peeling: https://developer.download.nvidia.com/SDK/10/opengl/src/dual_depth_peeling/doc/DualDepthPeeling.pdf

gary-sweet:
> The original version of depth peeling (link below) relies on both fixed-function Z and a depth texture read to "fake" having a second depth buffer .................

I was actually just writing to say that what I had suggested wasn't sensible having thought about it more (and that I realised only 2 depth buffers were in use) and then noticed your message.  I clearly wasn't having a good day yesterday! Apologies for that.

mdagois:
> > The original version of depth peeling (link below) relies on both fixed-function Z and a depth texture read to "fake" having a second depth buffer .................
> 
> I was actually just writing to say that what I had suggested wasn't sensible having thought about it more (and that I realised only 2 depth buffers were in use) and then noticed your message. I clearly wasn't having a good day yesterday! Apologies for that.

No problem at all!
I think you and @asuessenbach made actually good points with those other algorithms.
They should definitely be added through some dropbox in the GUI.

mdagois:
Is there any more issue to fix?
I think I can support more depth peeling algorithms, but I'd rather do it from a base that's already merged.

SaschaWillems (changes requested):
I'm happy with how the samples work.

I have a few remarks though:

By default, object opacity is at 1.0. So in order to actually see transparency, one needs to lower that first. Can you put that at e.g. 0.75 by default instead?

And the sample code does not have a single code comment. The readme does have some information, but it's light-weight. Since we want people to learn from our samples, at least some code comments should be added so people have an easier time following the code.

And a minor one: The readme for the sample has a "Tests" paragraph. Can you remove that? It doesn't fit into that readme.

mdagois:
> By default, object opacity is at 1.0. So in order to actually see transparency, one needs to lower that first. Can you put that at e.g. 0.75 by default instead?

I set it to 0.5.
The background is now visible from the start through the transparent geometry.

> And the sample code does not have a single code comment. The readme does have some information, but it's light-weight. Since we want people to learn from our samples, at least some code comments should be added so people have an easier time following the code.

I added comments in the fragment shaders and in the function that records the command buffer.
Please let me know if that's enough.

> And a minor one: The readme for the sample has a "Tests" paragraph. Can you remove that? It doesn't fit into that readme.

I removed the section.

gary-sweet:
One small thing I realised recently is that you don't actually have any intersecting geometry in the sample. Depth-peeling works even with intersecting geometry so I wonder if it's worth adding a separate object that intersects the first one to show that?

If you're planning to enhance the sample in a later PR it could wait until then. Maybe a checkbox that toggles an intersecting object for example.

mdagois:
> One small thing I realised recently is that you don't actually have any intersecting geometry in the sample. Depth-peeling works even with intersecting geometry so I wonder if it's worth adding a separate object that intersects the first one to show that?
> 
> If you're planning to enhance the sample in a later PR it could wait until then. Maybe a checkbox that toggles an intersecting object for example.

Oops, you are right!
And it is true for the other OIT sample too (oit_linked_lists).
I will make a new PR once this one is in to fix the issue for both sample.

asuessenbach (approved):

SaschaWillems (approved):
Thank you very much. LGTM with the recent changes :)

gary-sweet (approved):

marty-johnson59:
Merging - 3 approvals

PR #887, opened 2024-01-27
Add sample for dynamic rendering local read

SaschaWillems:
## Description

This PR adds a new sample that shows how to replace render passes + multiple sub passes with dynamic rendering and the new `VK_KHR_dynamic_rendering_local_read` extension. The sample contains code paths for both those setups which can be toggled via a pre-processor define.

The sample also comes with a small tutorial that explains some of the differences.

**Note:** This is a copy of the sample from the internal gitlab. With the extension now public, we can continue on this over here.

**Note:** Requires assets from https://github.com/KhronosGroup/Vulkan-Samples-Assets/pull/23, so that PR needs to be merged before or right after this PR has been merged.

Tested on Windows 11 with an NVIDIA RTX 4070 and the latest public Vulkan developer driver.

## General Checklist:

## Sample Checklist

- [ ] Any dependent assets have been merged and published in downstream modules (see https://github.com/KhronosGroup/Vulkan-Samples-Assets/pull/23)


SaschaWillems:
I'm having trouble with getting the clang format right. I'm on VS 2022 (latest version) and I don't know why our clangformat does fail.  

It fails on this:

```cpp
	VkPipelineRenderingCreateInfoKHR pipeline_rendering_create_info{VK_STRUCTURE_TYPE_PIPELINE_RENDERING_CREATE_INFO_KHR};
	pipeline_create_info.pNext = &pipeline_rendering_create_info;
```

Which looks totally fine to me and is the same as the already existing dynamic rendering sample. I can only imagine that our script can't cope with code inside of pre-processor blocks.

clang_format.py only works for me in WSL but it tells me that everything is fine and does not change any of the files...

I noticed that our documentation still says "clang-format 8", my WSL setup is already at 14. I'd rather not downgrade.

Any ideas @tomadamatkinson?

Maybe we can discuss this at our next meeting. Tbh. the clang format stuff is driving me mad for pretty much every sample I write :(

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
```diff
@@ -0,0 +1,1052 @@
...
+		draw_scene(scenes.opaque, cmd, scene_opaque_pass.pipeline_layout);
+
+		// We want to read the input attachments in the next pass, with dynamic rendering local read this requires use of a barrier with the "by region" flag set
+
+		// A new feature of the dynamic rendering local read extension is the ability to use pipeline barriers in the dynamic render pass
+		// to allow framebuffer-local dependencies (i.e. read-after-write) between draw calls using the "by region" flag
+		// So with this barrier we can use the output attachments from the draw call above as input attachments in the next call
+		VkMemoryBarrier2KHR memoryBarrier{};
+		memoryBarrier.sType         = VK_STRUCTURE_TYPE_MEMORY_BARRIER_2_KHR;
+		memoryBarrier.srcStageMask  = VK_PIPELINE_STAGE_2_FRAGMENT_SHADER_BIT;
```
cforfang: Feels like this should be `srcStageMask = VK_PIPELINE_STAGE_2_COLOR_ATTACHMENT_OUTPUT_BIT` and `VK_ACCESS_2_COLOR_ATTACHMENT_WRITE_BIT` in this use-case (unlike the blog, which talks about write/read to storage images -- not write/read of color attachments).

SaschaWillems: Thanks you very much :) Will change that on the internal version first.

cforfang: Similarly, is it possible the `dstAccessMask` here should be `VK_ACCESS_2_COLOR_ATTACHMENT_READ_BIT` instead?

SaschaWillems: Not sure, I think `VK_ACCESS_2_INPUT_ATTACHMENT_READ_BIT` would be correct in here. Will try to have this solved internally by people that know more about this than me ;)

framework/api_vulkan_sample.cpp
```diff
@@ -566,7 +566,10 @@ ApiVulkanSample::~ApiVulkanSample()
 			vkDestroyDescriptorPool(device->get_handle(), descriptor_pool, nullptr);
 		}
 		destroy_command_buffers();
-		vkDestroyRenderPass(device->get_handle(), render_pass, nullptr);
+		if (render_pass != VK_NULL_HANDLE)
```
asuessenbach, line 596: No need to check for `(render_pass != VK_NULL_HANDLE)` here, as vkDestroyRenderPass can handle that.

samples/README.adoc
```diff
@@ -281,6 +281,13 @@ Enables overestimation to generate fragments for every pixel touched instead of
 Demonstrates how to use Dynamic Rendering.
 Read the blog post here for discussion: (https://www.khronos.org/blog/streamlining-render-passes)
 
+=== xref:./extensions/dynamic_rendering_local_read/README.adoc[Dynamic Rendering local read]
+
+*Extension*: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VK_KHR_dynamic_rendering_local_read[`VK_KHR_dynamic_rendering_local_read`]
+
+Demonstrates how to use Dynamic Rendering with local reads to fully replace render passses with multiple subpasses.
+See this https://www.khronos.org/blog/streamlining-subpasses[this blogpost].
```
asuessenbach: One surplus 'this' here.

samples/extensions/dynamic_rendering_local_read/README.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+== Replacing subpasses for local reads
+
+=== Input attachments
+
+Just like local reads in subpasses, dynamic rendering local read also makes use of input attachments. That should make it easy to convert existing code to this new extension. So unless you do advanced things like input attachment reordering, the changes required to add pixel local reads to dynamic rendering are minimal and only affect the application side. There are no changes to the shader interface, so shaders that have been used with renderpasses + subpasses can be used without any changes. Even with dynamic rendering and local reads you use `subpassInput` and `subpassLoad`.
+
+=== Self-Dependencies
+
+With the `dynamicRenderingLocalReads` feature enabled, it#s now possible to use pipeline barriers within dynamic rendering if they include the `VK_DEPENDENCY_BY_REGION_BIT`. Such a barrier makes attachments before the barrier readable as input attachments afterwards. The extension also introduces the new image layout `VK_IMAGE_LAYOUT_RENDERING_LOCAL_READ_KHR` that can be used for storage images and attachments to make writes to those visible via input attachments.
```
asuessenbach: typo: it#s -> it's

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
```diff
@@ -0,0 +1,1052 @@
...
+	set_api_version(VK_API_VERSION_1_2);
+	add_instance_extension(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+	add_device_extension(VK_KHR_DYNAMIC_RENDERING_EXTENSION_NAME);
+	add_device_extension(VK_KHR_DYNAMIC_RENDERING_LOCAL_READ_EXTENSION_NAME);
+	// To simplify barrier setup used for dynamic rendering, we use sync2
+	add_device_extension(VK_KHR_SYNCHRONIZATION_2_EXTENSION_NAME);
+
+	// Dynamic rendering doesn't use render passes
+	// To make sure that framework related classes like the user interface are aware of this, we explicitly st the base class' renderpass to a null handle
+	render_pass = VK_NULL_HANDLE;
```
asuessenbach, line 43: Does that make sense?
The render_pass is created at a much later point in time, no matter what you're setting it here.
Actually, it should be initialized with VK_NULL_HANDLE from the very beginning.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
```diff
@@ -0,0 +1,1052 @@
...
+
+	LOGI("Using dynamic rendering with local read");
+#endif
+}
+
+DynamicRenderingLocalRead::~DynamicRenderingLocalRead()
+{
+	if (device)
+	{
+		for (Pass pass : {scene_opaque_pass, scene_transparent_pass, composition_pass})
```
asuessenbach, line 53: Should probably be `Pass & pass`. Otherwise you're copying those structures.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
```diff
@@ -0,0 +1,1052 @@
...
+		{
+			vkDestroyPipeline(get_device().get_handle(), pass.pipeline, nullptr);
+			vkDestroyPipelineLayout(get_device().get_handle(), pass.pipeline_layout, nullptr);
+			vkDestroyDescriptorSetLayout(get_device().get_handle(), pass.descriptor_set_layout, nullptr);
+		}
+		for (FrameBufferAttachment attachment : {attachments.albedo, attachments.normal, attachments.positionDepth})
```
asuessenbach, line 59: Should probably be: `FrameBufferAttachment & attachment`, which does not compile, and `FrameBufferAttachment const & attachment` does not work with `destroy_attachment`, which takes a non-const reference.
So, you could either change `destroy_attachment` to take a `const &`, or just call it three times without a loop.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
```diff
@@ -0,0 +1,1052 @@
...
+}
+
+void DynamicRenderingLocalRead::request_gpu_features(vkb::PhysicalDevice &gpu)
+{
+	if (gpu.get_features().samplerAnisotropy)
+	{
+		gpu.get_mutable_requested_features().samplerAnisotropy = true;
+	}
+#if defined(USE_DYNAMIC_RENDERING)
+	auto &requested_dynamic_rendering_features            = gpu.request_extension_features<VkPhysicalDeviceDynamicRenderingFeaturesKHR>(VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DYNAMIC_RENDERING_FEATURES_KHR);
```
asuessenbach, line 74: Does that make sense?
We get the `VkPhysicalDeviceDynamicRenderingFeaturesKHR` here, with all supported flags set (well, it's just one). No matter what, we set the `dynamicRendering` flag.
Shouldn't we make it similar to how we handle the `samplerAnisotropy` flag above?

asuessenbach (commented):
Some first comments...

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
asuessenbach: This is not specific to this sample, though.
Might be worth to file an issue on it...

SaschaWillems:
Just a quick note, that this code isn't 100% matching the internal version of the sample. Thanks for the feedback though, I'll take a look at it once the internal issues are solved.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
```diff
@@ -0,0 +1,1052 @@
...
+		{
+			vkDestroyPipeline(get_device().get_handle(), pass.pipeline, nullptr);
+			vkDestroyPipelineLayout(get_device().get_handle(), pass.pipeline_layout, nullptr);
+			vkDestroyDescriptorSetLayout(get_device().get_handle(), pass.descriptor_set_layout, nullptr);
+		}
+		for (FrameBufferAttachment attachment : {attachments.albedo, attachments.normal, attachments.positionDepth})
+		{
+			destroy_attachment(attachment);
+		}
+		vkDestroySampler(get_device().get_handle(), textures.transparent_glass.sampler, nullptr);
```
asuessenbach, line 63: If the model could not be loaded, the sampler here is still undefined, crashing the app.
We would need to initialize it to VK_NULL_HANDLE in api_vulkan_sample.h, line 61.

asuessenbach (commented):
Any reason, you made the dynamic rendering a compile-time switch instead of a run-time switch?

SaschaWillems:
> Any reason, you made the dynamic rendering a compile-time switch instead of a run-time switch?

Makes it easier to maintain the sample. The render pass path is just there so people can easily see the differences in code.

samples/extensions/dynamic_rendering_local_read/README.adoc
SaschaWillems: Fixed in the gitlab version of the sample.

samples/README.adoc
SaschaWillems: Fixed in the gitlab version of the sample.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
SaschaWillems: I'm probably not proficient enough in C++, but that doesn't seem to work.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
SaschaWillems: Same as above. But is it really a problem? It's the destructor anyway, so the copy shouldn't hurt.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
SaschaWillems: I see. It looks like the way this works in the framework, all properties of that struct are set to true, which is probably not what we want. Another thing that needs fixing in the framework :(

I'll open an issue for that.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
SaschaWillems: That's correct, but that's also the case for some of the other samples, so something to consider for framework updates.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
SaschaWillems: See #938

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
SaschaWillems: Fixed in the gitlab version of the sample.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
SaschaWillems: Yes, it makes sense. This is required for the non-render pass part or else the UI will not properly work and trigger lots of validation errors.

I also prefer to have this explicitly set in the code (instead of initializing) so people actually see it when reading through the code.

asuessenbach: Usually the `render_pass` is created in `setup_render_pass` (as it's done in this sample as well, if `USE_DYNAMIC_RENDERING` is not defined), right?
Therefore, I would expect this comment and code in that function.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
asuessenbach: Oops, you're right.
Would need to be `Pass const & pass`.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
asuessenbach: Indeed, same as above.
And it might be a bad habit of mine, but I don't like code that unnecessarily wastes cycles. Even if it's "just" the destructor. Especially if it's relatively easy to resolve. And besides that, that `destroy_attachment` is not only called in the destructor.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
asuessenbach: That is, you're going to file an issue against the framework to initialize the sampler?

SaschaWillems:
@tomadamatkinson : I'm struggling to get this to pass all checks. The pre-commit clang format step did something but it still fails, and I don't have a clue how to fix it. Running the clang format script manually shows some files, but nothing is actually changed.

I also don't know why the doxgen checks are failing now :(

Also no idea why Android is failing all of a sudden :(

SaschaWillems:
And it looks like I made the mistake of committing and pushing files with merge conflict info still present :/

Maybe I should refrain from adding new samples until we are done with the framework rework.

framework/api_vulkan_sample.cpp
```diff
@@ -1,3 +1,4 @@
+/* Copyright (c) 2019-2024, Sascha Willems
```
asuessenbach: You've doubled the Copyright comment here, resulting in doxygen failing.

And the Android build seems to struggle with the same issue.

And so does the Clang-Tidy check.

SaschaWillems: Thanks 👍🏻  Most probably happened while trying to merge

SaschaWillems:
Still fails with clang format and I have no clue why. I guess it's because I'm making heavy use of compiler directives? Running the script locally lists some files, but doesn't help with the problem sadly:

image

This list of files doesn't match with the ones from the CI failure.

It looks like it did something, but no local changes or commit are generated.

SaschaWillems:
Copyright checks now also fail for a file that isn't even touched by this PR 😟

tomadamatkinson:
Hey @SaschaWillems i tried locally with `pre-commit run clang-format --files $(git diff main --name-only)`

image

Did you install pre-commit to validate your commits? It will pull and run the correct version of clang-format for you

SaschaWillems:
Five of the six files in your screenshot aren't even touched by this PR.

And I actually did disable pre-commit due to another problem, will try and check what happens with pre-commit enabled.

samples/README.adoc
```diff
@@ -44,494 +44,3 @@ include::./extensions/README.adoc[]
 
 [[tooling-samples]]
 include::./tooling/README.adoc[]
-
```
asuessenbach, line 47: You removed more or less the complete readme?!?

SaschaWillems:
@tomadamatkinson : Stupid question, but is it possible to manually run pre-commit? I don't see any mention of this in our docs.

SaschaWillems:
@tomadamatkinson: Okay, so I need to enable pre-commit explicitly for each folder. I work on branches in separate folders. Doing that and running your command I see it changes at least one of the files from the failure.

BUT, the changes make no sense to me, e.g. this. Here it wants to remove the whole alignment of the assignments that is almost 100% similar to the other path in the define

image

This doesn't look correct either:

image

(Compare above code with the one below it wants to change)

It seems like it has troubles with code in defines.

That doesn't feel right :/

SaschaWillems:
The main issue I have with pre-commit: It won't let me push my manual changes anymore with adding it's own changes. That's probably the reason I had to disable it for this branch :/

jeroenbakker-atmind:
Have tried this on linux, but there is only support for this example on NVIDIA platforms at this moment. Will try to test on windows later this week.

When building I had to turn off the opencl interop example due to out of sync changes.

SaschaWillems:
Sorry, totally forgot to merge main into this. I've merged and pushed the branch, so it should compile fine now.

And thanks for testing :)

jeroenbakker-atmind:
Works without any issue on: 

```
Operating system: Windows-10-10.0.19045-SP0 64 Bits
Graphics card: AMD Radeon RX 5700 ATI Technologies Inc. 24.1.1.240110
```
```
[info] Using dynamic rendering with local read
[info] Vulkan debug utils enabled (VK_EXT_debug_utils)
[info] Extension VK_KHR_get_physical_device_properties2 found, enabling it
[info] Extension VK_KHR_win32_surface found, enabling it
[info] Extension VK_EXT_debug_utils found, enabling it
[info] Found GPU: AMD Radeon RX 5700
[info] Selected GPU: AMD Radeon RX 5700
[info] Dedicated Allocation enabled
[info] Device supports the following requested extensions:
[info]          VK_KHR_get_memory_requirements2
[info]          VK_KHR_dedicated_allocation
[info]          VK_KHR_dynamic_rendering
[info]          VK_KHR_synchronization2
[info]          VK_KHR_dynamic_rendering_local_read       
[info]          VK_KHR_swapchain
[info] Surface supports the following surface formats:
[info]          R8G8B8A8Unorm, SrgbNonlinear
[info]          B8G8R8A8Unorm, SrgbNonlinear
[info]          R8G8B8A8Srgb, SrgbNonlinear
[info]          B8G8R8A8Srgb, SrgbNonlinear
[info] Surface supports the following present modes:
[info]          Immediate
[info]          Fifo
[info]          FifoRelaxed
[warning] (HPPSwapchain) Image extent (0, 0) not supported. Selecting (1280, 720).
[warning] (HPPSwapchain) Surface format (Undefined, SrgbNonlinear) not supported. Selecting (B8G8R8A8Srgb, SrgbNonlinear).
[info] (HPPSwapchain) Image usage flags: TransferSrc ColorAttachment
[warning] (HPPSwapchain) Composite alpha 'Inherit' not supported. Selecting 'Opaque.
[warning] (HPPSwapchain) Present mode 'Mailbox' not supported. Selecting 'Fifo'.
[info] (HPPSwapchain) Surface format selected: B8G8R8A8Srgb, SrgbNonlinear
[info] Depth format selected: VK_FORMAT_D32_SFLOAT
[info] Depth format selected: VK_FORMAT_D32_SFLOAT
[info] Depth format selected: VK_FORMAT_D32_SFLOAT
[info] Depth format selected: VK_FORMAT_D32_SFLOAT
[info] Time spent loading images: 0.000756 seconds across 16 threads.
[info] Time spent loading images: 0.000330 seconds across 16 threads.
```

If you need more tests on other platforms/architectures, please let me know

asuessenbach:
Silly question... where do I find the assets used with this sample (`scenes/subpass_scene_opaque.gltf`, `scenes/subpass_scene_transparent.gltf`, and `textures/transparent_glass_rgba.ktx`)?

jeroenbakker-atmind:
@asuessenbach you can checkout the next PR https://github.com/KhronosGroup/Vulkan-Samples-Assets/pull/23

asuessenbach:
@jeroenbakker-atmind Thanks, now it runs fine here as well: Win10, NVIDIA RTX A3000 Laptop

asuessenbach:
The gui is only available when `USE_DYNAMIC_RENDERING` is not defined. Is that for a specific reason?

When I just prepare the gui even if `USE_DYNAMIC_RENDERING` is defined, I get a VL error:
`Validation Error: [ VUID-VkGraphicsPipelineCreateInfo-renderPass-06055 ] | MessageID = 0xcfde2b97 | vkCreateGraphicsPipelines(): pCreateInfos[0].pColorBlendState->attachmentCount is 1, but VkPipelineRenderingCreateInfo::colorAttachmentCount is zero because the pNext chain was not included. The Vulkan spec states: If renderPass is VK_NULL_HANDLE, pColorBlendState is not dynamic, and the pipeline is being created with fragment output interface state, pColorBlendState->attachmentCount must be equal to VkPipelineRenderingCreateInfo::colorAttachmentCount (https://vulkan.lunarg.com/doc/view/1.3.283.0/windows/1.3-extensions/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-renderPass-06055)`
Will that be resolved and the GUI could be available with some newer validation layer?

SaschaWillems:
Yes, that layer message is the reason the GUI is disabled by default. For the GUI to work with that sample I would have to do some extensive changes to the GUI framework class. I may do that once this sample has been merged in a separate PR.

asuessenbach (dismissed):

jeroenbakker-atmind:
I can still test on NVIDIA GTX 980 + 760 IIRC only the 980 will support this.

pdaniell-nv (commented):
I noticed the following running this with Vulkan SDK 1.3.290:
```
[error] 269944751 - VUID-RuntimeSpirv-NonWritable-06340: Validation Error: [ VUID-RuntimeSpirv-NonWritable-06340 ] Object 0: handle = 0xc25f26000000009c, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0x101707af | vkCreateGraphicsPipelines(): pCreateInfos[0].pStages[1] SPIR-V (VK_SHADER_STAGE_FRAGMENT_BIT) uses descriptor [Set 0, Binding 3, variable "LightsBuffer"] (type VK_DESCRIPTOR_TYPE_STORAGE_BUFFER or VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC) which is not marked with NonWritable, but fragmentStoresAndAtomics was not enabled. The Vulkan spec states: If fragmentStoresAndAtomics is not enabled, then all storage image, storage texel buffer, and storage buffer variables in the fragment stage must be decorated with the NonWritable decoration (https://vulkan.lunarg.com/doc/view/1.3.290.0/windows/1.3-extensions/vkspec.html#VUID-RuntimeSpirv-NonWritable-06340)
```

SaschaWillems:
Thank you very much for testing. That validation message seems to have been introduced by a recent SDK. I added the readonly attribute to the buffer mentioned in that message.

asuessenbach (approved):

pdaniell-nv (approved):

gpx1000 (approved):

marty-johnson59:
Merging - 3 approvals

PR #888, opened 2024-01-31
use correct address computation for image descriptor buffer

4DA:
Image descriptor storage pointer uses size and offset from uniform binding descriptor. Use correct image descriptor offset & size instead.


SaschaWillems:
Thanks for your PR. Build fails though due to copyright checks. If you update a file, you also need to update the copyright date to the current year. Once you fix that we'll start reviewing it.

SaschaWillems (approved):
LGTM. Thanks for spotting this.

Issue #860, opened 2023-12-04
Store Assets in Raw and Compressed formats


tomadamatkinson:
It may be an idea to store assets for the samples as raw alongside their compressed counter parts. Or to move the compression of assets into the compilation step of the project. This would allow us to customise how we prepare assets for a project or sample and reduce the time  we spend decompressing ASTC texture for incompatible platforms

SaschaWillems:
We already have a sample for basis universal compression, and the KTX library supports that. So we could store textures in that format and then transcode at runtime. That's much faster than the current ASTC decoding on desktop.

SaschaWillems:
We should def take a look at this. Esp. when doing a batch run in debug mode, some samples that load scenes using ASTC compressed images take ages to load, even on a fairly fast CPU. On my setup, this makes a full batch run take more than 20 minutes, which is a lot since it's something we require when doing framework changes.

SaschaWillems:
Will be fixed with upcoming PR for #1202

PR #894, opened 2024-02-07
Move vkb::filesystem to components

tomadamatkinson:
## Description

Simplifies our filesystem implementation by using std::filesystem underneath. This is the filesystem used in #846 which can be easily used in different executables to perform small tasks (See `shader_reflector` and `variant_reflector`. Moved to components and can be used using `vkb__filesystem`. Filesystem is influenced by the decisions made on the framework/2.0 branch

The existing "legacy" api is kept to maintain compatibility with  samples. In the future asset loaders may choose to use the filesystem directly with `vkb::filesystem::get()`

Merging this will help simplify the review on #846

Tested on MacOS (M1), Linux (Ubuntu 22.04), Windows (11) and Android (Pixel 3). Samples work as expected

## General Checklist:

## Sample Checklist

- [ ] For new samples, I have added a paragraph with a summary to the appropriate chapter in the [samples readme](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/samples/README.adoc)


SaschaWillems (dismissed):
Thank you very much for reworking this. Looks good and works fine on my setup 👍🏻

gary-sweet:
I'm seeing a couple of build problems. The first is a redefinition of LOGE:

```
[100%] Building CXX object app/CMakeFiles/vulkan_samples.dir/main.cpp.o
In file included from Vulkan-Samples/components/filesystem/include/filesystem/filesystem.hpp:26,
                 from Vulkan-Samples/app/main.cpp:23:
Vulkan-Samples/components/core/include/core/util/logging.hpp:28: warning: "LOGE" redefined
   28 | #define LOGE(...) spdlog::error("{}", fmt::format(__VA_ARGS__));
      | 
In file included from Vulkan-Samples/app/main.cpp:18:
Vulkan-Samples/framework/common/logging.h:35: note: this is the location of the previous definition
   35 | #define LOGE(...) spdlog::error("[{}:{}] {}", __FILENAME__, __LINE__, fmt::format(__VA_ARGS__));
```

gary-sweet:
The second issue is that `std::filesystem` isn't available before gcc8 unless you explicitly include `stdc++fs` as a link library.

We've had to work around that a number of times using something like:
```
if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
   target_link_libraries(${TGT} PRIVATE stdc++fs)
endif()

```

gary-sweet (changes requested):
See my front-page comments

components/filesystem/include/filesystem/filesystem.hpp
```diff
@@ -0,0 +1,84 @@
...
+	virtual const Path &external_storage_directory() const = 0;
+	virtual const Path &temp_directory() const             = 0;
+
+	void write_file(const Path &path, const std::string &data);
+
+	// Read the entire file into a string
+	std::string read_file_string(const Path &path);
+
+	// Read the entire file into a vector of bytes
+	std::vector<uint8_t> read_file_binary(const Path &path);
```
jherico, line 66:
```suggestion
	std::vector<uint8_t> read_file_binary(const Path &path);
	void with_file_contents(const Path& path, const std::function<void(const uint8_t*, size_t)>& handler);
```
In my experience, most applications only need the asset data for the scope of the call where it's read. While read_binary_file returns a std::vector and thus always allocates heap space, the implementation of something like with_file_contents is free to give a pointer to a memory mapped file, or something like the result of an AAsset_getBuffer(...) call on Android, which can lower the amount of actual heap space used in the application. Obviously anything going into a staging buffer will consume host memory anyway, but at least this would potentially cut that in half.

components/filesystem/src/std_filesystem.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+	if (ec)
+	{
+		throw std::runtime_error("Failed to create directory");
+	}
+
+	return !ec;
+}
+
+std::vector<uint8_t> StdFileSystem::read_chunk(const Path &path, size_t start, size_t offset)
```
jherico: I find the naming here very confusing.  `start` is being used as an offset into the file, while `offset` is being used how I would expect something named `size` or `count`.

framework/platform/filesystem.cpp
jherico, line 1: You've removed this file, but not the corresponding `filesystem.hpp` file.  You could also get rid of most of the other tiny changes in this PR by changing `platform/filesystem.hpp` to just contain `filesystem/legacy.h`.

components/filesystem/include/filesystem/filesystem.hpp
jherico: I tried this on a local branch starting from this PR branch. I updated the GLTF loader code and the image loading code to both use loading from memory mapped files instead of from vectors.  I then ran `time ./build/app/bin/release/AMD64/vulkan_samples.exe --stop-after-frame 100 sample wait_idle` on both my branch and on the main branch.  

However, load performance was not appreciably different between the two branches.  Therefore unless there are actually issues with memory constrained platforms, I'm guessing that the increased complexity of using memory mapping APIs isn't going to be worth it.

I may try some other platforms later, but it looks as if the file IO streams API has significantly improved in performance on Windows (relative to memory mapped files) since the last time I did work like this.

components/core/src/logging.cpp
```diff
@@ -0,0 +1,52 @@
...
+#endif
+
+namespace vkb
+{
+namespace logging
+{
+void init()
+{
+	// Taken from "spdlog/cfg/env.h" and renamed SPDLOG_LEVEL to VKB_LEVEL
+	auto env_val = spdlog::details::os::getenv("VKB_LEVEL");
```
asuessenbach: Maybe should be named `VKB_LOG_LEVEL`, or so, telling that it's about logging.

tomadamatkinson:
> I'm seeing a couple of build problems. The first is a redefinition of LOGE:
> 
> ```
> [100%] Building CXX object app/CMakeFiles/vulkan_samples.dir/main.cpp.o
> In file included from Vulkan-Samples/components/filesystem/include/filesystem/filesystem.hpp:26,
>                  from Vulkan-Samples/app/main.cpp:23:
> Vulkan-Samples/components/core/include/core/util/logging.hpp:28: warning: "LOGE" redefined
>    28 | #define LOGE(...) spdlog::error("{}", fmt::format(__VA_ARGS__));
>       | 
> In file included from Vulkan-Samples/app/main.cpp:18:
> Vulkan-Samples/framework/common/logging.h:35: note: this is the location of the previous definition
>    35 | #define LOGE(...) spdlog::error("[{}:{}] {}", __FILENAME__, __LINE__, fmt::format(__VA_ARGS__));
> ```

Due to a conflict with `logging.h` and `logging.hpp` I will remove the original version in favour of the one that is ported to the components folders

components/filesystem/include/filesystem/filesystem.hpp
tomadamatkinson: Thank you for benchmarking the differences! From what i have found in the past most of our slow loading times comes from decompressing ASTC if a platform doesnt support this format. See #860 as a reference to this topic

tomadamatkinson: Resolving this for now but definitely open to adding a memory mapped alternative

components/core/src/logging.cpp
tomadamatkinson: Agreed, resolved!

components/filesystem/src/std_filesystem.cpp
tomadamatkinson: Ah this is me rushing. Resolved with `read_chunk(const Path &path, size_t offset, size_t count)`

framework/platform/filesystem.cpp
tomadamatkinson: I'm a bit torn on this point. I agree that this will reduce the changes on this PR. However, the end goal is to encourage future contributions to be made in the more modular version of the framework. Using more of the modular framework. Explicitly pointing out that the existing code using the older filesystem is legacy may help here

Happy to change... Dreading reverting those changes. I dont think i separated them into an isolated commit :face_exhaling:

tomadamatkinson:
Thanks all for taking the time to review! Branch is updated

asuessenbach (dismissed):

gary-sweet (dismissed):
Thanks

components/filesystem/src/std_filesystem.cpp
```diff
@@ -0,0 +1,154 @@
...
+{
+	auto stat = stat_file(path);
+	return stat.is_directory;
+}
+
+bool StdFileSystem::exists(const Path &path)
+{
+	auto stat = stat_file(path);
+	return stat.is_file || stat.is_directory;
+}
```
jherico, line 72: Probably not critical since the usage is confined to a pretty small set of use cases within this repository, but...

All of these end up calculating the size which isn't used.  It seems like they might be better implmented using `std::filesystem::status` directly.

in `exists` in particular, we could get the status and check explicitly for `type() == std::filesystem::file_type::not_found`, which would then return the proper result if the path given were a symlink or other non-file, non-directory type.  As it stands, this will just return false for devices, sockets, symlinks, etc.

components/filesystem/README.adoc
SaschaWillems, line 1: Is this intentionally empty?

samples/performance/multithreading_render_passes/multithreading_render_passes.cpp
SaschaWillems, line 1: This file is displayed as completely changed for me. Is that intentional?

SaschaWillems (approved):
Code LGTM.

I tested on Windows 11 with an RTX 4070 and on Android with a Google Pixel 4a.

I only have two minor comments not directly related to the code itself.

asuessenbach (approved):

components/filesystem/README.adoc
tomadamatkinson: Happy to follow up with in depth documentation on why VFS are useful and discuss some open source alternatives or examples of good VFS. I did originally have some docs in here but I think they where too opinionated. It would be good to have something based on more thorough research

samples/performance/multithreading_render_passes/multithreading_render_passes.cpp
tomadamatkinson: Not sure why this is diffed like this... I only changed the headers. Shall i check this out from main and try again?

jherico: That's one of 6 files that are currently checked in as CRLF, only 1 of which should _actually_ be CRLF

image

components/filesystem/README.adoc
SaschaWillems: Perfectly fine. Just noticed the empty file and was wondering why it was there.

gary-sweet (approved):

Issue #896, opened 2024-02-07
CMake configuration order of operations is incorrect


jherico:
The [code](https://github.com/KhronosGroup/Vulkan-Samples/blob/5e4e3d20b3a86f7814e026042341cf34ef530edd/CMakeLists.txt#L31) near the beginning of the root CMakeLists.txt reads as follows

```cmake
if(DEFINED CMAKE_C_COMPILER_LAUNCHER)
if(WIN32)
    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
        string(REPLACE "/Zi" "/Z7" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}")
        string(REPLACE "/Zi" "/Z7" CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG}")
    elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
        string(REPLACE "/Zi" "/Z7" CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE}")
        string(REPLACE "/Zi" "/Z7" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
    elseif(CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
        string(REPLACE "/Zi" "/Z7" CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
        string(REPLACE "/Zi" "/Z7" CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO}")
    endif()
endif()
endif()
```

However, these lines are all placed before `project(vulkan_samples)`, which is when the compiler is detected and all of these `CMAKE_<LANG>_FLAGS_<TYPE>` flags are initialized.

Additionally, MSVC is a multi-config build system, so `CMAKE_BUILD_TYPE` is almost never populated at configure time.  When the project is actually built, for instance with `cmake --build`, this code won't even be processed because it's only triggered at configure time.

SaschaWillems:
Fixed with #897.

@jherico : If you put a line like "Fixes #896" into the PRs description, any related issue will be closed automatically when the respective PR is merged.

PR #897, opened 2024-02-07
Fix sccache usage in Windows CI builds

jherico:
## Description

This corrects a couple problems preventing the Windows CI builds from using sccache.  

Fix for #896 and reduces windows build significantly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
         name: Configure and build
         shell: pwsh
         run: |
-          cmake -B"build/${{ matrix.platform }}" -G"NMake Makefiles" -DVKB_BUILD_TESTS=ON -DVKB_BUILD_SAMPLES=ON -DCMAKE_C_COMPILER_LAUNCHER=C:\\Users\\runneradmin\\.cargo\\bin\\sccache -DCMAKE_CXX_COMPILER_LAUNCHER=C:\\Users\\runneradmin\\.cargo\\bin\\sccache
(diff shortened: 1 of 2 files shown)


jherico:
@gpx1000 You mentioned having trouble getting sccache working in Windows CI builds.  I was able to track it down after I noticed something odd in the beginning of the CMake config.  

The basic issue was twofold...  

* Although sccache doesn't work with the Visual Studio generator, some sources online seem to suggest it works with the "NMake Makefiles" generator, but as far as I can tell it does not.  Switching to Ninja works though.
* The online commentary about the need to use `/Z7` instead of `/Zi` may be accurate, but the mechanism that was being used didn't work for the reasons I noted in #896

tomadamatkinson (dismissed):
Nice! Looks like this cuts the Windows build down from ~30minutes to ~7minutes in the CI

jherico:
It occurred to me that I should revert the change adding messaging about lack of ccache or sccache since they don't work with the most commonly used Windows generators.  I could gate them behind a `if (CMAKE_GENERATOR MATCHES "Visual Studio")`, but I figured it's best to just leave it closer to as is.

gpx1000:
I always kept it in Ninja build for Windows as I wanted to ensure the windows build system was using at least NMake.  SCCache was supposed to work with NMake but I never got that working right.  I think it's far better to just bite the bullet and use Ninja as you have here.  I don't think at first blush that the macos build failure in CI is due to the changes here.  lemme know if you need help changing it.
Thanks for the contribution; greatly appreciated!  And great catches of bugs.

SaschaWillems (approved):
Thank you very much for this PR. Works fine for me on Windows 11 👍🏻

gpx1000 (approved):
lgtm; thanks for the contribution!

tomadamatkinson (approved):

PR #901, opened 2024-02-08
Update to most recent production release of VMA

jherico:
## Description

This updates the VMA submodule commit to the one tagged as release version 3.0.1, and makes changes to the codebase to account for the difference in the VMA API between the prior 3.0.0-dev version and the current release version.  In particular

* Initialization of VMA now only requires the instance and device `vkGetXXXProcAddress` functions
* The use of `VMA_ALLOCATION_CREATE_MAPPED_BIT` doesn't guarantee mappable memory unless combined with `VMA_ALLOCATION_CREATE_HOST_ACCESS_RANDOM_BIT` or `VMA_ALLOCATION_CREATE_HOST_ACCESS_SEQUENTIAL_WRITE_BIT`, so the default flags for the Buffer types are correspondingly updated
* There are a number of additional extensions VMA can make use of, if they're available and enabled, so these were added to the corresponding Device classes during initialization.  Most of these flags have no impact if the requested Vulkan version is 1.2 or 1.3 and the corresponding extension has been folded into core.  In those cases the functionality will be enabled by default.
* The `vmaCalculateStats` function is now `vmaCalculateStatistics` and the result structure is slightly changed
* The include path for the VMA has changed from `<VMA_DIR>/src` to `<VMA_DIR>/include`.  In order to suppress a large number of clang-tidy warnings from the header, the include path has now been marked as `SYSTEM` in CMake.

Fixes #899

## General Checklist:

Please ensure the following points are checked:

- [X] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Code-Style)
- [X] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Copyright-Notice-and-License-Template)
- [X] I have commented any added functions (in line with Doxygen)
- [X] I have commented any code that could be hard to understand
- [X] My changes do not add any new compiler warnings
- [X] My changes do not add any new validation layer errors or warnings
- [X] I have used existing framework/helper functions where possible
- [X] My changes do not add any regressions
- [x] I have tested every sample to ensure everything runs correctly
- [X] This PR describes the scope and expected impact of the changes I am making

 Note: The Samples CI runs a number of checks including:
 - [X] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
[...]

PR #906, opened 2024-02-11
Allocated refactor

### LINKED ISSUE #900: Suggestion: Refactor (HPP)Image & (HPP)Buffer to improve VMA usage and reduce duplicated code.
The `Buffer` and `Image` classes (along with their corresponding HPP versions) have a few things that might benefit from refactoring.

* They share a significant amount of duplicated code.
* The `Buffer` classes have some odd choices for the defaults in their constructors
  * The Buffer classes don't have any default for the `VmaMemoryUsage` but the do default `VmaAllocationCreateFlags` to `VMA_ALLOCATION_CREATE_MAPPED_BIT` as mentioned.  This is actually inverted.  The `VmaAllocationCreateFlags` should probably NOT have a default while the `VmaMemoryUsage` should almost always default to `VMA_MEMORY_USAGE_AUTO`.  
  * This may be the underlying cause of #848
* The `Image` constructors don't expose the `VmaAllocationCreateFlags` but are already unwieldy in terms of the number of arguments
* _Maybe_ Images that act have attachment usage in their usage flags should get the `VMA_ALLOCATION_CREATE_DEDICATED_MEMORY_BIT` flag automatically? 

I propose a change.

First, create a shared base class for anything that uses `VmaAllocation` for storage. Since VMA doesn't have a C++ set of bindings, the base class could be shared by both the C++ and  C classes (the class would include a template argument to specify whether it derived from `HPPVulkanResource` or `VulkanResource`).  All of the VMA specific members could move into this base class, including all the mapping, unmapping and flushing related code.

Second, fixing the `Buffer` constructor to reverse the arguments to give one a new default and take the other away would create a large burden for a single change, and adding any arguments to the `Image` ctors would probably be a mistake.

As an alternative, I suggest that a builder pattern be added.  Instead a long list of ctor arguments, a set of reasonable defaults inside a builder object could be created, and then customized as much or as little as desired in use cases.  The `Image` and `Buffer` classes could then accept as their primary constructor a builder object.

Again, a common base class could be utilized, so that you'd have builder functionality that was VMA specific centralized, looking something like this

```c++
template <typename BuilderType>
struct AllocatedBuilder
{
	VmaAllocationCreateInfo alloc_create_info;

	AllocatedBuilder()
	{
		alloc_create_info       = {};
		alloc_create_info.usage = VMA_MEMORY_USAGE_AUTO;
	};

	BuilderType &set_vma_flags(VmaAllocationCreateFlags flags)
	{
		alloc_create_info.flags = flags;
		return *static_cast<BuilderType *>(this);
	}
        ...
```

And then for each subclass you'd have an individual builder like this for for example for the `HPPImage`

```c++
struct HPPImageBuilder : public AllocatedBuilder<HPPImageBuilder>
{
	vk::ImageCreateInfo create_info;

	HPPImageBuilder(vk::Extent3D const &extent, vk::Format format = vk::Format::eR8G8B8A8Unorm)
	{
		create_info.extent = extent;
		create_info.format = format;
		// Better reasonable defaults than vk::ImageCreateInfo default ctor
		create_info.mipLevels   = 1;
		create_info.arrayLayers = 1;
		create_info.imageType   = vk::ImageType::e2D;
	}

	HPPImageBuilder(vk::Extent2D const &extent, vk::Format format = vk::Format::eR8G8B8A8Unorm) :
	    HPPImageBuilder(vk::Extent3D{extent.width, extent.height, 1}, format)
	{
	}

	HPPImageBuilder &set_image_type(vk::ImageType type)
	{
		create_info.imageType = type;
		return *this;
	}

	HPPImageBuilder &set_array_layers(uint32_t layers)
	{
		create_info.arrayLayers = layers;
		return *this;
	}
	...
```

The main `HPPImage` constructor could then look like `HPPImage(HPPDevice &device, HPPImageBuilder const & builder);`

A given use case for the object might change from this...

```c++
		auto color_image = vkb::core::HPPImage{device,
		                                       vk::Extent3D{surface_extent.width, surface_extent.height, 1},
		                                       DEFAULT_VK_FORMAT,        // We can use any format here that we like
		                                       vk::ImageUsageFlagBits::eColorAttachment | vk::ImageUsageFlagBits::eTransferSrc,
		                                       VMA_MEMORY_USAGE_GPU_ONLY};
```

to this

```c++
		// We can use any format here that we like
		core::HPPImageBuilder builder(surface_extent, DEFAULT_VK_FORMAT);
		builder.set_usage(vk::ImageUsageFlagBits::eColorAttachment | vk::ImageUsageFlagBits::eTransferSrc);
		auto color_image = vkb::core::HPPImage{device, builder};
```

In this case I've included a `core::HPPImageBuilder` ctor that accepts an `Extent2D` and automatically extends it to the `Extent3D` inside the builder class.

### LINKED ISSUE #941: Missing memory flushes
While the `Buffer::update` functions combine `map`, `flush` and `unmap` operations for moving system memory into mapped memory, there are a number of places in the code that use `map` and `unmap` with no call to `flush`.  For example:

https://github.com/KhronosGroup/Vulkan-Samples/blob/6fbdb437e2202c7b8a6f453dae017081310a0938/samples/extensions/descriptor_indexing/descriptor_indexing.cpp#L432

AFAIK, there's no guarantee of any coherent memory pools in Vulkan so unless the flag is checked there should always be a `flush` preceding an `unmap`

jherico:
## Description

~This PR is currently built on top of #907 and #901 and should not be merged until they're approved and merged.~

Migrate the code common to the objects using the VMA library to a single location, mostly the new `allocated.h` header and template types.  The four classes, `vkb::core::Buffer`, `vkb::core::Image`, `vkb::core::HPPBuffer` and ``vkb::core::HPPImage` are now all implemented in terms of common base classes.

The new classes prefer the use of a builder pattern for construction, with reasonable defaults for some values set in the image builder types, reducing the cognitive load of using them as well as the complexity of the image constructors. In a future PR the existing ctors will be marked deprecated and all the samples referencing them will be updated to use the builder pattern.

The buffer types have new static methods intended to be used to create short-lived staging buffers containing data.  These static methods include convenience signatures to make it easy to upload containers full of data... so for instance, this code

```
core::Buffer stage_buffer{device,
                          vertex_data.size() * sizeof(Vertex),
                          VK_BUFFER_USAGE_TRANSFER_SRC_BIT,
                          VMA_MEMORY_USAGE_CPU_ONLY};

stage_buffer.update(vertex_data.data(), vertex_data.size() * sizeof(Vertex));
```

reduces to 

```
core::Buffer stage_buffer = core::Buffer::create_staging_buffer(device, aligned_vertex_data);
```

Fixes #900 

I'm struggling to 100% test the samples as `fragment_shading_rate_dynamic` currently crashes on my desktop, even on the main branch and the hlsl shader examples both produce validation warnings.

Fixed #941 

Partial fix the the example noted in this bug and at least one other place, but there should probably be an assert if unmap is called without flush ever having been called (regardless of whether the memory is coherent).

## General Checklist:


gary-sweet (changes requested):
I'm seeing these validation errors multiple times when I run virtually any sample, followed by a driver crash.

`[error] [framework/core/instance.cpp:50] 357932245 - VUID-VkImageSubresourceRange-levelCount-01720: Validation Error: [ VUID-VkImageSubresourceRange-levelCount-01720 ] Object 0: handle = 0xa000000000a, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x15559cd5 | vkCreateImageView(): pCreateInfo->subresourceRange.levelCount is zero. The Vulkan spec states: If levelCount is not VK_REMAINING_MIP_LEVELS, it must be greater than 0 (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageSubresourceRange-levelCount-01720)`

`[error] [framework/core/instance.cpp:50] 1022674482 - VUID-VkImageSubresourceRange-layerCount-01721: Validation Error: [ VUID-VkImageSubresourceRange-layerCount-01721 ] Object 0: handle = 0xa000000000a, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x3cf4c632 | vkCreateImageView(): pCreateInfo->subresourceRange.layerCount is zero. The Vulkan spec states: If layerCount is not VK_REMAINING_ARRAY_LAYERS, it must be greater than 0 (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageSubresourceRange-layerCount-01721)`

`[error] [framework/core/instance.cpp:50] -1968073940 - VUID-VkImageViewCreateInfo-imageViewType-04973: Validation Error: [ VUID-VkImageViewCreateInfo-imageViewType-04973 ] Object 0: handle = 0xa000000000a, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x8ab1932c | vkCreateImageView(): pCreateInfo->subresourceRange.layerCount (0) must be 1 when using viewType VK_IMAGE_VIEW_TYPE_2D (try looking into VK_IMAGE_VIEW_TYPE_*_ARRAY). The Vulkan spec states: If viewType is VK_IMAGE_VIEW_TYPE_1D, VK_IMAGE_VIEW_TYPE_2D, or VK_IMAGE_VIEW_TYPE_3D; and subresourceRange.layerCount is not VK_REMAINING_ARRAY_LAYERS, then subresourceRange.layerCount must be 1 (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageViewCreateInfo-imageViewType-04973)`

SaschaWillems:
This sadly doesn't build for me, so I can't test it:

image

I'm on Windows with VS2022 17.8.7

jherico:
@SaschaWillems @gary-sweet should be ready for testing and review.

SaschaWillems (changes requested):
Builds fine for me now, but it breaks some samples. E.g. the ray tracing extended and reflection samples now crashes for me following this error:

```
[error] [framework\core\instance.cpp:50] -330527817 - VUID-vkMapMemory-memory-00682: Validation Error: [ VUID-vkMapMemory-memory-00682 ] Object 0: handle = 0x84785e00000001b9, type = VK_OBJECT_TYPE_DEVICE_MEMORY; | MessageID = 0xec4c8bb7 | vkMapMemory():  Mapping memory without VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT set. Memory has type 0 which has properties VkMemoryPropertyFlags(0). The Vulkan spec states: memory must have been created with a memory type that reports VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkMapMemory-memory-00682)
[error] [framework\core/allocated.h:207] Detected Vulkan error: ERROR_MEMORY_MAP_FAILED
```

Probably need to to test all samples and also on different platforms to make sure we don't break anything with this.

jherico:
@SaschaWillems I've fixed the ray tracing examples to use the builder pattern but retain the usage flags used in the old code.

Regarding the buffers in question `raygen_shader_binding_table`, `miss_shader_binding_table`, `hit_shader_binding_table`, they're currently created, and populated in a way that requires them to be host visible, but they're only written to once. The current logic uses the deprecated VMA usage enum of `VMA_MEMORY_USAGE_CPU_TO_GPU`, which I believe tries to allocate memory that is device local but host visible if the card provides such a type, but will revert to host visible otherwise.  That's typical for things like uniform buffers that get written every frame, but seems strange here.

I guess my question is, since these buffers are only written once, why not just use straight up `DEVICE_LOCAL` memory and a staging buffer to populate them?  

> Probably need to to test all samples and also on different platforms to make sure we don't break anything with this.

I've now tested all samples and verified that they all produce output and no new validation errors as far as I can tell (the HLSL examples produce validation errors on main, and the `fragment_shader_rate_dynamic` sample crashes on main).  However, I only really have my main system of Windows 10 and an nVidia card to test on.

gary-sweet:
This all seems to work ok on my platform now, but I notice there is a clang format check failure so I can't approve it yet.

samples/extensions/ray_tracing_basic/ray_tracing_basic.cpp
```diff
@@ -79,46 +78,18 @@ void RaytracingBasic::request_gpu_features(vkb::PhysicalDevice &gpu)
...
-	color_image_view.format                          = VK_FORMAT_B8G8R8A8_UNORM;
-	color_image_view.subresourceRange                = {};
-	color_image_view.subresourceRange.aspectMask     = VK_IMAGE_ASPECT_COLOR_BIT;
-	color_image_view.subresourceRange.baseMipLevel   = 0;
-	color_image_view.subresourceRange.levelCount     = 1;
-	color_image_view.subresourceRange.baseArrayLayer = 0;
-	color_image_view.subresourceRange.layerCount     = 1;
-	color_image_view.image                           = storage_image.image;
-	VK_CHECK(vkCreateImageView(get_device().get_handle(), &color_image_view, nullptr, &storage_image.view));
+	vkb::core::ImageBuilder image_builder{{width, height, 1}};
```
SaschaWillems: I'd prefer to roll this back. The whole reason I made this (sample) so explicit, was so that people could easily adopt that code. These changes add a new layer of abstraction (on top of the framework) that would make it harder for people to follow/adopt this sample.

samples/extensions/ray_tracing_basic/ray_tracing_basic.cpp
```diff
@@ -129,65 +100,6 @@ void RaytracingBasic::create_storage_image()
 	get_device().flush_command_buffer(command_buffer, queue);
 }
 
-/*
-    Gets the device address from a buffer that's needed in many places during the ray tracing setup
-*/
-uint64_t RaytracingBasic::get_buffer_device_address(VkBuffer buffer)
```
SaschaWillems: Same as above. There is a reason I made all of this very explicit.

SaschaWillems (changes requested):
I added some comments to one of the samples I did. In general I'm fine with refactoring to make things better/easier, but such refactors shouldn't do fundamental changes to existing samples mainly for the sake of simplifying/unifying things. With many of the samples I write I'm deliberately explicit so people don't need to poke around the framework to see how things are done. And this PR undoes at least some of that, which I'm not too happy with.

SaschaWillems:
And while I'm in general a fan of the builder pattern, I'm kinda skeptical. Afaik we don't use it elsewhere and it kinda feels odd to introduce (yet) another pattern for an already convoluted and hard to maintain framework. Not sure on this, what do the other reviewers think of that approach?

samples/api/oit_linked_lists/oit_linked_lists.cpp
```diff
@@ -250,9 +250,14 @@ void OITLinkedLists::create_gather_pass_objects(const uint32_t width, const uint
 void OITLinkedLists::create_fragment_resources(const uint32_t width, const uint32_t height)
 {
 	{
-		const VkExtent3D image_extent = {width, height, 1};
-		linked_list_head_image        = std::make_unique<vkb::core::Image>(get_device(), image_extent, VK_FORMAT_R32_UINT, VK_IMAGE_USAGE_STORAGE_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT, VMA_MEMORY_USAGE_GPU_ONLY, VK_SAMPLE_COUNT_1_BIT);
-		linked_list_head_image_view   = std::make_unique<vkb::core::ImageView>(*linked_list_head_image, VK_IMAGE_VIEW_TYPE_2D, VK_FORMAT_R32_UINT);
+		vkb::core::ImageBuilder builder{width, height};
```
SaschaWillems: I don't see why or how this is better than before? I personally prefer the way it was.

samples/extensions/fragment_shading_rate_dynamic/fragment_shading_rate_dynamic.cpp
```diff
@@ -115,19 +115,19 @@ void FragmentShadingRateDynamic::create_shading_rate_attachment()
 		image_extent.depth  = 1;
 
 		auto create_shading_rate = [&](VkImageUsageFlags image_usage, VkFormat format) {
-			return std::make_unique<vkb::core::Image>(*device,
-			                                          image_extent,
-			                                          format,
-			                                          image_usage,
-			                                          VMA_MEMORY_USAGE_GPU_ONLY,
-			                                          VK_SAMPLE_COUNT_1_BIT);
+			vkb::core::ImageBuilder builder{image_extent};
```
SaschaWillems: Same here. At least for me, the builder pattern seems out of place in here.

samples/api/oit_linked_lists/oit_linked_lists.cpp
jherico: The current image constructor looks like this:

```c++
	Image(
	    Device const         &device,
	    const VkExtent3D     &extent,
	    VkFormat              format,
	    VkImageUsageFlags     image_usage,
	    VmaMemoryUsage        memory_usage       = VMA_MEMORY_USAGE_AUTO,
	    VkSampleCountFlagBits sample_count       = VK_SAMPLE_COUNT_1_BIT,
	    uint32_t              mip_levels         = 1,
	    uint32_t              array_layers       = 1,
	    VkImageTiling         tiling             = VK_IMAGE_TILING_OPTIMAL,
	    VkImageCreateFlags    flags              = 0,
	    uint32_t              num_queue_families = 0,
	    const uint32_t       *queue_families     = nullptr);
```

Even if you ignore the `Device` argument as being standard to pretty much all the VK resource ctors, that's still a dozen parameters with more than half of them being defaulted.  Most style checkers I know start flagging at anything more than 7 ctor paramers.

if you consider `Image(device, extent, format, VK_IMAGE_USAGE_SAMPLED_BIT, VMA_MEMORY_USAGE_AUTO, VK_SAMPLE_COUNT_1_BIT, 1, 6)`, it's not immediately obvious to a downstream consumer of these examples looking at the code what `1` and `6` mean in this context, or why the sample count is set when it hasn't been in prior image creation code.  It's only by already knowing the layout of the constructor that you can understand these values.

The builder pattern allows...

* reasonable defaults for values that are infrequently changed, such as the tiling or sample count
* access to ALL of the possible values you might want to modify, which the current ctor does not
  * In particular, the `VmaMemoryUsage` flag is the only part of `VmaAllocationCreateInfo` that can be customized through this parameter, while `VmaAllocationCreateFlags`, which is likely something consumers would want to change far more often, is not
* The builder pattern makes it perfectly clear to readers of the examples what the meaning of the values being set are, while constructor param doesn't unless you're looking at the constructor definition.

I have no doubt that you can find places in the code where the builder pattern ends up appearing "just as complex" or even consumes more lines of code as the existing constructor based approach, but IMO, in places like this we're sacrificing brevity for readability and clarity to consumers of the example code as a learning tool.  

Granted I made this change before I added the `build` methods to the builder and the `with_vma_usage` call is almost certainly superfluous, so this could be re-written as

```c++
		linked_list_head_image = vkb::core::ImageBuilder{width, height}
		    .with_format(VK_FORMAT_R32_UINT)
		    .with_usage(VK_IMAGE_USAGE_STORAGE_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT)
		    .build_unique(device);
```

Which would be slightly more compact and avoid the temporary named `builder` variable.  Would that make you happier?

samples/extensions/fragment_shading_rate_dynamic/fragment_shading_rate_dynamic.cpp
jherico: I could replace the entire block with

```c++
		shading_rate_image = vkb::core::ImageBuilder{image_extent}
			    .with_format(VK_FORMAT_R8_UINT)
			    .with_usage(VK_IMAGE_USAGE_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR | VK_IMAGE_USAGE_TRANSFER_DST_BIT)
			    .build_unique(*device);
		    
		shading_rate_image_compute = vkb::core::ImageBuilder{image_extent}
			    .with_format(VK_FORMAT_R8_UINT)
			    .with_usage(VK_IMAGE_USAGE_STORAGE_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT)
			    .build_unique(*device);
```

and eliminate the need for a lambda completely, making the code shorter overall.  Is that easier to follow?

jherico:
@SaschaWillems I have removed all use of the builder pattern from the code (except the internal usage in the image and buffer classes to simplify constructors. I have left, and expanded the use of convenience functions for creating staging buffers, as well as noticed and corrected some minor bugs in populating staging buffers where they fail to flush before unmapping.

gary-sweet (dismissed):

framework/core/allocated.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+AllocatedBase::AllocatedBase(AllocatedBase &&other) noexcept :
+    alloc_create_info(std::exchange(other.alloc_create_info, {})),
+    allocation(std::exchange(other.allocation, {})),
+    mapped_data(std::exchange(other.mapped_data, {})),
+    coherent(std::exchange(other.coherent, {})),
+    persistent(std::exchange(other.persistent, {}))
+{
+	assert(other.allocation == VK_NULL_HANDLE);
+	assert(other.mapped_data == nullptr);
```
asuessenbach: I don't get why you would want to have those `assert`s here. You set those value to default just two lines above.

SaschaWillems:
Any idea why the profiles sample cpp is displayed as completely changed (all lines) in this PR? It only looks like the copyright has changed, but I'm not able to see any other change due to the whole file being displayed as different.

SaschaWillems:
It's the same for a lot of other files too, which makes it nigh impossible to review these changes:

image

image

image

image

jherico:
@SaschaWillems It looks like some of the files got converted from LF to CRLF.  Most of the files in the repository are LF, but a few are CRLF (prior to this change).  I've updated the commit to revert the line ending changes.  I must have misconfigured something in git, because I thought I had my system configured to checkout and commit LF, always (autocrlf=false).  I'll review my settings and try to figure out how this happened.

@tomadamatkinson might be worthwhile to add to the pre-commit hook something that would verify that none of the line-endings have changed for existing files.  These are the only 5 files on `main` that are currently crlf:

```
i/crlf  w/crlf  attr/                   framework/scene_graph/components/orthographic_camera.cpp
i/crlf  w/crlf  attr/                   framework/vulkan_sample.h
i/crlf  w/crlf  attr/                   samples/performance/multithreading_render_passes/CMakeLists.txt
i/crlf  w/crlf  attr/                   samples/performance/multithreading_render_passes/multithreading_render_passes.cpp
i/crlf  w/crlf  attr/                   samples/performance/multithreading_render_passes/multithreading_render_passes.h
```

gary-sweet (dismissed):

framework/core/allocated.cpp
jherico: Yeah, I don't know why I did that.  It doesn't seem to be sourced from any of the code I ported.  It's possible I was migrating code that didn't use `std::exchange`, like the old image or buffer move constructor and added these asserts and then failed to remove them when I converted to use of `std::exchange`.  

I don't want to fix it right now though because I don't want to invalidate the approvals for something that doesn't have any negative impact.  If anything comes along that does force me to update the commits, I'll try to get this fixed too.

SaschaWillems: It's okay to resolve. Just ping us and we'll approve again.

jherico: fixed

gary-sweet (approved):

SaschaWillems (approved):

asuessenbach (approved):

kanavin:
@jherico I'm a bit baffled, but this seems to have regressed 32 bit builds (both arm and x86), where 64 bit build without issues:

```
| In file included from /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/core/allocated.h:22,
|                  from /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/core/buffer.h:22,
|                  from /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/api_vulkan_sample.h:36,
|                  from /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/gltf_loader.cpp:32:
| /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/core/vulkan_resource.h:50:24: error: redefinition of 'constexpr VkObjectType vkb::core::get_object_type(const HandleType&) [with HandleType = long long unsigned int; VkObjectType = VkObjectType]'
|    50 | constexpr VkObjectType get_object_type(const VkImageView &handle)
|       |                        ^~~~~~~~~~~~~~~
| /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/core/vulkan_resource.h:44:24: note: 'constexpr VkObjectType vkb::core::get_object_type(const HandleType&) [with HandleType = long long unsigned int; VkObjectType = VkObjectType]' previously declared here
|    44 | constexpr VkObjectType get_object_type(const VkImage &handle)
|       |                        ^~~~~~~~~~~~~~~
| /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/core/vulkan_resource.h:56:24: error: redefinition of 'constexpr VkObjectType vkb::core::get_object_type(const HandleType&) [with HandleType = long long unsigned int; VkObjectType = VkObjectType]'
|    56 | constexpr VkObjectType get_object_type(const VkRenderPass &handle)
|       |                        ^~~~~~~~~~~~~~~
| /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/core/vulkan_resource.h:44:24: note: 'constexpr VkObjectType vkb::core::get_object_type(const HandleType&) [with HandleType = long long unsigned int; VkObjectType = VkObjectType]' previously declared here
|    44 | constexpr VkObjectType get_object_type(const VkImage &handle)
|       |                        ^~~~~~~~~~~~~~~
| /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/core/vulkan_resource.h:62:24: error: redefinition of 'constexpr VkObjectType vkb::core::get_object_type(const HandleType&) [with HandleType = long long unsigned int; VkObjectType = VkObjectType]'
|    62 | constexpr VkObjectType get_object_type(const VkSampler &handle)
|       |                        ^~~~~~~~~~~~~~~
| /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/core/vulkan_resource.h:44:24: note: 'constexpr VkObjectType vkb::core::get_object_type(const HandleType&) [with HandleType = long long unsigned int; VkObjectType = VkObjectType]' previously declared here
|    44 | constexpr VkObjectType get_object_type(const VkImage &handle)
|       |                        ^~~~~~~~~~~~~~~
| /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/core/vulkan_resource.h:68:24: error: redefinition of 'constexpr VkObjectType vkb::core::get_object_type(const HandleType&) [with HandleType = long long unsigned int; VkObjectType = VkObjectType]'
|    68 | constexpr VkObjectType get_object_type(const VkBuffer &handle)
|       |                        ^~~~~~~~~~~~~~~
| /srv/storage/alex/yocto/build-arm-32-alt/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/vulkan-samples/git/git/framework/core/vulkan_resource.h:44:24: note: 'constexpr VkObjectType vkb::core::get_object_type(const HandleType&) [with HandleType = long long unsigned int; VkObjectType = VkObjectType]' previously declared here
|    44 | constexpr VkObjectType get_object_type(const VkImage &handle)
|       |                        ^~~~~~~~~~~~~~~
| ninja: build stopped: subcommand failed.
```

jherico:
@kanavin I'll try to locally reproduce and find a fix today.

jherico:
ok, so the problem here is that on 32 bit architectures the handles still need to be 64 bit.  So while on 64 bit architectures `VkImage` resolves to `struct objectVkImage_T *`, on 32 bit platforms it resolves to `uint64_t`.  

```
#ifndef VK_DEFINE_NON_DISPATCHABLE_HANDLE
    #if (VK_USE_64_BIT_PTR_DEFINES==1)
        #define VK_DEFINE_NON_DISPATCHABLE_HANDLE(object) typedef struct object##_T *object;
    #else
        #define VK_DEFINE_NON_DISPATCHABLE_HANDLE(object) typedef uint64_t object;
    #endif
#endif
```

This means that all the template specializations like this:

```c++
template <>
constexpr VkObjectType get_object_type(const VkImage &handle)
{ ... }
template <>
constexpr VkObjectType get_object_type(const VkImageView &handle)
{ ... }
```

have the exact same signature and can't be differentiated by the compiler the way they can in 64-bit land.  

@tomadamatkinson should a 32-bit build be added to the CI configs?  Two distinct PR's of mine appear to have caused regressions for `cmake -A Win32`.

jherico:
So, when I run `cmake -A Win32` and build the framework I get a LOT of errors, many of which appear way outside the scope of the changes in this PR and in the ASTC PR.  For instance, I get a lot of errors building KTX, which hasn't been touched in 3 years.

Looking at the readme, I'm unclear on if 32-bit builds are supported since all of the instructions seem to explicitly specify 64-bit arch.  

I'm working on a branch that resolves some of the 32-bit issues I've seen so far, but I'm not touching the third-party build issues other than for astc.

kanavin:
> So, when I run `cmake -A Win32` and build the framework I get a LOT of errors, many of which appear way outside the scope of the changes in this PR and in the ASTC PR. For instance, I get a lot of errors building KTX, which hasn't been touched in 3 years.
> 
> Looking at the readme, I'm unclear on if 32-bit builds are supported since all of the instructions seem to explicitly specify 64-bit arch.
> 
> I'm working on a branch that resolves some of the 32-bit issues I've seen so far, but I'm not touching the third-party build issues other than for astc.

I was worried that in the absence of 32 bit CI that gates all changes, fixing one issue will uncover another (with no clear end in sight). Please do fix what's in your 'jurisdiction' and I'll see what can be done about the rest (my interest is primarily 32 bit linux, especially arm and risc-v which are still used to ship products).

PR #907, opened 2024-02-11
Clean up some top level projects

jherico:
## Description

This makes some small tweaks to the CMake config to migrate some of the projects that are cluttering the top level into proper subfolders.  The glslangValidator and the cli11 projects both create things at the top level or in folders under the top level that should be moved into the "Third Party" folder in IDEs supporting hierarchies.  Additionally, some targets that have been created by `add_custom_target` which are useful for bundling, but which are useless when navigating projects in an IDE have been moved to "Cmake/CustomTargets" (the current offline shader compilation WIP branch generates a TON of these)

The only caveat here is that the CLI11 project is generated as an `interface_library` target type, which has a limited set of properties that can be set on it prior to CMake version 3.19, so the minimum required has to be bumped in order to resolve that.  The minimum required CMake version can be left alone if needed but that one specific change will have to be reverted in that case.

## General Checklist:

This only modifies some meta properties for the project targets in CMake and the resulting location of the projects inside a hierarchical build IDE such as MSVC.  It has no impact on compilation or linking properties, nor changes any sample code, so most of these checks are superfluous in this instance.

- [ ] ~I have commented any added functions (in line with Doxygen)~
- [ ] ~I have commented any code that could be hard to understand~
- [ ] ~My changes do not add any new validation layer errors or warnings~
- [ ] ~I have used existing framework/helper functions where possible~
- [ ] ~My changes do not add any regressions~
- [ ] ~I have tested every sample to ensure everything runs correctly~
- [ ] ~This PR describes the scope and expected impact of the changes I am making~

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2023, Arm Limited and Contributors
+# Copyright (c) 2019-2024, Arm Limited and Contributors
 #
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.19)
 
 project(third_party LANGUAGES C CXX)
 
@@ -83,7 +83,7 @@ elseif(UNIX)
     else()
         message(FATAL_ERROR "Unknown WSI")
     endif()
-endif() 
+endif()
 
 # vma
 add_library(vma INTERFACE)
@@ -95,7 +95,7 @@ target_link_libraries(vma INTERFACE vulkan)
 # libktx
 set(KTX_DIR ${CMAKE_CURRENT_SOURCE_DIR}/ktx)
 
-set(KTX_SOURCES  
+set(KTX_SOURCES
     ${KTX_DIR}/lib/checkheader.c
     ${KTX_DIR}/lib/dfdutils/createdfd.c
     ${KTX_DIR}/lib/dfdutils/colourspaces.c
@@ -136,7 +136,7 @@ set(KTX_SOURCES
     ${KTX_DIR}/lib/basisu/transcoder/basisu_transcoder.h
     ${KTX_DIR}/lib/basisu/transcoder/basisu.h
     ${KTX_DIR}/lib/basisu/zstd/zstd.c
-   
+
     # KT1
     ${KTX_DIR}/lib/texture1.c
     ${KTX_DIR}/lib/texture1.h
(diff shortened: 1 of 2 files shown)


third_party/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.19)
```
SaschaWillems: We have CMake 3.12+ as min. requirements in our build instructions. We should decide if we really want to bump that version and if yes, we need to adjust our documentation accordingly.

jherico: The top level project currently has `cmake_minimum_required(VERSION 3.16)`, and I was actually surprised that I found `cmake_minimum_required` declarations all over the project, since it's almost always set once, at the top level and that's it.  

Trying to revert to the `3.12` requirement as documented would probably be pretty tricky, although I can understand if you want avoid updating beyond 3.16, since out of the box Ubuntu 20.04 only supports up to 3.16 I think.  I do understand that as maintainers of an example repository it's important for you to strike a balance between wanting the features and improvements of more recent CMake versions and needing to support people who are pinned on older versions for whatever reason (usually the Android NDK CMake version in my experience).

Like I said, the only specific change that requires this version bump is CLI11, so if it's decided that it's undesirable to make this change I can remove that one.  For now I'll look try to bring the docs in alignment with this value.

SaschaWillems: Yeah, it's kinda messy right now. I think we need to discuss this internally, as we have people compiling our samples on some very specific platforms. Not sure if those platforms do have access to recent CMake versions.

SaschaWillems:
Thanks a lot for doing this. Having some of the projects at top-level in the project browser was always kinda annoying. Aside from the CMake bump, which we need to discus internally @marty-johnson59 this looks good.

jherico:
I've reverted the minimum required declaration, but left the folder setting for now.  I believe what happens now is that if you're using CMake < 3.19 it emit a warning that you can't set the folder property on interface libraries, but doesn't prevent the CMake configuration from finishing.

gary-sweet (approved):
LGTM. CMake 3.19 would have been fine for me (since the question was asked on the samples call).

marty-johnson59:
Thanks!

SaschaWillems (approved):

asuessenbach (approved):

marty-johnson59:
Merging - 3 approvals

SaschaWillems:
Looks like this breaks compatibility with older CMake versions. We currently state CMake 3.12 or newer in our build document. So we either need to fix the CMakeList or update our requirements (I prefer the latter).

jherico:
I agree with Sascha that updating the minimum CMake version should be strongly considered, perhaps even going so far as to including instructions on how to bootstrap a private CMake version if you're stuck on something older.

While this is an example repository and should be aimed at the widest audience, it's essentially a top level project and doesn't have downstream consumers who might be constrained in their development environment.  While some example creators might be so constrained, I think it's reasonable that this should be prioritized lower relative to the ability to use more advanced CMake features.  

Related: https://www.scivision.dev/cmake-changelog/

PR #910, opened 2024-02-13
Unify class VulkanSample and HPPVulkanSample.

asuessenbach:
## Description

First step to unifying C- and C++-bindings of vulkan in the framework: Unify the central classes `VulkanSample` and `HPPVulkanSample`.
Conceptually does the following:
- replaces `VulkanSample` and `HPPVulkanSample` by a templated class `template <bool CppBindings> VulkanSample`.
- internally, that new class uses the C++-bindings.
- controlled by the bool template parameter `CppBindings`, it provides some interface functions using the C-bindings data types or the C++-bindings data types.
- for now, all the other classes with two flavors (`Device` and `HPPDevice`, ...) still exist and their usage in this class' interface are also controlled by the bool template parameter `CppBindings`.
- to gracefully handle that dichotomy, all members of that class are private now, and some accessor functions are provided.
- especially due to that latest bullet point, literally all samples needed to be touched to use those accessor functions!

## General Checklist:

## Sample Checklist


framework/vulkan_sample.h
```diff
@@ -99,244 +101,274 @@ namespace vkb
  * - Core classes: Classes in vkb::core wrap Vulkan objects for indexing and hashing.
  */
 
+template <bool CppBindings>
 class VulkanSample : public Application
 {
+	/// <summary>
+	/// PUBLIC INTERFACE
+	/// </summary>
```
jherico, line 129:
```suggestion
	/// </summary>
	template <typename CppType, typename CType>
	conditional_type : public std::conditional<CppType, CType, CType>{};

	using RenderContextType = conditional_type<vkb::rendering::HPPRenderContext, vkb::RenderContext>::type;
	using DeviceType =  conditional_type<vkb::core::HPPDevice, vkb::Device>::type;
	using InstanceType = conditional_type<vkb::core::HPPInstance, vkb::Instance>::type;
```

framework/vulkan_sample.h
```diff
@@ -99,244 +101,274 @@ namespace vkb
...
+	/// PUBLIC INTERFACE
+	/// </summary>
   public:
 	VulkanSample() = default;
+	~VulkanSample() override;
 
-	virtual ~VulkanSample();
+	Configuration                                                                                            &get_configuration();
+	typename std::conditional<CppBindings, vkb::rendering::HPPRenderContext, vkb::RenderContext>::type       &get_render_context();
+	typename std::conditional<CppBindings, vkb::rendering::HPPRenderContext, vkb::RenderContext>::type const &get_render_context() const;
```
jherico:
```suggestion
	RenderContextType       &get_render_context();
	RenderContextType const &get_render_context() const;

```

framework/vulkan_sample.h
```diff
@@ -99,244 +101,274 @@ namespace vkb
...
+	virtual void draw_gui();
 
-	void finish() override;
+	/**
+	 * @brief Starts the render pass, executes the render pipeline, and then ends the render pass
+	 * @param command_buffer The command buffer to record the commands to
+	 * @param render_target The render target that is being drawn to
+	 */
+	virtual void draw_renderpass(vkb::core::HPPCommandBuffer &command_buffer, vkb::rendering::HPPRenderTarget &render_target);
+	virtual void draw_renderpass(CommandBuffer &command_buffer, RenderTarget &render_target);
```
jherico: having two functions with different signatures like this might lead to confusion.  Why not template specialization with more conditional types?

Someone could then write 

```
template<>
VulkanSample<true>::draw_renderpass(CommandBufferType& command_buffer, RenderTargetType& render_target)`
```

That way only one signature type is visible to downstream classes for calling and overriding.

framework/vulkan_sample.h
asuessenbach: Good point! Makes the conditional code less verbose.

asuessenbach: Done.

framework/vulkan_sample.h
asuessenbach: Done.

framework/vulkan_sample.h
asuessenbach: Done.

framework/CMakeLists.txt
```diff
@@ -64,14 +63,12 @@ set(FRAMEWORK_FILES
     resource_cache.cpp
     resource_record.cpp
     resource_replay.cpp
-    vulkan_sample.cpp
     api_vulkan_sample.cpp
     timer.cpp
     camera.cpp
+	hpp_api_vulkan_sample.cpp
```
jherico:
```suggestion
    hpp_api_vulkan_sample.cpp
```

framework/hpp_api_vulkan_sample.h
```diff
@@ -60,7 +60,7 @@ struct HPPVertex
  *
  * See vkb::ApiVulkanSample for documentation
  */
-class HPPApiVulkanSample : public vkb::HPPVulkanSample
+class HPPApiVulkanSample : public vkb::VulkanSample<true>
 {
```
jherico: I typically rely on a convention like this..

```suggestion
class HPPApiVulkanSample : public vkb::VulkanSample<true>
{
    using Parent = vkb::VulkanSample<true>;
```

so that in my code I can always write `Parent::parent_method()` instead of having to write the particular base class out each time.

framework/hpp_vulkan_sample.h
jherico, line 1: You mgiht consider retaining this file and having it contain mostly just a `using HPPVulkanSample = VulkanSample<true>` declaration, which should reduce the amount of downstream changes to files that do nothing but remove this header inclusion and change the name of their base class.

samples/api/hpp_compute_nbody/hpp_compute_nbody.cpp
```diff
@@ -32,9 +32,9 @@ HPPComputeNBody::HPPComputeNBody()
 
 HPPComputeNBody::~HPPComputeNBody()
 {
-	if (get_device() && get_device()->get_handle())
+	if (has_device() && get_device().get_handle())
 	{
-		vk::Device device = get_device()->get_handle();
+		vk::Device device = get_device().get_handle();
```
jherico, line 37: if you added an `bool operator()` and an `DeviceType& operator ->()` to the device classes you'd be able to avoid having to do a lot of these changes right now. That would make the salient points of these changes easier to review, and then a later PR could mark those helper functions as `[[deprecated]]` and do all the mechanical `get_device()->` to `get_device().` changes without any "meaningful" changes mixed in.

For the `bool operator()` to work you'd have to have a stub device that was present until the real device was created to return false I guess.

framework/vulkan_sample.h
```diff
@@ -99,244 +101,281 @@ namespace vkb
...
 	 */
-	virtual void create_instance();
+	virtual std::unique_ptr<InstanceType> create_instance(bool headless);
 
 	/**
-	 * @brief Main loop sample events
+	 * @brief Override this to customise the creation of the render_context
 	 */
-	void update(float delta_time) override;
+	virtual void create_render_context();
```
jherico, line 176: You might consider having the `RenderContext` type be a singleton and moving the `create_device` and `create_instance` over to it.  I've noticed that there are some odd dependencies from things like the gui classes to the sample base classes, and typically those exist strictly to get access to the render context.  

If the render context type is a singleton and any class has access to it directly, a lot of complexity and interdependencies fall away.

framework/vulkan_sample.h
```diff
@@ -99,244 +101,281 @@ namespace vkb
  * - Core classes: Classes in vkb::core wrap Vulkan objects for indexing and hashing.
  */
 
+template <bool CppBindings>
```
jherico: possibly rename this class to `VulkanSampleBase` and then create `using VulkanSample = VulkanSampleBase<false>; using HPPVulkanSample = VulkanSampleBase<true>;` at the bottom of this file.

framework/vulkan_sample.h
```diff
@@ -348,5 +387,918 @@ class VulkanSample : public Application
...
+inline std::unique_ptr<typename VulkanSample<CppBindings>::InstanceType> VulkanSample<CppBindings>::create_instance(bool headless)
+{
+	if constexpr (CppBindings)
+	{
+		return std::make_unique<vkb::core::HPPInstance>(get_name(), get_instance_extensions(), get_validation_layers(), headless, api_version);
+	}
+	else
+	{
+		return std::make_unique<vkb::Instance>(get_name(), get_instance_extensions(), get_validation_layers(), headless, api_version);
+	}
```
jherico:
```suggestion
	return std::make_unique<InstanceType>(get_name(), get_instance_extensions(), get_validation_layers(), headless, api_version);
```

samples/tooling/profiles/profiles.cpp
```diff
@@ -109,16 +107,18 @@ void Profiles::create_device()
 	}
 
 	// Post device setup required for the framework
-	device = std::make_unique<vkb::Device>(gpu, vulkan_device, surface);
-	device->add_queue(0, queue_create_info.queueFamilyIndex, queue_family_properties[selected_queue_family], true);
-	device->prepare_memory_allocator();
-	device->create_internal_command_pool();
-	device->create_internal_fence_pool();
+	auto device = std::make_unique<vkb::Device>(gpu, vulkan_device, get_surface());
+	get_device().add_queue(0, queue_create_info.queueFamilyIndex, queue_family_properties[selected_queue_family], true);
```
SaschaWillems: Shouldn't this access `device` instead? See a line above, where device is created as a local variable.

framework/api_vulkan_sample.h
```diff
@@ -99,7 +99,7 @@ struct Meshlet
  *
  * See vkb::VulkanSample for documentation
  */
-class ApiVulkanSample : public vkb::VulkanSample
+class ApiVulkanSample : public vkb::VulkanSample<false>
```
SaschaWillems: Mostly out of curiosity, but what toes `<false>` mean in this regard and why is it needed now? Can't this be a default value that only needs to specified when ddifferent?

jherico: The base class `VulkanSample` uses a boolean template parameter to determine whether to use the C++ or C types for various members, using `std::conditional` inside `VulkanSample`.

asuessenbach: As @jherico said.
One could make, for example, the C-binding type the default, reducing `VulkanSample<false>` to `VulkanSample<>` at lots of places. I don't think, that that would give you much.
I could imagine, though, to introduce some `enum class BindingType { CBinding, CppBinding };`, or so, and use that instead of just a bool. Would result in `VulkanSample<BindingType::CBinding>` and `VulkanSample<BindingType::CppBinding>`, which arguably is a bit more descriptive.

framework/hpp_api_vulkan_sample.h
asuessenbach: Great approach!
I think, we should introduce such a `Parent` (or `ParentType`?). But with subsequent PRs.

framework/hpp_vulkan_sample.h
asuessenbach: That might be true.
But I think, most of the downstream changes are due to making previously `public` members `private` and introduce the (binding-specific) accessor-functions.

samples/api/hpp_compute_nbody/hpp_compute_nbody.cpp
asuessenbach: Agreed, that would have been a good approach to limit the size of this PR.
But as all those changes have been done already, I'm a bit reluctant to revert all that and re-do it later on again.

framework/vulkan_sample.h
asuessenbach: Interesting approach... but definitely not as part of this PR.
Maybe you could file an issue on that?

framework/vulkan_sample.h
asuessenbach: This might have reduced the downstream changes a bit.
But, again, as all the changes have been done already, I would prefer to not revert that again.
Besides that, it keeps the notion of `HPPVulkanSample` being some derivative of `VulkanSample`, which isn't the case.
Maybe, instead one could introduce some `using VulkanSampleC = VulkanSample<false>` and `using VulkanSampleCpp = VulkanSample<true>`, or so?

framework/hpp_vulkan_sample.h
jherico: I was thinking about the sample class changes, not the structural changes to the accessors on the helper base class, but your point elsewhere about the changes being a _fait accompli_ is taken.

framework/vulkan_sample.h
jherico: https://github.com/KhronosGroup/Vulkan-Samples/issues/921

samples/tooling/profiles/profiles.cpp
asuessenbach: Fixed.

framework/CMakeLists.txt
asuessenbach: Done.

framework/vulkan_sample.h
asuessenbach: Done.

asuessenbach:
Somehow other peoples changes has sneaked into this one. Maybe due to some rebasing?
And I don't know, how to get them out of the PR again. Can anybody help me with that?

tomadamatkinson:
> Somehow other peoples changes has sneaked into this one. Maybe due to some rebasing?
And I don't know, how to get them out of the PR again. Can anybody help me with that?

@asuessenbach as it has been a few days it likely wont be achievable with git reflog. Something you might try is

- Checkout your branch
- Checkout a new branch (temp to test the changes without editing your current work)
- Squash your commits
- Rebase again on main - override any changes present in your branch with those in main

I tend to be able to fix these changes locally if it happens to me but its quite hard to recall the precise steps i normally take. The steps above may not work for you but its something to try at least

tomadamatkinson:
I will take a more thorough review once the rebase issues are sorted. I reviewed `vulkan_sample.h` and the sample changes. Unsure which changes have leaked in from other commits.

Im happy with the overall interface. I would have suggested something like `get_device` and `get_c_device` to keep things ultra simple, but I didn't realise just how large the `VulkanSample` interface had grown. Is the idea to slowly phase out the C interfaces using this same technique?

asuessenbach:
> I will take a more thorough review once the rebase issues are sorted. I reviewed vulkan_sample.h and the sample changes. Unsure which changes have leaked in from other commits.

It seems, all the alien changes have gone with my latest force-push. So feel free to review in more depth ;)

tests/system_test/test_framework/vulkan_test.h
```diff
@@ -23,7 +23,7 @@
 
 namespace vkbtest
 {
-class VulkanTest : public vkb::VulkanSample
```
tomadamatkinson, line 26: Proof i made it too the bottom :wink:

tomadamatkinson (commented):
In general very happy with the change. Following this pattern in the future should let us condense a lot of the duplicate interfaces. In time this should help simplify things again but also mean that we transition to the C++ api seamlessly.

Will approve when ive tested locally. Hopefully tomorrow! If i don't end up reviewing soon please @ me

SaschaWillems (dismissed):
Tested this on Windows 11 with an RTX 4070 and on Android 13 with a Google Pixel 4a (build from Windows 11). Compiles and runs fine on both.

I'm not much of a C+++ expert, but this looks like a nice solution and a step in the right direction and gets a thumbs up from me :)

Thank you very much for this PR!

third_party/vulkan
JoseEmilio-ARM, line 1: was this updated intentionally?

JoseEmilio-ARM (dismissed):

third_party/vulkan
asuessenbach: Yes, that was intentional.
For some unknown reason, one of the compiler in the CI complained about some template magic in Vulkan-Hpp. Updating that to a newer version resolved that.

tomadamatkinson (dismissed):
LGTM! Thanks for all the work on this

SaschaWillems (approved):

marty-johnson59:
Merging per agreement on the call - now that conflicts are resolved..

Issue #913, opened 2024-02-15
fragment_shading_rate_dynamic crashes


jherico:
Building from main on Windows 10 (MSVC 2022) the fragment_shading_rate_dynamic sample currently crashes after a set of validation errors.  

Output after swap chain setup is as follows:

```
UNASSIGNED-GeneralParameterError-ExtensionNotEnabled(ERROR / SPEC): msgNum: -383040211 - Validation Error: [ UNASSIGNED-GeneralParameterError-ExtensionNotEnabled ] Object 0: handle = 0x2321cce5530, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xe92b452d | Attempted to call vkCmdSetFragmentShadingRateKHR() but its required extension VK_VERSION_1_1 has not been enabled

    Objects: 1
        [0] 0x2321cce5530, type: 1, name: NULL
UNASSIGNED-GeneralParameterError-ExtensionNotEnabled(ERROR / SPEC): msgNum: -383040211 - Validation Error: [ UNASSIGNED-GeneralParameterError-ExtensionNotEnabled ] Object 0: handle = 0x2321cce5530, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xe92b452d | Attempted to call vkCmdSetFragmentShadingRateKHR() but its required extension VK_VERSION_1_2 has not been enabled

    Objects: 1
        [0] 0x2321cce5530, type: 1, name: NULL
UNASSIGNED-GeneralParameterError-ExtensionNotEnabled(ERROR / SPEC): msgNum: -383040211 - Validation Error: [ UNASSIGNED-GeneralParameterError-ExtensionNotEnabled ] Object 0: handle = 0x2321cce5530, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xe92b452d | Attempted to call vkCmdSetFragmentShadingRateKHR() but its required extension VK_VERSION_1_1 has not been enabled

    Objects: 1
        [0] 0x2321cce5530, type: 1, name: NULL
[...]

PR #914, opened 2024-02-15
Allow explicit skips in batch mode

### LINKED ISSUE #936: CLI11 parsing issue with `ManyValues`
several of the flags available for batch mode (`--tag`, `--category`) are marked with the `ManyValues` enum, but any attempt to start an application with multiple values for these fails at the argument parsing stage with an error that 1 value was expected but N were given.  This happens regardless of whether the command is invoked as `vulkan-samples batch --tags foo --tag bar` or `vulkan-samples batch --tags foo bar`

This appears to be a bug in the wrapping code for the CLI11 library, where the [code](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/framework/platform/parsers/CLI11.cpp#L108) looking at the flags and translating them into CLI11 structures does nothing different for `OneValue` vs `ManyValues`

jherico:
## Description

Batch mode is extremely useful for testing a change hasn't impacted the behavior of any samples (say, but crashing or producing a black screen).

However, if there are known crash issues in examples (that are present on the main branch) such as #913  then batch mode only becomes useful until you hit that broken example.

This change adds a flag that allows batch mode to skip examples by id that would normally be included otherwise.

## General Checklist:

Fixes #936

### DIFF
--- a/app/plugins/batch_mode/batch_mode.h
+++ b/app/plugins/batch_mode/batch_mode.h
@@ -32,11 +32,11 @@ using BatchModeTags = vkb::PluginBase<vkb::tags::Entrypoint, vkb::tags::FullCont
 
 /**
  * @brief Batch Mode
- * 
+ *
  * Run a subset of samples. The next sample in the set will start after the current sample being executed has finished. Using --wrap-to-start will start again from the first sample after the last sample is executed.
- * 
+ *
  * Usage: vulkan_samples batch --duration 3 --category performance --tag arm
- * 
+ *
  */
 class BatchMode : public BatchModeTags
 {
@@ -62,7 +62,9 @@ class BatchMode : public BatchModeTags
 
 	vkb::FlagCommand categories_flag{vkb::FlagType::ManyValues, "category", "C", "Filter samples by categories"};
 
-	vkb::SubCommand batch_cmd{"batch", "Enable batch mode", {&duration_flag, &wrap_flag, &tags_flag, &categories_flag}};
+	vkb::FlagCommand skip_flag{vkb::FlagType::ManyValues, "skip", "", "Skip a sample by id"};
+
+	vkb::SubCommand batch_cmd{"batch", "Enable batch mode", {&duration_flag, &wrap_flag, &tags_flag, &categories_flag, &skip_flag}};
 
   private:
 	/// The list of suitable samples to be run in conjunction with batch mode
--- a/app/plugins/batch_mode/batch_mode.cpp
+++ b/app/plugins/batch_mode/batch_mode.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2020-2023, Arm Limited and Contributors
+/* Copyright (c) 2020-2024, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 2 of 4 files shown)


app/plugins/batch_mode/batch_mode.h
```diff
@@ -62,7 +62,9 @@ class BatchMode : public BatchModeTags
 
 	vkb::FlagCommand categories_flag{vkb::FlagType::ManyValues, "category", "C", "Filter samples by categories"};
 
-	vkb::SubCommand batch_cmd{"batch", "Enable batch mode", {&duration_flag, &wrap_flag, &tags_flag, &categories_flag}};
+	vkb::FlagCommand skip_flag{vkb::FlagType::ManyValues, "skip", "", "Skip a sample by id"};
```
gary-sweet, line 65: How is this intended to work?
ManyValues implies that you can give a list of samples, the text says "Skip a sample by id" which implies a single sample.

If it's supposed to be many, how do I specify the list on the command line? Nothing I've tried has worked.

jherico: I was patterning it on the `--tags` parameter, which I assumed was functional and could accept multiple values, but that appears to have been incorrect.  It looks like there's a bug in the CLI11 parsing wrapper [here](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/framework/platform/parsers/CLI11.cpp#L108) where the code doesn't do anything different for `OneValue` vs `ManyValues`.  I will add a commit to fix this.

app/plugins/batch_mode/batch_mode.h
```diff
@@ -62,7 +62,9 @@ class BatchMode : public BatchModeTags
 
 	vkb::FlagCommand categories_flag{vkb::FlagType::ManyValues, "category", "C", "Filter samples by categories"};
 
-	vkb::SubCommand batch_cmd{"batch", "Enable batch mode", {&duration_flag, &wrap_flag, &tags_flag, &categories_flag}};
```
gary-sweet, line 65: Should we add "S" as a shortcut here so it's similar to the other options?

app/plugins/batch_mode/batch_mode.h
gary-sweet: This does work now, providing you 'just know' that you have to give multiple "--skip" arguments. e.g "--skip a --skip b". I wonder if there's a way we can make that more obvious in the help text.

gary-sweet (commented):
I'm wondering if the CLI11 wrapper fix should be in a separate PR, since it doesn't just affect this change.

app/plugins/batch_mode/batch_mode.h
jherico: I don't think so.  While tags and categories have the short form, none of the other flags in any of the other plugins have a single-character form, and I suspect it would be too easy to accidentally assign the same short-form to more than one flag, such as the stretch flag or the screenshot flag, etc.

app/plugins/batch_mode/batch_mode.h
jherico: This flag is aimed primarily at example developers who want to test out that their changes don't break any of the samples (but might need to deal with a known broken sample), so I don't think it needs to be excessively user-friendly to people who are using the examples to learn techniques and therefore are less likely to run in batch-mode in the first place.

jherico:
The CLI11 fix is already a distinct commit.  I think making a whole distinct PR for it might be excessive.

app/plugins/batch_mode/batch_mode.h
tomadamatkinson: Agreed on this last point. I explicitly use batch mode to test for regressions! The help message here also makes sense to me in context of the rest of this plugins options

app/plugins/batch_mode/batch_mode.h
tomadamatkinson: Happy for this to be long form as the intended use case is for regression testing and not for a general user. In the future we may want to opt to use a json config to standardise this as our test framework over the existing system tests (which are not in use by Khronos but to my knowledge are still used by Arm)

tomadamatkinson (approved):
Welcomed change! :fire: 

Works locally. Opens batch mode up to be a more sustainable regression testing capability

gary-sweet (approved):

gpx1000 (approved):

PR #933, opened 2024-02-22
Update validation layer

JoseEmilio-ARM:
## Description

The current default version of the validation layers have a bug that prints an incorrect error message in a new sample I am working on. This change updates the template so that we fetch the most recent version of the layers available today (with this, and possibly a few other bugs, fixed).

These changes only affect the template, so for them to take effect in an existing Android build, you need to follow the instructions in https://github.com/KhronosGroup/Vulkan-Samples/blob/670f52353dc9cc422ab535fa5b66d2eeb04a70db/bldsys/cmake/template/gradle/download_vvl.gradle#L29-L32

Note that the Validation Layers repository does not consistently name their binaries or URL paths, so those had to be updated as well, and might need to be changed again in the future.

## General Checklist:

### DIFF
--- a/bldsys/cmake/template/gradle/app.build.gradle.in
+++ b/bldsys/cmake/template/gradle/app.build.gradle.in
@@ -2,7 +2,7 @@ plugins {
 	id 'com.android.application'
 }
 
-ext.vvl_version='1.3.231.1'
+ext.vvl_version='1.3.275.0'
 apply from: "./download_vvl.gradle"
 
 android {
--- a/bldsys/cmake/template/gradle/download_vvl.gradle
+++ b/bldsys/cmake/template/gradle/download_vvl.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'com.android.application'
 
 /*
- * Copyright (c) 2022, The Android Open Source Project.
+ * Copyright (c) 2022-2024, The Android Open Source Project.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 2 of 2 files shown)


jherico:
`bldsys/cmake/template/gradle/app.build.gradle.in` is already a template that gets configured [here](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/bldsys/cmake/create_gradle_project.cmake#L197).  If you changed 
`bldsys/cmake/template/gradle/download_vvl.gradle` to also be a template then in the future you'd only need to make changes to `create_gradle_project.cmake` to update the validation layer version.

gary-sweet:
This is an Android-only change, so I can't test it I'm afraid.

JoseEmilio-ARM:
> `bldsys/cmake/template/gradle/app.build.gradle.in` is already a template that gets configured [here](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/bldsys/cmake/create_gradle_project.cmake#L197). If you changed `bldsys/cmake/template/gradle/download_vvl.gradle` to also be a template then in the future you'd only need to make changes to `create_gradle_project.cmake` to update the validation layer version.

Hi @jherico , I looked into this, but I think it might not be worth having a template for `download_vvl.gradle`, because this is the only parameter that needs adjusting, and in reality it is never used, it is only a fallback, if the caller of the script does not define their own version: 

https://github.com/KhronosGroup/Vulkan-Samples/blob/670f52353dc9cc422ab535fa5b66d2eeb04a70db/bldsys/cmake/template/gradle/download_vvl.gradle#L39-L42

Perhaps we could remove this safeguard, to make it more explicit that the caller (`build.gradle`) needs to provide a layer version? 

Also they don't need to always match. Ideally I would not have needed to change the fallback version, but unfortunately the download URL changed and did not work with the previous version anymore 😞

SaschaWillems (approved):
LGTM

tomadamatkinson (approved):

Issue #938, opened 2024-02-23
Requesting extension features also enables all available features


SaschaWillems:
In a lot of samples we do something like this:

```cpp
auto &requested_extension_features = gpu.request_extension_features<VkPhysicalDeviceSomeExtensionFeaturesKHR>(VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SOME_EXTENSION_FEATURES_KHR);
requested_extension_features.someFeature = VK_TRUE;
```

But the call to `gpu.request_extension_features` already sets all supported features to true by calling `vkGetPhysicalDeviceFeatures2KHR`, which makes the second line (enabling a feature) superfluous.

This doesn't do any harm, but it might confuse people trying to write or debug samples.

asuessenbach:
One could introduce a new function `vkb::PhysicalDevice::get_extension_features<ExtensionFeatureStruct>` that could be used like this:
```
	if (gpu.get_extension_features<VkPhysicalDeviceHostQueryResetFeaturesEXT>().hostQueryReset)
	{
		gpu.request_extension_features<VkPhysicalDeviceHostQueryResetFeaturesEXT>().hostQueryReset= VK_TRUE;
	}
```
... and would need to adjust all usages of `request_extension_features` accordingly!
... and modify `request_extension_features` such that it returns either an empty features struct, or the one already requested before.

PR #945, opened 2024-02-25
Add new sample for shader debugprintf

### LINKED ISSUE #926: Add sample for shader debug printf
Having a sample to show how to use debug printf in shaders would be a nice addition for the tooling category of our samples. It's easy to implement and offers an opportunity for a nice tutorial, so it shouldn't be too much work. I'm already started working on it. This issue is meant to track progress and to avoid someone else doing duplicate work.

### LINKED ISSUE #985: No way to access instance from samples anymore
With #910 some public members were moved to private and can now only be accessed via functions, e.g. `has_device()` instead of `device`.

But it looks like there is no way to access the instance anymore. There is no `has_instance()`, but I need to access it for my shader debug printf sample like this:

```cpp
	if (instance)
	{
		vkDestroyDebugUtilsMessengerEXT(instance->get_handle(), debug_utils_messenger, nullptr);
	}
```

SaschaWillems:
## Description

This PR adds a new sample showing how to use debugprintf in shaders. It also includes a tutorial.

Note: This also does some changes to the framework required for samples that do custom instance creation.

Tested on Windows 11 with an NVIDIA RTX 4070

Fixes #926
Fixes #985

## General Checklist:

## Sample Checklist


gary-sweet (changes requested):
We don't support debugPrintfEXT but I'm still seeing a couple of issues when I run this PR.

I get a validation error at the start:
`VUID-vkCreateInstance-ppEnabledExtensionNames-01388(ERROR / SPEC): msgNum: -437968512 - Validation Error: [ VUID-vkCreateInstance-ppEnabledExtensionNames-01388 ] | MessageID = 0xe5e52180 | vkCreateInstance(): pCreateInfo->ppEnabledExtensionNames[0] Missing extension required by the instance extension VK_KHR_display: VK_KHR_surface. The Vulkan spec states: All required extensions for each extension in the VkInstanceCreateInfo::ppEnabledExtensionNames list must also be present in that list (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCreateInstance-ppEnabledExtensionNames-01388)`

It terminates with an error about `VK_KHR_get_physical_device_properties2`:
```
[error] [framework/platform/platform.cpp:169] Error Message: Couldn't request feature from device as VK_KHR_get_physical_device_properties2 isn't enabled!
[error] [framework/platform/platform.cpp:170] Failed when running application shader_debugprintf
Segmentation fault
```
and then segfaults in `vkDestroySurfaceKHR()` called from `VulkanSample::~VulkanSample()`

SaschaWillems:
Thanks for noticing. The framework seems to add `VK_KHR_surface` explicitly and via glfw (which looks wrong on e.g. Windows as it shows up twice). I guess on your platform the implicit addition via glfw doesn't happen, hence it's missing. Will fix that.

SaschaWillems:
I have added a tutorial for this sample. It's now ready for review.

samples/extensions/shader_debugprintf/README.adoc
```diff
@@ -0,0 +1,197 @@
...
+
+:pp: {plus}{plus}
+
+image::./images/sample.png[Sample]
+
+== Overview
+
+This sample demonstrates the usage of the https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_shader_non_semantic_info.html[VK_KHR_shader_non_semantic_info] for adding means to output debug messages from SPIR-V shaders using a print function at runtime.
+
+It allows you to output invocation related debug messages from a give shader stage and can help to find problems or bugs together with a graphics debugger like a https://www.renderdoc.org[RenderDoc]. But it's also possible to display that data in your application (e.g. a debug window or the command line), although this has some limitations like a fixed buffer size that all messages need to fit in.
```
asuessenbach: typo: give -> given
typo: like a -> like

samples/extensions/shader_debugprintf/README.adoc
```diff
@@ -0,0 +1,197 @@
...
+----
+VSOutput main(VSInput input) {
+	float4 output.Position = mul(ubo.view, mul(ubo.model, input.Position));
+	printf("Transformed position = %v4f", output.Position);
+}
+----
+
+== Setup
+
+Shader printf is a feature of the validation layers. So the firs step is to enable the `VK_LAYER_KHRONOS_validation` layer at instance creation. In addition to the layer(s) to be enabled we also need to provide `VkValidationFeaturesEXT` with the `VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT` feature enabled:
```
asuessenbach: typo: firs -> first

samples/extensions/shader_debugprintf/README.adoc
```diff
@@ -0,0 +1,197 @@
...
+----
+void ShaderDebugPrintf::create_instance()
+{
+    ...
+
+    // Shader printf is a feature of the validation layers that needs to be enabled
+    std::vector<VkValidationFeatureEnableEXT>  validation_feature_enables = {VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT};
+
+    VkValidationFeaturesEXT validation_features{VK_STRUCTURE_TYPE_VALIDATION_FEATURES_EXT};
+    validation_features.enabledValidationFeatureCount = 1;
```
asuessenbach, line 72: instead of plain `1`, we should use `static_cast<uint32_t>(validation_feature_enables.size())`
Or with just one feature enabled, maybe just use a plain `VkValidationFeatureEnableEXT enable_printf = VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT;`

samples/extensions/shader_debugprintf/README.adoc
```diff
@@ -0,0 +1,197 @@
...
+	debug_utils_messenger_create_info.messageSeverity = VK_DEBUG_UTILS_MESSAGE_SEVERITY_INFO_BIT_EXT;
+	debug_utils_messenger_create_info.messageType     = VK_DEBUG_UTILS_MESSAGE_TYPE_VALIDATION_BIT_EXT;
+	debug_utils_messenger_create_info.pfnUserCallback = debug_utils_message_callback;
+
+	VK_CHECK(vkCreateDebugUtilsMessengerEXT(vulkan_instance, &debug_utils_messenger_create_info, nullptr, &debug_utils_messenger));
+----
+
+[NOTE]
+====
+Shader printf messages are of `info` type, so in order for them to be passed to the callback you need to enable set the `VK_DEBUG_UTILS_MESSAGE_SEVERITY_INFO_BIT_EXT` `messageSeverity` flag.
```
asuessenbach: typo: to enable set -> to enable

samples/extensions/shader_debugprintf/README.adoc
```diff
@@ -0,0 +1,197 @@
...
+
+[,glsl]
+----
+// Output the vertex position using debug printf
+if (gl_VertexIndex == 0) {
+    debugPrintfEXT("Position = %v4f", outPos);
+}
+----
+
+And then gather those so they can be displayed in the graphical user interface. The sample also checks if it's actually a printf message (`pMessageIdName == WARNING-DEBUG-PRINTF`), which is something one could you to distinguish between actual validation messages and printf outputs:
```
asuessenbach: typo: one could you -> one could do

samples/extensions/shader_debugprintf/README.adoc
```diff
@@ -0,0 +1,197 @@
...
+		shader_message = shader_message.substr(shader_message.find("Position"));
+		debug_output.append(shader_message + "\n");
+	}
+	return VK_FALSE;
+}
+----
+
+== Printf output buffer size
+
+The default buffer size for debug outputs is limited. If you output a lot of messages, e.g. by doing a printf per fragment shader invocation, this may not suffice and you'll get a validation warning telling you about the buffer being to small. This buffer size can be increased via the VkConfig utility from the LunarG SDK. See https://www.lunarg.com/wp-content/uploads/2021/08/Using-Debug-Printf-02August2021.pdf[this whitepaper] for details on that.
```
asuessenbach, line 157: Is there no way to increase that size programmatically?

samples/extensions/shader_debugprintf/README.adoc
```diff
@@ -0,0 +1,197 @@
...
+
+[NOTE]
+====
+This **does not** apply to outputting messages in graphics debuggers.
```
asuessenbach, line 161: What is "This" here?
That the buffer size is limited? Or that you can increase it via VkConfig?

samples/extensions/shader_debugprintf/README.adoc
```diff
@@ -0,0 +1,197 @@
...
+====
+
+== Memory considerations
+
+Using debug printf will consume a descriptor set, so if you use every last descriptor set it may not work and you may need to increase set count at pool allocation. Thus in the sample we allocate an additional set from the pool. It also consumes device memory based on the no. and/or size fo the debug messages.
```
asuessenbach: Maybe: you may need to increase set count -> you may need to increase the set count
Maybe/Typo: on the no. and/or size fo -> on the number and/or size of

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -414,20 +414,18 @@ void ShaderDebugPrintf::create_instance()
...
 	VkApplicationInfo app_info{VK_STRUCTURE_TYPE_APPLICATION_INFO};
 	app_info.pApplicationName = "Shader debugprintf";
 	app_info.pEngineName      = "Vulkan Samples";
 	app_info.apiVersion       = VK_API_VERSION_1_1;
 
+	// Shader printf is a feature of the validation layers that needs to be enabled
+	std::vector<VkValidationFeatureEnableEXT> validation_feature_enables = {VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT};
+
+	VkValidationFeaturesEXT validation_features{VK_STRUCTURE_TYPE_VALIDATION_FEATURES_EXT};
+	validation_features.enabledValidationFeatureCount = 1;
```
asuessenbach: maybe: `1` -> `static_cast<uint32_t>(validation_feature_enables.size())`

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -0,0 +1,500 @@
...
+	build_command_buffers();
+	prepared = true;
+	return true;
+}
+
+// This sample overrides the instance creation part of the framework to chain in additional structures
+void ShaderDebugPrintf::create_instance()
+{
+	// Initialize Volk Vulkan Loader
+	VkResult result = volkInitialize();
```
asuessenbach: No need to `volkInitialize` here. It has already been called in `VulkanSample::prepare`.

asuessenbach (commented):
Great description on how to use it.

For some unknown reason, this sample "runs" incredibly slow on my machine (multiple seconds per frame). But when the Vulkan Configurator is running as well, I don't get any output anymore and the speed is as expected. Any idea on that?

SaschaWillems:
> Great description on how to use it.
> 
> For some unknown reason, this sample "runs" incredibly slow on my machine (multiple seconds per frame). But when the Vulkan Configurator is running as well, I don't get any output anymore and the speed is as expected. Any idea on that?

That seems to be a bug with the validation layers. I noticed this too, but couldn't reproduce :(

samples/extensions/shader_debugprintf/README.adoc
SaschaWillems: Not that I'm aware of. I didn't find anything in this regards in any of the SDK documents or the LunarG white papers.

samples/extensions/shader_debugprintf/README.adoc
SaschaWillems: Both. Graphics debuggers don't care for the buffer size and thus changing it with VkConfig has no effect.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SaschaWillems: Thanks. The multiple layers of abstraction in the framework always makes it hard to find those things :/

samples/extensions/shader_debugprintf/README.adoc
SaschaWillems: I'll use `size()`. For people learning from our samples it imo makes it easier if they see that this can be a vector of enables.

gary-sweet:
This no longer crashes for me, but I'm still seeing:

```
[error] [framework/platform/platform.cpp:175] Error Message: Couldn't request feature from device as VK_KHR_get_physical_device_properties2 isn't enabled!
[error] [framework/platform/platform.cpp:176] Failed when running application shader_debugprintf
    71:05:42.547 nxserverlib: Vulkan Display(0x7e5c000ee0) unregistered

```

SaschaWillems:
Thank you very much for testing :)

I guess that error is caused by #938

Currently the framework will always enable all features of an extension at the time it's requested. I'll see if I can do a workaround for that.

asuessenbach:
I think, @gary-sweet's issue here is, that `VK_KHR_get_physical_device_properties2` is not enabled (is it supported, at all?), even though it has been added in `ShaderDebugPrintf::create_instance`! Should have failed there, then.

gary-sweet:
> I think, @gary-sweet's issue here is, that `VK_KHR_get_physical_device_properties2` is not enabled (is it supported, at all?), even though it has been added in `ShaderDebugPrintf::create_instance`! Should have failed there, then.

We do support VK_KHR_get_physical_device_properties2, so it's not obvious what the issue is. I'll do some degugging.

gary-sweet:
> I'll do some debugging.

It looks like this sample is creating its own VkInstance which it then constructs the Instance() object from. This appears to bypass a bunch of the extension enabling that happens in the Instance main constructor (all the enable_extension() stuff). The end result is that the Instance object doesn't think the extension is enabled, even though it actually is in the contained VkInstance.

SaschaWillems:
Yeah, it's similar to the profiles example. The framework assumes some stuff, even if you create your own instance, which in turn causes problems like these. It also tries to enable all features of an ext structure (see #928). But since I don't want to change anything about the framework in this PR (would be out of it's scope) I'll see if I can workaround this.

SaschaWillems:
This is kinda hard to debug, and I feel like this is platform-specific. The only place where this error is thrown is in (hpp_)phyiscal_device.h and at least on Windows I never even enter that code when running my sample. So we probably do something different/specific when running on your platform that causes the assert to trigger. I'll somehow try to reproduce this on Windows.

I guess it's caused by this part of device creation: https://github.com/KhronosGroup/Vulkan-Samples/blob/main/framework/core/device.cpp#L109

It'll implicitly enable some extensions for all samples if they're supported. 

Sometimes I wish our framework wouldn't do so much implicit stuff. Often feels like the exact opposite of Vulkan :/

SaschaWillems:
Sorry, this currently fails after merging main. Lots of framework changes and at least something missing for me to continue working on this, see #985,

gary-sweet (changes requested):
I believe this is now missing an entry in samples/CMakeLists.txt. This means the sample isn't currently runnable and doesn't appear in the samples lists in the help.

SaschaWillems:
Welp, probably happened while merging with main. Thanks for letting me know, will fix.

SaschaWillems:
Can you check again? Afair the list in that cmake file is purely optional. I'm actually not sure why that list of samples in there even exist. CMake should do a folder traversal and adds all samples it can find, no matter if the sample is explicitly added in that list or not.

gary-sweet:
> Can you check again? Afair the list in that cmake file is purely optional. I'm actually not sure why that list of samples in there even exist. CMake should do a folder traversal and adds all samples it can find, no matter if the sample is explicitly added in that list or not.

Well that's odd then, because the sample didn't exist for me unless I hacked it into that file.

SaschaWillems:
That's interesting. I just reran cmake and deleted the cache beforehand and the sample is somehow picked up. Will have to investigate.

SaschaWillems:
Added it to the CMake file anyway. Can't hurt ;)

gary-sweet:
> That's interesting. I just reran cmake and deleted the cache beforehand and the sample is somehow picked up. Will have to investigate.

Hmm. So I trashed my build folders completely and rebuilt and then it appeared, without the entry in cmake. It's possible that the cmake generate hadn't run, which is what I was seeing.

gary-sweet (approved):
The fact that shader-printf 'just works' (provided that you support VK_KHR_shader_non_semantic_info) had kind of passed me by and I had to scratch my head for a little bit when I saw it working with my driver to figure out why!

What a very useful thing, and an excellent sample.

[If I had an extra button to auto-rotate the model it would be even better (since I have no other way to rotate API samples), but I can always add one later on.]

PR #946, opened 2024-02-25
Use more framework functionality in mobile_nerf

### LINKED ISSUE #940: Superfluous barriers in mobile_nerf
There are a couple of places where the mobile_nerf code uses a staging buffer to populate a device buffer, but also places a barrier right after the staging -> device buffer copy.

https://github.com/KhronosGroup/Vulkan-Samples/blob/6fbdb437e2202c7b8a6f453dae017081310a0938/samples/general/mobile_nerf/mobile_nerf.cpp#L1193

https://github.com/KhronosGroup/Vulkan-Samples/blob/6fbdb437e2202c7b8a6f453dae017081310a0938/samples/general/mobile_nerf/mobile_nerf.cpp#L1464

However, in every case, as far as I can tell, these barriers are all superfluous, because these command buffers are one-off executions created at startup time and each waits on a fence for the corresponding command buffer, no barrier should be required.  

Per the [understanding synchronization](https://www.khronos.org/blog/understanding-vulkan-synchronization) page...

> Semaphores are signaled only after all commands in the batch are complete. They make implicit memory guarantees, so we can access any memory following the semaphore without needing to think about adding memory barriers between them.

and

> Fences make the same implicit memory guarantee as semaphores

So if you know that an operation populating memory is going to be guarded by a fence, and it's not an image needing a layout transition, then you have no need of any explicit barriers.

### LINKED ISSUE #947: Allocated memory in mobile NERF sample is never freed
The mobile_nerf sample contains a call to `stbi_load` but no corresponding call to `stbi_image_free`.

https://github.com/KhronosGroup/Vulkan-Samples/blob/6fbdb437e2202c7b8a6f453dae017081310a0938/samples/general/mobile_nerf/mobile_nerf.cpp#L1519

jherico:
## Description

The mobile_nerf example makes minimal use of the framework code to offload boilerplate.  It also has some strange patterns in use in the code that uploads from staging buffers, such as putting the usage flags (vertex_buffer, index_buffer) onto the staging buffers, even though they're not required.

Fixes #940

Fixed #947

Tested on Windows 10, nVidia 2080 driver 546.01

## General Checklist:

## Sample Checklist

- [X] ~If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements)~

### DIFF
--- a/samples/general/mobile_nerf/mobile_nerf.cpp
+++ b/samples/general/mobile_nerf/mobile_nerf.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2023, Qualcomm Innovation Center, Inc. All rights reserved.
+/* Copyright (c) 2023-2024, Qualcomm Innovation Center, Inc. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -127,14 +127,7 @@ MobileNerf::~MobileNerf()
 			model.index_buffer.reset();
 
 			vkDestroySampler(get_device().get_handle(), model.texture_input_0.sampler, nullptr);
-			vkDestroyImageView(get_device().get_handle(), model.texture_input_0.view, nullptr);
-			vkDestroyImage(get_device().get_handle(), model.texture_input_0.image, nullptr);
-			vkFreeMemory(get_device().get_handle(), model.texture_input_0.memory, nullptr);
-
 			vkDestroySampler(get_device().get_handle(), model.texture_input_1.sampler, nullptr);
-			vkDestroyImageView(get_device().get_handle(), model.texture_input_1.view, nullptr);
-			vkDestroyImage(get_device().get_handle(), model.texture_input_1.image, nullptr);
-			vkFreeMemory(get_device().get_handle(), model.texture_input_1.memory, nullptr);
 
 			vkDestroyPipeline(device_ptr, model.pipeline_first_pass, nullptr);
 		}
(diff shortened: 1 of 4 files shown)


samples/general/mobile_nerf/mobile_nerf.cpp
```diff
@@ -1491,9 +1447,8 @@ void MobileNerf::create_texture(int model_index, int sub_model_index, int models
 	// Set up the input texture image
 
 	// TODO should load different scenes's feature map from command line
-	std::string assetBase      = vkb::fs::path::get(vkb::fs::path::Type::Assets);
```
jherico, line 1494: the `load_texture` function from the base class already prepends the asset path

samples/general/mobile_nerf/mobile_nerf.h
```diff
@@ -85,26 +85,6 @@ class MobileNerf : public ApiVulkanSample
 		}
 	};
 
-	struct VertexHash
```
jherico, line 88: This class is unused

samples/general/mobile_nerf/mobile_nerf.h
```diff
@@ -85,26 +85,6 @@ class MobileNerf : public ApiVulkanSample
...
-	{
-		size_t operator()(Vertex const &vertex) const
-		{
-			return std::hash<glm::vec3>()(vertex.position) ^
-			       (std::hash<glm::vec2>()(vertex.tex_coord) << 1);
-		}
-	};
-
-	struct Texture_Input
```
jherico, line 97: Replaced with the `Texture` class returned by the `ApiVulkanSample::load_texture` method.

RodrigoHolztrattner-QuIC (approved):
Looks good to me, tested and working as intended but with much more code clarity, thank you for your contribution!

samples/general/mobile_nerf/mobile_nerf.cpp
```diff
@@ -1167,39 +1133,41 @@ void MobileNerf::create_static_object_buffers(int model_index, int sub_model_ind
...
-		barrier.src_stage_mask  = VK_PIPELINE_STAGE_TRANSFER_BIT;
-		barrier.dst_stage_mask  = VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT;
-		barrier.src_access_mask = VK_ACCESS_TRANSFER_WRITE_BIT;
-		barrier.dst_access_mask = VK_ACCESS_SHADER_READ_BIT | VK_ACCESS_SHADER_WRITE_BIT;
-		cmd.buffer_memory_barrier(*output_buffer, 0, VK_WHOLE_SIZE, barrier);
-		return output_buffer;
-	};
-	model.vertex_buffer = copy(*staging_vertex_buffer, vertex_flags);
-	model.index_buffer  = copy(*staging_index_buffer, index_flags);
+	with_vkb_command_buffer([&](vkb::CommandBuffer &cmd) {
```
tomadamatkinson, line 1166: :fire:

tomadamatkinson (approved):

gary-sweet (approved):

SaschaWillems (approved):

PR #951, opened 2024-02-27
Update ASTC version

jherico:
## Description

* Update the astcenc library version to the most recent release.  
  * Detect SIMD support and build the corresponding atscenc variant
  * Fixes a number of potential memory leaks caused by the non-thread-safe code in the previous version of astcenc
* Update the mipmap generation code to allocate the required memory for all mips at once instead of mip by mip (less fragmentation, and should be faster)

## Performance

Ran the command_buffer_usage sample 5 times against this branch and against the current main branch, recording the _Time spent loading images_ log line for each run.  The target machine is a Ryzen Threadripper 1950X with 32 logical processor and for which the AVX2 ISA was detected and used.

| branch | average load time | standard deviation |
| - | - | - |
| main | 5.6638216 | 0.07 |
| astc | 3.3971104 | 0.06 |

This represents a 40% increase in ASTC image decompression performance.

Note that the only caveat for this is that the modules being used to detect the SIMD instruction set were only found under the license embedded in the added CMake files, which is a slight variant of the BSD-3 license.  Hopefully this shouldn't be a problem.

## General Checklist:

  - Strictly speaking there may be new or different compiler warnings inside the ASTC build, but that seems out of scope.

### DIFF
--- a/framework/CMakeLists.txt
+++ b/framework/CMakeLists.txt
@@ -78,7 +78,7 @@ set(COMMON_FILES
     common/ktx_common.h
     common/vk_common.h
     common/vk_initializers.h
-    common/glm_common.h 
+    common/glm_common.h
     common/resource_caching.h
     common/logging.h
     common/helpers.h
@@ -387,7 +387,7 @@ set(LINUX_D2D_FILES
     platform/unix/direct_window.h
     # Source Files
     platform/unix/unix_d2d_platform.cpp
-    platform/unix/direct_window.cpp) 
+    platform/unix/direct_window.cpp)
 
 source_group("\\" FILES ${FRAMEWORK_FILES})
 source_group("common\\" FILES ${COMMON_FILES})
@@ -514,7 +514,7 @@ target_link_libraries(${PROJECT_NAME} PUBLIC
     volk
     ktx
     stb
-    astc
+    ${ASTC_TARGET}
     imgui
     tinygltf
     glm
--- a/framework/scene_graph/components/image.cpp
+++ b/framework/scene_graph/components/image.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2023, Arm Limited and Contributors
+/* Copyright (c) 2018-2024, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -236,6 +236,24 @@ Mipmap &Image::get_mipmap(const size_t index)
 	return mipmaps[index];
 }
 
+// Note that this function returns the required size for ALL mip levels, *including* the base level.
+uint32_t get_required_mipmaps_size(const VkExtent3D &extent)
+{
+	constexpr uint32_t channels = 4;
+	auto               width    = std::max<uint32_t>(1, extent.width);
+	auto               height   = std::max<uint32_t>(1, extent.height);
+	auto               size     = width * height * channels;
+	auto               result   = size;
+	while (size != channels)
+	{
+		width  = std::max<uint32_t>(1u, width >> 1);
+		height = std::max<uint32_t>(1u, height >> 1);
+		size   = width * height * channels;
+		result += size;
+	}
+	return result;
+}
+
 void Image::generate_mipmaps()
 {
 	assert(mipmaps.size() == 1 && "Mipmaps already generated");
@@ -250,6 +268,10 @@ void Image::generate_mipmaps()
 	auto next_height = std::max<uint32_t>(1u, extent.height / 2);
 	auto channels    = 4;
 	auto next_size   = next_width * next_height * channels;
+
+	// Allocate for all the mips at once.  The function returns the total size needed for the
+	// existing base mip as well as all the mips that will be generated.
+	data.reserve(get_required_mipmaps_size(extent));
 
 	while (true)
 	{
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -290,44 +290,61 @@ endif()
 add_library(stb INTERFACE)
 target_include_directories(stb INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/stb")
 
+
+# On an x86_64 macbook, AVX2 is selected, but linking fails with this message,
+# `ld: object file .../libastcdec-avx2-static.a was built for different x86_64 sub-type (8) than link command line (3)`
+# so we fallback to native. If we're on an ARM based mac, there's no need for any SIMD extension, because they support
+# ASTC_LDR natively, so this library won't even be used.
+if (APPLE)
+    set(ASTC_ARCH NATIVE)
+else()
+    include( FindAVX2 )
+    include( FindSSE41 )
+    include( FindSSE2 )
+    include( FindNEON )
+    if (${AVX2_FOUND})
+        set(ASTC_ARCH AVX2)
+    elseif(${SSE41_FOUND})
+        set(ASTC_ARCH SSE41)
+    elseif(${SSE2_FOUND})
+        set(ASTC_ARCH SSE2)
+    elseif(${NEON_FOUND})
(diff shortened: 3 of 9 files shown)


framework/scene_graph/components/image.cpp
```diff
@@ -168,8 +168,7 @@ const std::vector<uint8_t> &Image::get_data() const
 
 void Image::clear_data()
 {
-	data.clear();
-	data.shrink_to_fit();
```
jherico: `shrink_to_fit` doesn't guarantee deallocation.

gary-sweet:
Is this PR also suffering from the CRLF issue? Whole files appear different.

framework/scene_graph/components/image.cpp
```diff
@@ -1,370 +1,396 @@
...
+		next_height = std::max<uint32_t>(1u, next_height / 2);
+		next_size   = next_width * next_height * channels;
+		if (next_width == 1 && next_height == 1)
+		{
+			break;
+		}
+	}
+
+	return result;
+}
```
asuessenbach: You're missing the 1x1 level.
Maybe better:
```
	constexpr uint32_t channels = 4;
	auto               width    = std::max<uint32_t>(1, extent.width);
	auto               height   = std::max<uint32_t>(1, extent.height);
	auto               size     = width * height * channels;
	auto               result   = size;
	while (size != channels)
	{
		width  = (width + 1) / 2;
		height = (height + 1) / 2;
		size   = width * height * channels;
		result += size;
	}

	return result;
```

framework/scene_graph/components/image.cpp
```diff
@@ -1,370 +1,396 @@
...
+	}
+
+	auto extent      = get_extent();
+	auto next_width  = std::max<uint32_t>(1u, extent.width / 2);
+	auto next_height = std::max<uint32_t>(1u, extent.height / 2);
+	auto channels    = 4;
+	auto next_size   = next_width * next_height * channels;
+
+	// Allocate for all the mips at once
+	data.reserve(get_required_mipmaps_size(extent));
```
asuessenbach: You probably wanted to say
`	data.reserve(data.size() + get_required_mipmaps_size(extent));
`

framework/scene_graph/components/image.cpp
jherico: Good point, but th suggested replacement fails as well for the degenerate case of a 1x1 image, so the while should be `do {...} while` I think.

framework/scene_graph/components/image.cpp
jherico: No, the mip size calculations above include lip #0.  I'll add a comment clarifying.

framework/scene_graph/components/image.cpp
asuessenbach: Ahm... no, I don't think it would fail.
`result` is initialized with the top-level `size`. If that top-level size is 1x1, it's correct.

jherico: Sorry, you're right, I missed that.

gary-sweet (dismissed):

tomadamatkinson (dismissed):
Good improvement. Thanks for performance testing!

SaschaWillems:
> Note that the only caveat for this is that the modules being used to detect the SIMD instruction set were only found under the license embedded in the added CMake files, which is a slight variant of the BSD-3 license. Hopefully this shouldn't be a problem.

We probably need to check this internally. @marty-johnson59: Maybe something to put on next calls agenda? Or forward to Legal at Khronos. It's a different license, so we need to make sure that we can use it and that we don't block any current user from using our samples by that new license.

SaschaWillems:
Tough tbh. I'd rather prefer moving all the assets that use astc to basisu instead. This is a nice stopgap, but probably becomes obsolete once we update the assets.

framework/CMakeLists.txt
```diff
@@ -532,6 +532,10 @@ if(${NEED_LINK_ATOMIC})
     target_link_libraries(${PROJECT_NAME} PUBLIC atomic)
 endif()
 
+# I'm uncertain why this has to be done explicitly, when the target_link_libraries call above should do it
```
SaschaWillems: Is this comment still current? If so we should tackle the issue before merging.

third_party/CMakeLists.txt
```diff
@@ -290,44 +290,61 @@ endif()
 add_library(stb INTERFACE)
 target_include_directories(stb INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/stb")
 
+
+# On an x86_64 macbook, AVX2 is selected, but then I get an error during the final link
```
SaschaWillems: Same as above. We either need to tackle this or reword the comment. Esp the "I..." part.

SaschaWillems (changes requested):
LGTM aside from the legal stuff. I only have some minor remarks regarding some comments.

framework/CMakeLists.txt
jherico: It is.  When I removed that I saw compile errors where `astcenc.h` could not be found.  In theory, the `target_link_libraries` call with the astc target _should_ automatically add the interface include paths.

SaschaWillems: So if it's not possible to fix this, could you rephrase the comment to not refer to you personally (the "I'm..." part). (can't explain it better, not a native English speaker).

jherico: it must have been some transient glitch.  I can't reproduce the error now.

third_party/CMakeLists.txt
jherico: fixed

tomadamatkinson (approved):

gary-sweet (approved):

SaschaWillems (approved):
Code LGTM.

We still need to check on licenses before merging.

tomadamatkinson:
@marty-johnson59 is a licence check needed on this?

marty-johnson59:
ATSC is under Apache 2 license, so we're OK with that.  Merging with 3..

SaschaWillems:
Looks like this broke builds on recent Apple devices (e.g. M1), see https://github.com/KhronosGroup/Vulkan-Samples/issues/965

kanavin:
@jherico @SaschaWillems there's another issue here: the newly added cmake modules check whether the compiler can produce the instructions, by trying various command line flags, and then forcibly enable them if so. 

This doesn't work when the build is intended for a machine that doesn't actually support those instructions; the checks should either try only the default compiler settings (e.g. no additional options), or have a switch to forcibly override ASTC_ARCH from cmake command line.

This was exposed by trying to build for 32 bit x86 (yes, it's still a valid architecture in 2024) - the checks happily enable AVX2 because 32bit x86 compiler can do it if you ask, even though the overall build is configured for a mid 2000s x86 CPU model that can't possibly have that.

jherico:
@kanavin is it critical that this be automatic, or would it be sufficient if you could force the ISA from the command line, like for instance `cmake -Bbuild -DASTC_ARCH=NONE`?  

If the latter is sufficient, then I have a PR up that should allow you to build: #978

kanavin:
> @kanavin is it critical that this be automatic, or would it be sufficient if you could force the ISA from the command line, like for instance `cmake -Bbuild -DASTC_ARCH=NONE`?
> 
> If the latter is sufficient, then I have a PR up that should allow you to build: #978

Something like this should work, yes. I've tested a local patch that does something similar, and was going to open a pull request:

https://github.com/kanavin/Vulkan-Samples/commit/41090ff333b2b4d888a15e3fba000229fc8740c3

While you're at it, SSE41 should become sse4.1 (with a dot), when lowercased, as that's how astc defines the target.

jherico:
Ugh, you're right... 🤦

image

I'm just going to exclude it for now.  It's a relatively narrow window of processors that support SSE 4.1 and NOT AVX2.

PR #956, opened 2024-02-28
Improve Help Logs

tomadamatkinson:
## Description

Displays help menu when running samples without a target app. This should help users new to the project get used to the command line options

image

Fixes #955

## General Checklist:

Please ensure the following points are checked:

- [x] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Code-Style)
- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Copyright-Notice-and-License-Template)
- [x] I have commented any added functions (in line with Doxygen)
- [x] I have commented any code that could be hard to understand
- [x] My changes do not add any new compiler warnings
- [x] My changes do not add any new validation layer errors or warnings
- [x] I have used existing framework/helper functions where possible
- [x] My changes do not add any regressions
- [x] I have tested every sample to ensure everything runs correctly
- [x] This PR describes the scope and expected impact of the changes I am making

 Note: The Samples CI runs a number of checks including:
 - [ ] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
 - [ ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements)

## Sample Checklist

If your PR contains a new or modified sample, these further checks must be carried out *in addition* to the General Checklist:
- [ ] I have tested the sample on at least one compliant Vulkan implementation
- [ ] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements)
- [ ] I have stated on what implementation the sample has been tested so that others can test on different implementations and platforms
- [ ] Any dependent assets have been merged and published in downstream modules
- [ ] For new samples, I have added a paragraph with a summary to the appropriate chapter in the readme of the folder that the sample belongs to [e.g. api samples readme](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/samples/api/README.adoc)
- [ ] For new samples, I have added a tutorial README.md file to guide users through what they need to know to implement code using this feature. For example, see [conditional_rendering](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/extensions/conditional_rendering)
[...]

PR #957, opened 2024-02-28
Don't ask to press return at the end of showing help on non-Windows platforms

gary-sweet:
## Description

Having to press return at the end of listing the samples or showing the help has been pretty annoying.
I suspect it's there for Windows users primarily to prevent to command window closing.
It was already excluded on Android platforms, I've also now excluded it on Linux platforms.

## General Checklist:

### DIFF
--- a/framework/platform/platform.cpp
+++ b/framework/platform/platform.cpp
@@ -248,7 +248,7 @@ void Platform::terminate(ExitCode code)
 	// Halt on all unsuccessful exit codes unless ForceClose is in use
 	if (code != ExitCode::Success && !using_plugin<::plugins::ForceClose>())
 	{
-#if !(defined PLATFORM__ANDROID || defined PLATFORM__LINUX)
+#ifdef PLATFORM__WINDOWS
 		std::cout << "Press return to continue";
 		std::cin.get();
 #endif
--- a/framework/platform/platform.cpp
+++ b/framework/platform/platform.cpp
@@ -248,7 +248,7 @@ void Platform::terminate(ExitCode code)
 	// Halt on all unsuccessful exit codes unless ForceClose is in use
 	if (code != ExitCode::Success && !using_plugin<::plugins::ForceClose>())
 	{
-#ifndef ANDROID
+#if !(defined PLATFORM__ANDROID || defined PLATFORM__LINUX)
 		std::cout << "Press return to continue";
 		std::cin.get();
 #endif
--- a/framework/platform/platform.cpp
+++ b/framework/platform/platform.cpp
@@ -245,14 +245,14 @@ void Platform::terminate(ExitCode code)
 
 	on_platform_close();
 
+#ifdef PLATFORM__WINDOWS
 	// Halt on all unsuccessful exit codes unless ForceClose is in use
 	if (code != ExitCode::Success && !using_plugin<::plugins::ForceClose>())
 	{
-#ifdef PLATFORM__WINDOWS
 		std::cout << "Press return to continue";
 		std::cin.get();
-#endif
 	}
+#endif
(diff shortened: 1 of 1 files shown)


tomadamatkinson (dismissed):
I think your right. This was there for the windows prompt

LGTM

gary-sweet:
> I think your right. This was there for the windows prompt

Should I change it only display on Windows in that case, and not for any other platform?

tomadamatkinson:
Actually yes that would be a better fix @gary-sweet

tomadamatkinson (dismissed):

framework/platform/platform.cpp
```diff
@@ -248,7 +248,7 @@ void Platform::terminate(ExitCode code)
 	// Halt on all unsuccessful exit codes unless ForceClose is in use
 	if (code != ExitCode::Success && !using_plugin<::plugins::ForceClose>())
 	{
-#ifndef ANDROID
+#ifdef PLATFORM__WINDOWS
 		std::cout << "Press return to continue";
 		std::cin.get();
 #endif
```
asuessenbach: Maybe the complete if-clause, not only the body, should be PLATFORM__WINDOWS-specific?

gary-sweet: Done

tomadamatkinson (approved):

asuessenbach (approved):

SaschaWillems:
Can we extend this to "Don't ask to press return at the end of showing help on terminals"? This might be fine if you run the binary from e.g. the explorer, but it's acting very weird if you run it from e.g. Windows terminal. You need to press a key after the help shows up, and the terminal prompt is then continuing in the middle of nowhere:

image

gary-sweet:
> Can we extend this to "Don't ask to press return at the end of showing help on terminals"? This might be fine if you run the binary from e.g. the explorer, but it's acting very weird if you run it from e.g. Windows terminal. You need to press a key after the help shows up, and the terminal prompt is then continuing in the middle of nowhere:

I'm not sure how anything I've changed here can affect this. The same code should be running on Windows platforms as it always was.

SaschaWillems:
Sorry, should've noted that this was caused by the change in #956. I think if we remove the key press requirement on Windows too, the odd behaviour goes away.

asuessenbach:
> Can we extend this to "Don't ask to press return at the end of showing help on terminals"?

To make it a bit more complex, this seems to depend on the terminal! I get different behaviours on different terminals (on Win10):
- with the simple command window, I have to enter return, no matter if std::cin.get() is called or not! But the next prompt is right after the help-output.
- with the PowerShell, I also have to enter return, no matter if std::cin.get() is called. And the next prompt is right after the previous prompt. That is: at the beginning of the help text!
- with a bash terminal, if std::cin.get() is not called, everything is as expected: you don't need to enter any key and the next prompt is right after the help-output.

That is, even though it seems you could reliably determine if the program has been started from any of those terminals, due to their different and partly strange behaviours it doesn't seem to be worth to make it so.

```
HWND consoleWnd = GetConsoleWindow();
DWORD consoleProcessId;
GetWindowThreadProcessId(consoleWnd, &consoleProcessId);
if (GetCurrentProcessId() == consoleProcessId)
```

gary-sweet:
I can't test anything on Windows, so if we believe this needs a different behaviour there I'm afraid someone else will have to implement it.  I don't see any reason that should prevent this PR going ahead though - any further change can be done on top of it.

SaschaWillems:
You're correct. Let's move the Windows related discussion to the PR or a new issue. Sorry for the noise.

SaschaWillems (approved):

Issue #920, opened 2024-02-19
Recreating resources in high level sample framework fundamentally broken (Subpass samples, etc.)


SaschaWillems:
While trying to fix the sync issues in the subpass sample (#455, #867) I noticed that it seems to be fundamentally broken/flawed. When enabling sync validation, the sample crashes the validation layers. After a few hours of debugging, it looks like it's trying to update the swap chain without properly making sure resources in use are waited on, resulting in validation crash. This happens right at startup. I was able to fix that one, but a similar error happens on resize. I'm looking into this, but may need some assistance as this is a sample done by Arm using the high level sample framework, so lots of indirections and implicit things happening that are very hard to debug.

![image](https://github.com/KhronosGroup/Vulkan-Samples/assets/10283127/ba24118f-32ed-4c39-a486-bbb4fca3d67e)

SaschaWillems:
This seems to affect several samples based on the Arm framework. I can reproduce this, albeit not with every single start/resize with e.g. the MSAA and also the command buffer usage sample. I think this is caused by how the framework recreates framebuffers/swapchains and as such may be a fundamental flaw in the Arm framework.

TzuChieh:
Sometimes (like 5% of the time) I would get this validation error when tweaking the subpasses sample:

[error] 732146062 - VUID-vkCmdDraw-None-08114: Validation Error: [ VUID-vkCmdDraw-None-08114 ] Object 0: handle = 0x1ee53b30478, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x2ba3a98e | vkCmdDraw():  the descriptor VkDescriptorSet 0x1ee53b30478[] [Set 0, Binding 0, Index 0] is using imageView VkImageView 0x0[] that is invalid or has been destroyed.

Could be related.

SaschaWillems:
One of the recentl merged PRs seemst to have fixed this. Haven't encountered this in recent times, even while doing a larger PR with lots of sample runs.

PR #963, opened 2024-03-06
Add Image Compression Control Sample (no-op merge, please ignore)

### LINKED ISSUE #882: New sample for VK_EXT_image_compression_control
I am currently working on a Performance/Extension sample for VK_EXT_image_compression_control, the plan is that it will illustrate:
* How to query for fixed rate compression support
* How to enable fixed rate compression support
* How to measure the bandwidth benefits of compression

Will use this issue for status updates, and will assign it to a PR in the near future, feedback welcome!

JoseEmilio-ARM:
:warning: Due to issues with GitHub, the PR was closed with the incorrect label 'merged' (the only commit is already present in `main`) and cannot be re-opened, so moved the review to a new PR: #979 

## Description

This sample showcases how to control image compression, in particular framebuffer attachments and the swapchain. The use case is a simple post-processing effect (chromatic aberration) applied to the output color of a forward-rendering pass.
The developer can select "Default" compression, "Fixed rate" compression, or "None". The sample shows the impact each option has on bandwidth (write bytes counter) and memory footprint (image sizes once allocated). More info in the tutorial: `samples/performance/image_compression_control/README.adoc`

image_compression_control

Fixes #882 

Also added a small change to the AFBC sample to use the new extension if available, to confirm that AFBC is being used.

#### Image class:

Added helper functions to query for compression support, apply compression, and check that compression was applied. Also to return image size (to display footprint).

#### Swapchain class:

Similarly, added helper functions to query for compression support, apply compression, and check that compression was applied, for the swapchain images. This also requires exposing this control to the Render Context class.

#### PhysicalDevice/Device classes:

Moved `is_extension_supported` responsibilities to the `PhysicalDevice` class, so that we can check if an extension is supported (and therefore if we can request its features) before the `Device` is created.

#### String utils:

Added a few more enum-to-string functions for image flags.

## General Checklist:

 - [x] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements). Tested on Windows and Android.

## Sample Checklist


SaschaWillems (changes requested):
This looks like an interesting example and a nice addition to our repo :)

Sadly for me on Windows 11 with an NVIDIA RTX 4070 this sample crashes with the following error message:

image

The assert happens in `RenderFrame &RenderContext::get_active_frame()`.

I also get these validation errors:

[error] [framework\core\instance.cpp:50] -1876993556 - VUID-VkDeviceCreateInfo-pNext-pNext: Validation Error: [ VUID-VkDeviceCreateInfo-pNext-pNext ] Object 0: handle = 0x20e8d305360, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x901f59ec | vkCreateDevice(): pCreateInfo->pNext<VkPhysicalDeviceImageCompressionControlSwapchainFeaturesEXT> includes a pointer to a VkPhysicalDeviceImageCompressionControlSwapchainFeaturesEXT, but when creating VkDevice, the parent extension (VK_EXT_image_compression_control_swapchain) was not included in ppEnabledExtensionNames. The Vulkan spec states: Each pNext member of any structure (including this one) in the pNext chain must be either NULL or a pointer to a valid struct for extending VkDeviceCreateInfo (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkDeviceCreateInfo-pNext-pNext)

[error] [framework\core\instance.cpp:50] -1876993556 - VUID-VkDeviceCreateInfo-pNext-pNext: Validation Error: [ VUID-VkDeviceCreateInfo-pNext-pNext ] Object 0: handle = 0x20e8d305360, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x901f59ec | vkCreateDevice(): pCreateInfo->pNext<VkPhysicalDeviceImageCompressionControlFeaturesEXT> includes a pointer to a VkPhysicalDeviceImageCompressionControlFeaturesEXT, but when creating VkDevice, the parent extension (VK_EXT_image_compression_control) was not included in ppEnabledExtensionNames. The Vulkan spec states: Each pNext member of any structure (including this one) in the pNext chain must be either NULL or a pointer to a valid struct for extending VkDeviceCreateInfo (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkDeviceCreateInfo-pNext-pNext)

samples/performance/image_compression_control/README.adoc
```diff
@@ -0,0 +1,291 @@
...
+
+image::./images/bandwidth.png[Bandwidth savings, 700, align="center"]
+
+Bandwidth savings coming from image compression depend on the pixels being compressed.
+Moving the camera and showing different distribution of colors in the frame changes the results.
+Be sure to profile your application and verify which compression scheme is optimal in each case.
+For instance, images with a high proportion of solid color (e.g. normals or material properties) may be more optimally compressed with variable bitrates than with fixed-rate.
+This is the case for the Space Module scene shown above.
+
+== VK_KHR_image_compression_control
```
SaschaWillems: Shouldn't this be `VK_EXT_image_compression_control`? There is no extension named `VK_KHR_image_compression_control`.

samples/performance/image_compression_control/README.adoc
```diff
@@ -0,0 +1,291 @@
...
+
+This sample enables the https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_image_compression_control.html[`VK_KHR_image_compression_control`] extension and requests the relevant device feature, https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceImageCompressionControlFeaturesEXT.html[`imageCompressionControl`]
```
SaschaWillems: Same as above. The text says `VK_KHR_image_compression_control` but it links to `VK_ET_image_compression_control`.

SaschaWillems:
The crash on my setup is most probably caused by the lack of support for  `VK_EXT_image_compression_control_swapchain`. But in that case, the sample shouldn't crash, but rather fail gracefully by e.g. reporting the extension as unsupported.

JoseEmilio-ARM:
> The crash on my setup is most probably caused by the lack of support for `VK_EXT_image_compression_control_swapchain`. But in that case, the sample shouldn't crash, but rather fail gracefully by e.g. reporting the extension as unsupported.

Thanks a lot for reviewing and testing 🙏 I tried it on Windows too (Quadro T2000) and I cannot reproduce the crash or the validation messages (SDK 1.3.250.1), do you get to see the logs before the crash? The extensions are marked as optional, so I get

```
[warning] Optional device extension VK_EXT_image_compression_control not available, some features may be disabled
[warning] Optional device extension VK_EXT_image_compression_control_swapchain not available, some features may be disabled
```

Then the sample should report that the extensions are not supported like this:

image

You think my drivers are more permissive, and perhaps there's a bug in [`request_extension_features`](https://github.com/KhronosGroup/Vulkan-Samples/pull/963/files#diff-42edc461419a276f27449b02303c10182c314f869b3080d049337ffc09da2a89R49)? or maybe we are expected to guard the calls in `request_gpu_features` with checks like `if (device->is_enabled(VK_EXT_IMAGE_COMPRESSION_CONTROL_EXTENSION_NAME))`?

JoseEmilio-ARM:
>or maybe we are expected to guard the calls in `request_gpu_features` with checks

~~Added a small patch to test this theory, @SaschaWillems please let me know if you have a chance to test again, thanks!~~

My bad, that seems to break things for me. Looking at other samples, they all request features this way, not sure what the issue is with this particular extension and only on some platforms 🤔 any ideas?

SaschaWillems:
I'm on a more recent build of the layers (built from source), and they report an error instead of a warning. This may cause the device to not get created.

Will check with layers from the latest SDK and report back.

SaschaWillems:
With SDK 1.3.275 I'm not getting any validation errors, I still get the assertion I posted above. If I ignore it, the sample will run though. So maybe something isn't properly initialized at first start? May be caused by what I described in #920.

JoseEmilio-ARM:
> With SDK 1.3.275 I'm not getting any validation errors, I still get the assertion I posted above. If I ignore it, the sample will run though. So maybe something isn't properly initialized at first start? May be caused by what I described in #920.

Thank you very much for checking, I'll look into it, although for some reason I cannot reproduce the assertion. Out of curiosity, do you get the same issue with other samples that require an extension that your platform does not support?

SaschaWillems:
Hard to test, as my GPU supports all extension we use in our samples. But "faking" this by requesting an unsupported extension in a sample looks to work properly. If I request an unsupported extension in an api based sample it displays an error and closes gracefully. Maybe it's something related to the ARM framework class after all?

This is where the error happens:

image

It triggers the assert, but only on the first frame of the sample. So I guess the first frame of the sample tries to access something not (yet) created. Hence why it works if I skip this. For the next frame, everything is correct.

JoseEmilio-ARM:
> It triggers the assert, but only on the first frame of the sample. So I guess the first frame of the sample tries to access something not (yet) created. Hence why it works if I skip this. For the next frame, everything is correct.

It could well be a framework issue, but finding it hard to debug without being able to reproduce. I did a clean-build on a different PC with a different GPU (RTX 3070) and the sample still works fine 🤔 

On the other hand I found that the new `Swapchain::query_supported_fixed_rate_compression` was using extension features, so added a new guard against it (see latest commit). Any chance you could try again? and if so, could you share the sample's LOGs up to the point of the assertion?

SaschaWillems:
The assert on the first frame is sadly still present. 

What's odd: The last three lines are

```
[info] Depth format selected: VK_FORMAT_D32_SFLOAT
[info] Depth format selected: VK_FORMAT_D32_SFLOAT
[info] Depth format selected: VK_FORMAT_D32_SFLOAT
```

Not sure why the sample (or framework) does that three times in a row.

I do see an error on the log though: 

[error] [framework\core\swapchain.cpp:645] (Swapchain) To query fixed-rate compression support, device extension VK_EXT_image_compression_control_swapchain must be enabled

Here is the log up to the first assert: [log_tex_compression.txt](https://github.com/KhronosGroup/Vulkan-Samples/files/14546543/log_tex_compression.txt)

JoseEmilio-ARM:
@SaschaWillems thank you very much for your help, the logs allowed me to find a few more issues. Also realized I was testing release builds and that's why I could not reproduce the assertion, apologies!

I think both the assertion and the validation layer error should now be fixed.

The repeated logs about the selected depth format are expected, I think they are related to the fact that we create a render target per RenderFrame in flight, which at the moment is one per swapchain image. Agree this should be improved as we probably do not need more than 2.

SaschaWillems (dismissed):
I can confirm that this now works as expected for me. Code and documentation also looking good to me 👍🏻

gary-sweet:
This works well for me. We only support one level of compression on our hardware, so the low/high selector in the UI just appears to do nothing.  Would it be sensible to hide that UI selector if only one level is available?

JoseEmilio-ARM:
> This works well for me. We only support one level of compression on our hardware, so the low/high selector in the UI just appears to do nothing. Would it be sensible to hide that UI selector if only one level is available?

Thank you very much for testing! Good point, certainly. Please let me know if my last commit works as expected.

gary-sweet (dismissed):
Thanks for the update, LGTM now.

framework/common/strings.cpp
```diff
@@ -1274,6 +1587,47 @@ const std::string color_component_to_string(VkColorComponentFlags flags)
 	                                            {VK_COLOR_COMPONENT_A_BIT, "A"}});
 }
 
+const std::string image_compression_flags_to_string(VkImageCompressionFlagsEXT flags)
+{
+	return 0 == flags ? "VK_IMAGE_COMPRESSION_DEFAULT_EXT" :
+	                    to_string<VkImageCompressionFlagsEXT>(flags,
+	                                                          {{VK_IMAGE_COMPRESSION_DEFAULT_EXT, "VK_IMAGE_COMPRESSION_DEFAULT_EXT"},
```
asuessenbach: No need to have `VK_IMAGE_COMPRESSION_DEFAULT_EXT` in the map and explicitly check for it (`0 == flags`)

framework/common/strings.cpp
```diff
@@ -1274,6 +1587,47 @@ const std::string color_component_to_string(VkColorComponentFlags flags)
...
+	                                                           {VK_IMAGE_COMPRESSION_FIXED_RATE_DEFAULT_EXT, "VK_IMAGE_COMPRESSION_FIXED_RATE_DEFAULT_EXT"},
+	                                                           {VK_IMAGE_COMPRESSION_FIXED_RATE_EXPLICIT_EXT, "VK_IMAGE_COMPRESSION_FIXED_RATE_EXPLICIT_EXT"},
+	                                                           {VK_IMAGE_COMPRESSION_DISABLED_EXT, "VK_IMAGE_COMPRESSION_DISABLED_EXT"}});
+}
+
+const std::string image_compression_fixed_rate_flags_to_string(VkImageCompressionFixedRateFlagsEXT flags)
+{
+	return 0 == flags ? "VK_IMAGE_COMPRESSION_FIXED_RATE_NONE_EXT" :
+	                    to_string<VkImageCompressionFixedRateFlagsEXT>(flags,
+	                                                                   {{VK_IMAGE_COMPRESSION_FIXED_RATE_NONE_EXT, "VK_IMAGE_COMPRESSION_FIXED_RATE_NONE_EXT"},
```
asuessenbach: As above: no need to have `VK_IMAGE_COMPRESSION_FIXED_RATE_NONE_EXT` in the map.

framework/core/image.cpp
```diff
@@ -251,5 +317,95 @@ std::unordered_set<ImageView *> &Image::get_views()
...
+		if (all_flags[i] & flags)
+		{
+			flags_vector.push_back(all_flags[i]);
+		}
+	}
+
+	return flags_vector;
+}
+
+VkImageCompressionPropertiesEXT Image::query_supported_fixed_rate_compression(const Device &device, VkImageCreateInfo &create_info)
```
asuessenbach: You should pass in a `VkImageCreateInfo const &`.
Or maybe even "just" the format, imageType, tiling, and usage, as all the other stuff from the create_info is not used.

framework/core/image.cpp
```diff
@@ -251,5 +317,95 @@ std::unordered_set<ImageView *> &Image::get_views()
...
+VkDeviceSize Image::get_image_size() const
+{
+	VkMemoryRequirements memory_requirements;
+
+	vkGetImageMemoryRequirements(device->get_handle(), handle, &memory_requirements);
+
+	return memory_requirements.size;
+}
+
+std::vector<VkImageCompressionFixedRateFlagBitsEXT> Image::fixed_rate_compression_flags_to_vector(VkImageCompressionFixedRateFlagsEXT flags)
```
asuessenbach: Isn't this a const function?

framework/core/swapchain.cpp
```diff
@@ -414,6 +444,35 @@ Swapchain::Swapchain(Swapchain                             &old_swapchain,
 	create_info.oldSwapchain     = properties.old_swapchain;
 	create_info.surface          = surface;
 
+	VkImageCompressionFixedRateFlagsEXT fixed_rate_flags[1] = {requested_compression_fixed_rate};
```
asuessenbach: An array of one element?
I think, you should either have just a single value and set the compressionControlPlaneCount below to 1
or have an array or a vector of values and set that count to the size of that array or vector.
As compressionControlPlaneCount is not obviously the size for the pFixedRateFlags, I would prefer to have the latter.

framework/core/swapchain.cpp
```diff
@@ -427,6 +486,32 @@ Swapchain::Swapchain(Swapchain                             &old_swapchain,
...
 
 	VK_CHECK(vkGetSwapchainImagesKHR(device.get_handle(), handle, &image_available, images.data()));
+
+	if (device.is_enabled(VK_EXT_IMAGE_COMPRESSION_CONTROL_SWAPCHAIN_EXTENSION_NAME) &&
+	    VK_IMAGE_COMPRESSION_FIXED_RATE_EXPLICIT_EXT == requested_compression)
+	{
+		// Check if fixed-rate compression was applied
+		const auto applied_fixed_rate_compression = core::Image::query_applied_compression(device.get_handle(), images[0]).imageCompressionFixedRateFlags;
+
+		if (applied_fixed_rate_compression != requested_compression_fixed_rate)
```
asuessenbach: I would use the same naming scheme: either `applied_fixed_rate_compression` and `requested_fixed_rate_compression`, or `applied_compression_fixed_rate` and `requested_compression_fixed_rate`.

JoseEmilio-ARM:
@jherico I updated the Image and Swapchain classes in this PR (so that their constructors can take the new structures, introduced by the compression control extensions, in their `pNext` chains)

I had to rebase the Image changes now that #906 was merged. What do you think of how I updated the Image using the new Builder? It seems that #973 will conflict too...

framework/core/allocated.cpp
```diff
@@ -180,6 +180,11 @@ void AllocatedBase::post_create(VmaAllocationInfo const &allocation_info)
 		}
 #endif
 
+	if (VK_IMAGE_COMPRESSION_DEFAULT_EXT != compression_control.flags)
+	{
+		create_info.pNext = &compression_control;
+	}
```
jherico: so, I feel like this code actually belongs in the `ImageBuilder` class.  i.e. when you create a builder and you call `with_compression`, that's the appropriate time to actually modify the `pNext` chain.

Similarly, I think we should avoid raw `pNext` modification like this.  Compression is unlikely to be the last extension that can be used in `VkImageCreateInfo::pNext` chains that we'll ever do, and correspondingly we should always interact with the chain in a manner that reflects that it might already be populated, and ensure we inject items into the chain, rather than simply overwriting it.

Here, this could be as simple as 
```suggestion
	if (VK_IMAGE_COMPRESSION_DEFAULT_EXT != compression_control.flags)
	{
	        compression_control.pNext = create_info.pNext;
		create_info.pNext = &compression_control;
	}
```

But it's something that should probably be farmed out to a helper function somewhere, because you want to deal with edge cases like both structures having their own chains that need to be merged and runtime checks that a given chain doesn't contain more than one element of the same structure type.

framework/core/allocated.h
```diff
@@ -260,7 +260,7 @@ class AllocatedBase
   protected:
 	virtual void           post_create(VmaAllocationInfo const &allocation_info);
 	[[nodiscard]] VkBuffer create_buffer(VkBufferCreateInfo const &create_info);
-	[[nodiscard]] VkImage  create_image(VkImageCreateInfo const &create_info);
+	[[nodiscard]] VkImage  create_image(VkImageCreateInfo &create_info, VkImageCompressionControlEXT const &compression_control = {});
```
jherico: Per my last comment, I think this function shouldn't need to be modified, because by the time we get here, the `pNext` chain for `create_info` should already be populated.

framework/core/hpp_image.cpp
```diff
@@ -81,7 +81,7 @@ HPPImage::HPPImage(HPPDevice              &device,
 HPPImage::HPPImage(HPPDevice &device, HPPImageBuilder const &builder) :
     HPPAllocated{builder.alloc_create_info, nullptr, &device}, create_info{builder.create_info}
 {
-	get_handle()           = create_image(create_info.operator const VkImageCreateInfo &());
+	get_handle()           = create_image(create_info.operator VkImageCreateInfo &());
```
jherico: moving the `pNext` manipulation into the builder logic would also remove the need for this change.

framework/core/image.cpp
```diff
@@ -78,18 +78,20 @@ ImagePtr ImageBuilder::build_unique(const Device &device) const
...
+             VmaMemoryUsage                      memory_usage,
+             VkSampleCountFlagBits               sample_count,
+             const uint32_t                      mip_levels,
+             const uint32_t                      array_layers,
+             VkImageTiling                       tiling,
+             VkImageCreateFlags                  flags,
+             uint32_t                            num_queue_families,
+             const uint32_t                     *queue_families,
+             VkImageCompressionFlagsEXT          compression_flags,
+             VkImageCompressionFixedRateFlagsEXT fixed_rate_flags) :
```
jherico: I'd argue against making this ctor any more complicated, and instead using the `ImageBuilder` logic directly in the example that's using it.

framework/core/image.cpp
```diff
@@ -194,5 +197,18 @@ std::unordered_set<ImageView *> &Image::get_views()
 	return views;
 }
 
+VkDeviceSize Image::get_image_size() const
```
jherico: minor quibble but I would call this `get_image_allocated_size` or something like that.  Anything to avoid someone mistaking this for being a useful value to, for instance, determine how big a staging buffer needs to be to copy from an uncompressed source to the image.

framework/core/image.h
```diff
@@ -119,8 +119,77 @@ struct ImageBuilder : public allocated::Builder<ImageBuilder, VkImageCreateInfo>
...
+					if (create_info.samples > 1)
+					{
+						LOGW("\t- It is a multi-sampled image (sample count = {})", to_string(create_info.samples));
+					}
+
+					compression_control.flags = VK_IMAGE_COMPRESSION_DEFAULT_EXT;
+				}
+			}
+		}
+
```
jherico: I would update the `create_info.pNext` chain here.  The local code knows that the `compression_control.pNext` is empty, so you could just do 

```c++
compression_control.pNext = create_info.pNext;
create_info.pNext = &compression_control;
```

But that assumes that no one will ever call `with_compression` more than once with the same builder (which isn't a great assumption).  Hence using a helper function is valuable.

In my C++ bindings examples I have the following:

```c++
template <typename T, typename N>
bool containsNext(const T& t, const N& n) {
    const vk::BaseInStructure* tp = &reinterpret_cast<const vk::BaseInStructure&>(t);
    const vk::BaseInStructure* np = &reinterpret_cast<const vk::BaseInStructure&>(n);
    while (tp != nullptr) {
        if (tp == np) {
            return true;
        }
        tp = tp->pNext;
    }
    return false;
}

template <typename T, typename N>
void injectNext(T& t, N& n) {
    vk::BaseOutStructure* tp = &reinterpret_cast<vk::BaseOutStructure&>(t);
    vk::BaseOutStructure* np = &reinterpret_cast<vk::BaseOutStructure&>(n);
    np->pNext = tp->pNext;
    tp->pNext = np;
}
```

If they're adapted to the C bindings then the logic would become 

```c++
if (!containsNext(create_info, compression_control)) 
{
  injectNext(create_info, compression_control)
}
```

framework/core/physical_device.h
```diff
@@ -158,6 +160,9 @@ class PhysicalDevice
 	// The features that this GPU supports
 	VkPhysicalDeviceFeatures features{};
 
+	// The extensions that this GPU supports
+	std::vector<VkExtensionProperties> device_extensions;
+
```
jherico: I would make this an `std::unordered_map<std::string, VkExtensionProperties>` so that the code for `is_extension_supported` is a simple `return device_extensions.contains(requested_extension);`

I get this is basically just moved code from `device.h`, but it seems like a little extra work up-front makes the properties easier to work with.

framework/core/swapchain.h
```diff
@@ -159,5 +185,9 @@ class Swapchain
 	    {VK_FORMAT_B8G8R8A8_SRGB, VK_COLOR_SPACE_SRGB_NONLINEAR_KHR}};
 
 	std::set<VkImageUsageFlagBits> image_usage_flags;
+
+	VkImageCompressionFlagsEXT requested_compression{VK_IMAGE_COMPRESSION_DEFAULT_EXT};
+
+	VkImageCompressionFixedRateFlagsEXT requested_compression_fixed_rate{VK_IMAGE_COMPRESSION_FIXED_RATE_NONE_EXT};
```
jherico: These seem like they'd belong in the `SwapchainProperties` class, rather than in the `SwapChain` class itself.  All the other properties that are configurable at runtime are there (excepting some weirdness with the `image_usage_flags`).

jherico: Also, migrating these into the SwapchainProperties should make resolving any conflicts with #973 easier to handle, I think.  

Feel free to move #973 into draft mode and/or comment on it if you want it to be deferred until after this is merged.

JoseEmilio-ARM: yes please, I'll address your comments but I would appreciate if you could then adapt this in #973 🙏

framework/core/image.h
```diff
@@ -119,6 +119,18 @@ struct ImageBuilder : public allocated::Builder<ImageBuilder, VkImageCreateInfo>
...
 
+	ImageBuilder &with_compression(VkImageCompressionControlEXT *image_compression_control)
+	{
+		if (image_compression_control)
+		{
+			image_compression_control->pNext = create_info.pNext;
+
+			create_info.pNext = image_compression_control;
+		}
+
```
jherico: I didn't expect you to move the compression code completely out of the image builder, but it's probably a good approach since it means that we can add lots of extension support without adding additional members that mostly get unused.

However, I would suggest that in this case. the function be changed to something more generic

```suggestion
	template<typename ExtensionType>
	ImageBuilder &with_extension(ExtensionType &extension)
	{
		extension.pNext = create_info.pNext;
		create_info.pNext = &extension;
	}
```

This still isn't perfect, since we should really ensure that we're not overwriting `extension.pNext`, but that can be fixed up later.

framework/core/allocated.cpp
JoseEmilio-ARM: Addressed in my last commit

framework/core/allocated.h
JoseEmilio-ARM: Addressed in my last commit

framework/core/hpp_image.cpp
JoseEmilio-ARM: Addressed in my last commit

framework/core/image.cpp
JoseEmilio-ARM: Addressed in my last commit

framework/core/image.cpp
JoseEmilio-ARM: Addressed in my last commit

framework/core/swapchain.h
jherico: Sure thing.

framework/core/image.h
JoseEmilio-ARM: I like this solution, but I think it's out of the scope of this PR 😅 I'd prefer to leave if for a future refactor if/when we have more structures to consider?

framework/core/physical_device.h
JoseEmilio-ARM: I tried this, but I think it requires C++ 20, and the samples project is 17 at the moment?

framework/core/device.cpp
```diff
@@ -75,21 +75,6 @@ Device::Device(PhysicalDevice                        &gpu,
...
-	// Display supported extensions
-	if (device_extensions.size() > 0)
-	{
-		LOGD("Device supports the following extensions:");
-		for (auto &extension : device_extensions)
-		{
-			LOGD("  \t{}", extension.extensionName);
-		}
-	}
-
```
jherico: I think if you're going to move this to `PhysicalDevice` you should make the corresponding change to `HPPDevice` and `HPPPhysicalDevice` so that they stay mostly in sync.

I'm less concerned with the corresponding changes to `Swapchain` and `Image`.

framework/core/physical_device.h
jherico: I don't think it requirest C++20 but if it's being problematic, don't let it slow down the actual sample.  

Honestly though I'd love C++20 because it would let us replace lots of functions that take `std::vector` to take a `std::span`, and allow us to pass vectors or `std::array` objects, thus allowing us to declare many current `std::vectors` as `constexpr std::array`.

framework/core/image.h
JoseEmilio-ARM: Addressed in newer commit

framework/core/device.cpp
JoseEmilio-ARM: Addressed in newer commit

jherico (approved):
LGTM, but bear in mind I'm not a Khronos member, so I don't think my review technically carries any weight.

JoseEmilio-ARM:
> LGTM, but bear in mind I'm not a Khronos member, so I don't think my review technically carries any weight.

Thank you very much! Your suggestions made it a much better PR 🙇

samples/performance/image_compression_control/image_compression_control.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+		VkFormat chosen_format{VK_FORMAT_UNDEFINED};
+		for (auto &candidate_format : format_list)
+		{
+			color_image_info.format = candidate_format;
+
+			if (!device->is_enabled(VK_EXT_IMAGE_COMPRESSION_CONTROL_EXTENSION_NAME))
+			{
+				break;
+			}
```
asuessenbach: Maybe a bit easier to follow:
```
VkFormat chosen_format{VK_FORMAT_UNDEFINED};
if (device->is_enabled(VK_EXT_IMAGE_COMPRESSION_CONTROL_EXTENSION_NAME))
{
	for (auto &candidate_format : format_list)
	...
}
```

samples/performance/image_compression_control/image_compression_control.cpp
```diff
@@ -0,0 +1,588 @@
...
+					chosen_format                    = candidate_format;
+					supported_fixed_rate_flags_color = fixed_rate_flags;
+				}
+
+				if (fixed_rate_flags.size() > 1)
+				{
+					chosen_format                    = candidate_format;
+					supported_fixed_rate_flags_color = fixed_rate_flags;
+					break;
+				}
```
asuessenbach: Maybe:
```
if ((VK_FORMAT_UNDEFINED == chosen_format) || (fixed_rate_flags.size() > 1))
{
	chosen_format                    = candidate_format;
	supported_fixed_rate_flags_color = fixed_rate_flags;
}

if (fixed_rate_flags.size() > 1)
{
	break;
}
```

samples/performance/image_compression_control/image_compression_control.cpp
```diff
@@ -0,0 +1,588 @@
...
+	                                 .with_vma_usage(VMA_MEMORY_USAGE_GPU_ONLY)};
+
+	vkb::core::ImageBuilder color_image_builder(color_image_info.extent);
+	color_image_builder.with_format(color_image_info.format)
+	    .with_usage(color_image_info.usage)
+	    .with_tiling(color_image_info.tiling)
+	    .with_vma_usage(VMA_MEMORY_USAGE_GPU_ONLY);
+
+	// Prepare compression control structure
+	VkImageCompressionFixedRateFlagsEXT color_fixed_rate_flags[1] = {compression_fixed_rate_flag_color};
```
asuessenbach: The same I said about a one-element array in swapchain.cpp

In fact, those two code sequences are nearly identical. Maybe worth a little helper function?

samples/performance/image_compression_control/image_compression_control.cpp
```diff
@@ -0,0 +1,588 @@
...
+		 * VkImageCompressionPropertiesEXT structure in its pNext chain.
+		 * This structure is then filled with the applied compression flags.
+		 */
+		LOGI("Applied fixed-rate compression for color ({}): {}", vkb::to_string(color_image_info.format),
+		     vkb::image_compression_fixed_rate_flags_to_string(color_image.get_applied_compression().imageCompressionFixedRateFlags));
+	}
+
+	// Update memory footprint values in GUI (displayed in MB)
+	footprint_swapchain = swapchain_image.get_image_allocated_size() / 1e6;
+	footprint_color     = color_image.get_image_allocated_size() / 1e6;
```
asuessenbach: Maybe divide by `(1024 * 1024)` to really get MB.

samples/performance/image_compression_control/image_compression_control.h
```diff
@@ -0,0 +1,189 @@
...
+	 * By default, many implementations support lossless compression.
+	 * Some also support fixed-rate compression, which is generally visually lossless.
+	 * This flag can also be used to disable all compression, which is not recommended.
+	 */
+	VkImageCompressionFlagsEXT compression_flag{VK_IMAGE_COMPRESSION_DEFAULT_EXT};
+
+	/**
+	 * @brief Possible compression schemes
+	 */
+	enum TargetCompression : int
```
asuessenbach: Maybe use an `enum class` here, to prevent polluting the namespace with the enum values.

samples/performance/image_compression_control/image_compression_control.h
```diff
@@ -0,0 +1,189 @@
...
+	/**
+	 * @brief Supported values for compression rate
+	 * These depend on the properties (format and usage) of the Swapchain image.
+	 */
+	std::vector<VkImageCompressionFixedRateFlagBitsEXT> supported_fixed_rate_flags_swapchain{};
+
+	/**
+	 * @brief Possible fixed-rate compression levels
+	 */
+	enum FixedRateCompressionLevel : int
```
asuessenbach: Again, maybe use an `enum class` here.

asuessenbach (commented):
Just a couple of more or less minor issues.
And a great usage of the builder pattern!

framework/core/image.cpp
JoseEmilio-ARM: it was `static`, but you might be looking at an old version of the diff, it was now moved to `vk_common.cpp`

samples/performance/image_compression_control/image_compression_control.cpp
JoseEmilio-ARM: I prefer to leave them separate (no helper) since compressing the image and the surface are two different use cases. I simplified the code as you suggested.

samples/performance/image_compression_control/image_compression_control.h
JoseEmilio-ARM: Done, although it meant adding many casts, wouldn't this have been cleaner?

framework/common/strings.cpp
JoseEmilio-ARM: Addressed in my last commit

framework/common/strings.cpp
JoseEmilio-ARM: Addressed in my last commit

framework/core/image.cpp
JoseEmilio-ARM: Addressed in my last commit

framework/core/image.cpp
JoseEmilio-ARM: Addressed in my last commit

framework/core/swapchain.cpp
JoseEmilio-ARM: Addressed in my last commit

samples/performance/image_compression_control/image_compression_control.cpp
JoseEmilio-ARM: Addressed in my last commit

samples/performance/image_compression_control/image_compression_control.cpp
JoseEmilio-ARM: Addressed in my last commit

samples/performance/image_compression_control/image_compression_control.h
JoseEmilio-ARM: Addressed in my last commit

samples/performance/image_compression_control/image_compression_control.cpp
JoseEmilio-ARM: Addressed in newer commit

framework/core/swapchain.cpp
JoseEmilio-ARM: Addressed in newer commit

JoseEmilio-ARM:
Apologies, for some reason, after the rebase, the PR kept picking up #910 as a change, even though it is the head of `main` right now. I tried another force-push with just that commit, and the PR was automatically closed. Does anyone know if it possible to re-open it? Otherwise, I'll create a new PR, sorry for the inconvenience!

PR #979, opened 2024-03-12
Add Image Compression Control Sample

### LINKED ISSUE #882: New sample for VK_EXT_image_compression_control

JoseEmilio-ARM:
:warning: Due to issues with GitHub, the original PR #963 was closed prematurely with the incorrect label 'merged' (the only commit is already present in `main`) and cannot be re-opened.

## Description

This sample showcases how to control image compression, in particular framebuffer attachments and the swapchain. The use case is a simple post-processing effect (chromatic aberration) applied to the output color of a forward-rendering pass.
The developer can select "Default" compression, "Fixed rate" compression, or "None". The sample shows the impact each option has on bandwidth (write bytes counter) and memory footprint (image sizes once allocated). More info in the tutorial: `samples/performance/image_compression_control/README.adoc`

image_compression_control

Fixes #882 

Also added a small change to the AFBC sample to use the new extension if available, to confirm that AFBC is being used.

#### Image class:

Added helper functions to query for compression support, apply compression, and check that compression was applied. Also to return image size (to display footprint).

#### Swapchain class:

Similarly, added helper functions to query for compression support, apply compression, and check that compression was applied, for the swapchain images. This also requires exposing this control to the Render Context class.

#### PhysicalDevice/Device classes:

Moved `is_extension_supported` responsibilities to the `PhysicalDevice` class, so that we can check if an extension is supported (and therefore if we can request its features) before the `Device` is created.

#### String utils:

Added a few more enum-to-string functions for image flags.

## General Checklist:

 - [x] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements). Tested on Windows and Android.

## Sample Checklist


framework/core/image.cpp
```diff
@@ -194,5 +194,18 @@ std::unordered_set<ImageView *> &Image::get_views()
 	return views;
 }
 
+VkDeviceSize Image::get_image_allocated_size() const
```
asuessenbach: Maybe `get_image_required_size` would be a better name.

framework/vulkan_sample.h
```diff
@@ -327,6 +327,13 @@ class VulkanSample : public vkb::Application
 	 */
 	static void set_viewport_and_scissor(CommandBufferType const &command_buffer, Extent2DType const &extent);
 
+	/**
+	 * @brief A list of surface formats in order of priority (vector[0] has high priority, vector[size-1] has low priority)
+	 */
+	std::vector<vk::SurfaceFormatKHR> default_surface_priority_list = {
```
asuessenbach: This has to be moved into the private section!
Would require a getter function then, with a `std::conditional` return value, similar to `get_device()` or `get_gui()`

samples/performance/image_compression_control/image_compression_control.cpp
```diff
@@ -0,0 +1,587 @@
...
+ * limitations under the License.
+ */
+
+#include "image_compression_control.h"
+
+#include "common/vk_common.h"
+#include "gltf_loader.h"
+#include "gui.h"
+#include "platform/platform.h"
+
```
asuessenbach: Is there a reason for this empty line?

samples/performance/image_compression_control/image_compression_control.cpp
```diff
@@ -0,0 +1,587 @@
...
+	                                 .with_vma_usage(VMA_MEMORY_USAGE_GPU_ONLY)};
+
+	vkb::core::ImageBuilder color_image_builder(color_image_info.extent);
+	color_image_builder.with_format(color_image_info.format)
+	    .with_usage(color_image_info.usage)
+	    .with_tiling(color_image_info.tiling)
+	    .with_vma_usage(VMA_MEMORY_USAGE_GPU_ONLY);
+
+	// Prepare compression control structure
+	auto                         color_fixed_rate_flags = compression_fixed_rate_flag_color;
```
asuessenbach: There's no need for this variable, you could use `compression_fixed_rate_flag_color` directly.

samples/performance/image_compression_control/image_compression_control.cpp
```diff
@@ -0,0 +1,587 @@
...
+	return true;
+}
+
+void ImageCompressionControlSample::create_render_context()
+{
+	/**
+	 * The framework expects a prioritized list of surface formats. For this sample, include
+	 * only those that can be compressed.
+	 */
+	std::vector<vk::SurfaceFormatKHR> surface_formats_that_support_compression;
```
asuessenbach: You should not use data types from the C++-bindings of vulkan in a C-bindings-sample

samples/performance/image_compression_control/image_compression_control.h
```diff
@@ -0,0 +1,189 @@
...
+
+	/* Helpers for managing GUI input */
+
+	int gui_target_compression{(int) TargetCompression::Default};
+
+	int last_gui_target_compression{gui_target_compression};
+
+	int gui_fixed_rate_compression_level{(int) FixedRateCompressionLevel::High};
+
+	int last_gui_fixed_rate_compression_level{gui_fixed_rate_compression_level};
```
asuessenbach: Why do you keep those four values as `int`?

asuessenbach (commented):
A couple of minor issues/questions again.

Note, though, that this review if by code-inspection only, as I don't have Hardware supporting those extensions.

framework/core/image.cpp
JoseEmilio-ARM: (I already renamed this once before) @jherico agree? :)

samples/performance/image_compression_control/image_compression_control.cpp
JoseEmilio-ARM: I had to do this after the rebase, the new header does not provide an alternative, does it?

samples/performance/image_compression_control/image_compression_control.h
JoseEmilio-ARM: ImGui requires it

samples/performance/image_compression_control/image_compression_control.cpp
JoseEmilio-ARM: ~it is passed to the function as an argument~ sorry mixing it up with the Swapchain case, I fixed it here in my last commit

samples/performance/image_compression_control/image_compression_control.cpp
JoseEmilio-ARM: yes, I think it's because clang-format sorts #includes alphabetically, but we have compilation issues in some cases, I can try again

framework/vulkan_sample.h
JoseEmilio-ARM: I had it as protected, but could not find the right place to move it after the rebase (with the unification changes)

I'll try to make it a conditional

samples/performance/image_compression_control/image_compression_control.cpp
JoseEmilio-ARM: OK I see we have

```
	void create_render_context(const std::vector<SurfaceFormatType> &surface_priority_list);
```

but maybe there's a bug, for some reason when building I was getting errors because the function sees a C++ binding. Might be related to the issue with the AFBC sample that crashes now when calling `update_swapchain`?

framework/core/image.cpp
JoseEmilio-ARM: Renamed in my last commit

framework/vulkan_sample.h
JoseEmilio-ARM: Done in my last commit, I see it now, thanks!

samples/performance/image_compression_control/image_compression_control.cpp
JoseEmilio-ARM: OK ignore my previous comment, I think I implemented what you suggested in my last commit, thanks!

samples/performance/image_compression_control/image_compression_control.h
JoseEmilio-ARM: Sorry got this mixed up as well, addressed in my last commit

samples/performance/image_compression_control/image_compression_control.cpp
JoseEmilio-ARM: Removed in my last commit

framework/vulkan_sample.h
```diff
@@ -253,21 +253,23 @@ class VulkanSample : public vkb::Application
...
-	bool                      has_device() const;
-	bool                      has_gui() const;
-	bool                      has_render_pipeline() const;
-	bool                      has_scene();
+	DeviceType                           &get_device();
+	DeviceType const                     &get_device() const;
+	GuiType                              &get_gui();
+	GuiType const                        &get_gui() const;
+	std::vector<SurfaceFormatType>       &get_surface_priority_list();
+	std::vector<SurfaceFormatType> const &get_surface_priority_list() const;
```
asuessenbach: As there's no inherent order for all those functions, I sorted them alphabetically.

samples/performance/image_compression_control/image_compression_control.cpp
```diff
@@ -153,13 +152,13 @@ void ImageCompressionControlSample::create_render_context()
 	else
 	{
 		// Filter default list to those formats that support compression
-		std::vector<vk::SurfaceFormatKHR> new_surface_priority_list;
+		std::vector<VkSurfaceFormatKHR> new_surface_priority_list;
 
-		for (size_t i = 0; i < default_surface_priority_list.size(); i++)
+		for (size_t i = 0; i < get_surface_priority_list().size(); i++)
```
asuessenbach: Maybe, to reduce the number of calls to `get_surface_priority_list`:
`auto const & default_surface_priority_list = get_surface_priority_list();`

asuessenbach (dismissed):

samples/performance/image_compression_control/image_compression_control.cpp
```diff
@@ -152,13 +152,13 @@ void ImageCompressionControlSample::create_render_context()
 	else
 	{
 		// Filter default list to those formats that support compression
+		auto const                     &default_surface_priority_list = get_surface_priority_list();
 		std::vector<VkSurfaceFormatKHR> new_surface_priority_list;
-
-		for (size_t i = 0; i < get_surface_priority_list().size(); i++)
+		for (size_t i = 0; i < default_surface_priority_list.size(); i++)
```
asuessenbach: Maybe an even nicer formulation would be
`for ( auto const & surface_priority : get_surface_priority_list() )`

asuessenbach (dismissed):

asuessenbach (approved):

gary-sweet (approved):

SaschaWillems (approved):
LGTM.

Only problem is the resize crash which is addressed by #982

marty-johnson59:
merging on 3

PR #983, opened 2024-03-14
Fix image transitions among Samples using Dynamic_Rendering

Mateusz-Kowalewski-Mobica:
With new versions of VVL for Samples which are using Dynamic_Rendering it appear synchronization issue described here:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7193#top

I identify all Samples which are using that extension and I fix it.

Note: For Shader_Object Sample I need to wait for fix regarding latest union of C or C++ framework (It can't be tested right now)

### DIFF
--- a/samples/extensions/dynamic_rendering/dynamic_rendering.cpp
+++ b/samples/extensions/dynamic_rendering/dynamic_rendering.cpp
@@ -372,6 +372,10 @@ void DynamicRendering::build_command_buffers()
 		{
 			vkb::image_layout_transition(draw_cmd_buffer,
 			                             swapchain_buffers[i].image,
+			                             VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT,
+			                             VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT,
+			                             0,
+			                             VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT,
 			                             VK_IMAGE_LAYOUT_UNDEFINED,
 			                             VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL,
 			                             range);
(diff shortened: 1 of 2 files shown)


samples/extensions/dynamic_rendering/dynamic_rendering.cpp
```diff
@@ -372,6 +372,10 @@ void DynamicRendering::build_command_buffers()
 		{
 			vkb::image_layout_transition(draw_cmd_buffer,
 			                             swapchain_buffers[i].image,
+			                             VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT,
```
jherico, line 375: Assuming the acquisition of the swapchain image is protected by a fence guarding it from the prior present, I don't see why this wouldn't be `VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT` (or better yet `VK_PIPELINE_STAGE_2_NONE`), especially as the source layout is "undefined", meaning we don't care about the contents of this image anymore.

Mateusz-Kowalewski-Mobica: From that issue I mentioned seems that changing layout will happen earlier than waiting for rasterization write to image.

I'm not that much experience Vulkan engineer, but this sentence seems reasonable:
`There should be additional execution dependency with COLOR_ATTACHMENT_OUTPUT stage, otherwise image transition can start before image acquire operation finished. The destination stage is set to COLOR_ATTACHMENT_OUTPUT and this guarantees that writes from rasterization will wait. But the image layout transition happens earlier, and by setting source stage to COLOR_ATTACHMENT_OUTPUT we create execution dependency with semaphore wait operation (we chain with pWaitDstStageMask which also specifies COLOR_ATTACHMENT_OUTPUT ).`

jherico: I'll have to defer to someone with a better understanding of this than myself then.  I don't quite get why the dstStage isn't sufficient to ensure the proper ordering.

gpx1000 (approved):

SaschaWillems (approved):

zkrx:
It looks like `samples/extensions/dynamic_rendering_local_read` was left behind:
https://github.com/KhronosGroup/Vulkan-Samples/blob/04f03b32b860ab25ab2731433597d9a23929c386/samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp#L837

SaschaWillems:
This PR predates the local read sample. I'll do a separate PR to fix it.

PR #984, opened 2024-03-14
Unify the two classes vkb::core::VulkanResource and vkb::core::HPPVulkanResource into one: vkb::core::VulkanResource<BindingType, HandleType>

asuessenbach:
## Description

The next unification step of C- and C++-bindings: vkb::core[HPP]VulkanResource.

## General Checklist:

## Sample Checklist


framework/core/vulkan_resource.h
```diff
@@ -18,149 +19,266 @@
...
-		device = other.device;
-		set_debug_name(other.debug_name);
+		using Type = T;
+	};
+	template <typename T>
+	struct DetermineResourceType<vkb::BindingType::C, T>
+	{
+		using Type = typename detail::HPPType<T>::Type;
+	};
+	using ResourceType = typename DetermineResourceType<bindingType, Handle>::Type;
```
jherico: I feel like the code would be simpler to deal with it the `Handle`  template parameter and the `handle` member were both always the C++ type and the `get_handle()` return type was dynamic in this way, rather than the member.

framework/core/vulkan_resource.h
```diff
@@ -18,149 +19,266 @@
...
-	constexpr VkObjectType get_object_type() const
+template <vkb::BindingType bindingType, typename Handle>
+inline VulkanResource<bindingType, Handle> &VulkanResource<bindingType, Handle>::operator=(VulkanResource &&other)
+{
+	handle = std::exchange(other.handle, {});
+	device = std::exchange(other.device, {});
+	set_debug_name(std::exchange(other.debug_name, {}));
+
+	return *this;
+}
```
jherico: I've been meaning to open an issue about this.  `set_debug_name` has no reason to be called on the assignment operator or the move constructor, since if the debug name is non-empty, then `set_debug_name` was called at some point against the same handle value in the `other` object.  `debug_name` should be set, but the `set_debug_name` and the corresponding call to `get_debug_utils().set_debug_name()` are superfluous.

framework/core/vulkan_resource.h
jherico: an alternative would be to discontinue the use of `get_handle` in favor of `get_c_handle` and `get_cpp_handle`.

asuessenbach: This construct is a little more complex just because it always stores the C++-handle.
But from C-based code, you tell it the C-handle and also get the C-handle back by just calling `get_handle()`.
And from C++-based code, you tell it the C++-handle and also get the C++-handle back by just calling `get_handle()`.
That is, you trade the complexity of this class for the simplicity of the code using it.

Actually, C-based code usually is not interested (and the coder don't even need to know about) C++-based types, and vice versa. That is, you can use this class by consistently use the types and handles you're used to in your code.

(Note: C-based and C++-based here of course means: C-based Vulkan bindings vs. C++-based Vulkan bindings. Of course, we all code in C++ here!)

framework/core/vulkan_resource.h
asuessenbach: I just tried to keep the current functionality as close as possible. And I don't thought about any possible implications of changing the code like you suggest.
But, of course, could be easily adjusted like that right here, or be done in a separate PR.

jherico: I've been trying to think of the minimum viable path to unification, and right now I'm leaning towards "Remove all the C wrapper classes and replace them with statements like `using Device = vkb::core::HPPDevice;`"

This should result in the minimum amount of changed / moved code IMO, although there are a few special cases where the `HPP` version of a class is extending the `C` bindings version, such as is the case with `HPPQueryPool`.  I may work on a PR towards that end to get feedback on this approach.

framework/core/vulkan_resource.h
```diff
@@ -18,149 +19,267 @@
...
+
+template <vkb::BindingType bindingType, typename Handle>
+inline void VulkanResource<bindingType, Handle>::set_debug_name(const std::string &name)
+{
+	debug_name = name;
+
+	if (device && !debug_name.empty())
+	{
+		get_device().get_debug_utils().set_debug_name(get_device().get_handle(), get_object_type(), get_handle_u64(), debug_name.c_str());
```
SaschaWillems, line 274: Mostly cosmetic, but this looks very awkward. So many `get_` functions in here. Makes it kinda hard to read. I somehow prefer being able to directly access things like device, instance, atc. without having to use a getter.

asuessenbach: Unfortunately, directly accessing members of a class unifying C- and C++-bindings would not work, as those members should behave differently from C-based vs. C++-based code.
From C-based code, `get_handle()` gets you a `VkHandle`, while from C++-based code `get_handle()` gets you a `vk::Handle`. Internally (privately!), the handle is always stored as a `vk::Handle`, but your C-based code does not need to know that.

SaschaWillems: Makes sense. Readability is still kinda odd though.

asuessenbach: One could introduce some `operator*()`, replacing the `get_handle()`, at least.
Would reduce the number of get-calls a little, but make it a bit less descriptive.

SaschaWillems: Yeah, maybe we should do that at a later point. Resolving this as I don't want to spam the PR ;)

framework/core/vulkan_resource.h
jherico: I've opened a separate PR / Issue for it.  thanks.

gpx1000:
looks like a few merge conflicts.

Looks good to me other than that.

SaschaWillems (approved):
LGTM and works fine for me.

Did a full batch run, but that ended with a crash in the fragment shading rate dynamic sample, which should be fixed with #992.

gary-sweet (approved):

PR #987, opened 2024-03-15
Fix UB in descriptor_set_layout.cpp

### LINKED ISSUE #986: reference to an out of scope variable
The `binding_flags_create_info` created here:

https://github.com/KhronosGroup/Vulkan-Samples/blob/0d0cf9b90541d349ceb49682d063d27a8c2432de/framework/core/descriptor_set_layout.cpp#L174

is then used by the code on this line via the create_info.pNext variable.

https://github.com/KhronosGroup/Vulkan-Samples/blob/0d0cf9b90541d349ceb49682d063d27a8c2432de/framework/core/descriptor_set_layout.cpp#L183

Using a pointer to a variable that is out of scope is not safe.  This instance is probably not triggering problems due to the lack of any additional variables being created on the stack between the end of the `if` scope and the use of the `create_info` struct, but this is an implementation detail and I believe that the current code path is technically UB.

jherico:
## Description

Move the scope of an extension structure so that it has the same call as the parent structure to which it may be attached via a `pNext` chain.

Fixes #986

## General Checklist:

### DIFF
--- a/framework/core/descriptor_set_layout.cpp
+++ b/framework/core/descriptor_set_layout.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2019-2020, Arm Limited and Contributors
+/* Copyright (c) 2019-2024, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Great catch. I think the reason this never really caused much of a problem is that it only seems to be used in a single sample.

SaschaWillems (approved):

asuessenbach (approved):

gpx1000 (approved):

marty-johnson59:
Merging - 3 approvals

PR #950, opened 2024-02-27
WIP: Offline Shaders (Without Variants)

tomadamatkinson:
## Description

As discussed on the 27/02/2024 working group call. The current offline shader approaches are overly complex due to the need to support for variants. Removing variants and simplifying the performance samples use of shaders may reduce the complexity of our shader compilation system.

This PR achieves offline compilation through a CMake helper. It also runs spirv-val to validate the spirv after it is created. For now the `.spv` files are not added to the repository, this may change if we agree on where they should be stored (build dir or next to source).

The PR enables offline shaders for existing ApiVulkanSamples by altering `load_shader`

After a configure and build the following commands have been tested on desktop

`<path/to>/vulkan_samples batch -C api`
`<path/to>/vulkan_samples batch -C extensions`
`<path/to>/vulkan_samples batch -C general`
`<path/to>/vulkan_samples batch -C tooling`

Performance samples work by attempting to find a precompiled shader before loading the source in ShaderSource. Variants have been removed by reworking shaders in samples

`<path/to>/vulkan_samples batch -C performance`

Todo:
- [ ] constant_data
- [ ] msaa

Fixes #<issue number>

## General Checklist:

Please ensure the following points are checked:

- [ ] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Code-Style)
- [ ] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Copyright-Notice-and-License-Template)
- [ ] I have commented any added functions (in line with Doxygen)
- [ ] I have commented any code that could be hard to understand
- [ ] My changes do not add any new compiler warnings
- [ ] My changes do not add any new validation layer errors or warnings
- [ ] I have used existing framework/helper functions where possible
- [ ] My changes do not add any regressions
- [ ] I have tested every sample to ensure everything runs correctly
- [ ] This PR describes the scope and expected impact of the changes I am making

 Note: The Samples CI runs a number of checks including:
 - [ ] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
 - [ ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements)

## Sample Checklist

[...]

PR #990, opened 2024-03-18
Add offline compilation for HLSL shaders with usage of DXC

Seweryn-Zielas-Mobica:
## Description

Add offline compilation for HLSL with usage of DXC that is bundled with Vulkan SDK.
HLSL shaders will be compiled only if DXC executable is found/provided, otherwise compilation will go as before, without any interruptions.

## General Checklist:

## Sample Checklist


CMakeLists.txt
```diff
@@ -30,6 +31,34 @@ endif()
...
+                        HINTS
+                        "$ENV{VULKAN_SDK}/Bin"
+                        )
+        else()
+            find_program(Vulkan_dxc_EXECUTABLE
+                        NAMES dxc
+                        HINTS
+                        "$ENV{VULKAN_SDK}/bin"
+                        )
+        endif()
```
asuessenbach: The two branches are very close!
Maybe something like this is easier to grasp:
```
if (WIN32)
    set( BinName Bin )
else()
    set( BinName bin )
endif()
find_program(... "$ENV{VULKAN_SDK}/${BinName})
```

bldsys/cmake/sample_helper.cmake
```diff
@@ -70,6 +73,14 @@ function(add_sample_with_tags)
         endforeach()        
     endif()
 
+    # Add HLSL shader files for this sample
+    if (TARGET_SHADER_FILES_HLSL)
+        list(APPEND SHADER_FILES_HLSL ${TARGET_SHADER_FILES_HLSL})
```
asuessenbach: Is this intermediate list `SHADER_FILES_HLSL` needed at all? You should be able to directly iterate over `${TARGET_SHADER_FILES_HSLS}`
Note: the same holds for the `SHADER_FILES_GLSL` above.

bldsys/cmake/sample_helper.cmake
```diff
@@ -135,15 +148,20 @@ function(add_project)
...
     source_group("\\" FILES ${TARGET_FILES})
 
-    # Add shaders to project group
+    # Add GLSL shaders to project group
     if (SHADERS_GLSL)
         source_group("\\Shaders" FILES ${SHADERS_GLSL})
     endif()
 
+    #Add HLSL shaders to project group
+    if (SHADERS_HLSL)
```
asuessenbach: I see, it works as it is... but shouldn't be `TARGET_SHADERS_HSLS` used here, instead of plain `SHADERS_HSLS`?
The same holds for the SHADERS_GLSL above.

bldsys/cmake/sample_helper.cmake
```diff
@@ -181,4 +199,36 @@ endif()
     if(VKB_DO_CLANG_TIDY)
         set_target_properties(${PROJECT_NAME} PROPERTIES CXX_CLANG_TIDY "${VKB_DO_CLANG_TIDY}")
     endif()
+
+    if(DEFINED Vulkan_dxc_EXECUTABLE AND DEFINED SHADERS_HLSL)
+        compile_hlsl_shaders()
+    endif()
+endfunction()
+
+function(compile_hlsl_shaders)
```
asuessenbach, line 204: Wouldn't it be cleaner, if the shaders to be handled would be passed into this function as arguments?

bldsys/cmake/sample_helper.cmake
```diff
@@ -181,4 +199,36 @@ endif()
...
+        elseif(${SHADER_FILE_HLSL} MATCHES "[^-]+.comp.hlsl")
+            set(DXC_PROFILE "cs_6_1")
+        elseif(${SHADER_FILE_HLSL} MATCHES "[^-]+.geom.hlsl")
+            set(DXC_PROFILE "gs_6_1")
+        elseif(${SHADER_FILE_HLSL} MATCHES "[^-]+.tesc.hlsl")
+            set(DXC_PROFILE "hs_6_1")
+        elseif(${SHADER_FILE_HLSL} MATCHES "[^-]+.tese.hlsl")
+            set(DXC_PROFILE "ds_6_1")
+        elseif(${SHADER_FILE_HLSL} MATCHES "[^-]+.mesh.hlsl")
+            set(DXC_PROFILE "cs_6_1")
```
SaschaWillems: Not 100% sure, but I think this should be `ms_6_6` instead. At least that's what I use for compiling mesh shaders in my samples.

CMakeLists.txt
```diff
@@ -30,6 +31,34 @@ endif()
...
+    endif()
+    if(DEFINED Vulkan_dxc_EXECUTABLE)
+        message("Found DirectX Shader Compiler under ${Vulkan_dxc_EXECUTABLE}")
+    elseif()
+        message("Couldn't find DirectX Shader Compiler executable, make sure it is present in Vulkan SDK or add it manually via Vulkan_dxc_EXECUTABLE env variable")
+    endif()
+else()
+    message("VULKAN_SDK path not found, HLSL shaders won't be compiled")
```
SaschaWillems, line 52: Is it possible to manually specify a DXC path, or does it always fail if no SDK is installed? I often use a different DXC version that's newer than the SDK, so having a way to optionally manually specify the path would be great.

SaschaWillems (changes requested):
Thank you very much for this PR. Would love to see this merged as soon as possible :)

Aside from Andreas' comments I only had some minor remarks.

CMakeLists.txt
SaschaWillems: Another option could be checking path. I do that in my samples, and at least on windows, the Vulkan SDK bin folder is added to path upon installation, so the DXC binary always shows up via path.

CMakeLists.txt
jherico: the Win32 filesystem is case preserving, but not case sensitive.  I believe that `"$ENV{VULKAN_SDK}/bin"` should work for both, and even if that's not the case you could also just as easily include both `"$ENV{VULKAN_SDK}/bin"` and `"$ENV{VULKAN_SDK}/Bin"` as hints in a single `find_program` call.

jherico:
You might want to consider making this a PR against the existing branch by @tomadamatkinson in his PR #950 as that's where most of the work towards this end is happening right now.

SaschaWillems:
We discussed this internally, hence the separate PR. It's the way we agreed on.

CMakeLists.txt
jherico: My reading is that you could specify `Vulkan_dxc_EXECUTABLE` on the cmake configure command line and it would be honored.  A more generic `DXC_EXECUTABLE` might be a preferred command line option (to distinguish between that and the bundled Vulkan DXC binary), but that would only require slight modifications to the current flow.

Also, `find_program` does in fact check the system path, unless `NO_DEFAULT_PATH` or `NO_SYSTEM_ENVIRONMENT_PATH` is passed to `find_program` (or `CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH` is set to false).

bldsys/cmake/sample_helper.cmake
jherico: Note that this step is also converting from relative paths to absolute paths, which has to happen at some stage in the process.

asuessenbach: But that conversion is done in the loop right after this line.
So, essentially, instead of
```
    if (TARGET_SHADER_FILES_GLSL)    
        list(APPEND SHADER_FILES_GLSL ${TARGET_SHADER_FILES_GLSL})
        foreach(SHADER_FILE_GLSL ${SHADER_FILES_GLSL})
            list(APPEND SHADERS_GLSL "${PROJECT_SOURCE_DIR}/shaders/${SHADER_FILE_GLSL}")
        endforeach()        
    endif()
```
One could just write (slightly more efficient as well):
```
    foreach(SHADER_FILE_GLSL ${TARGET_SHADER_FILES_GLSL})
        list(APPEND SHADERS_GLSL "${PROJECT_SOURCE_DIR}/shaders/${SHADER_FILE_GLSL}")
    endforeach()        
```

CMakeLists.txt
Seweryn-Zielas-Mobica: Thanks for suggestion @jherico, I included both hints in `find_program` call.

bldsys/cmake/sample_helper.cmake
Seweryn-Zielas-Mobica: Thanks for code snippet @asuessenbach, I did as you suggested and it seems to work the same.

bldsys/cmake/sample_helper.cmake
Seweryn-Zielas-Mobica: Changed to `TARGET_SHADERS_HSLS ` as you suggested.

bldsys/cmake/sample_helper.cmake
Seweryn-Zielas-Mobica: I made a change a you suggested, tell me if now it is okay.

bldsys/cmake/sample_helper.cmake
Seweryn-Zielas-Mobica: It was mistake from my side, fixed.

CMakeLists.txt
Seweryn-Zielas-Mobica: Also form cmake version 3.25, `Vulkan_dxc_EXECUTABLE` is also added by `FindVulkan`.
https://cmake.org/cmake/help/latest/module/FindVulkan.html

SaschaWillems (approved):

asuessenbach (approved):

PR #696, opened 2023-05-10
iOS Simulator and iOS platform support

gpx1000:
## Description

Add support for iOS

Fixes #627 

## General Checklist:

Please ensure the following points are checked:

- [ ] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Code-Style)
- [ ] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#Copyright-Notice-and-License-Template)
- [ ] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
- [ ] I have commented any added functions (in line with Doxygen)
- [ ] I have commented any code that could be hard to understand
- [ ] My changes do not add any new compiler warnings
- [ ] My changes do not add any new validation layer errors or warnings
- [ ] I have used existing framework/helper functions where possible
- [ ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [ ] My changes do not add any regressions
- [ ] I have tested every sample to ensure everything runs correctly
- [ ] This PR describes the scope and expected impact of the changes I am making

## Sample Checklist

If your PR contains a new or modified sample, these further checks must be carried out *in addition* to the General Checklist:
- [ ] I have tested the sample on at least one compliant Vulkan implementation
- [ ] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/master/CONTRIBUTING.md#General-Requirements)
- [ ] I have stated on what implementation the sample has been tested so that others can test on different implementations and platforms
- [ ] Any dependent assets have been merged and published in downstream modules
- [ ] For new samples, I have added a paragraph with a summary to the appropriate chapter in the [samples readme](./../samples/README.md)

PR #994, opened 2024-03-20
Fix #791: Add support for Xcode

### LINKED ISSUE #791: Running samples on macOS doesn't work out of the box
Here is the log of running samples app right after building:
```
➜  mac git:(main) ./app/bin/Debug/arm64/vulkan_samples sample hpp_hello_triangle
[info] Logger initialized
UNASSIGNED-CreateInstance-status-message(INFO / SPEC): msgNum: -2016116905 - Validation Information: [ UNASSIGNED-CreateInstance-status-message ] Object 0: handle = 0x15a861400, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x87d47f57 | Khronos Validation Layer Active:
    Settings File: Found at /Users/vodobesku/.local/share/vulkan/settings.d/vk_layer_settings.txt specified by VkConfig application override.
    Current Enables: None.
    Current Disables: None.

    Objects: 1
        [0] 0x15a861400, type: 1, name: NULL
VUID-VkDeviceCreateInfo-pProperties-04451(ERROR / SPEC): msgNum: 976972960 - Validation Error: [ VUID-VkDeviceCreateInfo-pProperties-04451 ] Object 0: handle = 0x600002e20120, type = VK_OBJECT_TYPE_PHYSICAL_DEVICE; | MessageID = 0x3a3b6ca0 | vkCreateDevice: VK_KHR_portability_subset must be enabled because physical device VkPhysicalDevice 0x600002e20120[] supports it The Vulkan spec states: If the VK_KHR_portability_subset extension is included in pProperties of vkEnumerateDeviceExtensionProperties, ppEnabledExtensionNames must include "VK_KHR_portability_subset" (https://vulkan.lunarg.com/doc/view/1.3.261.0/mac/1.3-extensions/vkspec.html#VUID-VkDeviceCreateInfo-pProperties-04451)
    Objects: 1
        [0] 0x600002e20120, type: 2, name: NULL
[error] [samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp:43] 976972960 - VUID-VkDeviceCreateInfo-pProperties-04451: Validation Error: [ VUID-VkDeviceCreateInfo-pProperties-04451 ] Object 0: handle = 0x600002e20120, type = VK_OBJECT_TYPE_PHYSICAL_DEVICE; | MessageID = 0x3a3b6ca0 | vkCreateDevice: VK_KHR_portability_subset must be enabled because physical device VkPhysicalDevice 0x600002e20120[] supports it The Vulkan spec states: If the VK_KHR_portability_subset extension is included in pProperties of vkEnumerateDeviceExtensionProperties, ppEnabledExtensionNames must include "VK_KHR_portability_subset" (https://vulkan.lunarg.com/doc/view/1.3.261.0/mac/1.3-extensions/vkspec.html#VUID-VkDeviceCreateInfo-pProperties-04451)
[error] [framework/platform/platform.cpp:169] Error Message: Failed to open file: shaders/triangle.vert
[error] [framework/platform/platform.cpp:170] Failed when running application hpp_hello_triangle
Press return to continue
```

Also if cmake generator is set to Xcode (let's say one wants to run it from Xcode IDE, maybe capture Metal API frame, possibly debug shaders out of curiosity) then one will encounter that final linking of an app fails:

```
error: /Applications/Xcode_14_3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: can't open file: /Users/vodobesku/projects/Vulkan-Samples/build/mac/build/framework.build/Debug/Objects-normal/arm64/camera.o (No such file or directory)
error: /Applications/Xcode_14_3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: can't open file: /Users/vodobesku/projects/Vulkan-Samples/build/mac/build/framework.build/Debug/Objects-normal/arm64/image.o (No such file or directory)
error: /Applications/Xcode_14_3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: can't open file: /Users/vodobesku/projects/Vulkan-Samples/build/mac/build/framework.build/Debug/Objects-normal/arm64/sampler.o (No such file or directory)
Command Libtool failed with a nonzero exit code
```

jeroenbakker-atmind:
## Description

Xcode projects were not able to link. The reason is that libtool (linker) doesn't support to sharing same name in a single library. Framework uses 4 files with the same name. One in the `framework` folder and one in `framework/scene_graph/components`.

These files are:
 - `camera.cpp`
 - `image.cpp`
 - `sampler.cpp`
 - `hpp_image.cpp`
 
There are two ways to solve this, rename these files so all files in a library is unique or split the library into 2 so within a library the compile unit names are still unique. This PR splits framework into two separate library (`framework` and `framework_scene_graph`).

After this issue has been merged we can update the documentation how to generate the Xcode project. eg.

```
cmake -G Xcode -B build/mac-xcode
```

Fixes #791

- [ ] framework_scene_graph library is compiled using stdcpp14 on windows, other libraries use stdcpp17. we should keep them in sync but I haven been able to find the location where stdcpp17 is set.

## General Checklist:

- [ ] I have tested every sample to ensure everything runs correctly **I didn't have access to hardware capable of running all samples.**

   - [x] Windows
   - [x] Linux
   - [x] macOS
   - [x] Android

## Sample Checklist

- [NA] I have tested the sample on at least one compliant Vulkan implementation
- [NA] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements)
- [NA] I have stated on what implementation the sample has been tested so that others can test on different implementations and platforms
- [NA] Any dependent assets have been merged and published in downstream modules
- [NA] For new samples, I have added a paragraph with a summary to the appropriate chapter in the readme of the folder that the sample belongs to [e.g. api samples readme](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/samples/api/README.adoc)
- [NA] For new samples, I have added a tutorial README.md file to guide users through what they need to know to implement code using this feature. For example, see [conditional_rendering](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/extensions/conditional_rendering)
- [NA] For new samples, I have added a link to the [Antora navigation](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/antora/modules/ROOT/nav.adoc) so that the sample will be listed at the Vulkan documentation site


gpx1000:
I fixed this in a different way in the #696.  See if that version works for you.

I do like this as a solution as well.  The problem with my solution is the names can never conflict, the advantage of this solution is it doesn't matter it will work longer term.
Thanks for the submission!

framework/CMakeLists.txt
```diff
@@ -450,70 +462,79 @@ endif()
 add_library(${PROJECT_NAME} OBJECT ${PROJECT_FILES})
 set_target_properties(${PROJECT_NAME} PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
+set(PUBLIC_COMPILE_OPTIONS)
+set(PRIVATE_COMPILE_OPTIONS)
+
 # compiler flags based on compiler type
 if(NOT MSVC)
-    target_compile_options(${PROJECT_NAME} PUBLIC -fexceptions)
+    list(APPEND ${PUBLIC_COMPILE_OPTIONS} -fexceptions)
```
jherico: When using `list` the second argument should be the name of the variable, not the evaluated contents of the variable... like so...

```suggestion
    list(APPEND PUBLIC_COMPILE_OPTIONS -fexceptions)
```

This applies to everywhere below where you have `${PUBLIC_COMPILE_OPTIONS}` and the like.. these aren't going to do what you think they're going to do.

jherico (changes requested):

jherico:
I called out one instance.  You need to search your branch for every instance of the string `list(APPEND ${` and fix them.

jeroenbakker-atmind:
@jherico thx for pointing that out. I also detected a logical error where the definitions where used as compile options.
I will try to spin on a Windows machine and see how to solve the linker error.

jeroenbakker-atmind:
@gpx1000 I glanced over your solution (also pointed that one out in the PR description. I wasn't aware of that it was already being addressed. So we need to check which one would be best suited.

My approach is that developers should not become aware of the unique file names issue. In larger applications you typically use prefixes or make the file names unique (by adding the `_core` as you did). IMO that is less readable for developers as the content of the file doesn't reflect the filename itself. eg the class `vbk::core::Image` would be easier to find when the file is still called `Image.cpp/h`.

Anyhow as I am still new to the project I am open for both solutions.

jeroenbakker-atmind:
Just a note that I had to force push. I saw that one of my machines used a different email address, which made it not possible to accept the CLA. I rewrote the history so one email address was used (for my changes). The other changes (and authors) were kept intact.

gpx1000:
> @gpx1000 I glanced over your solution (also pointed that one out in the PR description. I wasn't aware of that it was already being addressed. So we need to check which one would be best suited.
> 
> My approach is that developers should not become aware of the unique file names issue. In larger applications you typically use prefixes or make the file names unique (by adding the `_core` as you did). IMO that is less readable for developers as the content of the file doesn't reflect the filename itself. eg the class `vbk::core::Image` would be easier to find when the file is still called `Image.cpp/h`.
> 
> Anyhow as I am still new to the project I am open for both solutions.

I like your solution more.  If we can get this merged ahead of my PR getting merged then I'll back out my changes.  Otherwise, I'll create a new PR which reverts the renaming fix.  Thanks for the submission.  We really appreciate any and all help and this is a good fix.

framework/CMakeLists.txt
```diff
@@ -450,70 +463,78 @@ endif()
...
-        target_compile_options(${PROJECT_NAME} PRIVATE -Werror)
+        list(APPEND PRIVATE_COMPILE_OPTIONS -Werror)
     elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-        target_compile_options(${PROJECT_NAME} PRIVATE /W3 /WX)
+        list(APPEND PRIVATE_COMPILE_OPTIONS /W3 /WX)
     endif()
 endif()
 
+target_compile_definitions(${PROJECT_NAME} PUBLIC ${PUBLIC_COMPILE_DEFINITIONS})
+target_compile_options(${PROJECT_NAME} PUBLIC ${PUBLIC_COMPILE_OPTIONS})
```
jherico: These are superfluous.  If you're setting public compile options and defines, they will automatically be applied to `framework`, because being public means they're transitive to all downstream consumers.

framework/CMakeLists.txt
```diff
@@ -402,6 +402,21 @@ source_group("scene_graph\\components\\" FILES ${SCENE_GRAPH_COMPONENT_FILES})
...
 source_group("stats\\" FILES ${STATS_FILES})
 
+# Adding scene graph files to a seperate library. Libtool doens't support having multiple files sharing the same name.
+set(PROJECT_SCENE_GRAPH_FILES
+    ${SCENE_GRAPH_FILES}
+    ${SCENE_GRAPH_COMPONENT_FILES}
+    ${SCENE_GRAPH_SCRIPTS_FILES}
+)
+
+add_library(framework_scene_graph ${PROJECT_SCENE_GRAPH_FILES})
```
jherico: Do not rely on the behavior of `add_library` to default to static.  Be explicit:

```suggestion
add_library(framework_scene_graph STATIC ${PROJECT_SCENE_GRAPH_FILES})
```

If someone has `build_shared_libs` turned on when doing the CMake configure, this will fail to build, because the `framework_scene_graph` and `framework` have bidirectional dependencies, so building the former as a shared library will fail at link time.

jeroenbakker-atmind:
Solved code review feedback and fixed building on Linux

jeroenbakker-atmind:
Currently this PR would be a cleanup and revert parts of #696. Next steps:
- merge main back in the branch
- revert the renaming of the files in framework/core

framework/CMakeLists.txt
```diff
@@ -278,7 +278,6 @@ set(CORE_FILES
     core/physical_device.cpp
     core/device.cpp
     core/debug.cpp
-    core/image_core.cpp
```
jeroenbakker-atmind, line 280: Note: core/image_core.cpp was twice in the list.

jeroenbakker-atmind:
This PR is rebased with latest main. The renaming of CPP files done by #696 has been reverted. I also noticed a file to be listed twice. Any feedback is welcome!

jeroenbakker-atmind:
The issue is that VulkanSamples has a minimum requirement of stdcpp14, but recently added changes uses stdcpp17 syntax. I think it is better to stick to the syntax of stdcpp14.

Due to my lack of experience in CMake I couldn't detect the location inside the build script that sets the stdcpp to 17.
- In main the files are compiled with stdcpp17. This branch splits the framework library and `framework_scene_graph` is build using the stdcpp14. In clang this adds warnings, but on windows these warnings are errors.
  - `framework_scene_graph.dir` c++ calls contains `-std=gnu++14`. 
  - `framework.dir` c++ calls contains `-std=gnu++17`
- `global_options.cmake` sets the `CMAKE_CXX_STANDARD` to 14
- `project` macro doesn't change CXX flags, CXX flags doesn't contain reference to gnu++17

gpx1000 (dismissed):

gary-sweet (dismissed):

jeroenbakker-atmind:
Nothing changed, but github detected a possible merge conflict, which was automatically resolved when merging manually.

Seems like more CPP17 code style was added. I will fix them as well ...

jeroenbakker-atmind:
This task is currently
Blocked by: #1017

PR #1013, opened 2024-04-03
Refactor getting/adding extension features to create a VkDevice with.

### LINKED ISSUE #938: Requesting extension features also enables all available features

asuessenbach:
## Description

As described in #938, requesting and enabling extension features is currently incorrect. With this PR, this is done similar to how it's done for the simple PhysicalDeviceFeatures: one function to get the supported bits (`get_extension_features`) and one to actually set them into the structure chain (`add_extension_features`).
The perfect handling would look like this:
```
if (gpu.get_extension_features<VkExtensionFeaturesStruct>(VK_STRUCTURE_TYPE_EXTENSION_FEATURES_STRUCT).featureBit)
{
	gpu.add_extension_features<VkExtensionFeaturesStruct>(VK_STRUCTURE_TYPE_EXTENSION_FEATURES_STRUCT).featureBit = VK_TRUE;
}
else
{
	// some fallback or error handling!
}
```
To ease usage, `[HPP]PhysicalDevice` got two new template functions: `request_optional_feature` and `request_required_feature`, which log a message or throw an exception, respectively, if the requested feature is not supported.
Requesting a required feature would now look like
```
gpu.request_required_feature<VkExtensionFeaturesStruct>(VK_STRUCTURE_TYPE_EXTENSION_FEATURES_STRUCT,
                                                        &VkExtensionFeaturesStruct::featureBit,
                                                        "VkExtensionFeaturesStruct",
                                                        "featureBit");
```
Where the last two arguments to that functions are just for error reporting and maybe could omitted.

To ease usage even more, some macros have been added, so that requesting a required feature would now look like

`REQUEST_REQUIRED_FEATURE(gpu, VkExtensionFeaturesStruct, VK_STRUCTURE_TYPE_EXTENSION_FEATURES_STRUCT, featureBit);
`

Fixes #938

## General Checklist:

## Sample Checklist

### DIFF
--- a/framework/core/device.cpp
+++ b/framework/core/device.cpp
@@ -89,11 +89,17 @@ Device::Device(PhysicalDevice                        &gpu,
 	if (is_extension_supported("VK_KHR_performance_query") &&
 	    is_extension_supported("VK_EXT_host_query_reset"))
 	{
-		auto perf_counter_features     = gpu.request_extension_features<VkPhysicalDevicePerformanceQueryFeaturesKHR>(VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PERFORMANCE_QUERY_FEATURES_KHR);
-		auto host_query_reset_features = gpu.request_extension_features<VkPhysicalDeviceHostQueryResetFeatures>(VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_HOST_QUERY_RESET_FEATURES);
+		auto perf_counter_features =
+		    gpu.get_extension_features<VkPhysicalDevicePerformanceQueryFeaturesKHR>(VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PERFORMANCE_QUERY_FEATURES_KHR);
+		auto host_query_reset_features =
+		    gpu.get_extension_features<VkPhysicalDeviceHostQueryResetFeatures>(VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_HOST_QUERY_RESET_FEATURES);
 
 		if (perf_counter_features.performanceCounterQueryPools && host_query_reset_features.hostQueryReset)
 		{
+			gpu.add_extension_features<VkPhysicalDevicePerformanceQueryFeaturesKHR>(VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PERFORMANCE_QUERY_FEATURES_KHR)
+			    .performanceCounterQueryPools = VK_TRUE;
+			gpu.add_extension_features<VkPhysicalDeviceHostQueryResetFeatures>(VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_HOST_QUERY_RESET_FEATURES).hostQueryReset =
+			    VK_TRUE;
 			enabled_extensions.push_back("VK_KHR_performance_query");
 			enabled_extensions.push_back("VK_EXT_host_query_reset");
 			LOGI("Performance query enabled");
--- a/framework/core/hpp_device.cpp
+++ b/framework/core/hpp_device.cpp
@@ -89,11 +89,13 @@ HPPDevice::HPPDevice(vkb::core::HPPPhysicalDevice               &gpu,
 	// live in the same command buffer as beginQuery
 	if (is_extension_supported(VK_KHR_PERFORMANCE_QUERY_EXTENSION_NAME) && is_extension_supported(VK_EXT_HOST_QUERY_RESET_EXTENSION_NAME))
 	{
-		auto perf_counter_features     = gpu.request_extension_features<vk::PhysicalDevicePerformanceQueryFeaturesKHR>();
-		auto host_query_reset_features = gpu.request_extension_features<vk::PhysicalDeviceHostQueryResetFeatures>();
+		auto perf_counter_features     = gpu.get_extension_features<vk::PhysicalDevicePerformanceQueryFeaturesKHR>();
+		auto host_query_reset_features = gpu.get_extension_features<vk::PhysicalDeviceHostQueryResetFeatures>();
 
 		if (perf_counter_features.performanceCounterQueryPools && host_query_reset_features.hostQueryReset)
 		{
+			gpu.add_extension_features<vk::PhysicalDevicePerformanceQueryFeaturesKHR>().performanceCounterQueryPools = true;
+			gpu.add_extension_features<vk::PhysicalDeviceHostQueryResetFeatures>().hostQueryReset                    = true;
 			enabled_extensions.push_back(VK_KHR_PERFORMANCE_QUERY_EXTENSION_NAME);
 			enabled_extensions.push_back(VK_EXT_HOST_QUERY_RESET_EXTENSION_NAME);
 			LOGI("Performance query enabled");
--- a/framework/core/hpp_physical_device.h
+++ b/framework/core/hpp_physical_device.h
@@ -91,52 +91,62 @@ class HPPPhysicalDevice
 	vk::PhysicalDeviceFeatures &get_mutable_requested_features();
 
 	/**
-	 * @brief Requests a third party extension to be used by the framework
+	 * @brief Get an extension features struct
 	 *
-	 *        To have the features enabled, this function must be called before the logical device
-	 *        is created. To do this request sample specific features inside
-	 *        VulkanSample::request_gpu_features(vkb::HPPPhysicalDevice &gpu).
-	 *
-	 *        If the feature extension requires you to ask for certain features to be enabled, you can
-	 *        modify the struct returned by this function, it will propagate the changes to the logical
-	 *        device.
+	 *        Gets the actual extension features struct with the supported flags set.
+	 *        The flags you're interested in can be set in a corresponding struct in the structure chain
+	 *        by calling PhysicalDevice::add_extension_features()
 	 * @returns The extension feature struct
 	 */
 	template <typename HPPStructureType>
-	HPPStructureType &request_extension_features()
+	HPPStructureType get_extension_features()
 	{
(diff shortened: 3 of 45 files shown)


gary-sweet (changes requested):
I don't think we can just assert in all these cases can we?
It seems to prevent batch mode from running as soon as it finds an unsupported case.

asuessenbach:
> I don't think we can just assert in all these cases can we?

asuessenbach:
> I don't think we can just assert in all these cases can we?

You're kind of right. As I stated in the Description above, that should be replaced by some meaningful error or fall-back handling.
What would you suggest to do instead until then?

gary-sweet:
> What would you suggest to do instead until then?

Well throwing a `std::runtime_error` would be better, as it doesn't terminate batch mode, just the current sample I guess.

I'm also not really sure what was wrong with just having a single call to request_extension_features?  I can't see why unrolling that code everywhere is better. Maybe I'm just missing the issue that's being solved here.

asuessenbach:
> I'm also not really sure what was wrong with just having a single call to request_extension_features?

The problem, as described in #938, was, that currently you get the supported extension features bits, possibly set some of them to true, no matter if they are supported at all, and store that structure in the structure chain for device creation.
That is, there are two issues:
- All supported flags are requested, no matter if you actually need/want the support of all those flags.
- Potentially support of a not supported feature is requested.

With this PR, you just set the flags you're interested in and when they're supported at all.

asuessenbach:
> Well throwing a `std::runtime_error` would be better, as it doesn't terminate batch mode, just the current sample I guess.

Sounds good. I can change that accordingly.

gary-sweet:
I think you'll need to rebase this over https://github.com/KhronosGroup/Vulkan-Samples/pull/1031. A lot of the samples don't run without that, so testing this isn't really possible without it.

SaschaWillems:
Tbh. I'm kinda torn over the syntax that now needs to be used. Will that actually work with e.g. code completion/intellisense? Or will it fail at compile time when I e.g. request a member of a feature that isn't present or that I did misspell?

I'm referring to this:

```
REQUEST_REQUIRED_FEATURE(gpu,
VkPhysicalDeviceGraphicsPipelineLibraryFeaturesEXT,
VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_GRAPHICS_PIPELINE_LIBRARY_FEATURES_EXT,
graphicsPipelineLibrary);
```

The old syntax has you access that member (graphicsPipelineLibary) directly, which works fine with intellisense/code compleation and is easier to understand/follow:

```
auto &requested_vertex_input_features      = gpu.request_extension_features<VkPhysicalDeviceMeshShaderFeaturesEXT>(VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MESH_SHADER_FEATURES_EXT);
requested_vertex_input_features.meshShader = VK_TRUE;
```
(though I don't really like that syntax either)

I fear that this makes working on samples harder, hurts readability and foremost makes it harder to follow our samples.

Is there a different way to handle this that looks more similar to the old syntax?

asuessenbach:
> Will that actually work with e.g. code completion/intellisense?

As it is with all macros, you probably don't have any support by code completion or intellisense when you use `[HPP_]REQUEST_[OPTIONAL|REQUIRED]_FEATURE`. Any error will fire up at compile time.
When you use the helper functions `gpu.request_[optional|required]_feature` or the actual worker functions `gpu.[add|get]_extension_features`, you of course have all the help available by code completion and intellisense.

> The old syntax has you access that member (graphicsPipelineLibary) directly, which works fine with intellisense/code compleation and is easier to understand/follow:

But as described in #938 it's just wrong!

> Is there a different way to handle this that looks more similar to the old syntax?

I don't know of any.

Actually, I provided three different levels of complexity to handle that correctly. If you have an idea for an additional approach, I'd be happy to add that as well.

gary-sweet (dismissed):

SaschaWillems:
As discussed on the call, here is the way I do it in my samples:

- I declare the extension struct(s) in the sample
- I setup the whole extension struct(s) in the sample constructor
- I add the struct(s) to a pNext pointer of my device / instance

An example:

```cpp
class VulkanExample : public VulkanExampleBase
{
    VkPhysicalDeviceTimelineSemaphoreFeaturesKHR enabledTimelineSemaphoreFeaturesKHR{};

    VulkanExample() : VulkanExampleBase()
    {
        ...

        enabledTimelineSemaphoreFeaturesKHR.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_TIMELINE_SEMAPHORE_FEATURES_KHR;
        enabledTimelineSemaphoreFeaturesKHR.timelineSemaphore = VK_TRUE;

        pNextDevice = &enabledTimelineSemaphoreFeaturesKHR;
    }    

}
```

It might not be the best approach from a C++ language point of view (esp. how pNextDevice is implemented), but it makes it very easy for people to follow my samples and you can copy that 1:1 into your own application without having to follow any templates or macros.

It would probably be a more substantial change, so before we take this further we should agree on what way we want to follow.

asuessenbach:
@SaschaWillems I assume, that `pNextDevice` is the `pNext` of some `VkPhysicalDeviceFeatures2`, that you hold somewhere. And you're using a pointer to that `VkPhysicalDeviceFeatures2` to create a `VkDevice` using `vkCreateDevice()`.
That is, it seems that setting the requested features and actually using that is also distributed to (at least) two separate code sequences.
But, what's more important, I don't see that you first ask if the requested feature actually is supported. You would need a second structure chain to first ask for support using `vkGetPhysicalDeviceFeatures2()`.

From a sample point of view, I think, it's important that the reader of the sample can easily see what features are requested. But I would not consider the details on how that is handled to be important for a sample. Besides for a sample that actually demonstrates how feature requesting is done, of course.

SaschaWillems:
> But, what's more important, I don't see that you first ask if the requested feature actually is supported. You would need a second structure chain to first ask for support using vkGetPhysicalDeviceFeatures2().

I actually don't do that. If one requests a feature that isn't supported, instance or device creation will fail with an error code hinting at a feature not supported. I want to deliberately keep samples as simple as possible.

asuessenbach:
> If one requests a feature that isn't supported, instance or device creation will fail with an error code hinting at a feature not supported.

Interesting approach! That is, you would get some `VK_ERROR_FEATURE_NOT_PRESENT` on `vkCreateDevice` and bail out at that point. But you can't tell then, what feature is missing. And how would you handle some optionally used feature?
I would consider to first get the supported features and potentially react accordingly on their absence the "cleaner" and also more descriptive way to handle that.
And I don't think, that code is that much harder to read and understand.

SRSaunders:
This PR also addresses issue #1144.  It fixes the crashes I am seeing due to missing feature support on my AMD 6600XT GPU during `vkCreateDevice()` initialization on Windows 10 with SDK 1.3.290.  Thanks @asuessenbach for this work.

However, I would like to request a couple of small things:

1. Please update the `mutableComparisonSamplers ` portability feature activation in samples _async_compute_ and _multithreading_render_passes_.
2. If possible, please add line type optionality to the _dynamic_line_rasterization_ sample.  Some GPUs support a subset of the line types and it would be good to have partial support vs. all-or-nothing.  The implementation could be something like what I tried in the now-retracted PR #1145.

asuessenbach:
@SRSaunders I adjusted the two samples _async_compute_ and _multithreading_render_passes_. But the changes in _dynamic_line_rasterization_ would actually change that sample, which I'ld like to avoid with this already pretty big and long standing PR. Would be nice if you could change that after this PR has been merged.

SRSaunders:
> I adjusted the two samples async_compute and multithreading_render_passes

Thanks.

> But the changes in dynamic_line_rasterization would actually change that sample, which I'ld like to avoid with this already pretty big and long standing PR. Would be nice if you could change that after this PR has been merged.

No problem.  I'd be happy to come back to that once this PR is merged.

SaschaWillems (dismissed):
LGTM and works for me.

Though I can't do a full batch run. With the latest SDK I get a GPU crash in the ray tracing extended sample. Doesn't look like it's related to this PR though.

gary-sweet (dismissed):

SRSaunders:
I don't want to interfere with the approvals already granted, but two more samples have now been added using the old `gpu.request_extension_features<>()` method:  _dynamic_rendering_local_read_ and _host_image_copy_.  Applying this PR without fixing these will result in project build failures.  I ran into this today when trying to stack up a few pending changes for testing.

Is there any way to speed up approvals for these kinds of framework updates, or at least put in safeguards to restrict certain kinds of changes (e.g. new samples) until they are merged?  It appears as an outside user that there is a process weakness here.  This one was obvious, but I could see more subtle issues arising in the future.

SaschaWillems:
We usually meet twice a month to discuss PRs. That's the process. We'll discus this on our next call and make sure we don't break anything.

SaschaWillems:
@asuessenbach : Can you merge main into this so we can see if there are any CI failures as mentioned?

SaschaWillems (approved):
Thank you very much :)

gary-sweet (approved):

PR #1027, opened 2024-04-16
Make sampler creation optional when loading textures

### LINKED ISSUE #866: Make sampler creation in texture loader optional
As per the discussion in #PR850 we should adjust the texture loader so that creating a sampler is optional. That way a sample that uses a different sampler doesn't have to delete the one automatically created at texture loading.

SaschaWillems:
## Description

This PR adds a new argument to the texture loading functions of the api samples base class that allows to skip default sampler creation. With this change samples that use custom samplers for certain textures (shader object, terrain tessellation) no longer need to delete the sampler that would've been implicitly created when loading the texture. This results in cleaner, easier to understand code.

Fixes #866

## General Checklist:

### DIFF
--- a/framework/api_vulkan_sample.cpp
+++ b/framework/api_vulkan_sample.cpp
@@ -976,7 +976,7 @@ VkDescriptorImageInfo ApiVulkanSample::create_descriptor(Texture &texture, VkDes
 	return descriptor;
 }
 
-Texture ApiVulkanSample::load_texture(const std::string &file, vkb::sg::Image::ContentType content_type, bool create_sampler = true)
+Texture ApiVulkanSample::load_texture(const std::string &file, vkb::sg::Image::ContentType content_type, bool create_sampler)
 {
 	Texture texture{};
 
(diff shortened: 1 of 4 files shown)


gary-sweet (approved):
I can't actually run either of the two samples affected, but the change looks ok.

asuessenbach (commented):
You probably wanted to also call `load_texture[_array]` for `terrain_array_textures` and `heightmap_texture` in the `terrain_tessellation` sample with the non-default value of `false` for the new last argument?

Besides that, as the only difference between the default sampler and the locally generated samplers is the `addressMode`, you might as well add that as an optional argument to `load_texture[_array]`, similar to how it's done in the hpp-version of those functions?
Or otherwise change the hpp-version of those functions accordingly, to make them logically equivalent, again?

SaschaWillems:
Will have to check. I tend to forget that I need to change stuff like this in two places. Maybe I'll defer this to after we have unified the framework :/

gpx1000 (approved):

PR #1035, opened 2024-04-22
Unify vkb::Subpass and vkb::rendering::HPPSubpass into vkb::rendering::Subpass<bindingType>

asuessenbach:
## Description

## General Checklist:

## Sample Checklist


SaschaWillems:
This does work fine for me (tested a few samples with the batch run), but I',m kinda unsure about the actual interface.

Once we have unified the repo will we have to add the binding type to each and every class from the framework? If so, I fear that this will make the code hard to read.

e.g. comparing this:
```cpp
std::vector<std::unique_ptr<vkb::Subpass>> subpasses{};
```
with:
```cpp
std::vector<std::unique_ptr<vkb::rendering::Subpass<vkb::BindingType::C>>> subpasses{};
```

Imo readability suffers a lot. Is it possible to somehow omit the binding type unless one explicitly wants to use the CPP variants?

asuessenbach:
I agree, the resulting types are somewhat longer than before.
- I moved the class Subpass from namespace `vkb` to `vkb::rendering`, to make it clear where it resides. If it's consensus that that is a bad move, it would be easy to revert that.
- In order to have the very same class work for both, the C- and the C++-bindings, that BindingType template parameter is added. It is needed. One could, though, add some type aliases like
```
using SubpassC   = Subpass<vkb::BindingType::C>;
using SubpassCpp = Subpass<vkb::BindingType::Cpp>;
```
which would reduce the aforementioned complexity a bit:
`std::vector<std::unique_ptr<vkb::rendering::SubpassC>> subpasses{};`
If it's consensus, I could easily add something like that.

gpx1000 (dismissed):

framework/core/command_buffer.cpp
```diff
@@ -131,7 +131,11 @@ void CommandBuffer::flush(VkPipelineBindPoint pipeline_bind_point)
 	flush_descriptor_state(pipeline_bind_point);
 }
 
-void CommandBuffer::begin_render_pass(const RenderTarget &render_target, const std::vector<LoadStoreInfo> &load_store_infos, const std::vector<VkClearValue> &clear_values, const std::vector<std::unique_ptr<Subpass>> &subpasses, VkSubpassContents contents)
+void CommandBuffer::begin_render_pass(const RenderTarget                                           &render_target,
+                                      const std::vector<LoadStoreInfo>                             &load_store_infos,
+                                      const std::vector<VkClearValue>                              &clear_values,
+                                      const std::vector<std::unique_ptr<vkb::rendering::SubpassC>> &subpasses,
```
gpx1000, line 137: FWIW, some compilers will have issues with >> next to each other.  I don't know if it's been fixed in all compilers or not as I just got into a habit of putting a space between the two.  This compiles and works in Linux; and CI passes so this is probably a non-issue.

framework/rendering/subpass.h
```diff
@@ -40,217 +50,442 @@ struct alignas(16) Light
...
+	using RenderContextType = typename std::conditional<bindingType == vkb::BindingType::Cpp, vkb::rendering::HPPRenderContext, vkb::RenderContext>::type;
+	using RenderTargetType  = typename std::conditional<bindingType == vkb::BindingType::Cpp, vkb::rendering::HPPRenderTarget, vkb::RenderTarget>::type;
 
-	virtual ~Subpass() = default;
+  public:
+	Subpass(RenderContextType &render_context, ShaderSource &&vertex_shader, ShaderSource &&fragment_shader);
 
+	Subpass(const Subpass &)            = delete;
+	Subpass(Subpass &&)                 = default;
+	virtual ~Subpass()                  = default;
```
gpx1000, line 101: Shouldn't this be override/final instead of virtual here?

framework/rendering/subpass.h
```diff
@@ -40,217 +50,442 @@ struct alignas(16) Light
...
+inline const ShaderSource &Subpass<bindingType>::get_fragment_shader() const
+{
+	return fragment_shader;
+}
 
-	DepthStencilState depth_stencil_state{};
+template <vkb::BindingType bindingType>
+inline void Subpass<bindingType>::set_color_resolve_attachments(std::vector<uint32_t> color_resolve)
+{
+	color_resolve_attachments = color_resolve;
```
gpx1000, line 425: copy a vector on each call.  maybe move is better?

framework/rendering/subpass.h
```diff
@@ -40,217 +50,442 @@ struct alignas(16) Light
...
+		depth_stencil_resolve_mode = static_cast<vk::ResolveModeFlagBits>(mode);
+	}
+}
 
-	/// Default to no color resolve attachments
-	std::vector<uint32_t> color_resolve_attachments = {};
+template <vkb::BindingType bindingType>
+inline void Subpass<bindingType>::set_input_attachments(std::vector<uint32_t> input)
+{
+	input_attachments = input;
```
gpx1000, line 462: copy vector on each call.  check if move is correct.

framework/rendering/subpass.h
```diff
@@ -40,217 +50,442 @@ struct alignas(16) Light
...
+}
 
-	/// Default to no depth stencil resolve attachment
-	uint32_t depth_stencil_resolve_attachment{VK_ATTACHMENT_UNUSED};
-};
+template <vkb::BindingType bindingType>
+inline void Subpass<bindingType>::set_output_attachments(std::vector<uint32_t> output)
+{
+	output_attachments = output;
```
gpx1000, line 468: copy vector on each call.  check if move is correct.

framework/core/command_buffer.cpp
```diff
@@ -277,7 +281,7 @@ void CommandBuffer::bind_index_buffer(const core::Buffer &buffer, VkDeviceSize o
 	vkCmdBindIndexBuffer(get_handle(), buffer.get_handle(), offset, index_type);
 }
 
-void CommandBuffer::bind_lighting(LightingState &lighting_state, uint32_t set, uint32_t binding)
+void CommandBuffer::bind_lighting(vkb::rendering::LightingStateC &lighting_state, uint32_t set, uint32_t binding)
```
gpx1000, line 284: we're already in the vkb namespace.  It's safe to use rendering::LightingStateC

framework/core/command_buffer.cpp
asuessenbach: This was resolved with C++11!
I can't imagine that there's a C++11 compiler out in the wild, that does not support those double closing angle brackets.

framework/rendering/subpass.h
asuessenbach: Good catch!
Will make it a `std::vector<uint32_t> const &`

framework/rendering/subpass.h
asuessenbach: As above.

framework/rendering/subpass.h
asuessenbach: As above.

framework/core/command_buffer.cpp
asuessenbach: I tend to use the complete namespace. I think, it's easier to mentally locate the used entity.
If you want, I can remove that `vbk::` here, though.

gpx1000 (approved):

gary-sweet (approved):

SaschaWillems (approved):
Compiles and runs fine for me. Code changes look fine to me 👍🏻

marty-johnson59:
Merging with 3 approvals.  Thanks!

PR #1051, opened 2024-05-19
Fix and rework Android error notifications

### LINKED ISSUE #1050: Missing error handling for unsupported samples on Android
Afair we had proper error handling for samples with e.t. unsupported extensions on Android. Running such a sample would show an error message and gracefully returned to the samples list. With a current build, this is no longer the case. Instead unsupported samples just sit there with a blank screen.

Update: It looks like we miss a permission required with Android 13 and newer. The samples do have code to request permissions, but that's unused and non-working :(

SaschaWillems:
## Description

This PR completely overhauls the way our framework displays errors on Android. Initially the framework used a notification that would only display "something has crashed" with a link to the log file but no reason as to why a sample crashed or wasn't able to run. Displaying such notifications isn't a great way to display errors and also requires permissions starting with Android 13. Since we never requested those permissions, those messages weren't displayed and instead of gracefully closing, samples that don't work (e.g. because the device lacks a requested extension) would just result in a blank screen.

This PR replaces those notifications with a proper alert dialog, which is a better and cleaner way to display errors. The PR also adds a last error message member to the platform that is then displayed in this alert:

Note that error messages aren't optimal, but the way the framework currently handles errors is pretty convoluted, so this is the best I could do for now.

Fixes #1050

## General Checklist:


tomadamatkinson (dismissed):
Cant test on android but LGTM. This should provide better errors than what was there before!

JoseEmilio-ARM (approved):
Works well, thanks a lot! Noticed that for the same error (extension not present), after hitting 'close', some samples return to the launcher, while others crash the application.

JoseEmilio-ARM (dismissed):

gpx1000 (dismissed):

marty-johnson59:
Merging now that conflicts are resolved (per discussion on Monday's call)

PR #1072, opened 2024-06-15
Add availability checks for debug report extension usage

kocdemir:
## Description

Adding the missing condition checks before using the structures and functions from the VK_EXT_debug_report extension in the samples.

Normal functionality should not change when the extension is supported, tested on a device where the extension is missing.

### DIFF
--- a/samples/api/hello_triangle/hello_triangle.cpp
+++ b/samples/api/hello_triangle/hello_triangle.cpp
@@ -162,10 +162,10 @@ void HelloTriangle::init_instance(Context                         &context,
 {
 	LOGI("Initializing vulkan instance.");
 
-    if (volkInitialize())
-    {
-        throw std::runtime_error("Failed to initialize volk.");
-    }
+	if (volkInitialize())
+	{
+		throw std::runtime_error("Failed to initialize volk.");
+	}
 
 	uint32_t instance_extension_count;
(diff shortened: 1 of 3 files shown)


SaschaWillems:
Thank you very much for your contribution. Note that CI fails at the clang format step, which means you need to properly format your changes with clang-format, see https://github.com/KhronosGroup/Vulkan-Samples/actions/runs/9530177624/job/26297193969?pr=1072

kocdemir:
Thanks, I've updated the file now.

samples/api/hello_triangle/hello_triangle.cpp
```diff
@@ -176,7 +176,23 @@ void HelloTriangle::init_instance(Context                         &context,
...
+	bool has_debug_report = false;
+	for (const auto &ext : instance_extensions)
+	{
+		if (strcmp(ext.extensionName, VK_EXT_DEBUG_REPORT_EXTENSION_NAME) == 0)
+		{
+			has_debug_report = true;
+			break;
+		}
+	}
+	if (has_debug_report)
```
asuessenbach, line 188: Maybe you can replace this loop with something like
```
bool has_debug_report = std::any_of(instance_extensions.begin(),
                                    instance_extensions.end(),
                                    [](VkExtensionProperties const &ep) { return strcmp(ep.extensionName, VK_EXT_DEBUG_REPORT_EXTENSION_NAME) == 0; });
```

gary-sweet (approved):

Issue #1083, opened 2024-07-02
VK_ENABLE_BETA_EXTENSIONS and VK_KHR_portability_subset not properly enabled for Apple platforms


SRSaunders:
According to the Vulkan spec, the VK_KHR_portability_subset extension (different from the enumeration extension) must be enabled on implementations that support it, otherwise validation errors occur.

In addition, this is required for access to portability features that must be turned on for certain samples in the set (e.g. ones that require `mutableComparisonSamplers`).

I will submit a PR for this.

Issue #1082, opened 2024-07-02
shader_debugprintf and descriptor_indexing examples not working on macOS


SRSaunders:
Two issues with these two examples on macOS:

1. **shader_debugprintf** unnecessarily uses VK_API_VERSION_1_1 which for unknown reasons causes a slowdown, and does not enable `VKB_ENABLE_PORTABILITY` which prevents the sample from running.
2. **descriptor_indexing** does not enable Metal Argument Buffers which are required for this example.  In addition, there is currently another issue with variable descriptor counts on MoltenVK, which needs to be worked around until such time as MVK corrects the issue.

Will submit a PR for this.  As part of this PR I will submit a framework enhacement that enables access to the relatively new Vulkan layer settings API (for enabling Vulkan Validation layer features, and MoltenVK's layer features).

Issue #1081, opened 2024-07-02
CMakeLists.txt logic for finding dxc has deficiencies


SRSaunders:
Two issues found with the CMakeLists logic for finding dxc:

1. `project(vulkan_samples)` is called after `find_package(Vulkan)` which is incorrect.  This results in default platform search paths not being set up properly for `find_package()` and **dxc** is not found unless `VULKAN_SDK` is defined ahead of time in the shell.  This is awkward and unnecessary on Windows, and does not support system-wide installations on macOS and linux.
2. The test for `Vulkan_dxc_EXECUTABLE` should be `if(Vulkan_dxc_EXECUTABLE)` and not `if(DEFINED Vulkan_dxc_EXECUTABLE)`.  The latter test erroneously passes even if `Vulkan_dxc_EXECUTABLE-NOTFOUND` is the result from `find_package()` or `find_program()`.

I will submit a fix that corrects these as part of larger PR.

Issue #1080, opened 2024-07-01
Failures in batch mode cause early exit on macOS/iOS, error recovery not resilient


SRSaunders:
Running batch mode on macOS or iOS where the portability implementation is not complete results in early failures (e.g. when mvk throws errors due to SPIR-V to MSL conversion limitations, Vulkan feature limitations, etc).  This makes it difficult to run through the complete list of samples without manual intervention (e.g. sample "xxx") or adding custom code to skip the known failing samples in batch mode which is not ideal.  Instead, the error recovery framework should be more resilient and detect these issues, recover, and continue through the complete list of samples until the end.

I have prepared a pending PR with several changes that handles this problem in a general way, as well as solves several other issues observed on macOS.

PR #1084, opened 2024-07-02
Add Layer Settings API to framework, improve batch mode error recovery, fix macOS issues

SRSaunders:
## Description

This PR provides solutions to #1080, #1081, #1082, and #1083.

1. Fixes defects in CMakeLists.txt find dxc logic.  Supports system-level Vulkan SDK installations that include dxc.
2. Adds layer settings capability to framework with new `add_layer_setting(VkLayerSettingEXT layerSetting)`, and `get_layer_settings()` API calls in **vulkan_sample.h**.  Modified `Instance()` and `HPPInstance()` parameter lists to include `required_layer_settings` so they can be activated during instance creation.  Protected all changes with `#if defined(VK_EXT_layer_settings)` for backwards compatibility with Vulkan SDKs earlier than 1.3.272 where the extension may not be defined.
3. Modified **shader_debugprintf** sample to override `get_validation_layers()` and use layer settings to activate the Validation Layer's `VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT` feature using the standard framework.  The sample's custom `create_instance()` is now only used if `VK_EXT_layer_settings` is not available at runtime. 
4. Modified **descriptor_indexing** sample to use the layer settings API on macOS to activate MoltenVK's  `MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS` feature which is necessary to increase the available descriptors for the sample.  Also worked around a macOS/MoltenVK limitation in the variable descriptor count feature.  This may be fixed in future versions of MoltenVK, but the workaround will function for past, current and future MoltenVK versions. 
5. Enabled `VK_ENABLE_BETA_EXTENSIONS` on Apple platforms for access to the `VK_KHR_portability_subset` extension and features.  Activating `VK_KHR_portability_subset` is required on all portability implementations where the extension is defined and available.  Also activated the `mutableComparisonSamplers` portability feature for the **async_compute** and **multithreading_render_pass** samples.
6. Improved batch mode error recovery by: a) using `throw std::runtime_error()` vs. `ExitCode::FatalError` in `main_loop()` and `main_loop_frame()`, b) modifying `VK_CHECK` to use `throw std::runtime_error()` to gracefully recover from more subtle issues such as pipeline creation problems (e.g. SPIR-V to Metal cross-compilation errors due to feature limitations) as well as queue submission errors, c) resetting the shader compiler's target environment settings prior to each sample to avoid stickiness from earlier samples in batch mode, d) allowing `VK_SUBOPTIMAL_KHR` which avoids early termination of some samples when exiting while in batch mode, and e) fixing specific issues in **async_compute**, **texture_mipmap_generation**, and **fragment_shader_barycentric** samples to permit graceful failure and continuation while in batch mode.
7. Changed packaging to build a command line macOS executable (not an app bundle) to match the existing online build/run instructions on the web site (i.e. use `./build/mac/app/bin/Release/<x86_64|arm64>/vulkan_samples` vs. `./build/mac/app/bin/Release/<x86_64|arm64>/vulkan_samples.app/Contents/MacOS/vulkan_samples`).  App bundles are needed for iOS, but get in the way on macOS for command line applications like vulkan_samples.
8. Added .DS_Store to the .gitignore file for git cleanliness on Apple.

I have other improvements pending for Xcode project generation for macOS/iOS, running on iOS and the iOS simulator (x86_64 and arm64 hosts), and standalone MoltenVK support (i.e. without the vulkan loader, useful when working on the MoltenVK dev stream), but will submit those in a separate PR.

## General Checklist:

I have tested using batch runs on Windows and macOS.  ~~Depending on CI for now on linux - will test later~~ Batch mode testing on Manjaro now complete.  Cannot test on Android - don't have the environment.  I have also tested on iOS but have additional changes to add in a separate PR.

 - [X] I did a full batch run using the `batch` command line argument to make sure all samples still work properly


CMakeLists.txt
```diff
@@ -46,20 +48,19 @@ if(VULKAN_FOUND)
...
-        message("Couldn't find DirectX Shader Compiler executable, make sure it is present in Vulkan SDK or add it manually via Vulkan_dxc_EXECUTABLE env variable")
+    if(Vulkan_dxc_EXECUTABLE)
+        message(STATUS "Found DirectX Shader Compiler under ${Vulkan_dxc_EXECUTABLE}")
+    else()
+        message(WARNING "Couldn't find DirectX Shader Compiler executable, make sure it is present in Vulkan SDK or add it manually via Vulkan_dxc_EXECUTABLE env variable. HLSL shaders won't be compiled")
+        unset(Vulkan_dxc_EXECUTABLE CACHE)
     endif()
 else()
-    message("VULKAN_SDK path not found, HLSL shaders won't be compiled")
+    message(WARNING "VULKAN_SDK path not found, HLSL shaders won't be compiled")
```
SaschaWillems: I don't think we should put out a warning in this case. Having DXC available is fully optional, hence why we added compiled .spv for all HLSL shaders.

framework/common/error.h
```diff
@@ -81,15 +81,14 @@ class VulkanException : public std::runtime_error
 }        // namespace vkb
 
 /// @brief Helper macro to test the result of Vulkan calls which can return an error.
-#define VK_CHECK(x)                                                 \
```
SaschaWillems, line 84: This change will affect a large portion of the samples, so I'm unsure as to why this change is required. We recently fixed batch mode and I didn't notice any problems in that regard.

samples/extensions/shader_debugprintf/CMakeLists.txt
```diff
@@ -31,4 +31,4 @@ add_sample(
     SHADER_FILES_HLSL
         "shader_debugprintf/hlsl/scene.vert.hlsl"
         "shader_debugprintf/hlsl/scene.frag.hlsl"
-    DXC_ADDITIONAL_ARGUMENTS "-fspv-extension=SPV_KHR_non_semantic_info")
```
SaschaWillems: What's the change here?

samples/performance/async_compute/async_compute.cpp
```diff
@@ -38,6 +38,15 @@ AsyncComputeSample::AsyncComputeSample()
 	config.insert<vkb::BoolSetting>(1, double_buffer_hdr_frames, true);
 }
 
+void AsyncComputeSample::request_gpu_features(vkb::PhysicalDevice &gpu)
+{
+#ifdef VK_ENABLE_BETA_EXTENSIONS
+	// We need to enable the mutableComparisonSamplers feature of the VK_KHR_portability_subset extension
```
SaschaWillems: Can you elaborate on this? Why do we need to enable this in this case?

SaschaWillems (changes requested):
Thank you very much for this PR. Very much appreciated 👍🏻 

I did a quick first look, but with so many changes crammed into one PR it may take some time to properly review this ;)

Though I won't be able to test any Mac OS / iOS related changes due to a lack of such hardware.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -406,11 +439,14 @@ std::unique_ptr<vkb::Instance> ShaderDebugPrintf::create_instance(bool headless)
...
 	enabled_extensions.push_back(VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
 	enabled_extensions.push_back(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+#if (defined(VKB_ENABLE_PORTABILITY))
+	enabled_extensions.push_back(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
+#endif
 
 	VkApplicationInfo app_info{VK_STRUCTURE_TYPE_APPLICATION_INFO};
 	app_info.pApplicationName = "Shader debugprintf";
 	app_info.pEngineName      = "Vulkan Samples";
-	app_info.apiVersion       = VK_API_VERSION_1_1;
```
SaschaWillems, line 413: The reason behind requiring 1.1 here was a bug in the validation layers. On 1.0 this sample run with less than one frame per second for me during developement, and last time I checked that error was still present and the cause for that validation layer bug hasn't been found. I'd prefer to revert this change because of that.

CMakeLists.txt
SRSaunders: OK, that's fine.  Would you prefer this as a STATUS message, or no message at all?

On the topic of many changes at once, I realize the review problem but tried to select items in this PR that work together to achieve a specific outcome - that is a successful batch mode run across all samples on macOS without crashes or early exit prior to all samples completing.  As you can see this touches quite a few things.  However,  I did specifically leave out Xcode and iOS/iOS simulator-related changes for a follow-on PR.  I believe that one can be smaller and more focussed on a few areas.  In any case, please take your time and I am happy to provide explanations and/or make changes based on your feedback.

framework/common/error.h
SRSaunders: This turned out to be important for samples where pipeline construction can fail on macOS/iOS due to SPIR-V to MSL conversion errors caused by lack of support for certain SPIR-V language features, even with the extension showing as supported.  For instance, in the **fragment_shader_barycentric** sample, the Vulkan barycentrric extension is available on MoltenVK, but the SPIR-V language feature `PerVertexKHR` is not currently supported by MSL and `BaryCoordKHR` and `BaryCoordNoPerspKHR` cannot be activated at the same time.  Since SPIR-V to MSL compilation happens on the fly during pipeline construction, you need to check for failures here using `VK_CHECK` like this...

`VK_CHECK(vkCreateGraphicsPipelines(...))`

The previous VK_CHECK would fail out with a hard exit and cause batch mode to stop.  My suggested modification will trigger the error recovery code built into the framework to catch these, stop the current sample, and move on to the next.  Batch mode may work fine on other platforms that have full Vulkan support, but does not work well on portability implementations like MoltenVK where partial support is present not just at the extension level but at the SPIR-V feature level as well.  I did test this change both in batch mode and for individual samples with no observed negative effects on other platforms (Windows for now, linux coming later today).  I will continue to test and will report back if I find issues.

CMakeLists.txt
SaschaWillems: > OK, that's fine. Would you prefer this as a STATUS message, or no message at all?

The latter would be fine. Not having DXC only affects maintainers, a very small group of people.

> On the topic of many changes at once, I realize the review problem but tried to select items in this PR that work together to achieve a specific outcome

And that's totally fine. Just wanted to mention that we often take quite some time to review larger PRs ;)

samples/extensions/shader_debugprintf/CMakeLists.txt
SRSaunders: This spurious change is likely caused by a CRLF difference, where I initially made a change in one commit and then reverted.  I suspect editing on macOS resulted in an end-of-line terminator change.  You can either ignore this one or remove from the change list.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: This is quite strange, since I find the opposite problem with macOS when VK_API_VERSION_1_1 is activated - very slow operation.  This slowdown does not occur when VK_API_VERSION_1_0 is activated in the custom override, or as part of the standard framework (which is defaulted to VK_API_VERSION_1_0) when using layer settings instead.

I wil restest this on Windows and Linux using VK_API_VERSION_1_0 to see what happens with both the custom override and with the standard framework using layer settings.  Stay tuned.

SaschaWillems: Sorry, I mixed up. The slowdown ONLY happens with 1.1. I sadly can't remember why I forced 1.1 for this sample, but I think it was another problem with validation layers at that time.

As for your findings (slowdown), can you add that to this issue: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7562

SRSaunders: Thanks for the clarification.  And yes, I will add to the Validation Layers issue you mention above.
**UPDATE**: Comments have now been added to the issue.

samples/performance/async_compute/async_compute.cpp
SRSaunders: The portability feature `mutableComparisonSamplers` needs to be enabled for certain samples when running on MoltenVK on macOS/iOS, otherwise you get the following validation error...

_[error] 256275552 - VUID-VkDescriptorImageInfo-mutableComparisonSamplers-04450: Validation Error: [ VUID-VkDescriptorImageInfo-mutableComparisonSamplers-04450 ] | MessageID = 0xf467460 | vkUpdateDescriptorSets(): pDescriptorWrites[4] (portability error): sampler comparison not available. The Vulkan spec states: If the VK_KHR_portability_subset extension is enabled, and VkPhysicalDevicePortabilitySubsetFeaturesKHR::mutableComparisonSamplers is VK_FALSE, then sampler must have been created with VkSamplerCreateInfo::compareEnable set to VK_FALSE (https://vulkan.lunarg.com/doc/view/1.3.283.0/mac/1.3-extensions/vkspec.html#VUID-VkDescriptorImageInfo-mutableComparisonSamplers-04450)_

 The `VK_ENABLE_BETA_EXTENSIONS` define is currently synonymous with the `VK_KHR_portability_subset` extension (note this is not the same as `VK_KHR_portability_enumeration`), and as of this pull request is enabled only for Apple platforms.  If I am missing something and this is not correct, please let me know.

CMakeLists.txt
SRSaunders: Removed Vulkan_SDK not found WARNING message in commit https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/6342fc08678d634417068e21d005b4f1e91f235f below. Retained existing message (now STATUS) if Vulkan_SDK found but dxc executable not found.

framework/common/error.h
SRSaunders: I have just completed testing on linux (Manjaro) and the new VK_CHECK functionality seems to work fine on that platform, as well as Windows and macOS/iOS.  Cannot test on Android.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: After testing on linux, I have determined that simple compile-time #ifdef logic for selecting between `VK_EXT_layer_settings` and the original `create_instance()` override in **shader_debugprintf** is not sufficient.  `VK_EXT_layer_settings` can be defined in the included Vulkan Headers but may not be available at runtime depending on the SDK version one has installed.  This can result in runtime failures if the extension is not available.  To get around this I would have to perform a runtime check for `VK_EXT_layer_settings` and somehow revert to the `create_instance()` override if not available.  Let me think about this and see if that can be made simple enough.  ~~If not, I may have to revert the layer settings change in the **shader_debugprintf** sample.~~

SRSaunders: I added commit https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/65d0f7abc0461246e7dad7d0a45f5bce636cdea3 that checks for `VK_EXT_layer_settings ` at runtime and falls back to the override if not available.  Code change was relatively simple, and actually reduces the #ifdefs.  Tested on Windows, Linux, macOS.

SRSaunders:
I added commit https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/7c0195acae6536b1f6ac63f8922cb5c7a818d8f6 to fix the Doxygen and Copyright date issues found by CI.

~~I am not sure what to do with **clang-issues.diff** as it does not appear to be properly formatted git patch file.~~

I have also added commit https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/8caca90df210eb9c1c8c77c89e7141c2b54b8c92 to fix clang-format issues found by CI.

framework/common/error.h
tomadamatkinson: Agreed a throw here is better than an abort. This may cause some more crashes in batch mode from samples not correctly checking handles before deallocating. That is a good thing as it will highlight where samples can be improved to be more correct

SRSaunders:
The remaing CI issues (Doxygen, clang-format) appear to be upstream issues, not related to this PR.

framework/core/hpp_instance.cpp
```diff
@@ -347,6 +350,19 @@ HPPInstance::HPPInstance(const std::string                            &applicati
 	}
 #endif
 
+#if defined(VK_EXT_layer_settings)
+	vk::LayerSettingsCreateInfoEXT layerSettingsCreateInfo;
+
+	// If layer settings extension enabled by sample, then activate layer settings during instance creation
+	if (std::find(enabled_extensions.begin(), enabled_extensions.end(), VK_EXT_LAYER_SETTINGS_EXTENSION_NAME) != enabled_extensions.end())
+	{
+		layerSettingsCreateInfo.settingCount = uint32_t(required_layer_settings.size());
```
asuessenbach: Please use a `static_cast<uint32_t>` here.

framework/core/hpp_instance.cpp
```diff
@@ -186,8 +186,11 @@ bool enable_all_extensions(const std::vector<const char *>             required_
 HPPInstance::HPPInstance(const std::string                            &application_name,
                          const std::unordered_map<const char *, bool> &required_extensions,
                          const std::vector<const char *>              &required_validation_layers,
-                         bool                                          headless,
-                         uint32_t                                      api_version)
+#if defined(VK_EXT_layer_settings)
+                         const std::vector<VkLayerSettingEXT> &required_layer_settings,
```
asuessenbach: The HPPInstance should take a `std::vector<vk::LayerSettingEXT>`!

framework/core/instance.cpp
```diff
@@ -376,6 +379,19 @@ Instance::Instance(const std::string                            &application_nam
 	}
 #endif
 
+#if defined(VK_EXT_layer_settings)
+	VkLayerSettingsCreateInfoEXT layerSettingsCreateInfo{VK_STRUCTURE_TYPE_LAYER_SETTINGS_CREATE_INFO_EXT};
+
+	// If layer settings extension enabled by sample, then activate layer settings during instance creation
+	if (std::find(enabled_extensions.begin(), enabled_extensions.end(), VK_EXT_LAYER_SETTINGS_EXTENSION_NAME) != enabled_extensions.end())
+	{
+		layerSettingsCreateInfo.settingCount = uint32_t(required_layer_settings.size());
```
asuessenbach: Please use  `static_cast<uint32_t>` here.

samples/extensions/shader_debugprintf/CMakeLists.txt
asuessenbach: So, yes, you should remove this from the PR.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -393,9 +416,39 @@ bool ShaderDebugPrintf::prepare(const vkb::ApplicationOptions &options)
...
+	uint32_t instance_extension_count;
+	VK_CHECK(vkEnumerateInstanceExtensionProperties(nullptr, &instance_extension_count, nullptr));
+	std::vector<VkExtensionProperties> available_instance_extensions(instance_extension_count);
+	VK_CHECK(vkEnumerateInstanceExtensionProperties(nullptr, &instance_extension_count, available_instance_extensions.data()));
+
+	// When VK_EXT_layer_settings is available at runtime, the debugPrintfEXT layer feature is enabled using the standard framework
+	// Note this check is necessary in the short term for backwards compatibility with SDKs < 1.3.272 without VK_EXT_layer_settings
+	if (auto it = std::find_if(available_instance_extensions.begin(), available_instance_extensions.end(),
+	                           [](VkExtensionProperties extension) { return strcmp(extension.extensionName, VK_EXT_LAYER_SETTINGS_EXTENSION_NAME) == 0; });
+	    it != available_instance_extensions.end())
```
asuessenbach: Maybe somewhat simpler:
```
	if (std::any_of(available_instance_extensions.begin(),
	                [](VkExtensionProperties extension) { return strcmp(extension.extensionName, VK_EXT_LAYER_SETTINGS_EXTENSION_NAME) == 0; }))
```

samples/performance/async_compute/async_compute.cpp
asuessenbach: If this is a MoltenVK-specific thing, it should be guareded by a check that says so. Using some defined that currently is synonymous to some other that currently is only enabled on that platform is probably not future-safe.

samples/performance/multithreading_render_passes/multithreading_render_passes.cpp
```diff
@@ -39,6 +39,15 @@ MultithreadingRenderPasses::MultithreadingRenderPasses()
 	config.insert<vkb::IntSetting>(2, multithreading_mode, 2);
 }
 
+void MultithreadingRenderPasses::request_gpu_features(vkb::PhysicalDevice &gpu)
+{
+#ifdef VK_ENABLE_BETA_EXTENSIONS
+	// We need to enable the mutableComparisonSamplers feature of the VK_KHR_portability_subset extension
```
asuessenbach: ... same comment as in async_compute.cpp.

samples/performance/pipeline_cache/pipeline_cache.cpp
```diff
@@ -43,11 +43,11 @@ PipelineCache::~PipelineCache()
 	{
 		/* Get size of pipeline cache */
 		size_t size{};
-		VK_CHECK(vkGetPipelineCacheData(get_device().get_handle(), pipeline_cache, &size, nullptr));
+		vkGetPipelineCacheData(get_device().get_handle(), pipeline_cache, &size, nullptr);
```
asuessenbach: Why do you remove the VK_CHECK here?

samples/performance/pipeline_cache/pipeline_cache.cpp
```diff
@@ -43,11 +43,11 @@ PipelineCache::~PipelineCache()
...
 
 		/* Get data of pipeline cache */
 		std::vector<uint8_t> data(size);
-		VK_CHECK(vkGetPipelineCacheData(get_device().get_handle(), pipeline_cache, &size, data.data()));
+		vkGetPipelineCacheData(get_device().get_handle(), pipeline_cache, &size, data.data());
```
asuessenbach: As above.

asuessenbach (commented):
The VulkanSample class is one of the few classes (by now) that are unified with regards to C- and C++-bindings of vulkan. To keep it consistent, some changes to your PR would be needed, that you can find in the attached patch.
[0001-Adjust-LayerSetting-handling-in-VulkanSample-class.patch](https://github.com/user-attachments/files/16125853/0001-Adjust-LayerSetting-handling-in-VulkanSample-class.patch)

samples/extensions/shader_debugprintf/CMakeLists.txt
SRSaunders: Will do.

samples/performance/async_compute/async_compute.cpp
SRSaunders: Based on my understanding, I believe this is a Vulkan _portability_ thing and not a MoltenVK-specific thing.  MoltenVK just happens to be a leading example of portability.  Looking at the validation error message it states:

_If the VK_KHR_portability_subset extension is enabled, and VkPhysicalDevicePortabilitySubsetFeaturesKHR::mutableComparisonSamplers is VK_FALSE, then sampler must have been created with VkSamplerCreateInfo::compareEnable set to VK_FALSE_ 

The two examples (_async_compute_ and _multithreading_render_passes_) where I have enabled `mutableComparisonSamplers` create samplers with `compareEnable` set to VK_TRUE.  Thus the spec requires this portability feature to be enabled for _any_ portability implementation.

The current guard that enables the `VK_KHR_portability_subset` extension and its associated features is `VK_ENABLE_BETA_EXTENSIONS`.  This is already used in other places in the code.  For instance, look at the existing _portability_ sample, and also my addition to `vulkan_sample.h` which enables the `VK_KHR_portability_subset` extension which is also required by the spec for portability implementations.  I was just trying to stay consistent with the approach used by the existing framework and samples.

Note that `VKB_ENABLE_PORTABILITY` has a different purpose in the current framework, and that is to enable the `VK_KHR_portability_enumeration` extension, which is distinct from `VK_KHR_portability_subset`.

framework/core/hpp_instance.cpp
SRSaunders: First of all, thank you @asuessenbach for all your review feedback as well as the patch above.  This is very useful input and I see now this is required to remain consistent with your framework.  I apologize for not considering the unified bindings issue in my original submission and you are absolutely correct in catching this.  The only change I will make is the addition of a `#if defined(VK_EXT_layer_settings)` guard to the definition of `LayerSettingType`.  I don't want compilation to fail with earlier SDK versions that do not support layer settings.

framework/core/hpp_instance.cpp
SRSaunders: Solved with your patch.

framework/core/instance.cpp
SRSaunders: Agreed.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: Thanks for this suggestion.   Agreed.

samples/performance/multithreading_render_passes/multithreading_render_passes.cpp
SRSaunders: Answered in thread above.

samples/performance/pipeline_cache/pipeline_cache.cpp
SRSaunders: I removed this since I understand that it's never a good idea to `throw()` from a destructor.  I believe that modern C++ defaults for destructors is a non-throwing exception specification.  In addition my IDE (Xcode) warns me about this with its static code analysis.  If you are concerned about this sample's case (the only one I can find in the project that calls VK_CHECK in a destructor), I could instead grab the return code and if not success post a fatal error vs. a throw from the VK_CHECK macro.

I am happy to implement your preference here: a) leave it with the VK_CHECK removed, or b) add in a test for failure using the old VK_CHECK expansion code:

```
VkResult result = vkGetPipelineCacheData(...);
if (result != VK_SUCCESS)
{
	LOGE("Detected Vulkan error: {}", vkb::to_string(result));
	abort();
}      
```

samples/performance/pipeline_cache/pipeline_cache.cpp
SRSaunders: Same answer as above...

samples/performance/async_compute/async_compute.cpp
asuessenbach: Yep, sounds reasonable.
Thanks for the clarifications.

samples/performance/pipeline_cache/pipeline_cache.cpp
asuessenbach: Good catch!

I think, some error check for those cases would be good. Or maybe, you could even introduce something like a `VK_CHECK_NOTHROW`, or so?

framework/core/hpp_instance.cpp
SRSaunders: I would like to ask @asuessenbach one additional question before I resolve the final issues of this PR.  Is the `#if defined(VK_EXT_layer_settings)` guard truly necessary in the context of this project?  I placed these in the code to prevent build problems if compiled against an older SDK.  However, will that ever occur with this project?  Since Vulkan-Samples uses a fixed version of Vulkan Headers (currently 277) where  `VK_EXT_layer_settings` is defined, do we have to worry?  As you noticed, I recently added a runtime check for the layer settings extension in _shader_debugprintf_, so that case is covered off.  The other layer settings case is for the _descriptor_indexing_ sample which is Apple/MoltenVK-specific and things are rapidly evolving there regarding the use of Metal Argument buffers and default support for bindless - thus I am less concerned about it.

Can you please provide your advice on this?  Removing the `#if defined(VK_EXT_layer_settings)` conditionals would make the code cleaner if there is no risk of build failures.  Or are there cases where older versions of Vulkan-Headers will be used, or cases where the project is statically built against a specific SDK version?

asuessenbach: I think, having code guarded by checks against the extension that defines the used stuff would be very instructive... but even more distracting. Especially if you would introduce that for other extensions as well, the code would get unreadable very quickly.
And as (I think) this is the first compile-time check against the existence of an extension, I would vote for _not_ having those guards. It should suffice to have run-time checks.
@SaschaWillems, what do you think?

SaschaWillems: Yeah, I'm totally with @asuessenbach on this. We never guard against an extension, and we wanted to add `VK_EXT_layer_settings` support anyway (See #863).

And we also don't need backwards compat with old SDKs. Actually the samples work fine without.

SRSaunders: Thanks for your quick feedback @asuessenbach and @SaschaWillems.  I will remove them.

samples/performance/pipeline_cache/pipeline_cache.cpp
SaschaWillems: Exactly. Just ignoring the error isn't good practice either. So your proposal of log + abort looks good to me, but then again it's a sample, and since that change is not related to the PR maybe we should just revert this.

SRSaunders: Well, it's kind of related since calling `throw()` inside `VK_CHECK` makes it inappropriate for a destructor.  I don't think I should just revert to original since it would leave a new problem behind.  I had a closer look at what the code is actually doing and would like to suggest the following:

```
/* Get size of pipeline cache */
size_t size{};
VkResult result = vkGetPipelineCacheData(get_device().get_handle(), pipeline_cache, &size, nullptr);
if (result == VK_SUCCESS && size > 0)
{
	/* Get data of pipeline cache */
	std::vector<uint8_t> data(size);
	result = vkGetPipelineCacheData(get_device().get_handle(), pipeline_cache, &size, data.data());
	if (result == VK_SUCCESS && size > 0)
	{
		if (size < data.size())
		{
		  data.resize(size);
		}

		/* Write pipeline cache data to a file in binary format */
		vkb::fs::write_temp(data, "pipeline_cache.data");
	}
}

if (result != VK_SUCCESS || size == 0)
{
	LOGE("Detected Vulkan error: {}, pipeline cache not saved.", vkb::to_string(result));
}

/* Destroy Vulkan pipeline cache */
vkDestroyPipelineCache(get_device().get_handle(), pipeline_cache, nullptr);
```

This would detect any errors and prevent the write operation if either `vkGetPipelineCacheData()` fails.  It also would report any such failure to the log without doing an abort and then shutdown gracefully.  Note this is essentially the same code as in **vulkan_funcs.hpp**.

Frankly, I am not even sure if `LOGE()` is important here, given the **pipeline_cache.data** file is only for storing cache data from previous runs, and can be used for cache warmup on follow-on runs.  Once I hear back I will push this change and hopefully you can proceed with any final reviews.

SRSaunders: Submitted _pipeline_cache_ destructor error handling change.  Comments welcome on review.

SRSaunders:
I don't plan to make any more changes here unless there are additional review comments or issues found.

I will submit changes in a separate PR to remove macOS workarounds within the _descriptor_indexing_ sample once improvements are made to MoltenVK in this area (currently in development).

samples/extensions/descriptor_indexing/descriptor_indexing.cpp
```diff
@@ -31,6 +31,26 @@ DescriptorIndexing::DescriptorIndexing()
...
+	// On macOS use layer setting to configure MoltenVK for using Metal argument buffers (needed for descriptor indexing)
+	add_instance_extension(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME, /*optional*/ true);
+
+	VkLayerSettingEXT layerSetting;
+	layerSetting.pLayerName   = "MoltenVK";
+	layerSetting.pSettingName = "MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS";
+	layerSetting.type         = VK_LAYER_SETTING_TYPE_INT32_EXT;
+	layerSetting.valueCount   = 1;
+
+	static const int32_t useMetalArgumentBuffers = 1;
```
asuessenbach, line 46: Maybe add a short comment why this needs to be a static variable.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -393,9 +415,37 @@ bool ShaderDebugPrintf::prepare(const vkb::ApplicationOptions &options)
 	return true;
 }
 
+// This sample overrides the per-sample layer framework to force activation of the validation layer
+const std::vector<const char *> ShaderDebugPrintf::get_validation_layers()
+{
+	// Note validation layer is already enabled for debug builds, so set the default list to empty
+	std::vector<const char *> validation_layers = {};
```
asuessenbach: Maybe instead:
`std::vector<const char *> validation_layers = ApiVulkanSample::get_validation_layers();`

samples/extensions/descriptor_indexing/descriptor_indexing.cpp
SRSaunders: The layer settings API wants a pointer to the setting and not its value (i.e. not a copy), in this case:

`layerSetting.pValues = &useMetalArgumentBuffers`

Given the vector of layer settings is used outside of the sample's constructor context (i.e. after the constructor's completion), you need to keep the variable around so it has not been deallocated when referenced by `vkCreateInstance()`.  This would be made easier in ref counting implementations, but not the case here.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: Yes, I can do this if you would like.  Can you tell me if `ApiVulkanSample::get_validation_layers()` will ever return non-null?  i.e. could there be a time in the future when it will contain other values?

samples/extensions/descriptor_indexing/descriptor_indexing.cpp
asuessenbach: I mean... a short comment in the source ;)

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
asuessenbach: Are you asking me about the future???

It might be possible, yeah. Today, I can't imagine what might lead to such a situation.
But no matter what, I think, it's good practise to somehow explicitly handle the returned value from the base class when you override a function.

samples/extensions/descriptor_indexing/descriptor_indexing.cpp
SRSaunders: Sorry :)

I will do this when I get home tonight (EDT time).

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: Same as above,  will fix later tonight EDT.

framework/vulkan_sample.h
```diff
@@ -1044,6 +1087,11 @@ inline bool VulkanSample<bindingType>::prepare(const ApplicationOptions &options
 		}
 	}
 
+#ifdef VK_ENABLE_BETA_EXTENSIONS
```
SaschaWillems: Shouldn't this be guarded by some mac/ios specific define instead? It's possible that we enable beta extensions for future samples (afair we even did that at some point, maybe with ray tracing) and I guess this would fail for all samples then as it'll always request portability.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -393,9 +416,37 @@ bool ShaderDebugPrintf::prepare(const vkb::ApplicationOptions &options)
 	return true;
 }
 
+// This sample overrides the per-sample layer framework to force activation of the validation layer
```
SaschaWillems, line 419: Why not always simply return VK_LAYER_KHRONOS_validation?

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -375,6 +391,13 @@ bool ShaderDebugPrintf::prepare(const vkb::ApplicationOptions &options)
 		return false;
 	}
 
+	// Register debug utils callback here vs in ShaderDebugPrintf::create_instance() so it works with both override and layer settings
```
SaschaWillems, line 394: Can you elaborate on this comment and also check if the readme for this sample needs to be adjusted?

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -393,9 +416,37 @@ bool ShaderDebugPrintf::prepare(const vkb::ApplicationOptions &options)
...
+
 // This sample overrides the instance creation part of the framework to chain in additional structures
 std::unique_ptr<vkb::Instance> ShaderDebugPrintf::create_instance(bool headless)
 {
+	uint32_t instance_extension_count;
+	VK_CHECK(vkEnumerateInstanceExtensionProperties(nullptr, &instance_extension_count, nullptr));
+	std::vector<VkExtensionProperties> available_instance_extensions(instance_extension_count);
+	VK_CHECK(vkEnumerateInstanceExtensionProperties(nullptr, &instance_extension_count, available_instance_extensions.data()));
+
+	// When VK_EXT_layer_settings is available at runtime, the debugPrintfEXT layer feature is enabled using the standard framework
```
SaschaWillems, line 441: Can you elaborate on this? I read through that comment multiple times but I still don't get why it's needed. If it's really required please also add a note to the sample's readme.

SaschaWillems (changes requested):

framework/vulkan_sample.h
SRSaunders: @SaschaWillems I am away for a few days but should be able to respond tomorrow night EDT.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
asuessenbach: I think, for release builds, you usually don't want the validation layer to be active.

asuessenbach (dismissed):

framework/vulkan_sample.h
SRSaunders: @asuessenbach had exactly the same question.  This was my answer: https://github.com/KhronosGroup/Vulkan-Samples/pull/1084#discussion_r1669430006.  It's not a macOS/iOS thing, it's a _portability_ library thing, and I was trying to stay within the conventions of the project (see _portability_ sample which I have not touched).

However, given that @SaschaWillems and @asuessenbach have raised this, there is likely a semantics problem with using `VK_ENABLE_BETA_EXTENSIONS` as a dual-purpose thing within the project: i.e. to enable a) portability subset features and b) Vulkan Beta extensions (which includes portability).  Perhaps we should instead ~~be looking at expanding the meaning and use of `VKB_ENABLE_PORTABILITY` (i.e. to cover portability enumeration _and_ subset features)~~ create a new `VKB_ENABLE_PORTABILITY_SUBSET` define.  We would then set `VK_ENABLE_BETA_EXTENSIONS` as a dependency of `VKB_ENABLE_PORTABILITY_SUBSET` so the Vulkan library portability features are exposed, but **not** use it as a guard within the project.   That would solve the problem for possible future needs regarding beta extensions for other purposes.  Would you like me to push an additional commit that makes this change?

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: I did this so that `VK_LAYER_KHRONOS_validation` is not enabled twice when in Debug mode and the log does not show the same validation layer twice in a row.  Otherwise you get this:

```
[info] Vulkan debug utils enabled (VK_EXT_debug_utils)
[info] Extension VK_EXT_debug_utils found, enabling it
[info] Extension VK_KHR_get_physical_device_properties2 found, enabling it
[info] Extension VK_KHR_portability_enumeration found, enabling it
[info] Extension VK_EXT_metal_surface found, enabling it
[info] Extension VK_EXT_layer_settings found, enabling it
[info] 		VK_LAYER_KHRONOS_validation
[info] 		VK_LAYER_KHRONOS_validation
```

I figured this was the easiest/cleanest solution.  Otherwise you would have to check for uniqueness when assembling the validation layers in `Instance::Instance()`

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: There are two code paths now for enablement of the validation layer and the debug printf feature of the validation layer:  a) the existing approach that uses a custom override for `create_instance()`, and b) a new approach that enables the validation layer using a simple `get_validation_layers()` override (for release builds) combined with the layer settings method of enabling the debug printf feature of the validation layer.  The latter does not use the custom `create_instance()` override code, but instead calls the standard `VulkanSample::create_instance()` method if `VK_EXT_layer_settings` is available at runtime (see your next question).

The registration of the debug utils callback has to happen after either approach has finished, and the best location for that is at the beginning of `ShaderDebugPrintf::prepare()`.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: This runtime check for `VK_EXT_layer_settings` allows both `create_instance()` approaches to coexist and not break backwards compatibility for older SDKs or on platforms that don't yet support `VK_EXT_layer_settings`.  If backwards compatibility is not a concern, then the `create_instance()` override method would not be needed at all, and it could be removed from the code for 100% reliance on `VK_EXT_layer_settings` to enable debug printf.  Given that layer settings is relatively new, I felt that backwards compatibilty in this case was necessary.

If the test in the above code snippet returns true (i.e. finds `VK_EXT_layer_settings` at runtime), then the code runs the standard framework's `create_instance()` method (with layer settings) and returns, while if the test fails and `VK_EXT_layer_settings` is not available at runtime, the override's custom `create_instance()` code is run instead.

framework/vulkan_sample.h
SRSaunders: I looked at the code again, and found a way to use the existing `VKB_ENABLE_PORTABILITY` define to handle both the portability enumeration _and_ portability subset extensions in the framework.  The result is actually a bit better than before, with dynamic determination of `VK_KHR_portability_enumeration` in the runtime instance vs. reliance on static enablement in cmake config files.  This is important not just for support of older SDKs, but also to allow future support of MoltenVK standalone.  Note that `VK_KHR_portability_enumeration` is a Vulkan loader feature only, and not present in the actual driver.  When MoltenVK standalone is used, the extension is not present and the code needs to be flexible enough to handle this case.

This change also solves the concerns of @SaschaWillems regarding possible future enablement of Vulkan beta extensions (`VK_ENABLE_BETA_EXTENSIONS`) for purposes other than portability.

The changes are in commits https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/6e73f7c60eba57b9dab968cd5134d6e14cebc9f6 and https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/c756611c9e09c9ebc2a33e06db6c0dd08d9a370c.  Please review and let me know whether this helps from your perspective.  If this does not suit your plans, then it could always be reverted.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: If you are okay with this dual track approach for backwards compatability in this sample, I will come back and add info to the README explaining it.  Please let me know.

SaschaWillems: Totally fine with it 👍🏻

framework/vulkan_sample.h
SRSaunders: I am pretty sure this change is correct, so I am going to close this specific item and request a formal review later tonight.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: Thanks - I will update the README tonight and close this final issue before requesting a final re-review.

SRSaunders: README updated in commit https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/98acbe72e512438c41e5e19b819c4fdabbcd1b8a.  Closing this specific comment which resolves the last outstanding issue.  I will now request re-reviews of the PR, hopefully we are getting closer!

SRSaunders:
Hopefully all issues raised by @asuessenbach and @SaschaWillems are now resolved.  Again, I don't plan any more updates and re-review of recent changes / resolutions can now proceed.

CMakeLists.txt
```diff
@@ -46,20 +48,17 @@ if(VULKAN_FOUND)
                     "$ENV{VULKAN_SDK}/bin"
                     )
     endif()
-    if(DEFINED Vulkan_dxc_EXECUTABLE)
-        message("Found DirectX Shader Compiler under ${Vulkan_dxc_EXECUTABLE}")
-    elseif()
-        message("Couldn't find DirectX Shader Compiler executable, make sure it is present in Vulkan SDK or add it manually via Vulkan_dxc_EXECUTABLE env variable")
+    if(Vulkan_dxc_EXECUTABLE)
```
gpx1000, line 51: It's probably better to use Vulkan_dxc_exe_FOUND for the boolean check

CMakeLists.txt
```diff
@@ -46,20 +48,17 @@ if(VULKAN_FOUND)
...
+        message(STATUS "Found DirectX Shader Compiler under ${Vulkan_dxc_EXECUTABLE}")
+    else()
+        message(STATUS "Couldn't find DirectX Shader Compiler executable, make sure it is present in Vulkan SDK or add it manually via Vulkan_dxc_EXECUTABLE cmake variable. HLSL shaders won't be compiled.")
+        unset(Vulkan_dxc_EXECUTABLE CACHE)
```
gpx1000: Why would we unset this if it's not found?

app/CMakeLists.txt
```diff
@@ -37,17 +37,15 @@ if(ANDROID)
...
-        list(REMOVE_AT SRC 0)
-        list(APPEND SRC
-                ${CMAKE_CURRENT_SOURCE_DIR}/ios/main.mm
-                ${CMAKE_CURRENT_SOURCE_DIR}/ios/AppDelegate.h
-                ${CMAKE_CURRENT_SOURCE_DIR}/ios/AppDelegate.m
-                ${CMAKE_CURRENT_SOURCE_DIR}/ios/ViewController.h
-                ${CMAKE_CURRENT_SOURCE_DIR}/ios/ViewController.mm
-        )
-    endif ()
+elseif(IOS)
```
gpx1000, line 40: Is this logic change correct?  It looks like it went from adding ios specific files before defining the executable as a macosx bundle when it's an ios file but when it's just apple, it's a macosx bundle with the changes it looks like we're treating osx bundles as a win32.

bldsys/cmake/global_options.cmake
```diff
@@ -33,15 +33,10 @@ endif()
 
 if(APPLE)
     cmake_minimum_required(VERSION 3.24)
+    set(VKB_ENABLE_PORTABILITY ON CACHE BOOL "Enable portability enumeration and subset features in the framework.  This is required to be set when running on Apple platforms." FORCE)
```
gpx1000, line 36: I think the original method might be preferred.  If someone is using vulkan less than 1.3 then they are on a build that doesn't yet have the portability extension.  Which would mean their app would go from working to not working with this mandated.

CMakeLists.txt
SRSaunders: Thanks for reviewing the PR.  Yes, I can change `Vulkan_dxc_EXECUTABLE` to `Vulkan_dxc_exe_FOUND`.  The semantics are the same, but your suggestion reads better.

CMakeLists.txt
SRSaunders: I did this for performance reasons.  See the following code snippet from **sample_helper.cmake**:

```
    if(DEFINED Vulkan_dxc_EXECUTABLE AND DEFINED SHADERS_HLSL)
        compile_hlsl_shaders(
            SHADERS_HLSL ${TARGET_SHADERS_HLSL}
            DXC_ADDITIONAL_ARGUMENTS ${TARGET_DXC_ADDITIONAL_ARGUMENTS}
        )
    endif()
```

When dxc is **not** found, `${Vulkan_dxc_EXECUTABLE}` will equal `Vulkan_dxc_EXECUTABLE-NOTFOUND`, which is **DEFINED** in cmake semantics.  If I did not unset the variable, the logic above will run erroneously and slow down/timeout on a missing dxc executable.  A different solution might change the above code to either:

`if(Vulkan_dxc_EXECUTABLE` ... or `if(Vulkan_dxc_exe_FOUND ...`

I did not do either of these since I wanted to minimize the amount of code I touched in the PR.  If you don't like the `unset()`, then I could fix the problem within **sample_helper.cmake** as above.  Please advise.

app/CMakeLists.txt
SRSaunders: Yes, this is correct.  You only want an app bundle (`MACOSX_BUNDLE`) for iOS since it is required there.  On macOS using an app bundle is awkward since **vulkan_samples** is fundamentally a command line application with command line options.  Actually your current documentation assumes it's a command line app.  When making a command line application on macOS or linux, the cmake option WIN32 is ignored.  Note this cmake code path (without my change) currently supports linux, even though `add_executable()` specifies `WIN32`.

bldsys/cmake/global_options.cmake
SRSaunders: My change here was motivated by feedback from @SaschaWillems and @asuessenbach.  They both felt that using `VK_ENABLE_BETA_EXTENSIONS` as a guard in the code base was confusing and overloaded.  I tend to agree.  This define should be used as input for configuring Vulkan and not changing code in Vulkan-Samples, specifically around use of the `VK_KHR_portability_subset` device extension and features.  The project should be able to support future uses of Vulkan beta extensions other than for just portability.  I wanted to make a simple change to fix this without introducing a new define into the VKB_ configuration parameters.

What I ended up doing was using the existing `VKB_ENABLE_PORTABILITY` define to switch on both the `VK_KHR_portability_enumeration` instance extension and the `VK_KHR_portability_subset` device extension/features on Apple platforms (can also be used on any other portability platform).  However, I did this with a difference compared to the existing code base: Rather than relying on static config within cmake, I now test for existence at runtime in the code (see commits https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/6e73f7c60eba57b9dab968cd5134d6e14cebc9f6 and https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/c756611c9e09c9ebc2a33e06db6c0dd08d9a370c).  When running against an older SDK without the enumeration extension, it will not be found and therefore not enabled.  I have also tested this against MoltenVK standalone which does not implement the enumeration extension since that is a loader-only capability.  Note that build-time concerns are not a problem since Vulkan-Samples ships with Vulkan Headers 277, where the enumeration extension is defined.

Lastly, the old cmake test for Vulkan SDK 1.3.x was not strictly correct.  The `VK_KHR_portability_enumeration` extension was introduced in SDK 1.3.211, and made mandatory in 1.3.216.  Any 1.3 SDKs earlier than 211 will fail under the current implementation (i.e. extension not present).  I am not sure if this is much of a practical concern nowadays, but perhaps it's a detail worth noting for comparison purposes.

gpx1000: Perfect!  Thanks for the work and getting the right solution!  I'm sold on this.

app/CMakeLists.txt
gpx1000: I think this is correct for the current architecture.  I think that we should modify the OSX application logic to be useful in a MACOSX_BUNDLE rather than a standard add_executable() call.  That's more purely for the fact that it would be more idomatic.  However, I think upon review, I like your approach better.

CMakeLists.txt
gpx1000: It's probably better to go ahead and change it to if(Vulkan_dxc_exe_FOUND to be more idomatic as this is a sample of best practices repo then that path does hold weight with me.

SRSaunders: OK thanks very much for the feedback.  I will make the change, retest, and submit.

CMakeLists.txt
SRSaunders: A quick update:  Can't use `Vulkan_dxc_exe_FOUND` in all places.  This variable is only returned by the **FindVulkan.cmake** module used by `find_package(Vulkan)`.  If you need to fall back to `find_program()` this variable will not be set.  So the code will need to be:

```
# search for Vulkan SDK
find_package(Vulkan)

if(Vulkan_FOUND)
    if(NOT Vulkan_dxc_exe_FOUND)
        find_program(Vulkan_dxc_EXECUTABLE
                    NAMES dxc
                    HINTS
                    "$ENV{VULKAN_SDK}/Bin"
                    "$ENV{VULKAN_SDK}/bin"
                    )
    endif()
    if(Vulkan_dxc_EXECUTABLE)
        message(STATUS "Found DirectX Shader Compiler under ${Vulkan_dxc_EXECUTABLE}")
    else()
        message(STATUS "Couldn't find DirectX Shader Compiler executable, make sure it is present in Vulkan SDK or add it manually via Vulkan_dxc_EXECUTABLE cmake variable. HLSL shaders won't be compiled.")
    endif()
endif()
```

The second test can't use `Vulkan_dxc_exe_FOUND`, but must use `Vulkan_dxc_EXECUTABLE`.  Same goes for the change in **sample_helper.cmake** below.

SRSaunders: Fixed in https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/1e3e85b9ab95a2e04b2d0180452ce9b68f509bff (`Vulkan_dxc_exe_FOUND` used in first test only, and `unset()` removed, as shown above).

CMakeLists.txt
SRSaunders: Fixed in https://github.com/KhronosGroup/Vulkan-Samples/pull/1084/commits/1e3e85b9ab95a2e04b2d0180452ce9b68f509bff (`Vulkan_dxc_EXECUTABLE` used instead of `DEFINED Vulkan_dxc_EXECUTABLE` in **sample_helper.cmake**).

Cannot use `Vulkan_dxc_exe_FOUND` due to `find_program()` fallback in previous review point.

gpx1000 (approved):

asuessenbach (approved):

SRSaunders:
@SaschaWillems are there any further changes needed to proceed with this PR?  I have tried to respond to your review comments with resolutions but have not heard back.  If there is anything else please let me know.  I would like to submit some iOS fixes and Xcode improvements (in a new PR) that are dependent on this one.  Thanks.

SaschaWillems:
Two things: I'm currently lacking time for PRs and I don't own any Apple device so I can't actually test if this works. We discussed this on our recent call (how we'll review Apple related PRs in the future), so hopefully someone else can green light this.

SRSaunders:
> We discussed this on our recent call (how we'll review Apple related PRs in the future), so hopefully someone else can green light this.

Given your constraints, if you could nominate or assign an Apple-knowledgeable reviewer that would be great.  Thanks.

SaschaWillems:
We are already working on that and hopefully have another official Mac reviewer soon ;)

SaschaWillems (approved):
It compiles and works fine for me (did a full batch run) on Win 11 with an RT 4070, and I'm fine with the code changes. I don't want to block this, so I'll approve it from a code reviewers point-of-view.

SRSaunders:
Thanks @SaschaWillems for approving.

After this is merged, I may update the logic for picking the API version for _shader_debugprintf_.  Given that the VVL is now fixed and will be picked up in the next SDK, the API version should be set to 1.1 starting with the next SDK release.  I can add a bit of logic to detect this at runtime and dynamically set the correct API version based on the running SDK instance.  I would put this change in a supplementary PR.  What do you think?

gpx1000:
@marty-johnson59 ready to merge here. 

@SRSaunders we typically try to be careful about updating the API so older versions which might still be out there in the wild continue to work.  However, I certainly am on board with considering an update.  Ideally we try to keep it to just the iPhone or Mac platforms. 
Lemme know if you need any help with it.  I do Mac development and can be reliable for any help you might need.

SRSaunders:
@gpx1000 thanks for approving the merge.

Regarding the _shader_debugprintf_ sample, there has been a platform-independent bug in the Vulkan Validation Layers that manifests itself as an-almost unusable performance issue when using the debugPrintfEXT feature with the correct Vulkan API 1.1 setting.  This happens on all platforms, not just macOS.  See defect https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7562.  This PR changes the sample to Vulkan API 1.0 as a temporary workaround, but it should be moved back to Vulkan 1.1 when the fixed VVL ships, which should be in the next SDK version.  I was just offering to make a follow-on change to this PR that recognizes this and allows the _shader_debugprintf_ sample code to adapt to the SDK instance it is running.  Note the VVL fix came in after this PR was frozen for review so I did not change it here.

Your offer to assist with macOS issues is very welcome - thank you!  In fact, I do have a pending fix that I would like to get feedback on - relating to swapchain recreation with MoltenVK.  I will open a separate issue and ask for your feedback there.  Thanks again.

JoseEmilio-ARM:
Hi, I missed this PR, but recently I needed to test the settings extension and I have been trying to use the helper function `add_layer_setting` from here (thanks for this!), and I think I might have found a bug.

I was trying to enable the best practice validation, by simply adding these lines to the constructor of a different sample:

```
	const char*    layer_name = "VK_LAYER_KHRONOS_validation";
	const VkBool32 value      = VK_TRUE;

	add_layer_setting({layer_name, "validate_best_practices", VK_LAYER_SETTING_TYPE_BOOL32_EXT, 1, &value});
```

However, to make it work, I had to remove the requirement that the `VK_EXT_layer_settings` is enabled as an Instance extension. As it has been discussed here before, this is actually not a requirement, since the extension is implemented by the validation layer itself. 

If I remove these lines from `instance.cpp`, then the settings work and I can see the validation warnings as expected:

```
--- a/framework/core/instance.cpp
+++ b/framework/core/instance.cpp
@@ -383,13 +383,14 @@ Instance::Instance(const std::string                            &application_nam
        VkLayerSettingsCreateInfoEXT layerSettingsCreateInfo{VK_STRUCTURE_TYPE_LAYER_SETTINGS_CREATE_INFO_EXT};

        // If layer settings extension enabled by sample, then activate layer settings during instance creation
-       if (std::find(enabled_extensions.begin(), enabled_extensions.end(), VK_EXT_LAYER_SETTINGS_EXTENSION_NAME) != enabled_extensions.end())
-       {
                layerSettingsCreateInfo.settingCount = static_cast<uint32_t>(required_layer_settings.size());
                layerSettingsCreateInfo.pSettings    = required_layer_settings.data();
                layerSettingsCreateInfo.pNext        = instance_info.pNext;
                instance_info.pNext                  = &layerSettingsCreateInfo;
-       }
```

Enabling the extension from the sample as suggested in the comment above does not work for me, as the instance keeps reporting the extension as not supported.

(Update: is this the same bug as the one being addressed in #1187 ?)

SRSaunders:
> However, to make it work, I had to remove the requirement that the `VK_EXT_layer_settings` is enabled as an Instance extension. As it has been discussed here before, this is actually not a requirement, since the extension is implemented by the validation layer itself.

@JoseEmilio-ARM you are completely correct. In the process of solving the issues in #1187, and with the help of VVL engineers, I came to realize that the `VK_EXT_layer_settings` extension is typically provided by the layer, and only sometimes provided by the driver (e.g. MoltenVK).  As a result I made a change in that PR to do the following in **instance.cpp** and **hpp_instance.cpp**:

```
// If layer settings are defined, then activate the sample's required layer settings during instance creation
if (required_layer_settings.size() > 0)
{
	layerSettingsCreateInfo.settingCount = static_cast<uint32_t>(required_layer_settings.size());
	layerSettingsCreateInfo.pSettings    = required_layer_settings.data();
	layerSettingsCreateInfo.pNext        = instance_info.pNext;
	instance_info.pNext                  = &layerSettingsCreateInfo;
}
```

If you could retest after applying #1187 that would be great (note I have just merged in main to correct a recent conflict caused by changes to headless mode).  Please let me know if this works for you.

Hopefully this PR can be reviewed and merged relatively soon, as it fixes this general layer settings issue as well as the _shader_debugprintf_ sample.

PR #1090, opened 2024-07-08
Fix support for --headless.

### LINKED ISSUE #314: Headless mode crashes with API samples
`./vulkan_samples render_passes --headless` works, while
`./vulkan_samples instancing --headless` does not for example.

The crash appears to be at the end of ApiVulkanSample::submit_frame(). It calls device->get_queue_by_present(0) which fails because there are no queues that support 'present' since we're in headless mode.

### LINKED ISSUE #830: Headless mode seems to be broken
None of the samples seem to work in headless mode anymore. I get the following validation error when running with `--headless`:

```
[error] [framework\core\instance.cpp:50] -437968512 - VUID-vkCreateInstance-ppEnabledExtensionNames-01388: Validation Error: [ VUID-vkCreateInstance-ppEnabledExtensionNames-01388 ] Object 0: VK_NULL_HANDLE, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xe5e52180 | Missing extension required by the instance extension VK_EXT_headless_surface: VK_KHR_surface. The Vulkan spec states: All required extensions for each extension in the VkInstanceCreateInfo::ppEnabledExtensionNames list must also be present in that list (https://vulkan.lunarg.com/doc/view/1.3.261.1/windows/1.3-extensions/vkspec.html#VUID-vkCreateInstance-ppEnabledExtensionNames-01388)
```

As per the spec `VK_EXT_headless_surface` requires `VK_KHR_surface`, yet at instance creation we do this:

```cpp
	// Try to enable headless surface extension if it exists
	if (headless)
	{
		const bool has_headless_surface = enable_extension(VK_EXT_HEADLESS_SURFACE_EXTENSION_NAME,
		                                                   available_instance_extensions, enabled_extensions);
		if (!has_headless_surface)
		{
			LOGW("{} is not available, disabling swapchain creation", VK_EXT_HEADLESS_SURFACE_EXTENSION_NAME);
		}
	}
	else
	{
		enabled_extensions.push_back(VK_KHR_SURFACE_EXTENSION_NAME);
	}
```

Imo this is wrong and the else path should always be run. But even changing that will then cause the sample to fail at surface creation, as the base sample class always expects a valid surface:

```cpp
	if (!surface)
	{
		throw std::runtime_error("Failed to create window surface.");
	}
```

It looks like headless isn't properly handled, at least not on windows and possibly completely broken. 

Refs #314

iagoCL:
## Description

The main objective of this PR is to fix the `--headless` flag. Currently it does not work in the framework, due to not requesting a surface and swap chain.

This PR reworks `HeadlessWindow` to make it more similar to the other window classes. This will allow us to enable `VK_EXT_headless_surface` in `HeadlessWindow::get_required_surface_extensions` like the other windows types. Now we can remove the `headless `paramater from `Instance::Instance` and `VulkanSample<>::create_instance` simplifying the framework.

To keep the samples `hello_world` and `hpp_hello_world` as simple as possible, I did not add support for headless. Instead, an assert is run to ensure the are not enabling with headless in those samples.

## General Checklist:

Fixes #830 
Fixes #314

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -118,8 +118,12 @@ vulkan_samples sample swapchain_images
 # Run AFBC sample in benchmark mode for 5000 frames
 vulkan_samples sample afbc --benchmark --stop-after-frame 5000
 
-# Run bonza test offscreen
-vulkan_samples test bonza --headless
+# Run compute nbody using headless_surface and take a screenshot of frame 5 
+# Note: headless_surface uses VK_EXT_headless_surface.
+# This will create a surface and a Swapchain, but present will be a no op.
+# The extension is supported by Swiftshader(https://github.com/google/swiftshader).
+# It allows to quickly test content in environments without a GPU.
+vulkan_samples sample compute_nbody --headless_surface -screenshot 5
 
 # Run all the performance samples for 10 seconds in each configuration
 vulkan_samples batch --category performance --duration 10
--- a/app/android/java/com/khronos/vulkan_samples/SampleLauncherActivity.java
+++ b/app/android/java/com/khronos/vulkan_samples/SampleLauncherActivity.java
@@ -292,7 +292,7 @@ public class SampleLauncherActivity extends AppCompatActivity {
             arguments.add("--benchmark");
         }
         if (isHeadless) {
-            arguments.add("--headless");
+            arguments.add("--headless_surface");
         }
         String[] argArray = new String[arguments.size()];
         sendArgumentsToPlatform(arguments.toArray(argArray));
--- a/app/plugins/window_options/window_options.h
+++ b/app/plugins/window_options/window_options.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2020-2022, Arm Limited and Contributors
+/* Copyright (c) 2020-2024, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -47,7 +47,7 @@ class WindowOptions : public WindowOptionsTags
 	vkb::FlagCommand width_flag      = {vkb::FlagType::OneValue, "width", "", "Initial window width"};
 	vkb::FlagCommand height_flag     = {vkb::FlagType::OneValue, "height", "", "Initial window height"};
 	vkb::FlagCommand fullscreen_flag = {vkb::FlagType::FlagOnly, "fullscreen", "", "Run in fullscreen mode"};
-	vkb::FlagCommand headless_flag   = {vkb::FlagType::FlagOnly, "headless", "", "Run in headless mode"};
+	vkb::FlagCommand headless_flag   = {vkb::FlagType::FlagOnly, "headless_surface", "", "Run in headless surface mode. A Surface and swap-chain is still created using VK_EXT_headless_surface."};
 	vkb::FlagCommand borderless_flag = {vkb::FlagType::FlagOnly, "borderless", "", "Run in borderless mode"};
 	vkb::FlagCommand stretch_flag    = {vkb::FlagType::FlagOnly, "stretch", "", "Stretch window to fullscreen (direct-to-display only)"};
 	vkb::FlagCommand vsync_flag      = {vkb::FlagType::OneValue, "vsync", "", "Force vsync {ON | OFF}. If not set samples decide how vsync is set"};
--- a/framework/api_vulkan_sample.cpp
+++ b/framework/api_vulkan_sample.cpp
@@ -1,4 +1,5 @@
 /* Copyright (c) 2019-2024, Sascha Willems
+ * Copyright (c) 2024, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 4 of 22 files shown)


SaschaWillems (changes requested):
This crashes for me when trying to start any sample in headless mode.

The crash happens in `getSurfaceSupportKHR` in the vulkan hpp header and is coming from `get_suitable_gpu` in `hpp_instance.cpp`.

Callstack:

image

It works fine without headless.

System is Windows 11 with an NV RTX 4070 and recent Vulkan drivers. The `K_EXT_headless_surface` extension is supported.

I also have a switfshader ICD installed to test features not supported by my GPU, but the crash happens when calling `getSurfaceSupportKHR` for my discrete RTX 4070.

nv-ahamelin:
Change tested on:
- Ubuntu: 24.04 LTS
- GPU: RTX 4080
- Layer exposing `VK_EXT_headless_surface` (which is not currently supported natively on NVIDIA GPUs): https://gitlab.freedesktop.org/mesa/vulkan-wsi-layer

I have tested 3 different setups, 1. Display connected+GDM3 started, 2. Display connected+GDM3 shutdown and 3. Display disconnected.

1. Display connected+GDM3 started
The change appears to be functional on my side, I am able to run multiple samples with the `--headless` option without encountering errors.
 
2. Display connected+GDM3 shutdown
The samples will  fail with the following error message: 
```
[error] Error Message: vk::Device::acquireNextImageKHR: ErrorInitializationFailed
[error] Failed when running application Swapchain Images
```
A bit of investigation led me to a failure in the wsi-layer where the usage of vkImportSemaphoreFdKHR when found available fails, leading to the error we see above.
I made a local modification to the layer to bypass that usage and instead use the layer fallback path instead in the following file `[...]/vulkan-wsi-layer/wsi/swapchain_base.cpp:L434` . With that modification, I am able to run successfully the samples in this configuration.

3. Display disconnected
I have observed the same behavior as the one in 2. and required the same local modification to the layer to make it function.

iagoCL:
My current setup is using Swiftshader https://github.com/google/swiftshader "a high-performance CPU-based implementation of the Vulka 1.3 graphics AP" the headless flag seems to work correctly in all tested examples

I can reproduce SaschaWillems crash. Needs more investigation, but seems to be valid use of the API, and only reproducible if you enable ICDs for Nvidia and SwiftShader. As a workaround you can avoid the crash explicitily selecting the valid GPU using the --gpu flag
I also agreed with SaschaWillems to also add some small improvements to the code and extra documentation to better explain VK_EXT_headless_surface, its intention and how it is diferent from rendering without swapchain.
I have been a bit busy this last month, but hopefully I will submit a new PR with some improvements soon.

iagoCL:
Change the name of the flag and add some comments to better explain it.
I also added an explanation in the example in the readme.
Based on some feedback I also added a warning when running headless with multiple GPU candidates.
Tested on Windows and Linux using Swiftshader.

SaschaWillems:
Looks like this triggers a CI failure due to copyright headers not being up-to-date. Can you fix that?

iagoCL:
I did not modified those files, so unsure what trigered the copyright failure. 

I just rebased against main, hopefully that fixes the issue

gary-sweet:
I think I said I was going to test this at the last meeting, but it turns out that we don't actually support VK_EXT_headless_surface, so I 'm afraid I can't.

SaschaWillems:
It seems to work when explicitly selecting swiftshader. I don't see any visual output, which I guess is expected.

But I'm still getting a crash when not explicitly selecting swiftshader. I do get the warning first that tells me to explicitly select a gpu, but if wont' do that, I get the same error as above. Is there no way to avoid running with that flag on GPUs that don't have support for it? Or at least catch that error and exit gracefully. For me it's stuck in an endless loop of exceptions.

SaschaWillems:
And does this really fix #314?

iagoCL:
> But I'm still getting a crash when not explicitly selecting swiftshader. I do get the warning first that tells me to explicitly select a gpu, but if wont' do that, I get the same error as above. Is there no way to avoid running with that flag on GPUs that don't have support for it? Or at least catch that error and exit gracefully. For me it's stuck in an endless loop of exceptions.

If we have multiple GPUs, the correct way to check which one supports headless is by calling `vkGetPhysicalDeviceSurfaceSupportKHR` as the current code id doing. This is causing a crash on Nvidia.

I tried to catch and handle this crash more gratefully, but the crash is in the driver, so exceptions did not work, and I did not found a correct way to detect it from application side.

The alternative is to exit after printing the warning, and forcing user to manually select GPU.

> And does this really fix #314?

Not sure.

It mentions `instancing` and `render_passes`. When I mentioned that issue I just tested those issues, and both work correctly with headless after this change.

Now that I read the problem more carefully, I am confused.
At first glance the current Vulkan Samples will only work on devices that have a present queuee. We check this on `get_suitable_gpu` (with or without headless), so it should be imposible for `get_queue_by_present` to fail

So the crash is fixed, but we still have the `get_device().get_queue_by_present(0).wait_idle()` on `submit_frame`. We should remove the wait idle, but that is independent of headless, so maybe it will be better to update #314 or create a new issue for that.

SaschaWillems (approved):
As noted in the last call this PR looks great and fixes the initial problem. As for getting "real" headless support into the framework (samples that don't need any visual output at all) that's something we should do with a separate PR.

gpx1000 (approved):

marty-johnson59:
Per discussion, let's merge this with 2 (w/ Steve's approval)

PR #1101, opened 2024-07-15
FindVulkan fixes. Bump CMake minimum version.

### LINKED ISSUE #1100: cmake_minimum_required insufficient for used features
The root `CMakeLists.txt` specifies `cmake_minimum_required(VERSION 3.16)` right before calling `FindVulkan`. However `FindVulkan`, when finding a framework (such as for iOS), sets `IMPORTED_LOCATION` to the name of the framework directory because that is what ${Vulkan_LIBRARIES} expands to.

```
if(Vulkan_FOUND AND NOT TARGET Vulkan::Vulkan)
    add_library(Vulkan::Vulkan UNKNOWN IMPORTED)
    set_target_properties(Vulkan::Vulkan PROPERTIES
            IMPORTED_LOCATION "${Vulkan_LIBRARIES}"
            INTERFACE_INCLUDE_DIRECTORIES "${Vulkan_INCLUDE_DIRS}")
endif()
```

This is only supported in CMake 3.28+. Prior to that `IMPORTED_LOCATION` had to be set to the library just inside the framework directory. When using a pre 3.28 version of CMake this leads to a totally obscure error from ld

```
ld: file cannot be mmap()ed, errno=22 path=<...>/VulkanSDK/1.3.283.0/iOS/lib/vulkan.framework
```

3.28 is a very recent version of CMake so I think the appropriate fix is to change FindVulkan to handle versions pre and post 3.28.

Furthermore `app/CMakeLists.txt` also sets the minimum version to 3.16 but it uses `XCODE_EMBED_FRAMEWORKS_CODE_SIGN_ON_COPY` which was only introduced in CMake 3.20.

I will submit a PR to fix the above that changes the minimum version to 3.20 in both and fixes FindVulkan in the way described.

Note that many other places in this repo set even earlier minimum versions than 3.16. I do not know if any of those should be bumped.

MarkCallow:
## Description

Fixes 4 issues with FindVulkan:
1. Make it work with CMake 3.20+ as well as the currently required 3.28.
2. Set `CMAKE_FIND_FRAMEWORK` to `FIRST`, the only value for which the searches work.
3. Fix if() to prevent use of a variable that is undefined when seeking for macOS.
4. Remove an errant closing brace.

Bumps the minimum cmake version for `CMakeLists.txt` and `app/CMakeLists.txt` consistent with the CMake features used. In the case of `CMakeLists.txt `the version is bumped to 3.20 for consistency with app and to support use of `cmake_path` which is used in the accompanying fix to `FindVulkan.cmake` to support versions earlier than 3.28.

Fixes 1, 3 and 4 are useful to this project. All fixes are useful to people who are copying FindVulkan into their own projects. No. 2 is especially important to those using building for iOS using vcpkg which sets `CMAKE_FIND_FRAMEWORK` to `LAST`.

Fixes #1100

## General Checklist:

_I have left unchecked items which are not relevant to this change._

## Sample Checklist

### DIFF
--- a/bldsys/cmake/module/FindVulkan.cmake
+++ b/bldsys/cmake/module/FindVulkan.cmake
@@ -287,7 +287,7 @@ else()
             "$ENV{VULKAN_SDK}/lib"
     )
 endif()
-if(APPLE AND DEFINED ENV{VULKAN_SDK})
+if(APPLE AND DEFINED Vulkan_Target_SDK)
     list(APPEND _Vulkan_hint_include_search_paths
             "${Vulkan_Target_SDK}/macOS/include"
     )
@@ -301,7 +301,7 @@ if(APPLE AND DEFINED ENV{VULKAN_SDK})
         )
     else()
         list(APPEND _Vulkan_hint_library_search_paths
-                "${Vulkan_Target_SDK}}/lib"
+                "${Vulkan_Target_SDK}/lib"
         )
     endif()
 endif()
--- a/bldsys/cmake/module/FindVulkan.cmake
+++ b/bldsys/cmake/module/FindVulkan.cmake
@@ -668,7 +668,7 @@ find_package_handle_standard_args(Vulkan
 
 if(Vulkan_FOUND AND NOT TARGET Vulkan::Vulkan)
     add_library(Vulkan::Vulkan UNKNOWN IMPORTED)
-    cmake_path(GET Vulkan_LIBRARIES EXTENSION _Vulkan_lib_extension)
+    get_filename_component(_Vulkan_lib_extension ${Vulkan_LIBRARIES} LAST_EXT)
     if(_Vulkan_lib_extension STREQUAL ".framework" AND CMAKE_VERSION VERSION_LESS 3.28)
         set_target_properties(Vulkan::Vulkan PROPERTIES
                 # Prior to 3.28 must reference library just inside the framework.
--- a/bldsys/cmake/module/FindVulkan.cmake
+++ b/bldsys/cmake/module/FindVulkan.cmake
@@ -536,8 +536,6 @@ endif()
 if(DEFINED _Vulkan_saved_cmake_find_framework)
     set(CMAKE_FIND_FRAMEWORK ${_Vulkan_saved_cmake_find_framework})
     unset(_Vulkan_saved_cmake_find_framework)
-else()
-    unset(CMAKE_FIND_FRAMEWORK)
 endif()
 
 if(Vulkan_GLSLC_EXECUTABLE)
--- a/bldsys/cmake/module/FindVulkan.cmake
+++ b/bldsys/cmake/module/FindVulkan.cmake
@@ -240,9 +240,9 @@ endif()
 # are ever included in the SDK for macOS, the search mechanism will need
 # revisiting.
 if(DEFINED CMAKE_FIND_FRAMEWORK)
-    set( _Vulkan_saved_cmake_find_framework ${CMAKE_FIND_FRAMEWORK} )
+    set(_Vulkan_saved_cmake_find_framework ${CMAKE_FIND_FRAMEWORK})
+    set(CMAKE_FIND_FRAMEWORK FIRST)
 endif()
-set( CMAKE_FIND_FRAMEWORK FIRST )
 
 if(IOS)
     get_filename_component(Vulkan_Target_SDK "$ENV{VULKAN_SDK}/.." REALPATH)
@@ -534,7 +534,7 @@ if (dxc IN_LIST Vulkan_FIND_COMPONENTS)
 endif()
 
 if(DEFINED _Vulkan_saved_cmake_find_framework)
-    set(CMAKE_FIND_FRAMEWORK ${saved_cmake_find_framework})
+    set(CMAKE_FIND_FRAMEWORK ${_Vulkan_saved_cmake_find_framework})
     unset(_Vulkan_saved_cmake_find_framework)
 else()
     unset(CMAKE_FIND_FRAMEWORK)
--- a/bldsys/cmake/module/FindVulkan.cmake
+++ b/bldsys/cmake/module/FindVulkan.cmake
@@ -318,7 +318,7 @@ find_library(Vulkan_LIBRARY
         HINTS
         ${_Vulkan_hint_library_search_paths}
 )
-message(STATUS vulkan_library ${Vulkan_LIBRARY} search paths ${_Vulkan_hint_library_search_paths})
+message(STATUS "vulkan_library ${Vulkan_LIBRARY} search paths ${_Vulkan_hint_library_search_paths}")
 mark_as_advanced(Vulkan_LIBRARY)
 
 find_library(Vulkan_Layer_API_DUMP
@@ -652,9 +652,18 @@ find_package_handle_standard_args(Vulkan
 
 if(Vulkan_FOUND AND NOT TARGET Vulkan::Vulkan)
     add_library(Vulkan::Vulkan UNKNOWN IMPORTED)
+    cmake_path(GET Vulkan_LIBRARIES EXTENSION _Vulkan_lib_extension)
+    if(_Vulkan_lib_extension STREQUAL ".framework" AND CMAKE_VERSION VERSION_LESS 3.28)
+        set_target_properties(Vulkan::Vulkan PROPERTIES
+                # Prior to 3.28 must reference library just inside the framework.
+                IMPORTED_LOCATION "${Vulkan_LIBRARIES}/vulkan")
+    else()
+        set_target_properties(Vulkan::Vulkan PROPERTIES
+                IMPORTED_LOCATION "${Vulkan_LIBRARIES}")
+    endif()
     set_target_properties(Vulkan::Vulkan PROPERTIES
-            IMPORTED_LOCATION "${Vulkan_LIBRARIES}"
             INTERFACE_INCLUDE_DIRECTORIES "${Vulkan_INCLUDE_DIRS}")
+    unset(_Vulkan_lib_extension)
 endif()
 
 if(Vulkan_FOUND AND NOT TARGET Vulkan::Headers)
--- a/bldsys/cmake/module/FindVulkan.cmake
+++ b/bldsys/cmake/module/FindVulkan.cmake
@@ -233,6 +233,17 @@ if("FATAL_ERROR" IN_LIST Vulkan_FIND_COMPONENTS)
     list(REMOVE_ITEM Vulkan_FIND_COMPONENTS "FATAL_ERROR")
 endif()
 
+# FindVulkan only works correctly with the default CMAKE_FIND_FRAMEWORK
+# value: FIRST. If LAST it will find the macOS dylibs instead of the iOS
+# frameworks when IOS is true. If ALWAYS it will fail to find the macOS
+# dylibs. If NEVER it will fail to find the iOS frameworks. If frameworks
+# are ever included in the SDK for macOS, the search mechanism will need
+# revisiting.
+if(DEFINED CMAKE_FIND_FRAMEWORK)
+    set( _Vulkan_saved_cmake_find_framework ${CMAKE_FIND_FRAMEWORK} )
+endif()
+set( CMAKE_FIND_FRAMEWORK FIRST )
+
 if(IOS)
     get_filename_component(Vulkan_Target_SDK "$ENV{VULKAN_SDK}/.." REALPATH)
     list(APPEND CMAKE_FRAMEWORK_PATH "${Vulkan_Target_SDK}/iOS/lib")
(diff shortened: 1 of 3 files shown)


SaschaWillems (dismissed):
LGTM. Thanks for this PR, very much appreciated.

MarkCallow:
Just spotted a completely unnecessary `unset` in my changes which I have removed in https://github.com/KhronosGroup/Vulkan-Samples/pull/1101/commits/11f46534cf4cf8e8f18017e62e4861c09e18c09d. Very sorry this is post your review @SaschaWillems thus invalidating it.

SaschaWillems (dismissed):

bldsys/cmake/module/FindVulkan.cmake
```diff
@@ -233,6 +233,17 @@ if("FATAL_ERROR" IN_LIST Vulkan_FIND_COMPONENTS)
     list(REMOVE_ITEM Vulkan_FIND_COMPONENTS "FATAL_ERROR")
 endif()
 
+# FindVulkan only works correctly with the default CMAKE_FIND_FRAMEWORK
+# value: FIRST. If LAST it will find the macOS dylibs instead of the iOS
+# frameworks when IOS is true. If ALWAYS it will fail to find the macOS
+# dylibs. If NEVER it will fail to find the iOS frameworks. If frameworks
+# are ever included in the SDK for macOS, the search mechanism will need
+# revisiting.
```
gpx1000, line 241: I'm not sure this is correct.  When I run VULKAN_SDK/ios/setup_env.sh the current pathing seems to be correct.  Both CI and locally in my mac, I'm seeing the build get created successfully with current setup.  Could you confirm the correction here is needed?

gpx1000, line 241: Not sure this is correct.  When I run VULKAN_SDK/ios/setup_env.sh the find_vulkan.cmake resolves correctly as currently is deployed.  The CI environment also builds/resolves correctly and locally, it resolves correctly.  Could you confirm that your changes are necessary?

MarkCallow: In the case of iOS, FindVulkan sets the library search path it uses to `$VULKAN_SDK/lib; $VULKAN_SDK/../iOS/lib`. If you use VULKAN_SDK/ios/setup_env.sh to set VULKAN_SDK then it ends up with the search path `VULKAN_SDK/iOS/lib; VULKAN_SDK/iOS/lib`. In that case it will not be affected by the setting of CMAKE_FIND_FRAMEWORK, _unless_ you have installed the macOS libraries globally.<sup>1</sup>

However if you use VULKAN_SDK/setup-env.sh to set your environment then `$VULKAN_SDK` will be VULKAN_SDK/macOS and FindVulkan will use the search path `VULKAN_SDK/macOS/lib; VULKAN_SDK/iOS/lib`. CMAKE_FIND_FRAMEWORK is always in play in this case.

For anyone building for both macOS and iOS having to have different values of VULKAN_SDK does not work. It needs to be fixed. My choice is to use VULKAN_SDK/setup-env.sh as everything except the libraries and `share/vulkan` is common to both. This results in the searcy path in the previous paragraph. Setting CMAKE_FIND_FRAMEWORK ensures the correct library is found. It does not break the case when VULKAN_SDK/ios/setup_env.sh was used. In fact it makes it more robust by guaranteeing to find frameworks first<sup>1</sup>. For `share/vulkan` my cmake code uses `${Vulkan_Target_SDK}/${CMAKE_SYSTEM_NAME}/share/vulkan` to ensure it gets the correct files. `${Vulkan_Target_SDK}` is set by FindVulkan to the root of the SDK.

<sup>1</sup> CMake goes through its entire search sequence looking for a dylib or framework as indicated by CMAKE_FIND_FRAMEWORK. If it doesn't find anything it repeats the search from the beginning of the sequence. 
Thus, if CMAKE_FIND_FRAMEWORK is LAST it will find a dylib that appears anywhere in the sequence before finding a framework. So it will find /usr/local/lib/libvulkan.dylib if global SDK installation was chosen.

gpx1000: If you're building with macOS and iOS, then you can use source VULKAN_SDK/iOS/setup_env.sh and source VULKAN_SDK/macOS/setup_env.sh  in the build process.
This is kinda a mess sadly and will get worse when we get watchOS or visionOS also setup this way.  However, this is what the recommended path is towards getting it to work.  I would prefer to use your solution, but we need to demonstrate the "official" solution path.  We'll cover this in the samples meeting today and see what the group thinks of this.

MarkCallow: > If you're building with macOS and iOS, then you can use source VULKAN_SDK/iOS/setup_env.sh and source VULKAN_SDK/macOS/setup_env.sh in the build process.

You can't if you are using the CMake GUI and Xcode IDE which I do. A single value of VULKAN_SDK is essential. It is easy enough for FindVulkan.cmake to do the right thing with knowledge of where in the SDK a single $VULKAN_SDK points.

marty-johnson59:
Hi @MarkCallow , we discussed this MR on the Samples call today and have reservations about merging this w/o some changes.  There appears to be 2 main items covered here: 1) bumping the cmake version number; 2) changes related to FindVulkan.  The first item seems to only be required for iOS builds - but it is likely to cause problems for other platforms (especially Android) where folks may have good reasons to stay with the older version.  We suggest splitting this change out into a separate PR and scoping it to only iOS platforms.  We aren't quite sure about the FindVulkan changes - but we think it makes sense to sync up with LunarG on this - they have plans for future versions that might impact our directions here.  @gpx1000 is the most familiar with this on the team and has offered to help - Steve, please chime in as needed.  Thanks!

MarkCallow:
> The first item seems to only be required for iOS builds - but it is likely to cause problems for other platforms (especially Android) where folks may have good reasons to stay with the older version. We suggest splitting this change out into a separate PR and scoping it to only iOS platforms.

How to scope to only iOS platforms? As far as I know `cmake_minimum_required` must appear before the `project` command and I know that the IOS variable is only set _after_ the project() command. Perhaps an explicit check of the version after `project`, `if(IOS)`with a `message(FATAL_ERROR ...)`. Suggestions welcome.

What is the minimum CMake version for which support is desired? There is no question that the `XCODE_EMBED` items in `app/CMakeLists.txt`require CMake 3.20. My question above relates to this file. For `CMakeLists.txt`, the workaround in `FindVulkan.cmake` for versions earlier than 3.28 can be modified to use the deprecated `get_filename_component` instead of `cmake_path` thus removing the need to up its minimum version to 3.20.

> We aren't quite sure about the FindVulkan changes - but we think it makes sense to sync up with LunarG on this - they have plans for future versions that might impact our directions here. @gpx1000 is the most familiar with this on the team and has offered to help - Steve, please chime in as needed. Thanks!

I intend to sync with LunarG but currently FindVulkan is not part of the Vulkan SDK. @gpx1000's blog on developing an iOS app with the Vulkan SDK says to copy the file from here. Thus the problems need to be fixed here now not at some future time. The fixes I have provided do not break any other usage. The code to ensure CMAKE_FIND_FRAMEWORK is FIRST within FindVulkan does not affect the value outside FindVulkan. The other fixes are vanilla bug fixes.

I see no reason to split this into 2 PRs.

bldsys/cmake/module/FindVulkan.cmake
gpx1000: FindVulkan.cmake like other Find*.cmake solutions are meant to be project specific the way cmake architecture works.  So if the standard method doesn't work locally, it's totally fine to make changes for your work setup.

The reason we're pushing back here is because this is meant to be a standard solution.  If we make changes to FindVulkan.cmake then it gets those same changes to ALL other platforms.  *maybe* something that makes sense is to pickup a define from an option that is passed to cmake and if that option is set; have it update the environment path variable (or even check both the environment variable and the define?).  However, we can't make changes *just* for one platform without testing all platforms.  This requirement also includes testing on distributions that don't necessarily come from LunarG as the pathing wont always be the same; but the environment variables will be (that's why there's the environment variables).

To be blunt, this problem goes away when the config variation of find_package gets included with the SDK.  That's a larger task that will require much more than changes to the samples repo.  However, eventually FIndVulkan.cmake will no longer exist and vulkanconfig.cmake will be included in the SDK (and hopefully all SDKs).

gpx1000:
> How to scope to only iOS platforms? As far as I know `cmake_minimum_required` must appear before the `project` command and I know that the IOS variable is only set _after_ the project() command. Perhaps an explicit check of the version after `project`, `if(IOS)`with a `message(FATAL_ERROR ...)`. Suggestions welcome.
> 
> What is the minimum CMake version for which support is desired? There is no question that the `XCODE_EMBED` items in `app/CMakeLists.txt`require CMake 3.20. My question above relates to this file. For `CMakeLists.txt`, the workaround in `FindVulkan.cmake` for versions earlier than 3.28 can be modified to use the deprecated `get_filename_component` instead of `cmake_path` thus removing the need to up its minimum version to 3.20.
> 

I think you're right about the idea of if(IOS and version less than 3.20 in findVulkan.cmake with an assert saying minimum cmake of 3.20 is required is best.  That way we're keeping the check to the file we're recommending people copy from.  It also keeps the project wide changes to where they currently are for all other platforms.  This reduces the amount of testing we need to do.

> I intend to sync with LunarG but currently FindVulkan is not part of the Vulkan SDK. @gpx1000's blog on developing an iOS app with the Vulkan SDK says to copy the file from here. Thus the problems need to be fixed here now not at some future time. The fixes I have provided do not break any other usage. The code to ensure CMAKE_FIND_FRAMEWORK is FIRST within FindVulkan does not affect the value outside FindVulkan. The other fixes are vanilla bug fixes.
> 
> I see no reason to split this into 2 PRs.

FindVulkan.cmake should hopefully never be included in the SDK.  FindVulkan.cmake is included with cmake and that is woefully out of date; and cmake is trying to remove it.  Ideally, we'd transition to vulkanconfig.cmake and include that with the SDK so it's not a project only solution.  However, I think FindVulkan.cmake in the current iteration should be thought of as a template.
I think if you do the path I mentioned above of leaving the version bump off and inlcude an assert in FindVulkan.cmake when it's being used by cmake less than 3.20, then one PR makes sense.  Otherwise, the version bump would benefit from being in a separate PR so we can make sure it works in all platforms by itself as that is a fairly big change.

MarkCallow:
@gpx1000 I think you misunderstood what I wrote. By modifying my FindVulkan pre-CMake 3.28 workaround to use `get_filename_component` I can revert the cmake_required_minimum in the root `CMakeLists.txt` back to 3.16.

It is app/CMakeLists.txt that needs the fatal error if IOS and CMake version < 3.20 in order to remove the version bump there. I'll update the PR tomorrow.

> Otherwise, the version bump would benefit from being in a separate PR so we can make sure it works in all platforms by itself

Isn't CI checking all desired supported platforms? If so they are all already passing with the PR as is.

> as that is a fairly big change.

The 2 minor bug fixes in FindVulkan aside, the change there only affects searches on iOS and the Vulkan::Vulkan target on iOS as that is the only SDK with frameworks. It will not affect other platforms or tests of the version bumps on them. But this is moot I will remove them.

Please ask the samples team to consider bumping the minimum version. The ones you have now are long in the tooth and as incident has shown people are starting to use new features without realizing they are not in the minimum required version.

gpx1000:
> @gpx1000 I think you misunderstood what I wrote. By modifying my FindVulkan pre-CMake 3.28 workaround to use `get_filename_component` I can revert the cmake_required_minimum in the root `CMakeLists.txt` back to 3.16.
> 
> It is app/CMakeLists.txt that needs the fatal error if IOS and CMake version < 3.20 in order to remove the version bump there. I'll update the PR tomorrow.
> 

Fair enough; thanks for doing the update!  I think it makes a lot of sense and communicates the limitation correctly to end users.

> 
> Isn't CI checking all desired supported platforms? If so they are all already passing with the PR as is.
> 

Sadly no.  Vulkan is supported on a LOT of platforms.  We don't even have full coverage of all supported Apple desktop platforms ;-) (M2 hardware isn't covered as an example).  While CI does provide a measure of strong indication things work in all platforms to be fair, there's plenty of platforms ecosystems that can't easily be covered for various reasons.  It's a best effort which is meant to cover the most standard way of doing things (i.e. getting the SDK from LunarG instead of other locations/distributions and installing system level cmake etc).  

> Please ask the samples team to consider bumping the minimum version. The ones you have now are long in the tooth and as incident has shown people are starting to use new features without realizing they are not in the minimum required version.

Yep, we do need to look at bumping the minimum version; it is pretty old.  It would make a lot of sense to bring it up to the minimums each platform ships with for cmake; ensure that everyone can get access to them; then test it to ensure it doesn't break anything etc.

MarkCallow:
I reverted the cmake version bumps in https://github.com/KhronosGroup/Vulkan-Samples/pull/1101/commits/dd24aac589d5bb7a618bdf8a4f3354abb147eb8a.

gpx1000 (approved):
Thanks this looks good to me !

SaschaWillems (approved):

gary-sweet (approved):

PR #1120, opened 2024-07-31
Fixes to iOS render loop and HUD overlay.

### LINKED ISSUE #1121: Samples appear different on iOS when VkClearValue::color value contains zero alpha
Many samples do not specify a value in `VkClearValue::color[3]`, which leaves the alpha value for the clear color as `0.0`.

For most platforms, this does not matter, but on iOS, the platform does honour the alpha value when blending the rendering onto the platform view. The result is that the background color for the sample displays whatever color the view is showing.

Unless it will somehow interfere with other platforms, I recommend that all samples should specify a value of `1.0` for the alpha value in `VkClearValue`.

I was going to introduce a PR for this, but I have no way of testing anything other than Apple platforms, and I did not want to introduce issues on other platforms, in case leaving alpha at `0.0` was intentional for some platforms.

### LINKED ISSUE #1131: Regression in Platform::main_loop_frame() that causes all samples to only render one frame on iOS
`Platform::main_loop_frame()` has code at the beginning:

```
	if (!app_requested())
	{
		return ExitCode::NoSample;
	}
```

That halts execution after the first frame, because `Platform::start_app()` sets `requested_app` to `nullptr`, and `Platform::main_loop_frame()` is called on every frame.

Removing the code above fixes the issue.

I've submitted PR #1120 that fixes this, and also consolidates `Platform::main_loop_frame()` into `Platform::main_loop()` to eliminate the difficult-to-maintain code-duplication. However, I'm having trouble passing the CI code formatting tests on that PR, so in case that PR doesn't make it in, or if there is a better solution, I'm adding this here, so someone more familiar with this environment can fix it.

billhollings:
* Refactor Platform::main_loop_frame() and main_loop(),
  to reduce fragility, and avoid unnecessary code duplication.

* Avoid initial app_requested() test in main_loop_frame(),
  as it was causing looping to stop after one frame.

* Since main_loop() is not a virtual function, remove
  declaration comment indicating it can be overridden.

* Set iOS view to have a black background, to ensure 
  transparency in HUD overlay displays in full on device.
  
* Fix rendering scaling in IosWindow::get_dpi_factor().
  Previous value was causing HUD overlay to render magnitudes too large.
  
* Add macOS files to .gitignore.

I created this from a branch on this repo instead of a separate fork, because I was not expecting to need to make a fix.

Fixes #1121
Fixes #1131

## General Checklist:

## Sample Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,10 +9,12 @@ tests/system_test/tmp
 output/*
 .mypy_cache
 tags
-.DS_Store
 
 # vim backup and temp files
 *~
 .*.sw*
 .sw*
 .tracy
+
+# macOS
+.DS_Store
--- a/framework/platform/ios/ios_window.mm
+++ b/framework/platform/ios/ios_window.mm
@@ -69,7 +69,7 @@ void IosWindow::close()
 
 float IosWindow::get_dpi_factor() const
 {
-    return [[UIScreen mainScreen] nativeScale] * (([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad) ? 132 : 163);
+	return 1.0;
 }
 
 std::vector<const char *> IosWindow::get_required_surface_extensions() const
--- a/framework/platform/platform.cpp
+++ b/framework/platform/platform.cpp
@@ -121,55 +121,52 @@ ExitCode Platform::initialize(const std::vector<Plugin *> &plugins)
 
 ExitCode Platform::main_loop_frame()
 {
-	if (!app_requested())
+	try
 	{
-		return ExitCode::NoSample;
-	}
-
-	if (!window->should_close() && !close_requested)
-	{
-		try
+		// Load the requested app
+		if (app_requested())
 		{
-			// Load the requested app
-			if (app_requested())
+			if (!start_app())
 			{
-				if (!start_app())
-				{
-					throw std::runtime_error("Failed to load requested application");
-				}
-
-				// Compensate for load times of the app by rendering the first frame pre-emptively
-				timer.tick<Timer::Seconds>();
-				active_app->update(0.01667f);
+				throw std::runtime_error("Failed to load requested application");
 			}
 
-			update();
-
-			if (active_app && active_app->should_close())
-			{
-				std::string id = active_app->get_name();
-				on_app_close(id);
-				active_app->finish();
-			}
-
-			window->process_events();
+			// Compensate for load times of the app by rendering the first frame pre-emptively
+			timer.tick<Timer::Seconds>();
+			active_app->update(0.01667f);
 		}
-		catch (std::exception &e)
+
+		if (!active_app)
 		{
-			LOGE("Error Message: {}", e.what());
-			LOGE("Failed when running application {}", active_app->get_name());
+			return ExitCode::NoSample;
+		}
 
-			on_app_error(active_app->get_name());
+		update();
 
-			if (app_requested())
-			{
-				LOGI("Attempting to load next application");
-			}
-			else
-			{
-				set_last_error(e.what());
-				return ExitCode::FatalError;
-			}
+		if (active_app->should_close())
+		{
+			std::string id = active_app->get_name();
+			on_app_close(id);
+			active_app->finish();
+		}
+
+		window->process_events();
+	}
+	catch (std::exception &e)
+	{
+		LOGE("Error Message: {}", e.what());
+		LOGE("Failed when running application {}", active_app->get_name());
+
+		on_app_error(active_app->get_name());
+
+		if (app_requested())
+		{
+			LOGI("Attempting to load next application");
+		}
+		else
+		{
+			set_last_error(e.what());
+			return ExitCode::FatalError;
 		}
 	}
 
@@ -183,54 +180,13 @@ ExitCode Platform::main_loop()
 		return ExitCode::NoSample;
 	}
 
-	while (!window->should_close() && !close_requested)
+	ExitCode exit_code = ExitCode::Success;
+	while ((exit_code == ExitCode::Success) && !window->should_close() && !close_requested)
 	{
-		try
-		{
-			// Load the requested app
-			if (app_requested())
-			{
-				if (!start_app())
-				{
-					throw std::runtime_error("Failed to load requested application");
-				}
-
-				// Compensate for load times of the app by rendering the first frame pre-emptively
-				timer.tick<Timer::Seconds>();
-				active_app->update(0.01667f);
-			}
-
-			update();
-
-			if (active_app && active_app->should_close())
-			{
-				std::string id = active_app->get_name();
-				on_app_close(id);
-				active_app->finish();
-			}
-
-			window->process_events();
-		}
-		catch (std::exception &e)
-		{
-			LOGE("Error Message: {}", e.what());
-			LOGE("Failed when running application {}", active_app->get_name());
-
-			on_app_error(active_app->get_name());
-
-			if (app_requested())
-			{
-				LOGI("Attempting to load next application");
-			}
-			else
-			{
-				set_last_error(e.what());
-				return ExitCode::FatalError;
-			}
-		}
+		exit_code = main_loop_frame();
 	}
 
-	return ExitCode::Success;
+	return exit_code;
 }
 
 void Platform::update()
--- a/framework/platform/platform.h
+++ b/framework/platform/platform.h
@@ -64,11 +64,17 @@ class Platform
 	virtual ExitCode initialize(const std::vector<Plugin *> &plugins);
 
 	/**
-	 * @brief Handles the main loop of the platform
-	 * This should be overriden if a platform requires a specific main loop setup.
+	 * @brief Handles the main update and render loop
 	 * @return An exit code representing the outcome of the loop
 	 */
 	ExitCode main_loop();
+
+	/**
+	 * @brief Handles the update and render of a frame.
+	 * Called either from main_loop(), or from a platform-specific
+	 * frame-looping mechanism, typically tied to platform screen refeshes.
+	 * @return An exit code representing the outcome of the loop
+	 */
 	ExitCode main_loop_frame();
 
 	/**
--- a/app/ios/Storyboard.storyboard
+++ b/app/ios/Storyboard.storyboard
@@ -34,7 +34,7 @@ limitations under the License.
                         <rect key="frame" x="0.0" y="0.0" width="393" height="852"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <viewLayoutGuide key="safeArea" id="vDu-zF-Fre"/>
-                        <color key="backgroundColor" systemColor="systemBackgroundColor"/>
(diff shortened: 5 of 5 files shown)


framework/platform/platform.h
```diff
@@ -130,6 +128,8 @@ class Platform
 	static const uint32_t MIN_WINDOW_HEIGHT;
 
   protected:
+	virtual bool uses_internal_looping() { return true; }
```
asuessenbach: Maybe make this function `const`?

framework/platform/ios/ios_platform.h
```diff
@@ -33,6 +33,7 @@ class IosPlatform : public Platform
     void* view;
 
 protected:
+	bool uses_internal_looping() override { return false; }
```
asuessenbach: What does this `internal_looping` mean? Why is it different on IOS than everywhere else?

asuessenbach:
You should at least resolve the failing Quality Check.

framework/platform/platform.h
billhollings: Fixed.

framework/platform/ios/ios_platform.h
billhollings: On other platforms, `main_loop()` is called once. On iOS, `main_loop()` is called on each frame by the platform display synchronization loop. See `ViewController::renderLoop`. This is generally how rendering is done on iOS.

This was apparently why `Platform::main_loop_frame()` was added separately from `Platform::main_loop()`, but this code replication is now showing it's maintainability problems in that maintaining one loop code and copying into the other loop code will sometimes cause unexpected behaviour.

Code had recently been added to the top of `main_loop_frame()` that assumed it would only be called once, which then broke on iOS.

Another option would be to separate the updating/rendering code inside the loop, from `main_loop()`, so it can be called separately on iOS. I decided a loop check would be a more minimal change, but I can easily take that approach, since it's probably a better factoring.

billhollings:
> You should at least resolve the failing Quality Check.

If tried many times.

1. For some reason, `clang-format` works slightly differently on macOS than on other platforms, so running `./scripts/clang_format.py main` ends up changing more than expected, and breaks the CI test here.
2. The [CI error messages](https://github.com/KhronosGroup/Vulkan-Samples/actions/runs/10461719961/job/28970635867?pr=1120) are incomprehensible to me, so I'm not able to decipher what it's complaining about, in order to simply fix it manually.

gpx1000:
https://github.com/KhronosGroup/Vulkan-Samples/actions/runs/10461719961?pr=1120

If you scroll to the bottom you can grab the diff that can be directly run and hopefully fix the issue.

framework/platform/ios/ios_platform.h
asuessenbach: ok... what do you think about having the one-time code in `main_loop()` and then calling into `main_loop_frame()` in a loop? Would that work?

asuessenbach:
> For some reason, clang-format works slightly differently on macOS than on other platforms

Indeed, it should remove white spaces at the end of a line, and it should replace spaces with tabs.
This patch shows the difference I would get when running clang-format: [0001-formatting.patch](https://github.com/user-attachments/files/16670591/0001-formatting.patch)

framework/platform/ios/ios_platform.h
billhollings: Good idea! I've implemented this now.

billhollings:
@gpx1000 @asuessenbach 

Thanks for the help on code formatting. I've got this formatting correctly now.

Plus I've added a couple of other iOS fixes, and updated the title and description above.

framework/platform/ios/ios_window.mm
```diff
@@ -69,7 +69,7 @@
 
 float IosWindow::get_dpi_factor() const
 {
-    return [[UIScreen mainScreen] nativeScale] * (([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad) ? 132 : 163);
+	return 1.0;
```
gpx1000, line 72: This is almost nothing more than a question; Is this considered the "right way" to get the dpi for text for all iOS devices?  It seems that a 1-1 ratio might be wrong in some cases?  I guess not given the displays themselves are small and the pixel pitch is small.

gpx1000 (dismissed):

framework/platform/ios/ios_window.mm
SRSaunders: I think @billhollings and I found the same issue that was causing UI Overlay scaling problems on iOS.  The way I solved it was the following.  This works for real iOS devices and the iOS Simulator:

In ViewController.mm:

```
-->	self.vulkan_view.contentScaleFactor = UIScreen.mainScreen.nativeScale;
        context = create_platform_context((int)args.count, const_cast<char**>(argv));
        _context = (vkb::IosPlatformContext*)context.get();
        _context->view = self.vulkan_view;
        _code = (vkb::ExitCode)platform_main(*context);
        uint32_t fps = 60;
```

In ios_window.mm

`return [[UIScreen mainScreen] nativeScale];`

framework/platform/platform.cpp
```diff
@@ -185,49 +177,7 @@ ExitCode Platform::main_loop()
...
-			{
-				LOGI("Attempting to load next application");
-			}
-			else
-			{
-				set_last_error(e.what());
-				return ExitCode::FatalError;
-			}
-		}
+		main_loop_frame();
```
asuessenbach: You're missing any failure exit code from main_loop_frame.
Maybe something like
```
ExitCode Platform::main_loop()
{
	ExitCode exit_code = ExitCode::Success;
	if (!app_requested())
	{
		exit_code = ExitCode::NoSample;
	}

	while ((exit_code == ExitCode::Success) && !window->should_close() && !close_requested)
	{
		exit_code = main_loop_frame();
	}

	return exit_code;
}

```

framework/platform/ios/ios_window.mm
SRSaunders: Looking at this again, I think the following is a bit more semantically correct:

In **ViewController.mm**:

```
...
-->	self.vulkan_view.contentScaleFactor = UIScreen.mainScreen.nativeScale;
        context = create_platform_context((int)args.count, const_cast<char**>(argv));
        _context = (vkb::IosPlatformContext*)context.get();
...
```

In **ios_window.mm**:

```
float IosWindow::get_dpi_factor() const
{
	return 1.0;
}

float IosWindow::get_content_scale_factor() const
{
	return [[UIScreen mainScreen] nativeScale];
}
```

I think @billhollings is correct to return 1.0 for `get_dpi_factor()`.  My changes for setting and getting the view's `contentScaleFactor` are also okay, but they offset each other for the imgui UI Overlay and the code would still work without adding this part.

framework/platform/platform.cpp
billhollings: D'oh! That's my bad. Thanks for catching it. Fixed now.

I've left the initial `!app_requested()` test as an immediate return, to clarify it's use as a one-time pre-check, and keep it disentangled from the looping logic.

framework/platform/ios/ios_window.mm
billhollings: > I think @billhollings is correct to return 1.0 for `get_dpi_factor()`. My changes for setting and getting the view's `contentScaleFactor` are also okay, but they offset each other for the imgui UI Overlay and the code would still work without adding this part.

Okay. How about I leave it as it is now (returning `1.0`), and adding additional logic around screen attributes can be left for another PR?

It is currently displaying nicely on both iPhone and iPad. On an iPad, the app display Idiom seems to stay as iPhone, so it uses the standard iPad PiP when an app is only designed for iPhone. As a result, it looks the same on both iPhone and iPad.

SRSaunders: > How about I leave it as it is now (returning 1.0), and adding additional logic around screen attributes can be left for another PR?

That's fine for now, but the iOS view `contentScaleFactor` and `get_content_scale_factor()` need to be setup for other samples in the project (e.g. _swapchain_recreation_).  I can submit these changes after your PR is merged.

I will work on your review request tomorrow.  There's an additional error path that needs to be handled in `main_loop_frame()` when the sample is not properly specified (e.g. cmd line typo). Should be a simple fix.

framework/platform/platform.cpp
```diff
@@ -121,55 +121,47 @@ ExitCode Platform::initialize(const std::vector<Plugin *> &plugins)
...
+			active_app->update(0.01667f);
+		}
 
-			if (active_app && active_app->should_close())
-			{
-				std::string id = active_app->get_name();
-				on_app_close(id);
-				active_app->finish();
-			}
+		update();
```
SRSaunders, line 144: To handle the case when `app_requested()` returns false right at the start (e.g. due to a typo in the sample name), you should probably do the following:

```
...
			active_app->update(0.01667f);
		}

-->		if (!active_app)
-->		{
-->			return ExitCode::NoSample;
-->		}

		update();

-->		if (active_app->should_close())
		{
			std::string id = active_app->get_name();
...
```

This will prevent things from crashing on an undefined `active_app` inside `update()`.  Note this is not an issue with non-iOS builds since an active sample is already validated in `platform_main()`.  However, for iOS the return code is not checked in `ViewController::viewDidLoad` and that is the root cause.  Nonetheless, the `main_loop_frame()` code should be tolerant of this, and at least not crash inside `update()` with an undefined `active_app`.  The iOS `viewDidLoad` code could be fixed later, along with a way to actually terminate the iOS app if the sample never loaded (or a batch run finishes).

.gitignore
```diff
@@ -16,3 +16,6 @@ tags
 .*.sw*
 .sw*
 .tracy
+
+# macOS
+.DS_Store
```
SRSaunders, line 20: .DS_Store now appears twice in the .gitignore file.  You could either remove the earlier instance, or revert your change - either approach is fine by me.

SRSaunders (changes requested):

asuessenbach (approved):
By code inspecting and running on Windows: LGTM.

gpx1000 (approved):

SRSaunders (approved):
LGTM.  Did full batch runs for iOS + macOS and both `main_loop_frame()` and `main_loop()` worked correctly.  Still need fixes for non-success return codes in iOS `viewDidLoad()` and `renderLoop()`, but these can be dealt with in a future PR.

billhollings:
> LGTM. Did full batch runs for iOS + macOS and both `main_loop_frame()` and `main_loop()` worked correctly. Still need fixes for non-success return codes in iOS `viewDidLoad()` and `renderLoop()`, but these can be dealt with in a future PR.

@SRSaunders Would you mind giving this a formal approval, so it can be merged, please?

Once the approvals are in, what is the process for merging? Do I do that? Or is there a designated person responsible for merging?

gpx1000:
We usually ask @marty-johnson59 to handle the merging.  We usually merge on 3; but Marty, I think this is safe to merge on 2.

SRSaunders:
> @SRSaunders Would you mind giving this a formal approval, so it can be merged, please?

@billhollings I don't have approval rights for this project.  I am happy to provide feedback, but I cannot mark the PR as ready for merge.  As you see above, @gpx1000 has that ability.

marty-johnson59:
Thanks.  Given this is Mac (we have limited folks who can test), I'm happy to merge w/ 2.  We can revert if needed.

PR #1130, opened 2024-08-16
Fix hlsl for ray tracing samples.

gpx1000:
Fix the build for hlsl in ray tracing samples.  They require SPV_KHR_ray_query extension be set in order to build.


SaschaWillems:
Not sure about removing the precompiled shader files. We deliberately added them so people that don't have access to DXC can still run our samples. We wan't to do the same with GLSL shaders once we have offline compilation up and running.

gpx1000:
Hmmm.  We should probably create a cmake rule to not touch them unless their shaders have changed and dxc is detected then.  Otherwise git is going to think they need updating when they don't

SaschaWillems:
Yeah, that's correct. I actually though that was the case with Mobica's PR that added DXC compilation :/

gpx1000:
Okay, I'll try to craft something for this PR that handles it then.

gpx1000:
I switched it from a function call to cmake tasks so cmake correctly will detect when updates to the shader source code should have a shader recompiled.   We might also setup a find_dxc.cmake package file so this more generic solution could be found anywhere and allow for a compile SPIR-V function for cmake that can be packaged with DXC.

SaschaWillems (approved):

gary-sweet (approved):

marty-johnson59:
I'll merge this on 2 approvals as it prevents running the Batch script...LMK if we need to revert

PR #1133, opened 2024-08-20
shader_debugprintf: Use API 1.2 for SDKs <= 1.3.290, API 1.1 for SDKs > 1.3.290

### LINKED ISSUE #1132: shader_debugprintf fails starting with Vulkan SDK 1.3.290
The _shader_debugprintf_ Vulkan API version performance workaround that I submitted for the Vulkan Validation Layer defect https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7562 no longer works when using SDK 1.3.290.

In PR #1084 I changed the API version from Vulkan 1.1 to Vulkan 1.0 which worked around the performance problem for prior SDKs as documented in the VVL defect above.  However, something changed in SDK 1.3.290 that correctly enforces the API 1.1 version for the `debugPrintfEXT` layer feature even without fixing the performance issue.  Note the fix for the VVL performance issue is expected in the next Vulkan SDK following 1.3.290.

To fix this regression, I can submit a change that detects the SDK instance version at runtime and sets the API version appropriately for this sample:
a) Set API 1.2 for SDKs <= 1.3.290 to avoid the performance defect and satisfy the Vulkan API version test in SDK 1.3.290
b) Set API 1.1 for SDKs > 1.3.290 to leverage the VVL fix https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8323 (which I have verified as working with this sample via a private build)

The only concern I have is that while setting API 1.2 for SDKs <= 1.3.290 works around the performance defect, it rules out running the sample on earlier Vulkan 1.1 implementations.  However, given that the VVL performance issue made the sample almost unusable with API 1.1 prior to the coming fix, I am not sure this really matters.

@SaschaWillems since you submitted the original VVL defect, do you have any recommendations here?  Do you agree with my proposed solution or can you suggest any alternatives?  I guess we could just set the API version back to 1.1 for all SDK versions and just live with the performance problem for SDKs <= 1.3.290.  Your thoughts?

SRSaunders:
## Description

This fixes the _shader_debugprintf_ sample to work correctly with Vulkan SDKs >= 1.3.290 where the Validation Layer requires Vulkan API >= 1.1 to enable the `debugPrintfEXT`feature.  Earlier SDKs with older Validation Layers did not enforce this and could run with Vulkan API 1.0.  This change also respects the upcoming Validation Layer performance fix for issue [7562](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7562) which caused slowdowns on SDKs <= 1.3.290 when using the `debugPrintfEXT` feature with the correct APi 1.1 version. VVL fix [8323](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8323) will be available in Vulkan SDKs > 1.3.290 and thus Vulkan API 1.1 can be used by this sample for these future SDKs without performance problems.

**UPDATE**: This PR also contains a workaround for SDKs not yet containing the VVL performance fix above.  For SDKs <= 1.3.290, this sample will now use Vulkan API 1.2 which does not exhibit the performance problem, and also satisfies the `debugPrintfEXT` layer feature which requires Vulkan API level >= 1.1 to function.

Tested on macOS Ventura using: a) Vulkan SDKs 1.3.283 and 1.3.290 with their stock Validation Layers, and b) using a private build of the most recent Validation Layer containing the fix above.

Fixes #1132

## General Checklist:

- [ ] I have tested every sample to ensure everything runs correctly.  _Not applicable._

## Sample Checklist

### DIFF
--- a/samples/extensions/shader_debugprintf/shader_debugprintf.cpp
+++ b/samples/extensions/shader_debugprintf/shader_debugprintf.cpp
@@ -433,6 +433,9 @@ const std::vector<const char *> ShaderDebugPrintf::get_validation_layers()
 // This sample overrides the instance creation part of the framework to chain in additional structures
 std::unique_ptr<vkb::Instance> ShaderDebugPrintf::create_instance(bool headless)
 {
+	uint32_t instanceApiVersion;
+	VK_CHECK(vkEnumerateInstanceVersion(&instanceApiVersion));
+
 	uint32_t instance_extension_count;
 	VK_CHECK(vkEnumerateInstanceExtensionProperties(nullptr, &instance_extension_count, nullptr));
 	std::vector<VkExtensionProperties> available_instance_extensions(instance_extension_count);
@@ -444,9 +447,9 @@ std::unique_ptr<vkb::Instance> ShaderDebugPrintf::create_instance(bool headless)
 	                available_instance_extensions.end(),
 	                [](VkExtensionProperties const &extension) { return strcmp(extension.extensionName, VK_EXT_LAYER_SETTINGS_EXTENSION_NAME) == 0; }))
 	{
-		// debugPrintfEXT layer feature requires Vulkan API 1.1, but has performance issues due to VVL defect in SDKs <= 1.3.290
+		// debugPrintfEXT layer feature requires Vulkan API 1.1, but use API 1.2 until VVL performance fix is available in SDKs > 1.3.290
 		// See VVL issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7562 for defect and fix information
-		set_api_version(VK_API_VERSION_1_1);
+		set_api_version(instanceApiVersion <= VK_MAKE_API_VERSION(0, 1, 3, 290) ? VK_API_VERSION_1_2 : VK_API_VERSION_1_1);
 
 		// Run standard create_instance() from framework (with set_api_version and layer settings support) and return
 		return VulkanSample::create_instance(headless);
@@ -477,7 +480,7 @@ std::unique_ptr<vkb::Instance> ShaderDebugPrintf::create_instance(bool headless)
 	VkApplicationInfo app_info{VK_STRUCTURE_TYPE_APPLICATION_INFO};
 	app_info.pApplicationName = "Shader debugprintf";
 	app_info.pEngineName      = "Vulkan Samples";
-	app_info.apiVersion       = VK_API_VERSION_1_1;
+	app_info.apiVersion       = instanceApiVersion <= VK_MAKE_API_VERSION(0, 1, 3, 290) ? VK_API_VERSION_1_2 : VK_API_VERSION_1_1;
 
 	// Shader printf is a feature of the validation layers that needs to be enabled
 	std::vector<VkValidationFeatureEnableEXT> validation_feature_enables = {VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT};
--- a/samples/extensions/shader_debugprintf/shader_debugprintf.cpp
+++ b/samples/extensions/shader_debugprintf/shader_debugprintf.cpp
@@ -442,9 +442,13 @@ std::unique_ptr<vkb::Instance> ShaderDebugPrintf::create_instance(bool headless)
 	// For backwards compatibility with SDKs < 1.3.272 without VK_EXT_layer_settings, the remainder of this custom override is required
 	if (std::any_of(available_instance_extensions.begin(),
 	                available_instance_extensions.end(),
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Doesn't work for me. The sample now seems to be completely broken in any case :(

* With `VK_API_VERSION_1_0`: Good framerate, but no output of vertex positions visible
* With `VK_API_VERSION_1_1`: Unusable framerate (less than 1 fps), but output of vertex positions visible

Tested with SDK 1.3.290.

Pinging @spencer-lunarg

spencer-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8323 should have fixed this, I'll pull down this change and test on the various things today to see what is going on

SRSaunders:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8323 is not included in SDK 1.3.290, and will only fix the performance issue in future SDKs when it is included.

@SaschaWillems that is why I asked if you wanted me to put in a workaround in _shader_debugprintf_ for previous SDKs.  You answered no, so that is why this PR reverts to API 1.1 for all previous and current SDKs, and is the reason for the performance issue with SDKs <= 290.  My workaround was to use Vulkan API 1.2 for previous SDKs <= 290, and API 1.1 for SDKs > 290.  This would avoid performance issues for SDKs that don't have the above fix.

Please advise if you want this workaround added here.

Note that API 1.0 was always the wrong choice, since it should never have worked, and only worked for SDKs < 290 by chance.  Even though 290 does not include the performance fix, it adds a check for debugPrintfEXT to be at least API 1.1.  So that is why my workaround proposes using API 1.2 for SDKs up to and including 290, and API 1.1 for all future SDKs.

Perhaps it would be easier if I just add the workaround in an additional commit here, and you both can review.

SaschaWillems:
> This fixes the shader_debugprintf sample to work correctly with Vulkan SDKs >= 1.3.290 where the Validation Layer requires Vulkan API 1.1 to enable the debugPrintfEXTfeature

I guess this mislead me then?

Tbh. I'm a bit out of the loop and I'm having a hard time answering to requests like this ad hoc.

Can't we simply revert back to the initial state of the sample (when I wrote it)?

SRSaunders:
> I guess this mislead me then?

Well, I guess I should have said >290 (and not >= 290).  Sorry about that - my bad.

> Can't we simply revert back to the initial state of the sample (when I wrote it)?

That won't help.  Your original unmodified sample used Vulkan API 1.1 which will exhibit the same VVL performance issues as you are seeing now.  You can blame me for moving back to API 1.0.  I should have moved to API 1.2 at least until the VVL fix was made by @spencer-lunarg.

I am sorry if I have been commenting too frequently or asking too many questions given your other commitments.  Perhaps you did not have time to appreciate the details of what I was asking.

I think it would be best for me to push the workaround that will fix this once and for all.

SRSaunders:
@SaschaWillems and @spencer-lunarg:  Commit https://github.com/KhronosGroup/Vulkan-Samples/pull/1133/commits/c209e6b86f4a31342eac9a2fa017d54ed64efd08 should fix perf issues for sample _shader_debugprintf_.

In my testing this fixes performance and vertex visibility when using SDK 1.3.290 (and earlier).  If you see otherwise, please let me know including the test conditions (e.g. SDK version, platform, etc.) that cause it.

SaschaWillems:
I feel kinda dumb, but this only works for me when NOT using the debug printf setting via vkconfig. If I use that preset, I don't see any vertex position output. It's very much possible that I'm doing something wrong. but I'm not sure what. I'm not sure why the code has an if clause for the layer setting extension at all.

I'm kinda torn on whether we want to add temporary workarounds for bugs in the SDK. 

Can't we simply wait for the fixed SDK to be released?

spencer-lunarg:
So I am still trying to grasp the conversation (and background here)

From my 1000-foot pole view, I would just have this requires Vulkan 1.2 (since there is enough 1.2 devices in the wild now) and when the next SDK comes out, just bump it down to 1.1

SRSaunders:
> From my 1000-foot pole view, I would just have this requires Vulkan 1.2 (since there is enough 1.2 devices in the wild now) and when the next SDK comes out, just bump it down to 1.1

That’s effectively what this PR does.  It just automates it at runtime so that you don’t have to go back later to change the code when a new SDK comes out, or worry about cases when running against a non-current SDK.

However if you are not happy with a runtime-based approach and instead prefer a static fix using API 1.2, please let me know and I can make the change.

spencer-lunarg (approved):
seems like a valid approach to me (from a VVL engineer point of view)

SaschaWillems (approved):

Issue #393, opened 2021-11-09
`raytracing_extended` sample fails with a scratch buffer alignment problem


expenses:
Hello. I'm currently unable to run the `raytracing_extended` sample on my AMD Radeon RX 6900 XT and Windows 10.

I did a validation capture and picked up the 3 following errors:

`VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710(ERROR / SPEC): msgNum: 1677691140 - Validation Error: [ VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710 ] Object 0: handle = 0x299797ac020, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x63ff8904 | vkCmdBuildAccelerationStructuresKHR:For each element of pInfos, its scratchData.deviceAddress member must be a multiple of VkPhysicalDeviceAccelerationStructurePropertiesKHR::minAccelerationStructureScratchOffsetAlignment. The Vulkan spec states: For each element of pInfos, its scratchData.deviceAddress member must be a multiple of VkPhysicalDeviceAccelerationStructurePropertiesKHR::minAccelerationStructureScratchOffsetAlignment (https://vulkan.lunarg.com/doc/view/1.2.189.2/windows/1.2-extensions/vkspec.html#VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710)`

`VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710(ERROR / SPEC): msgNum: 1677691140 - Validation Error: [ VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710 ] Object 0: handle = 0x299797ac020, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x63ff8904 | vkCmdBuildAccelerationStructuresKHR:For each element of pInfos, its scratchData.deviceAddress member must be a multiple of VkPhysicalDeviceAccelerationStructurePropertiesKHR::minAccelerationStructureScratchOffsetAlignment. The Vulkan spec states: For each element of pInfos, its scratchData.deviceAddress member must be a multiple of VkPhysicalDeviceAccelerationStructurePropertiesKHR::minAccelerationStructureScratchOffsetAlignment (https://vulkan.lunarg.com/doc/view/1.2.189.2/windows/1.2-extensions/vkspec.html#VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710)`

[...]

PR #1134, opened 2024-08-21
Add Mobile NeRF Ray Query Sample

RodrigoHolztrattner-QuIC:
## Description

Adds a new sample, Mobile NeRF Ray Query, which is an upgraded version of the original Mobile NeRF sample. This one uses rayqueries to accelerate the NeRF processing.

## General Checklist:

## Sample Checklist


samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.h
```diff
@@ -0,0 +1,204 @@
...
+// The third layer weights' size is changed from 48 to 64 to make sure a 16 bytes alignement
+// #define WEIGHTS_2_COUNT (48)
+#define WEIGHTS_2_COUNT (64)
+#define BIAS_0_COUNT (16)
+#define BIAS_1_COUNT (16)
+// The third layer bias' size is changed from 3 to 4 to make sure a 16 bytes alignement
+#define BIAS_2_COUNT (4)
+
+	// some typedef for each model
+	struct mlp_weights
```
asuessenbach: I think, the naming conventions here start types (like structs) with an upper-case letter. Why not just keep it with MLP_Weights?

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.h
```diff
@@ -0,0 +1,204 @@
...
+
+	uint64_t get_buffer_device_address(VkBuffer buffer);
+	void     create_top_level_acceleration_structure();
+	void     create_bottom_level_acceleration_structure(int model_entry);
+
+	void create_texture(int model_index, int sub_model_index, int models_entry);
+	void create_texture_helper(std::string const &texturePath, Texture &texture_input);
+};
+
+std::unique_ptr<vkb::VulkanSample<vkb::BindingType::C>> create_mobile_nerf_rayquery();
```
asuessenbach: If you like, you could use `VulkanSampleC`, instead of `VulkanSample<vkb::BindingType::C>`.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1395 @@
...
+	camera.set_rotation(glm::eulerAngles(orientation) * glm::pi<float>() / 180.f);
+	camera.set_translation(translation);
+}
+
+MobileNerfRayQuery::MobileNerfRayQuery()
+{
+	title = "Mobile Nerf Ray Query";
+
+	// Scalar Block Layout Extension requires Vulkan 1.2
+	set_api_version(VK_API_VERSION_1_2);
```
asuessenbach: I thought, `VK_EXT_scalar_block_layout` would depend on `VK_KHR_get_physical_device_properties2` and `VK_VERSION_1_1`.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1395 @@
...
+
+	// Required by VK_KHR_acceleration_structure
+	add_device_extension(VK_KHR_ACCELERATION_STRUCTURE_EXTENSION_NAME);
+	add_device_extension(VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME);
+	add_device_extension(VK_KHR_DEFERRED_HOST_OPERATIONS_EXTENSION_NAME);
+	add_device_extension(VK_EXT_DESCRIPTOR_INDEXING_EXTENSION_NAME);
```
asuessenbach, line 85: `VK_KHR_acceleration_structure` depends on `VK_KHR_deferred_host_operations` and either `VK_VERSION_1_2` or (`VK_VERSION_1_1` and `VK_EXT_descriptor_indexing` and `VK_KHR_buffer_device_address`)

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1395 @@
...
+	std::vector<VkImageView> views;
+	views.resize(2);
+	views[0] = depth_stencil.view;
+
+	// Depth/Stencil attachment is the same for all frame buffers
+	VkFramebufferCreateInfo framebuffer_create_info = {};
+	framebuffer_create_info.sType                   = VK_STRUCTURE_TYPE_FRAMEBUFFER_CREATE_INFO;
+	framebuffer_create_info.pNext                   = NULL;
+	framebuffer_create_info.renderPass              = render_pass_nerf;
+	framebuffer_create_info.attachmentCount         = views.size();
```
asuessenbach: `static_cast<uint32_t>` missing?

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1395 @@
...
+	clear_values[0].depthStencil = {1.0f, 0};
+	clear_values[1].color        = {{1.0f, 1.0f, 1.0f, 1.0f}};
+
+	VkRenderPassBeginInfo render_pass_begin_info    = vkb::initializers::render_pass_begin_info();
+	render_pass_begin_info.renderPass               = render_pass_nerf;
+	render_pass_begin_info.renderArea.offset.x      = 0;
+	render_pass_begin_info.renderArea.offset.y      = 0;
+	render_pass_begin_info.renderArea.extent.width  = width;
+	render_pass_begin_info.renderArea.extent.height = height;
+	render_pass_begin_info.clearValueCount          = clear_values.size();
```
asuessenbach: `static_cast<uint32_t>` missing?

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1395 @@
...
+	submit_info.commandBufferCount = 1;
+	submit_info.pCommandBuffers    = &draw_cmd_buffers[current_buffer];
+
+	// Submit to queue
+	VK_CHECK(vkQueueSubmit(queue, 1, &submit_info, VK_NULL_HANDLE));
+
+	ApiVulkanSample::submit_frame();
+}
+
+std::unique_ptr<vkb::VulkanSample<vkb::BindingType::C>> create_mobile_nerf_rayquery()
```
asuessenbach: Could be `VulkanSampleC`, again.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.h
RodrigoHolztrattner-QuIC: Missed this on my final review, thanks for pointing out.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.h
RodrigoHolztrattner-QuIC: Adjusted.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
RodrigoHolztrattner-QuIC: That's true, I changed it back to version 1.1 and added the device properties 2 as a dependency.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
RodrigoHolztrattner-QuIC: By changing back to `VK_VERSION_1_1 `, the other dependencies make sense now. Thank you!

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
RodrigoHolztrattner-QuIC: Added this and a few more that were missing. Thanks.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
RodrigoHolztrattner-QuIC: Added this and a few more that were missing. Thanks.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
RodrigoHolztrattner-QuIC: Updated, thank you!

RodrigoHolztrattner-QuIC:
Moving sample out from draft (for review).

Note: There are some _VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710_ validation errors on AMD devices when using acceleration structures. The _ray_tracing_extended_ sample seems to trigger this same errors as well. This might be device/driver dependent and not really an issue on more up-to-date HW/drivers. 

An old and similar issue [`raytracing_extended` sample fails with a scratch buffer alignment problem · Issue #393 · KhronosGroup/Vulkan-Samples · GitHub](https://github.com/KhronosGroup/Vulkan-Samples/issues/393) was supposedly tracking this problem and closed as resolved, but for me at least it, it is still present.

These errors do not appear on NVIDIA's or mobile devices in general.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1397 @@
...
+}
+
+MobileNerfRayQuery::MobileNerfRayQuery()
+{
+	title = "Mobile Nerf Ray Query";
+
+	set_api_version(VK_API_VERSION_1_1);
+
+	// Required by VK_EXT_scalar_block_layout
+	add_device_extension(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
```
SaschaWillems: This is wrong. VK_KHR_get_physical_device_properties2 is an instance extension.

samples/general/mobile_nerf_rayquery/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+get_filename_component(FOLDER_NAME ${CMAKE_CURRENT_LIST_DIR} NAME)
+get_filename_component(PARENT_DIR ${CMAKE_CURRENT_LIST_DIR} PATH)
+get_filename_component(CATEGORY_NAME ${PARENT_DIR} NAME)
+
+add_sample_with_tags(
+    ID ${FOLDER_NAME}
+    CATEGORY ${CATEGORY_NAME}
+    AUTHOR "Qualcomm"
+    NAME "Mobile NeRF Ray Query"
+    DESCRIPTION "A Mobile Neural Radiance Field synthesizer sample using ray query, based on textured polygons."
```
SaschaWillems, line 27: Can you add the shaders for this sample in here? That way people can open those from the IDE. See CMakeLists.txt from other samples.

SaschaWillems (changes requested):
Unlike the existing mobile NeRF sample, this one does not work for me. I get this error at startup:

```
[error] Required device extension VK_KHR_get_physical_device_properties2 not available, cannot run
[error] Error Message: Extensions not present : ERROR_EXTENSION_NOT_PRESENT
[error] Failed when running application Mobile NeRF Ray Query
```

This is caused, by VK_KHR_get_physical_device_properties2 being added as a device extension, while it actually is an instance level extension.

SaschaWillems:
Also mouse camera controls feel odd. Like they're inverted, compared to the other mobile NeRF sample.

SaschaWillems:
The sample also trigger a lot of warnings about uninitialized variables (~75). Those need to be fixed too.

SaschaWillems:
I also get these validation messages with the latest SDK:

```
[error] 269944751 - VUID-RuntimeSpirv-NonWritable-06340: Validation Error: [ VUID-RuntimeSpirv-NonWritable-06340 ] Object 0: handle = 0x59f7450000000038, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0x101707af | vkCreateGraphicsPipelines(): pCreateInfos[0].pStages[1] SPIR-V (VK_SHADER_STAGE_FRAGMENT_BIT) uses descriptor [Set 2, Binding 0, variable "indices_set"] (type VK_DESCRIPTOR_TYPE_STORAGE_BUFFER or VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC) which is not marked with NonWritable, but fragmentStoresAndAtomics was not enabled. The Vulkan spec states: If fragmentStoresAndAtomics is not enabled, then all storage image, storage texel buffer, and storage buffer variables in the fragment stage must be decorated with the NonWritable decoration (https://vulkan.lunarg.com/doc/view/1.3.290.0/windows/1.3-extensions/vkspec.html#VUID-RuntimeSpirv-NonWritable-06340)
[error] 269944751 - VUID-RuntimeSpirv-NonWritable-06340: Validation Error: [ VUID-RuntimeSpirv-NonWritable-06340 ] Object 0: handle = 0x59f7450000000038, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0x101707af | vkCreateGraphicsPipelines(): pCreateInfos[0].pStages[1] SPIR-V (VK_SHADER_STAGE_FRAGMENT_BIT) uses descriptor [Set 1, Binding 0, variable "vertices_set"] (type VK_DESCRIPTOR_TYPE_STORAGE_BUFFER or VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC) which is not marked with NonWritable, but fragmentStoresAndAtomics was not enabled. The Vulkan spec states: If fragmentStoresAndAtomics is not enabled, then all storage image, storage texel buffer, and storage buffer variables in the fragment stage must be decorated with the NonWritable decoration (https://vulkan.lunarg.com/doc/view/1.3.290.0/windows/1.3-extensions/vkspec.html#VUID-RuntimeSpirv-NonWritable-06340)
```

RodrigoHolztrattner-QuIC:
Hi @SaschaWillems , I got the updated code from the team and update the PR, thanks for your feedback!

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1261 @@
...
+	{
+		LOGI("Using VK_FORMAT_R16G16B16A16_SFLOAT for feature texture");
+		feature_map_format = VK_FORMAT_R16G16B16A16_SFLOAT;
+	}
+	else if (textureType == "32bit")
+	{
+		LOGI("Using VK_FORMAT_R32G32B32A32_SFLOAT for feature texture");
+		feature_map_format = VK_FORMAT_R32G32B32A32_SFLOAT;
+	}
+	else if (textureType == "8bit")
```
asuessenbach: You already had the "8bit" case above.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1261 @@
...
+	}
+
+	// Read camera position
+	if (asset_map["camera"].is_array() && asset_map["camera"].size() == 3)
+	{
+		camera_pos = glm::vec3(asset_map["camera"][0].get<float>(), asset_map["camera"][1].get<float>(), asset_map["camera"][2].get<float>());
+	}
+	else
+	{
+		LOGW("Fail to read camera position. Use defualt value.");
```
asuessenbach: typo: defualt -> default

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1261 @@
...
+
+	LOGI("Parsing mlp data {}", mlpJsonPath);
+	json data = json::parse(f);
+
+	// Record a index of the first sub-model
+	const auto first_sub_model = models.size();
+	int        obj_num         = data["obj_num"].get<int>();
+
+	// Here we know the actual number of sub models
+	const auto next_sub_model_index = models.size();
```
asuessenbach: Is there a reason to have identical variables `first_sub_model` and `next_sub_model_index`?

asuessenbach (commented):
I have to admit, I can't follow all that mlp-stuff...
But there are a few minor issues, and the camera movement is not consistent: moving left/right is correct but up/down is inverted.

RodrigoHolztrattner-QuIC:
Hi @asuessenbach, I just got the updated code from the team and updated the PR, the remaining issues should all have been addressed.

asuessenbach (dismissed):

samples/general/mobile_nerf_rayquery/README.adoc
```diff
@@ -0,0 +1,33 @@
...
+= Mobile NeRF Ray Query
+
+ifdef::site-gen-antora[]
+TIP: The source for this sample can be found in the https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/general/mobile_nerf_rayquery[Khronos Vulkan samples github repository].
+endif::[]
+
+NeRF is a new 3D representation method in Computer Vision that creates images of a 3D scene using several 2D pictures taken from different viewpoints.
+This method constructs a representation of the 3D volume. Various adaptations of NeRF target different use cases, including MobileNeRF, which focuses on rendering NeRF efficiently on mobile phones by leveraging existing traditional graphic hardware.
+
+This version enhances the https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/general/mobile_nerf[previous MobileNeRF implementation] by using the Vulkan Ray Query feature, which leverages the hardware ray tracing capabilities of the Adreno GPU.
```
JoseEmilio-ARM: Thanks for the sample! In this case, we could keep this line more generic, to account for other GPUs with RQ support, e.g. "capabilities of mobile GPUs that support it'?

SaschaWillems:
With the recent merge of #1013, this probably needs to be updated to master and the `request_gpu_features` function needs to be changed to how features are now requested since #1013.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1252 @@
...
+	create_descriptor_sets();
+	prepare_pipelines();
+	build_command_buffers();
+
+	prepared = true;
+	LOGI("Prepare Done!");
+	return true;
+}
+
+void MobileNerfRayQuery::request_gpu_features(vkb::PhysicalDevice &gpu)
```
SaschaWillems, line 182: With #1013 now merged, this needs to be adjusted accordingly. Probably best to update this to master and then update this part of the code.

Other than that I'm happy with this sample :)

SaschaWillems (changes requested):

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1236 @@
...
+#include "gltf_loader.h"
+#include "platform/platform.h"
+#include "rendering/subpasses/forward_subpass.h"
+#include "scene_graph/components/material.h"
+#include "scene_graph/components/mesh.h"
+#include "scene_graph/components/perspective_camera.h"
+
+namespace
+{
+constexpr uint32_t MIN_THREAD_COUNT = 1;
```
SaschaWillems: This variable doesn't seem to be used anywhere in this sample.

SaschaWillems (changes requested):
Good looks good (aside from a very minor thing, see my comment), sample runs fine.

But I do get this validation error at startup:

```
[error] 115483881 - VUID-VkShaderModuleCreateInfo-pCode-08740: Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08740 ] | MessageID = 0x6e224e9 | vkCreateShaderModule():  SPIR-V Capability UniformBufferArrayNonUniformIndexing was declared, but one of the following requirements is required (VkPhysicalDeviceVulkan12Features::shaderUniformBufferArrayNonUniformIndexing). The Vulkan spec states: If pCode is a pointer to SPIR-V code, and pCode declares any of the capabilities listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied (https://vulkan.lunarg.com/doc/view/1.3.290.0/windows/1.3-extensions/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-08740)
```

Looks like that feature isn't enabled.

samples/general/mobile_nerf_rayquery/README.adoc
RodrigoHolztrattner-QuIC: Sure, I edited that part to your suggestion, thank you.

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
RodrigoHolztrattner-QuIC: Removed, thanks for pointing out.

RodrigoHolztrattner-QuIC:
@SaschaWillems I was trying to resolve the issue below before pushing the changes to fix the validation layer error, at this point I'm starting to think it's a problem with the framework and not actually the sample, but just letting you know the latest version shouldn't have any validation errors anymore.

```
Gralloc4                com.khronos.vulkan_samples           E  isSupported(1, 1, 59, 1, ...) failed with 7
GraphicBufferAllocator  com.khronos.vulkan_samples           E  Failed to allocate (4 x 4) layerCount 1 format 59 usage b00: 2
AHardwareBuffer         com.khronos.vulkan_samples           E  GraphicBuffer(w=4, h=4, lc=1) failed (Unknown error -2), handle=0x0
```

RodrigoHolztrattner-QuIC:
It seems other samples have the errors I mentioned above, it doesn't seem specific to this one.

asuessenbach (dismissed):

SaschaWillems (dismissed):

samples/general/mobile_nerf_rayquery/README.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+ifdef::site-gen-antora[]
+TIP: The source for this sample can be found in the https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/general/mobile_nerf_rayquery[Khronos Vulkan samples github repository].
+endif::[]
+
+NeRF is a new 3D representation method in Computer Vision that creates images of a 3D scene using several 2D pictures taken from different viewpoints.
+This method constructs a representation of the 3D volume. Various adaptations of NeRF target different use cases, including MobileNeRF, which focuses on rendering NeRF efficiently on mobile phones by leveraging existing traditional graphic hardware.
+
+This version enhances the https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/general/mobile_nerf[previous MobileNeRF implementation] by using the Vulkan Ray Query feature, which leverages the hardware ray tracing capabilities of mobile GPUs that support it.
+This enhancement greatly boosts performance in most use cases. Additionally, the Vulkan API provides great flexibility for modifying and optimizing the rendering pipeline and shaders, enabling more functionalities while delivering optimal performance.
```
JoseEmilio-ARM, line 30: Hi, sorry for the delay in reviewing. I was wondering if you have an estimate of how much improvement to expect? I tested a couple of devices and there was little difference, could it be related to the notes on the `USE_OPAQUE` flag?

JoseEmilio-ARM (changes requested):
Tested on Arm GPUs and there seems to be an improvement (with respect to no-RQ) with the `USE_OPAQUE` flag, but a regression without it. Could `USE_OPAQUE` be the default?

shaders/mobile_nerf_rayquery/rayquery_morpheus_combo.frag
```diff
@@ -0,0 +1,343 @@
...
+ * built for WebGL)
+ * Contributor: (Qualcomm) Rodrigo Holztrattner - quic_rholztra@quicinc.com
+ */
+#version 460
+#extension GL_EXT_scalar_block_layout : enable
+#extension GL_EXT_ray_query : enable
+#extension GL_EXT_nonuniform_qualifier : enable
+
+// Uncomment this to use opaque mode
+// Should be faster, but would get VK_ERROR_DEVICE_LOST on some AMD devices
```
JoseEmilio-ARM: you mention the device_lost issue here, but below you mention it's due to compatibility with the original models?

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
```diff
@@ -0,0 +1,1235 @@
...
+
+	update_uniform_buffer();
+}
+
+void MobileNerfRayQuery::read_json_map()
+{
+	std::string assetBase = vkb::fs::path::get(vkb::fs::path::Type::Assets);
+	LOGI("Base assets path: {}", assetBase);
+
+#if defined(NERF_JSON_FILE)
```
JoseEmilio-ARM, line 212: is the JSON format/options documented? maybe it could be explained in the README? wondering if the code could be simplified to avoid duplication between combo and non-combo modes?

JoseEmilio-ARM (changes requested):

SaschaWillems:
Can we defer minor remarks/questions? This PR has been around for quite some time now and I think we should make sure that PRs as these get merged as fast as possible. Adding in late minor changes would require everyone to re-approve.

JoseEmilio-ARM:
In this case the sample goal is to show the performance advantage of using ray query, but this is not the case (in the devices I tested), so it might require major changes, including a better solution for the device_lost issue mentioned in the comments

RodrigoHolztrattner-QuIC:
@JoseEmilio-ARM the performance really depends on the number of objects and which device was used for testing, let me cover each topic and the device issue problem:

One or just a few objects are much faster when using the ray-query approach, we simplified the draw calls on this sample to make it easier to consume by doing as few draw commands as possible. Multiple draws, each with its own model, would significantly lower cache misses and consequentially, increase performance, while complicating the sample code.

While testing on our devices, Snapdragon 8 Gen 2 is almost in pair with the non-ray-query version, while Gen 3 and Gen 4 offers significant increased FPS then the previous generations.

For the USE_OPAQUE mode: The original models from the paper have transparent textures, but the ones we made available to the public doesn't. To ensure compatibility with some AMD devices the non-opaque mode was left there just to avoid the possible crash if one decides to use the original models. We don't mind removing the non-opaque code.

For the items where there could be actionable changes, I would prefer to create a new PR for addressing them instead of adding more changes to this one, as it's been active for a while now. I assume performance is the main point, but not only we do get faster performance with ray-query on more recent devices, it's a modern approach and likely to mature even more in the next generations of devices, providing even greater performance for NeRF processing on mobile.

gpx1000 (dismissed):

samples/general/mobile_nerf_rayquery/README.adoc
JoseEmilio-ARM: Answered below, happy to address in a future PR

shaders/mobile_nerf_rayquery/rayquery_morpheus_combo.frag
JoseEmilio-ARM: Answered below, happy to address in a future PR

samples/general/mobile_nerf_rayquery/mobile_nerf_rayquery.cpp
JoseEmilio-ARM: Answered below, happy to address in a future PR

JoseEmilio-ARM (approved):

gpx1000 (approved):

marty-johnson59:
Merging per discussion today.  Thanks everyone!

PR #1150, opened 2024-09-03
Work-around macOS/iOS blank screen issues on swapchain recreation

### LINKED ISSUE #1149: Apple devices can display blank screens following swapchain recreation
Blank screens and `VK_SUBOPTIMAL_KHR` swapchain results can occur on macOS and iOS following swapchain recreation when either `image_count`, `image_usage_flags`, or `transform` parameters are changed without changing the swapchain extent.  This issue manifests itself when activating options for the following samples: _afbc_, _msaa_, _swapchain_images_, _hpp_swapchain_images_, and _surface_rotation_.

I am not sure why this happens, but it may have something to do with MoltenVK's possible reuse of swapchain image and/or imageview resources, and the mechanism that Vulkan Samples uses when managing those resources on recreation.  If the swapchain extent is changed (i.e. resized), then recreation happens correctly and blank screens do not occur. Alternatively, this may be a defect in MoltenVK or somewhere else.

There is a workaround to avoid this problem: recreate the swapchain in `[HPP]RenderContext::begin_frame()` specifically in cases when `vkAcquireNextImageKHR()` returns `VK_SUBOPTIMAL_KHR` following swapchain parameter updates.  I will push a PR that implements this change for Apple devices only.  

However, it would be good to get feedback from @billhollings and @gpx1000 regarding the root cause of the problem, or if this is expected behaviour with MoltenVK.

SRSaunders:
## Description

This PR does ~~two~~ three things:

1. Adds workarounds on **macOS/iOS only** for swapchain recreation when only `image_count`, `image_usage_flags`, or `transform` parameters are changing (i.e. without resizing).  This workaround recreates the swapchain a second time within `[HPP]RenderContext::begin_frame()` specifically in cases when `vkAcquireNextImageKHR()` returns `VK_SUBOPTIMAL_KHR`. This fixes the blank screen problem when options are selected in certain samples (_afbc, msaa, swapchain_images, hpp_swapchain_images, and surface_rotation_).
2. Reverts the previous workarounds from PR #1084 which permitted `VK_SUBOPTIMAL_KHR` return codes within `[HPP]RenderContext::begin_frame()`.  These `VK_SUBOPTIMAL_KHR` return codes were due to swapchain recreation problems in item 1 above.  With these now addressed, the code allowing `VK_SUBOPTIMAL_KHR` is no longer needed.
3. **UPDATE**: Fixes a defect in the way `[HPP]RenderContext::begin_frame()` calls `swapchain->acquire_next_image()` a second time using the same semaphore from the first call. This is not correct since that semaphore will already be signaled and cannot be used again until cleared by a queue wait operation. To address this, I added code to destroy the semaphore and recreate it in a non-signaled state.

Fixes #1149 

Tested on macOS Ventura 13.6.6, iOS 17.6.1, and Windows 10 using an AMD 6600XT GPU.

## General Checklist:

 - [X] I did a full batch run using the `batch` command line argument to make sure all samples still work properly

## Sample Checklist

### DIFF
--- a/framework/rendering/hpp_render_context.cpp
+++ b/framework/rendering/hpp_render_context.cpp
@@ -322,8 +322,7 @@ void HPPRenderContext::begin_frame()
 			}
 		}
 
-		// eSuboptimalKHR can legitimately occur in batch mode when exiting certain samples (e.g. afbc, msaa, surface_rotation, swapchain_images)
-		if (result != vk::Result::eSuccess && result != vk::Result::eSuboptimalKHR)
+		if (result != vk::Result::eSuccess)
 		{
 			prev_frame.reset();
 			return;
--- a/framework/rendering/hpp_render_context.cpp
+++ b/framework/rendering/hpp_render_context.cpp
@@ -119,6 +119,10 @@ void HPPRenderContext::update_swapchain(const uint32_t image_count)
 	device.get_handle().waitIdle();
 
 	swapchain = std::make_unique<vkb::core::HPPSwapchain>(*swapchain, image_count);
+#if defined(PLATFORM__MACOS)
+	// Workaround on macOS/iOS to avoid blank screen when only changing image_count
+	swapchain = std::make_unique<vkb::core::HPPSwapchain>(*swapchain, image_count);
+#endif
 
 	recreate();
 }
@@ -134,6 +138,10 @@ void HPPRenderContext::update_swapchain(const std::set<vk::ImageUsageFlagBits> &
 	device.get_resource_cache().clear_framebuffers();
 
 	swapchain = std::make_unique<vkb::core::HPPSwapchain>(*swapchain, image_usage_flags);
+#if defined(PLATFORM__MACOS)
+	// Workaround on macOS/iOS to avoid blank screen when only changing image_usage_flags
+	swapchain = std::make_unique<vkb::core::HPPSwapchain>(*swapchain, image_usage_flags);
+#endif
 
 	recreate();
 }
@@ -157,6 +165,10 @@ void HPPRenderContext::update_swapchain(const vk::Extent2D &extent, const vk::Su
 	}
 
 	swapchain = std::make_unique<vkb::core::HPPSwapchain>(*swapchain, vk::Extent2D{width, height}, transform);
+#if defined(PLATFORM__MACOS)
+	// Workaround on macOS/iOS to avoid blank screen when only changing transform
+	swapchain = std::make_unique<vkb::core::HPPSwapchain>(*swapchain, vk::Extent2D{width, height}, transform);
+#endif
 
 	// Save the preTransform attribute for future rotations
 	pre_transform = transform;
--- a/framework/rendering/hpp_render_context.cpp
+++ b/framework/rendering/hpp_render_context.cpp
@@ -119,10 +119,6 @@ void HPPRenderContext::update_swapchain(const uint32_t image_count)
 	device.get_handle().waitIdle();
 
 	swapchain = std::make_unique<vkb::core::HPPSwapchain>(*swapchain, image_count);
-#if defined(PLATFORM__MACOS)
-	// Workaround on macOS/iOS to avoid blank screen when only changing image_count
-	swapchain = std::make_unique<vkb::core::HPPSwapchain>(*swapchain, image_count);
-#endif
 
 	recreate();
 }
@@ -138,10 +134,6 @@ void HPPRenderContext::update_swapchain(const std::set<vk::ImageUsageFlagBits> &
 	device.get_resource_cache().clear_framebuffers();
 
 	swapchain = std::make_unique<vkb::core::HPPSwapchain>(*swapchain, image_usage_flags);
-#if defined(PLATFORM__MACOS)
-	// Workaround on macOS/iOS to avoid blank screen when only changing image_usage_flags
-	swapchain = std::make_unique<vkb::core::HPPSwapchain>(*swapchain, image_usage_flags);
-#endif
 
 	recreate();
 }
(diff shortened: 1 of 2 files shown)


asuessenbach:
Creating the very same handle a second time seems a bit weird! Even as a workaround.
Do we really want to have such a strange pattern in this repository?

SRSaunders:
> Creating the very same handle a second time seems a bit weird! Even as a workaround.
> Do we really want to have such a strange pattern in this repository?

@asuessenbach I understand your perspective.  My approach here was to log the problem and ask for assistance from a few knowledgeable experts regarding macOS-specifics (see defect #1149).  I spent some time tracing around in the debugger and it seems that swapchain old resource destruction and new resource creation can happen out-of-order.  It seems to be the final destroy of old image resources (after creation of new ones) that causes the blank screen.  While this is a possible clue, I don't know enough about MoltenVK internals to determine the root cause. I felt that putting this workaround up may provide additional information that could help diagnose the problem.  Note that the second call does create a new handle instance, but with no memory of image resources from the previous configuration.

SRSaunders:
@asuessenbach - I'm following up here with some new information and a code update.

It seems like MoltenVK behaves differently than native Vulkan implementations when rebuilding the swapchain for reasons other than resizing.  On Windows and Linux, swapchain image count, image usage flags, and transform changes seem to be transparent to `vkAcquireNextImageKHR()` which returns `VK_SUCCESS` on the next acquire operation.  However, with MoltenVK, the next call to `vkAcquireNextImageKHR()` following a swapchain change as above (i.e. not a resize) results in `VK_SUBOPTIMAL_KHR`.  The existing logic in Vulkan-Samples does not handle this properly, other than the temporary band-aid I applied in an earlier PR which allowed the suboptimal return code without bailing out.  However that was not really the right approach and is something I am trying to correct here.

In my latest changes I have removed the awkward "double handle create" approach you rightly commented on, and instead added code to `[HPP]RenderContext::begin_frame()` which calls `handle_surface_changes(true)` on macOS/iOS for a forced update when `VK_SUBOPTIMAL_KHR` or `VK_ERROR_OUT_OF_DATE_KHR` is returned by `vkAcquireNextImageKHR()`.  This recreates the swapchain again and cleans up the blank screen issue.  Hopefully this approach is a bit more logical and clear.

I also noticed a defect in the way `[HPP]RenderContext::begin_frame()` calls `swapchain->acquire_next_image()` a second time using the same semaphore from the first call.  This is not correct since that semaphore will already be signaled and cannot be used again until cleared by a queue wait operation.  To address this, I added code to destroy the semaphore and recreate it in a non-signaled state.  This corrects the problem and I have tested to make sure it works properly.

framework/rendering/hpp_render_context.cpp
```diff
@@ -302,16 +302,24 @@ void HPPRenderContext::begin_frame()
...
+			// eSuboptimalKHR may occur on macOS/iOS following changes to swapchain other than extent/size
+			bool swapchain_updated = handle_surface_changes(true);
+#else
 			bool swapchain_updated = handle_surface_changes(result == vk::Result::eErrorOutOfDateKHR);
+#endif
 
 			if (swapchain_updated)
 			{
+				// Need to destroy and reallocate acquired_semaphore since it may have already been signaled
+				device.get_handle().destroySemaphore(acquired_semaphore);
```
asuessenbach, line 316: Don't know exactly under what circumstances this code path is taken.
But simply destroying the semaphore that you've got via `prev_frame.request_semaphore_with_ownership()` seems odd. Maybe you should call `prev_frame.release_owned_semaphore()` here?

asuessenbach (commented):
Looks much, well, more logical. Thanks.
Just one remaining issue.

framework/rendering/hpp_render_context.cpp
SRSaunders: > Don't know exactly under what circumstances this code path is taken

This is existing code that is run on any platform where `VK_ERROR_OUT_OF_DATE_KHR` is returned by `swapchain->acquire_next_image()`.  I am simply extending that case on Apple platforms to also cover when `VK_SUBOPTIMAL_KHR` is returned.

> Maybe you should call prev_frame.release_owned_semaphore() here?

Funny thing, I initially tried exactly that until I realized that all it does is return the signaled semaphore to the `released_semaphores` vector in `SemaphorePool` - it does not unsignal  the semaphore.  When that signaled semaphore is picked up again it will cause the same problem inside `[HPP]RenderContext::begin_frame()`.  I am not aware of any mechanism to manually clear a Vulkan binary semaphore (other than wait on a queue), so my only option was to destroy it and get another.  Fortunately the semantics of `request_semaphore_with_ownership()` means that it is manually-managed, and not present in any private vector within `SemaphorePool` until such time as you return it to the pool via `release_owned_semaphore()` in `[HPP]RenderContext::end_frame()`.  In other words you can delete it and regenerate without any issues.  But if you don't delete it, you will have a leak.

FYI, here is the Validation error message if you don't use an unsignaled semaphore:

`[error] -370888023 - VUID-vkAcquireNextImageKHR-semaphore-01286: Validation Error: [ VUID-vkAcquireNextImageKHR-semaphore-01286 ] Object 0: handle = 0xdf43f6000000012b, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0xe9e4b2a9 | vkAcquireNextImageKHR():  Semaphore must not be currently signaled. The Vulkan spec states: If semaphore is not VK_NULL_HANDLE it must be unsignaled (https://vulkan.lunarg.com/doc/view/1.3.290.0/mac/1.3-extensions/vkspec.html#VUID-vkAcquireNextImageKHR-semaphore-01286)`

SRSaunders: @asuessenbach I believe your point is resolved with explanation above.  Please let me know if otherwise.  Thx.

asuessenbach: Ah, yes, you're right. Thanks a lot for your explanations.

That is, from Windows-point-of-view, it looks good to me. Can't tell anything about the Mac-issue, though.

asuessenbach (approved):

gary-sweet (approved):

gpx1000 (approved):

marty-johnson59:
Merging - 3 approvals

PR #1170, opened 2024-09-20
Fix folder creation on Android

jloehr:
## Description

`root` doesn't has a trailing slash, so on Android a folder named `filesoutput/` was created instead of `files/output/`.
This PR adds the missing directory separator when creating folders.

## General Checklist:

### DIFF
--- a/components/filesystem/src/legacy.cpp
+++ b/components/filesystem/src/legacy.cpp
@@ -99,7 +99,7 @@ void create_path(const std::string &root, const std::string &path)
 	for (auto it = path.begin(); it != path.end(); ++it)
 	{
 		it = std::find(it, path.end(), '/');
-		create_directory(root + std::string(path.begin(), it));
+		create_directory(root + '/' + std::string(path.begin(), it));
 	}
 }
 


asuessenbach:
Under what circumstances do you get `root` without a trailing slash?
Maybe this function should be able handle both cases, `root` with and without a trailing slash.

jloehr:
Under Android `fs->external_storage_directory().string()` (which is passed into `create_path` as `root`) returns something along `/data/data/com.khronos.vulkan_samples/files` (don't have the exact path at hand now) without the trailing space.

https://github.com/KhronosGroup/Vulkan-Samples/blob/4c911f774c4069cafa215d61bdb9a9a2dfa18b22/components/filesystem/src/legacy.cpp#L74 seems to be the only place where `create_path` is called. We ran samples with that change on Android, Windows and Linux and didn't encounter any issues.

Maybe we can replace it with `std::filesystem::create_directories` altogether. Though haven't tested that. We ran with the simple added slash.

gpx1000 (approved):

marty-johnson59:
Hi @tomadamatkinson, can you take a quick look here?  Looks reasonable, but wanted to double check with you before merging.  Thanks

tomadamatkinson:
Hey @jloehr, this approach likely works, but I was a little unsure if this would affect other platforms.

> Maybe we can replace it with std::filesystem::create_directories altogether. Though haven't tested that. We ran with the simple added slash.

I think this is a great suggestion! It also pointed out some bugs that we could have had with vkb::fs in the long run. I also didn't know how the standard library would behave so i added some tests in here too. Please take a look at #1214 as an alternative approach

That said i am happy with either PR to merge :)

tomadamatkinson (approved):

marty-johnson59:
Merging w/ Tom's approval - per discussion on Monday's call - LMK if this causes issues and we can revert

PR #1173, opened 2024-09-23
iOS runtime fixes

SRSaunders:
## Description

This PR fixes multiple runtime issues for iOS physical device and iOS Simulator targets:

1. General: Fixes error handling for `main_loop_frame()` and eliminates redundancy with `main_loop()`.  Moved window close logic from `main_loop()` to `main_loop_frame()` so iOS can handle this condition properly.  This does not change behaviour for other platforms since `main_loop()` calls `main_loop_frame()`.
2. General: Asserts for valid render context before calling `render()` within `[HPP]ApiVulkanSample::update()`.  This is primarily defensive code but the issue was visible before improving error detection in `main_loop_frame()`.
3. iOS only: Adds `main_loop_frame()` return code handling within iOS `renderLoop()` - this was previously missing.  Also adds exit handling when an error is encountered or batch mode is complete.  The latter is unconventional for iOS, but is justified here since this is an Xcode-controlled developer application.
4. iOS only: Adds support for setting and getting iOS window content scale factor, which is important for the rendering of certain samples on retina displays.
5. General: Fixes `[HPP]Gui::draw()` to work with iOS Simulator with its `vkCmdDrawIndexed()` vertex offset limitations.  The iOS Simulator does not support vertex offets > 0, so I made a modification that rebinds the vertex buffer with an offset when running on the iOS Simulator.  Functionality is unchanged when running on real iOS targets or other platforms.
6. General: Initializes pipeline variables to null handles for better startup error recovery in _oit*_ samples.  This issue shows up when a platform does not support the required sample features, and teardown crashes with uninitialized pipelines.
7. ~~General: Use Vulkan API 1.0 support funcs for Vulkan API 1.0 sample _host_image_copy_.  The sample originally used `vkGetPhysicalDeviceFormatProperties2()` (API 1.1) but should use `vkGetPhysicalDeviceFormatProperties2KHR()`.  This issue shows up when using MoltenVK which is more sensitive to the Vulkan API version than the Vulkan loader.~~ Moved to a separate PR.
8. ~~iOS Simulator only: Disable MoltenVK's Metal Argument Buffers on the iOS Simulator for samples: _layout_transitions, pipeline_barriers, subpasses_.  Otherwise the samples can display blank screens on the iOS Simulator.~~  Reverted for now until issue resolved in MoltenVK project.

Tested on iOS 17 with an iPhone 15, on the iOS Simulator, and macOS Ventura.

## General Checklist:

## Sample Checklist

### DIFF
--- a/framework/api_vulkan_sample.cpp
+++ b/framework/api_vulkan_sample.cpp
@@ -92,7 +92,11 @@ void ApiVulkanSample::update(float delta_time)
 		view_changed();
 	}
 
-	render(delta_time);
+	if (has_render_context())
+	{
+		render(delta_time);
+	}
+
 	camera.update(delta_time);
 	if (camera.moving())
 	{
--- a/framework/api_vulkan_sample.cpp
+++ b/framework/api_vulkan_sample.cpp
@@ -92,11 +92,8 @@ void ApiVulkanSample::update(float delta_time)
 		view_changed();
 	}
 
-	if (has_render_context())
-	{
-		render(delta_time);
-	}
-
+	assert(has_render_context());
+	render(delta_time);
 	camera.update(delta_time);
 	if (camera.moving())
 	{
--- a/framework/gui.cpp
+++ b/framework/gui.cpp
@@ -446,7 +446,7 @@ BufferAllocationC Gui::update_buffers(CommandBuffer &command_buffer)
 
 	if (!draw_data)
 	{
-		return;
+		return BufferAllocationC{};
 	}
 
 	size_t vertex_buffer_size = draw_data->TotalVtxCount * sizeof(ImDrawVert);
@@ -454,7 +454,7 @@ BufferAllocationC Gui::update_buffers(CommandBuffer &command_buffer)
 
 	if ((vertex_buffer_size == 0) || (index_buffer_size == 0))
 	{
-		return;
+		return BufferAllocationC{};
 	}
 
 	std::vector<uint8_t> vertex_data(vertex_buffer_size);
@@ -590,8 +590,11 @@ void Gui::draw(CommandBuffer &command_buffer)
 	{
 		// Save vertex buffer allocation in case we need to rebind with vertex_offset, e.g. for iOS Simulator
 		auto vertex_allocation = update_buffers(command_buffer);
-		vertex_buffers.push_back(vertex_allocation.get_buffer());
-		vertex_offsets.push_back(vertex_allocation.get_offset());
+		if (!vertex_allocation.empty())
+		{
+			vertex_buffers.push_back(vertex_allocation.get_buffer());
+			vertex_offsets.push_back(vertex_allocation.get_offset());
+		}
 	}
 	else
 	{
@@ -657,8 +660,11 @@ void Gui::draw(CommandBuffer &command_buffer)
 		}
 #if defined(PLATFORM__MACOS) && TARGET_OS_IOS && TARGET_OS_SIMULATOR
 		// iOS Simulator does not support vkCmdDrawIndexed() with vertex_offset > 0, so rebind vertex buffer instead
-		vertex_offsets.back() += cmd_list->VtxBuffer.Size * sizeof(ImDrawVert);
-		command_buffer.bind_vertex_buffers(0, vertex_buffers, vertex_offsets);
+		if (!vertex_offsets.empty())
+		{
+			vertex_offsets.back() += cmd_list->VtxBuffer.Size * sizeof(ImDrawVert);
+			command_buffer.bind_vertex_buffers(0, vertex_buffers, vertex_offsets);
+		}
 #else
 		vertex_offset += cmd_list->VtxBuffer.Size;
 #endif
--- a/framework/gui.cpp
+++ b/framework/gui.cpp
@@ -440,7 +440,7 @@ bool Gui::update_buffers()
 	return updated;
 }
 
-void Gui::update_buffers(CommandBuffer &command_buffer, RenderFrame &render_frame)
+BufferAllocationC Gui::update_buffers(CommandBuffer &command_buffer)
 {
 	ImDrawData *draw_data = ImGui::GetDrawData();
 
@@ -478,6 +478,8 @@ void Gui::update_buffers(CommandBuffer &command_buffer, RenderFrame &render_fram
 	index_allocation.update(index_data);
 
 	command_buffer.bind_index_buffer(index_allocation.get_buffer(), index_allocation.get_offset(), VK_INDEX_TYPE_UINT16);
+
+	return vertex_allocation;
 }
 
 void Gui::resize(const uint32_t width, const uint32_t height) const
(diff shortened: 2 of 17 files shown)


framework/api_vulkan_sample.cpp
```diff
@@ -92,7 +92,11 @@ void ApiVulkanSample::update(float delta_time)
 		view_changed();
 	}
 
-	render(delta_time);
+	if (has_render_context())
```
asuessenbach: Maybe, as it is supposed to have a render context here, just use an assertion?
Having a check here looks like the samples should do something reasonable even without a render context...

framework/gui.cpp
```diff
@@ -440,7 +440,7 @@ bool Gui::update_buffers()
 	return updated;
 }
 
-void Gui::update_buffers(CommandBuffer &command_buffer, RenderFrame &render_frame)
+BufferAllocationC Gui::update_buffers(CommandBuffer &command_buffer)
```
asuessenbach, line 443: Two returns in this function are missing the return value (as some failing CI checks already tell you).

framework/hpp_api_vulkan_sample.cpp
```diff
@@ -87,7 +87,11 @@ void HPPApiVulkanSample::update(float delta_time)
 		view_changed();
 	}
 
-	render(delta_time);
+	if (has_render_context())
```
asuessenbach: Same comment as with api_vulkan_sample.cpp.

framework/hpp_gui.cpp
```diff
@@ -407,7 +407,7 @@ bool HPPGui::update_buffers()
 	return updated;
 }
 
-void HPPGui::update_buffers(vkb::core::HPPCommandBuffer &command_buffer) const
+BufferAllocationCpp HPPGui::update_buffers(vkb::core::HPPCommandBuffer &command_buffer) const
```
asuessenbach, line 410: Same comments as with gui.cpp.

framework/platform/platform.cpp
```diff
@@ -151,6 +151,11 @@ ExitCode Platform::main_loop_frame()
 		}
 
 		window->process_events();
+
+		if (window->should_close() || close_requested)
```
asuessenbach, line 155: Shouldn't the `active_app` be finished on `window->should_close()` as well?

asuessenbach (commented):
I can't tell anything about the iOS Simulator stuff.
But it seems, you have three identical code sequences introduced in layout_transition.cpp, pipeline_barriers.cpp, and subpasses.cpp. Maybe introduce a little helper function, instead?

samples/performance/subpasses/subpasses.cpp
```diff
@@ -50,6 +50,27 @@ Subpasses::Subpasses()
 	config.insert<vkb::IntSetting>(3, configs[Config::RenderTechnique].value, 0);
 	config.insert<vkb::IntSetting>(3, configs[Config::TransientAttachments].value, 0);
 	config.insert<vkb::IntSetting>(3, configs[Config::GBufferSize].value, 1);
+
+#if defined(PLATFORM__MACOS) && TARGET_OS_IOS && TARGET_OS_SIMULATOR
```
SaschaWillems: This looks like the code above. Can you move this e.g. to a function and put it somewhere central? The iOS simulator is a very niche use case, and this could confuse people.

samples/extensions/host_image_copy/host_image_copy.cpp
```diff
@@ -100,7 +100,7 @@ void HostImageCopy::load_texture()
 	VkFormatProperties2 format_properties_2{};
 	format_properties_2.sType = VK_STRUCTURE_TYPE_FORMAT_PROPERTIES_2;
 	format_properties_2.pNext = &format_properties_3;
-	vkGetPhysicalDeviceFormatProperties2(get_device().get_gpu().get_handle(), image_format, &format_properties_2);
```
SaschaWillems: Can you move this into a separate PR? Would make the review process a lot easier and (at least for this change) also faster.

framework/hpp_gui.cpp
```diff
@@ -653,7 +668,13 @@ void HPPGui::draw(vk::CommandBuffer command_buffer) const
 			command_buffer.drawIndexed(cmd->ElemCount, 1, index_offset, vertex_offset, 0);
 			index_offset += cmd->ElemCount;
 		}
+#if defined(PLATFORM__MACOS) && TARGET_OS_IOS && TARGET_OS_SIMULATOR
```
SaschaWillems, line 677: I'm a bit torn on "polluting" code with a special case as the iOS simulator. Can this be solved in a more elegant way? I wouldn't mind if we'd simply disable the UI on the iOS simulator instead of adding specific code for it.

SaschaWillems (changes requested):

framework/api_vulkan_sample.cpp
SRSaunders: Thanks for reviewing quickly.  Changed to assert() in https://github.com/KhronosGroup/Vulkan-Samples/pull/1173/commits/431ccf206b3bde281ade0549056fc60621bd3cb6.

framework/hpp_api_vulkan_sample.cpp
SRSaunders: Changed to assert() in https://github.com/KhronosGroup/Vulkan-Samples/commit/431ccf206b3bde281ade0549056fc60621bd3cb6.

framework/gui.cpp
SRSaunders: Saw the CI failures but was too late to fix last night :)

Fixed in https://github.com/KhronosGroup/Vulkan-Samples/pull/1173/commits/cf483bf4e0b8b8cdc1a1742a7d8bc50d856be096.

framework/hpp_gui.cpp
SRSaunders: Fixed in https://github.com/KhronosGroup/Vulkan-Samples/pull/1173/commits/cf483bf4e0b8b8cdc1a1742a7d8bc50d856be096.

framework/platform/platform.cpp
SRSaunders: All I did was move the `window->should_close()` check _as-is_ from `main_loop()` to `main_loop_frame()`.  This check is typically used at the end of a batch run to shut things down.  Given that iOS never calls `main_loop()` and instead uses its own `renderLoop()` containing `main_loop_frame()`, this logic was missing on that platform.  By placing the check inside `main_loop_frame()`, all platforms now behave the same way on batch mode shutdown.

I am not sure I understand the condition you are asking about in your question.  Would you elaborate?  Thanks.

samples/performance/subpasses/subpasses.cpp
SRSaunders: Thanks very much for reviewing this PR quickly.  I understand your concern here as I would prefer not to require  this code either.  I already have an issue logged with the MoltenVK project requesting that Metal Argument Buffers be disabled by default on the iOS Simulator. See https://github.com/KhronosGroup/MoltenVK/issues/2285.

Given the above I have reverted this change in https://github.com/KhronosGroup/Vulkan-Samples/pull/1173/commits/2af1107e5dedd25710cfbdd1e3c7c533abf0c3d9, and instead will hopefully discuss with @billhollings about the right thing to do re configuring Metal Argument Buffers for the iOS Simulator case.

samples/extensions/host_image_copy/host_image_copy.cpp
SRSaunders: Agreed.  Reverted in https://github.com/KhronosGroup/Vulkan-Samples/pull/1173/commits/ae5e98d0389a2ce85180d48a2ee630e416a143d1.  I will raise a new PR.

framework/hpp_gui.cpp
SRSaunders: I debated this myself, and actually tried the UI disable route initially.  However, I didn't like the experience, especially in batch mode where you really can't tell what is going on as the sample runs through its various configurations.  And also in the near future I would to add gesture and key input handling for iOS and without the UI present on the Simulator that would be completely inaccessible.

The reason I like the iOS Simulator is that it is easy to run and does not require an Apple Development Certificate nor a unqiue bundle identifier for iOS.  I figured this would make your project easily accessible for iOS testing without adding overhead for potential users.  In addition, the Simulator allows me to try different devices that I don't own (e.g. larger screen and various generation iOS devices).

If you don't like the **#ifdef** special case handling, the change that I made is completely generic and could be applied to all platforms, i.e. simply don't use vertex_offset > 0 and use my rebinding code for all platforms.  Not sure if you would like that but it would make the code generic.  Please let me know.

gpx1000: Maybe if we combine them into a new macro that's APPLE specific.  Say SAMPLES_PLATFORM_APPLE or similar.  I think the path of having macros is fine but there are other platforms in the Apple ecosystem that we haven't added support for yet.

framework/platform/platform.cpp
asuessenbach: Ahm, you're right.
You're change does not break any of the current code paths. In fact, the code path I was wondering about was the one a few lines above: `if (active_app->should_close())`. That one is (at least on Windows) not used.
That is, your change here is reasonable.

marty-johnson59:
Hi @gpx1000, I think this just needs testing on iOS.  Can you help with that so we can merge?  Thx

gpx1000 (approved):

PR #1174, opened 2024-09-23
Use Vulkan API 1.0 support funcs for Vulkan API 1.0 sample host_image_copy

SRSaunders:
## Description

This PR replaces Vulkan 1.1 API support function `vkGetPhysicalDeviceFormatProperties2()` with Vulkan 1.0 equivalent `vkGetPhysicalDeviceFormatProperties2KHR()` for Vulkan API 1.0 sample _host_image_copy_. This prevents runtime failures (undefined functions) on Vulkan drivers that enforce the specified API version.

I did not up-version the sample to Vulkan API 1.1 since it appears that Vulkan API 1.0 was the intended API version, with extensions enabled by `VK_KHR_get_physical_device_properties2`.

Tested on macOS Ventura and iOS 17 using both the Vulkan loader and MoltenVK standalone.

## General Checklist:

## Sample Checklist

### DIFF
--- a/samples/extensions/host_image_copy/host_image_copy.cpp
+++ b/samples/extensions/host_image_copy/host_image_copy.cpp
@@ -93,11 +93,11 @@ void HostImageCopy::load_texture()
 	// Note: All formats that support sampling are required to support this flag
 	// So for the format used here (R8G8B8A8_UNORM) we could skip this check
 	// The flag we need to check is an extension flag, so we need to go through VkFormatProperties3
-	VkFormatProperties3 format_properties_3{};
+	VkFormatProperties3KHR format_properties_3{};
 	format_properties_3.sType = VK_STRUCTURE_TYPE_FORMAT_PROPERTIES_3_KHR;
 	// Properties3 need to be chained into Properties2
-	VkFormatProperties2 format_properties_2{};
-	format_properties_2.sType = VK_STRUCTURE_TYPE_FORMAT_PROPERTIES_2;
+	VkFormatProperties2KHR format_properties_2{};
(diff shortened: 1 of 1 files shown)


asuessenbach (dismissed):

SaschaWillems:
Shouldn't we also use `VkFormatProperties2KHR` instead of `VkFormatProperties2` then?

SRSaunders:
> Shouldn't we also use VkFormatProperties2KHR instead of VkFormatProperties2 then?

That's a natural question, but the Vulkan Spec seems to say otherwise.  See https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPhysicalDeviceFormatProperties2.html.  Both versions of the function (API 1.1 and 1.0) reference `VkFormatProperties2`, and this seems to be borne out by the header files.

However, for readability and consistency, I could make the change since the following is also true (see https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkFormatProperties2.html):

```
 // Provided by VK_KHR_get_physical_device_properties2
typedef VkFormatProperties2 VkFormatProperties2KHR;
```

The current change works as is, but please let me know what you would prefer.

SaschaWillems:
If we use the KHR suffix in one place, we should use it in the other places too to make it consistent.

SaschaWillems:
Btw. VK_EXT_host_image_copy requires 1.1. ,so upping the api version of that sample would be just fine. Since I wrote that sample I can take care of that.

asuessenbach (approved):

gary-sweet (approved):

SaschaWillems (approved):

PR #1187, opened 2024-10-09
shader_debugprintf: support new VVL-DEBUG-PRINTF message and fix VVL version check for API selection

### LINKED ISSUE #1184: shader_debugprintf problems with new VulkanSDK 1.3.296
Two problems have arisen with the new SDK 1.3.296:

1. The debug messages from the validation layer are now `VVL-DEBUG-PRINTF` instead of `WARNING-DEBUG-PRINTF`.  This prevents the current callback logic from capturing the debug output values and the vertex output area is blank.
2. I made an incorrect assumption about the value of `instanceApiVersion` returned from `vkEnumerateInstanceVersion()`.  It seems that this value does not match the SDK version on all platforms (e.g. matches on macOS, but not on Windows).  Given this failed assumption, the logic I put in place to automatically use `VK_API_VERSION_1_1` for SDKs > 1.3.290 does not work properly on Windows. This causes the sample to bog down and run very slowly with SDK 1.3.296.

I will submit a PR to fix this (again).  I'm sorry about the mistake I made in 2 above, but I could not have anticipated item 1.  It seems the interaction with the Validation Layer is sensitive to change and breakage.

SRSaunders:
## Description

Fixes two issues that arose with Vulkan SDK 1.3.296:

1. Supports new `VVL-DEBUG-PRINTF` callback message.  Previous SDKs used `WARNING-DEBUG-PRINTF` or `UNKNOWN-DEBUG-PRINTF`.  Without this fix the debug data is not available in the UI Overlay.
2. Fixes my incorrect assumption that the Vulkan instance version matched the SDK version for all platforms - true on macOS but not true for Windows and Linux.  This version is used to set the API level for the sample, which is important for performance and to avoid a previous defect in the Vulkan Validation layer.  I have replaced the instance version check with a Validation Layer version check which is portable across all platforms: Win, Linux, macOS.  Without this fix, performance is poor on Windows and Linux when using Vulkan SDK 1.3.296.

Fixes #1184.

Tested on Windows 10, Manjaro Linux, and macOS Ventura using Vulkan SDKs 1.3.290 and 1.3.296.

I hope this is the last time I have to fix this.  It seems that VVL changes can easily break this sample.

## General Checklist:

## Sample Checklist


samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -433,23 +433,30 @@ const std::vector<const char *> ShaderDebugPrintf::get_validation_layers()
 // This sample overrides the instance creation part of the framework to chain in additional structures
 std::unique_ptr<vkb::Instance> ShaderDebugPrintf::create_instance(bool headless)
 {
-	uint32_t instanceApiVersion;
-	VK_CHECK(vkEnumerateInstanceVersion(&instanceApiVersion));
+	uint32_t layer_property_count;
```
SaschaWillems, line 430: Can you add a bit of an explanation in here as to why this is needed? That would make it easier to follow the code.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -433,23 +433,30 @@ const std::vector<const char *> ShaderDebugPrintf::get_validation_layers()
...
 
 	// When VK_EXT_layer_settings is available at runtime, the debugPrintfEXT layer feature is enabled using the standard framework
-	// For backwards compatibility with SDKs < 1.3.272 without VK_EXT_layer_settings, the remainder of this custom override is required
+	// For this case set Vulkan API version and return via parent class, otherwise the remainder of this custom override is required
 	if (std::any_of(available_instance_extensions.begin(),
 	                available_instance_extensions.end(),
 	                [](VkExtensionProperties const &extension) { return strcmp(extension.extensionName, VK_EXT_LAYER_SETTINGS_EXTENSION_NAME) == 0; }))
 	{
 		// debugPrintfEXT layer feature requires Vulkan API 1.1, but use API 1.2 until VVL performance fix is available in SDKs > 1.3.290
 		// See VVL issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7562 for defect and fix information
```
SaschaWillems: The issue referenced here is closed. Does this comment still apply?

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -433,23 +433,30 @@ const std::vector<const char *> ShaderDebugPrintf::get_validation_layers()
...
-		set_api_version(instanceApiVersion <= VK_MAKE_API_VERSION(0, 1, 3, 290) ? VK_API_VERSION_1_2 : VK_API_VERSION_1_1);
+		set_api_version(vvl_properties != layer_properties.end() && vvl_properties->specVersion <= VK_MAKE_API_VERSION(0, 1, 3, 290) ? VK_API_VERSION_1_2 : VK_API_VERSION_1_1);
```
SaschaWillems: A minor nitpick, but this one-liner is very hard to read. Could you make this easier to read by putting it outside of set_api_version and dividing and commenting it?

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -480,7 +487,7 @@ std::unique_ptr<vkb::Instance> ShaderDebugPrintf::create_instance(bool headless)
 	VkApplicationInfo app_info{VK_STRUCTURE_TYPE_APPLICATION_INFO};
 	app_info.pApplicationName = "Shader debugprintf";
 	app_info.pEngineName      = "Vulkan Samples";
-	app_info.apiVersion       = instanceApiVersion <= VK_MAKE_API_VERSION(0, 1, 3, 290) ? VK_API_VERSION_1_2 : VK_API_VERSION_1_1;
+	app_info.apiVersion       = vvl_properties != layer_properties.end() && vvl_properties->specVersion <= VK_MAKE_API_VERSION(0, 1, 3, 290) ? VK_API_VERSION_1_2 : VK_API_VERSION_1_1;
```
SaschaWillems: Same as above, IMO that statement is too complex and should not be a one-liner on the right side of the assignment.

SaschaWillems (changes requested):
Thank you very much for this PR. I do have some remarks though, mostly related to comment and code structure. I think it's important that people can easily follow understand the changes ;)

SRSaunders:
Thanks @SaschaWillems for the feedback.  I am away on vacation this week, but will make the requested changes when I am back.

UPDATE:  Back now and changes submitted in https://github.com/KhronosGroup/Vulkan-Samples/pull/1187/commits/0dc49635f09d33687b3999869d2196d48fe06789.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: Done

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: Yes, the referenced issue is still useful as a more detailed explanation of why API version selection is needed for this example.  I have added a bit of text to indicate when the VVL issue was fixed.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: Simplified the code here.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: As above, simplified the code here.

asuessenbach (dismissed):

SaschaWillems:
No idea why, but with this PR and the latest SDK (1.3.296) and in windows, this sample is now again running with less than 1 fps. Forcing it to use VK 1.2 is somehow even slower (0 or inf fps).

If I force VK 1.0 performance is fine, but I don't get any debug output.

Not sure what is happening here and why this sample is so problematic. The debug printf sample from m own samples repo works just fine no matter the api version :/

SRSaunders:
> No idea why, but with this PR and the latest SDK (1.3.296) and in windows, this sample is now again running with less than 1 fps. Forcing it to use VK 1.2 is somehow even slower (0 or inf fps).

Very strange.  Can I ask you to recheck before and after this PR, but being careful with your SDK version selection and project gen/build?  I did a lot of testing with old and new SDKs on Windows 10, Linux and macOS before submitting originally.  I will go back and test again to see if I can somehow duplicate what you are seeing.

> If I force VK 1.0 performance is fine, but I don't get any debug output.

Debug PrintF requires Vulkan 1.1 or later.  So no surprise that you are not getting debug output with API 1.0.

> The debug printf sample from my own samples repo works just fine no matter the api version

I suspect your repo's sample relies on the instrinsic Debug PrintF capability at the shader level on Windows.  However, this is not cross-platform portable.  Whereas the Vulkan-Samples one uses the VVL version of the feature all the time.  Perhaps that is why you are seeing a difference at least on Windows.  Again, I will so back and see if I can verify this.

SaschaWillems:
It also happens with the old code (before this PR). I only have SDK 1.3.296 installed. 

So probably a regression in the validation layers?

SRSaunders:
Ok, I have rechecked this PR on Windows 10, and even fast-forwarded my local branch to current main HEAD just to make sure.  I am using Vulkan SDK 1.3.296.0 with my Radeon RX6600XT GPU.  My Vulkan Configurator has been reset to default settings.

Before this PR I get:
main only

After this PR I get:
shader_debugprintf FF

Is it possible that your Vulkan Configurator has a custom setting that is interfering with the sample?  Or possibly a difference between AMD and nVidia GPUs?  Just grasping at straws since I cannot duplicate your issue and the 1.3.296 VVL seems to be working correctly using API 1.1 for debug printf.

asuessenbach (commented):
With this change, I have to distinguish two cases:
1. VulkanConfigurator is running
        VK_EXT_LAYER_SETTINGS_EXTENSION_NAME is available
        instance creation is done by VulkanSample::create_instance (line 469)
        render speed is high
        debug_utils_message_callback is never called, thus no debugprintf output
2. VulkanConfigurator is not running
        VK_EXT_LAYER_SETTINGS_EXTENSION_NAME is not available
        instance creation is done locally (line 523)
        render speed is extremely low
        debug_utils_message_callback is called, with higher rate than the frame rate

Note, in case 2, you're using `VkValidationFeaturesEXT`, which is part of `VK_EXT_VALIDATION_FEATURES_EXTENSION_NAME`. But you don't ask for it in the `ShaderDebugPrintf` constructor (or anywhere else). And in fact, that extension is not supported on my machine. Strange, that the VVL doesn't cry there.

SaschaWillems:
That would explain why it's so slow for me. I never ran that sample with the VulkanConfigurator running. That's case 2.

SRSaunders:
Thanks @asuessenbach for pointing out the missing `VK_EXT_validation_features` extension.  I have made a few changes that might make a difference as follows:

1. Moved layer settings out of the constructor, and into `ShaderDebugPrintf::create_instance()`.  Now it will run only if the `VK_EXT_layer_settings` extension is available.  This part is for encapsulation only and will not change behaviour.
2. Added and enabled the `VK_EXT_validation_features` extension when the `VK_EXT_layer_settings` extension is not available at runtime.  This **might** change behaviour, but I am concerned about @asuessenbach's comment that the extension is not available on his machine.  I'm not sure how that is possible.
3. Fixed an incorrect string comparison operation for `VK_EXT_layer_settings` in `[HPP]Instance::[HPP]Instance()`.  This was my mistake from an earlier PR.  This could have prevented proper specification of the validation layer feature settings when `VK_EXT_layer_settings` is active.  Again, this could change behaviour.

These changes may not be the final solution as I have observed the following when testing:

1. Linux (Manjaro) using Vulkan 1.3.295 (from pkg mgr) and VVL 1.3.290 (from pkg mgr): this PR works properly (good frame rate, debug data available) when running with **vkconfig** and without. `VK_EXT_layer_settings` is only available when **vkconfig** is active. In this case the debug data is available both in the UI and in the stdout console.  No performance issues are visible in either case.
2. macOS (Ventura) using Vulkan SDK 1.3.296: this PR works properly (good frame rate, debug data available) when running with **vkconfig** and without. `VK_EXT_layer_settings` is available both when **vkconfig** is inactive **and** active - this is a difference vs Linux. In the latter case (**vkconfig** active) the debug data is available both in the UI and in the stdout console.  No performance issues are present.  Also tested with Vulkan SDK 1.3.290 and the results are the same - no performance problems.  The only issue is that vkconfig does not appear to recognize the repeated message limit for the new VVL-* messages (vs. the previous INFO-* or WARNING-* messages, etc).  A minor issue but likely a bug.
3. Windows 10 using Vulkan SDK 1.3.296 with my AMD 6600XT GPU: this PR works properly (good frame rate, debug data available) when running without **vkconfig** only. When **vkconfig** is active, the sample will not start and complains about an unsupported extension during `vkCreateInstance()`.  However, `VK_EXT_layer_settings` **is** available during enumeration when **vkconfig** is active. Something very strange is going on here - either a bug on the Windows side or something I do not understand.  I am not sure how `VK_EXT_layer_settings` can be enumerated but not supported.  See my console output in this case:

nolayerext

In summary:

1. Linux: works properly using VVL 1.3.290 with and without **vkconfig**.  Can't test VVL 1.3.296 since it is not yet available as a package for my Manjaro distro.
2. macOS: works properly using VVL 1.3.290 and 1.3.296 with and without **vkconfig**.
3. Windows 10 on AMD 6600XT GPU: works properly using VVL 1.3.296 without **vkconfig** only.

Lastly, I thought `VK_EXT_layer_settings` was meant to replace and deprecate `VK_EXT_validation_features`.  I don't understand why `VK_EXT_layer_settings` is available all the time on macOS, but for Windows and Linux seems to be enabled only when **vkconfig** is running.  This seems incorrect to me.  Can you explain this?

SRSaunders:
Ok, I think I have finally figured it out.  It appears that you _don't_ need to actually enable the `VK_EXT_layer_settings` extension in order to use it.  I’m not sure if this is a feature or a bug.  In any case, I have updated the sample and `[HPP]Instance::[HPP]Instance()` to check for _availability_ of the extension vs. enablement.  This approach works across all platforms and behaviours appear to be consistent now:

1. Sample is tolerant of Vulkan SDK versions:  tested against VVL 1.3.290 (Win, Linux, macOS) and 1.3.296 (Win, macOS)
2. Sample is tolerant of **vkconfig** running or not running.  The only thing to be careful of when running **vkconfig** is to make sure "Limit Duplicated Messages" is turned off - otherwise debug callback messages will be suppressed and the debug output UI will be blank.

asuessenbach:
AFAIK, those two extensions (`VK_EXT_layer_settings` and `VK_EXT_validation_features`) are not supported by any NVIDIA GPU, but are provided by a layer injected by for example the VulkanConfigurator. That might explain why it's that slow.

Besides that, just to make sure it has been noted: As `VK_EXT_validation_features` is deprecated in favour of `VK_EXT_layer_settings`, using `VK_EXT_validation_features` would just be a fallback solution. Don't know, if it's worth to have that. And you should bail out in a friendly way, if none of those extensions is available, maybe with a hint to the VulkanConfigurator.

SaschaWillems:
Welp, still sub 1 fps for me with latest SDK and vkconfig NOT running.

Just let me know when it's in a state were I should test.

If we can't get this to work, we may simply go back to the initial version and maybe remove the debug output and tell people to attach a graphics debugger.

SaschaWillems:
Oddly I can't get the sample to run at acceptable frame rates anymore. Even when removing the debug output. The sample from my repo (which this was initially based on) works fine. I'm officially confused right now -.-

SRSaunders:
Thanks @asuessenbach for the info re nVidia GPUs.  I have an AMD card and I guess this is the difference here.

@SaschaWillems would you please test using this PR with vkconfig _running_ and let me know the result?  I presume you are using an nVidia GPU - please confirm.

If this works, and as @asuessenbach suggests, I will try to detect this condition and offer a message to nVidia users.

SaschaWillems:
> If this works, and as @asuessenbach suggests, I will try to detect this condition and offer a message to nVidia users.

If we get to a point where we have to show a message under certain conditions to users of a certain vendor we're not heading where I'd like our samples to head. I'd rather remove the output debug stuff then.

SRSaunders:
@SaschaWillems I understand.  However I’d still like to track this down if possible and you testing on Nvidia with vkconfig active would give more information.  I can’t do this test myself.  Thx.

SaschaWillems:
Windows 11 23H2, nvidia RTX 4070, latest Vulkan developer driver, SDK 1.3.296.

And I get <1 fps even with vkconfig up and running:

image

I'm pretty sure that the sample ran fine when I initially wrote it, but not sure why it no longer does.

Can't rule out a configuration issue on my side 100%, but not sure where to start looking.

SRSaunders:
I just added a minor hygiene change to use `vk::ExtensionProperties` vs. `VkExtensionProperties` in **HPPInstance()**.  Also updated some comments and decided to explicitly request required GPU features for **debugPrintfEXT** as per docs.

More importantly, I was able to find an nVidia GPU to test this.  I have narrowed down what causes the slowdown and am now convinced it is a VVL debugPrintfEXT defect on that GPU platform.  Simply by disabling the following debugPrintfEXT feature enablement lines I can restore FPS performance on nVidia machines for both **vkconfig** running and not running cases.  Unfortunately this drops the debug info, but hopefully this is a temporary thing until this issue can be addressed.

```
...
	//add_layer_setting(layerSetting);
...
	instance_create_info.pNext = nullptr; //&validation_features;
...
```

I will respond on the other thread to @spencer-lunarg to see if he can help.

spencer-lunarg:
@SRSaunders before we had the Slow Down on for Vulkan 1.1 and 1.2/1.3 were good... is that still the case or is it now for all versions?

SRSaunders:
When using an nVidia GPU with SDK 1.3.296, it slows down for all API versions.  When using SDK 1.3.290 with the same setup (nVidia GPU), the sample works properly when using API 1.2 - as expected per previous discussion.

For AMD GPUs (and Apple Silicon on macOS) with SDK 1.3.296 everything works properly when using API 1.1

spencer-lunarg:
ok, so the problem has be isolated down to an NVIDIA GPU (I was testing on Intel and found no issues)... Later tonight I will be back at my desk and can try again on my NVIDIA machine

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -27,8 +27,8 @@ VKAPI_ATTR VkBool32 VKAPI_CALL ShaderDebugPrintf::debug_utils_message_callback(
     const VkDebugUtilsMessengerCallbackDataEXT *pCallbackData,
     void                                       *pUserData)
 {
-	// Look for Validation Layer message id names: WARNING-DEBUG-PRINTF or UNASSIGNED-DEBUG-PRINTF (have observed UNASSIGNED with older Vulkan SDKs)
-	if (strcmp(pCallbackData->pMessageIdName, "WARNING-DEBUG-PRINTF") == 0 || strcmp(pCallbackData->pMessageIdName, "UNASSIGNED-DEBUG-PRINTF") == 0)
+	// Look for Validation Layer message id names: VVL-DEBUG-PRINTF or WARNING-DEBUG-PRINTF or UNASSIGNED-DEBUG-PRINTF (have observed WARNING and UNASSIGNED with older Vulkan SDKs)
+	if (strcmp(pCallbackData->pMessageIdName, "VVL-DEBUG-PRINTF") == 0 || strcmp(pCallbackData->pMessageIdName, "WARNING-DEBUG-PRINTF") == 0 || strcmp(pCallbackData->pMessageIdName, "UNASSIGNED-DEBUG-PRINTF") == 0)
```
spencer-lunarg, line 34: sorry this was me (didn't realize people where name string matching this), We got rid of "UNASSIGNED" and this was a "info" not a "warning" so hopefully we never have to change that name again

spencer-lunarg:
@SRSaunders I saw

>  It appears that you don't need to actually enable the VK_EXT_layer_settings extension in order to use it

So DebugPrintf I came out in 2020 and Layer Settings came out in 2023, so it was designed to be use the "old" way, 2023 we made it possible to use with Layer Settings (which is what vkconfig uses as well)

in 2024 we have been slowly pouring more effort into GPU-AV and DebugPrintf was there so we merged it in, added features, fixed bugs, so lots have recently was touching it. I tried to update the VVL docs for it https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/debug_printf.md

Not in the 1.3.296 SDK (didn't want to rush it) but the next future SDK will have to go `set VK_LAYER_PRINTF_ONLY_PRESET=1` for people who "just want to turn it on now" and not fuss with settings

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: This was not a difficult change to incorporate, so no worries.  Also happy to hear this should be more stable going forward.

SRSaunders:
> I tried to update the VVL docs for it https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/debug_printf.md

@spencer-lunarg thanks for updating the docs.  This is the same doc I referred back to today when digging into the issue we are facing here.

> ... but the next future SDK will have to go set VK_LAYER_PRINTF_ONLY_PRESET=1 for people who "just want to turn it on now" and not fuss with settings

Good to know there will be an even simpler way to enable debugPrintfEXT going forward.

While we are on the topic of `VK_EXT_layer_settings`, I would like to ask about the different behaviours of this extension cross-platform.  For instance, on Windows and Linux, it seems this extension is not advertized during enumeration unless **vkconfig** is running.  And in addition, if you try to enable it on Windows (even with **vkconfig** running), it will bail out with an extension not found error during `vkCreateInstance()` - kind of unexpected.  This is in contrast to what happens on macOS where the extension is available at all times (not just when **vkconfig** is running), and can be enabled during instance creation without bailing out.  Perhaps I am off base or don't understand it properly, but I thought that `VK_EXT_layer_settings` was meant to allow setting of layer features under program control during instance creation, and should not depend on whether **vkconfig** is running or not.  Would you help clarify this for me?

spencer-lunarg:
@SRSaunders 

> but I thought that VK_EXT_layer_settings was meant to allow setting of layer features under program control during instance creation, and should not depend on whether vkconfig is running or not

That is correct... something is wrong here (and of course it involves 3 different things) ... do me a favor, for this thing, create a new [VVL issue](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues) and I can work with Christope (who maintains both VkConfig and wrote the `VK_EXT_layer_settings` extension) to figure out what is going on.

The layer has to define its settings in way that can be used, for example you can set [printf_to_stdout](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/9935cdd6f9524b1da1e140393fa69a0a1b72125d/layers/VkLayer_khronos_validation.json.in#L849) by setting `VK_EXT_layer_settings::pSettingName` to `"printf_to_stdout"`. We (VVL) also use the [layer setting code in VUL](https://github.com/KhronosGroup/Vulkan-Utility-Libraries?tab=readme-ov-file#vulkanlayersettings) so that it will automatically allow people to set `set VK_LAYER_PRINTF_TO_STDOUT=1` to do the same thing

What gets ugly is that we have legacy ways to do settings sadly...until the commit right after we branched for 1.3.296, DebugPrintf and GPU-AV both did Shader Instrumentation, but would step over each other, so we needed a way to prevent both used at the same time. This was done with this very ugly `GPU_BASED_DEBUG_PRINTF` enum thing and honestly I want no one to waste time thinking about it, the next SDK will have a simple `printf_enable` setting to turn off/on DebugPrintf

spencer-lunarg:
Also want to clarify that `enables=VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT` is correct what you have, it is the "original" way of setting things, seems the issue 

So tested, I have a laptop with  AMD Radeon 780M integrated and NVIDIA 4060 (using env variables to turn on, not vkconfig, shouldn't really matter in theory)

- Samples ToT | AMD - 1.3.290 SDK - **slow** (1 FPS)
- Samples ToT | AMD - 1.3.296 SDK - **normal**
- Samples ToT | AMD - ToT VVL - **normal**
- Samples PR 1187 | AMD - 1.3.290 SDK - **slow** (1 FPS)
- Samples PR 1187 | AMD - 1.3.296 SDK - **normal**
- Samples PR 1187 | AMD - ToT VVL - **normal**

With NVIDIA

- Samples ToT | NVIDIA - 1.3.290 SDK - **slow** (1 FPS)
- Samples ToT | NVIDIA - 1.3.296 SDK -  **slow**
- Samples ToT | NVIDIA - ToT VVL -  **slow**
- Samples PR 1187 | NVIDIA - 1.3.290 SDK - **slow** (1 FPS)
- Samples PR 1187 | NVIDIA - 1.3.296 SDK -  **slow**
- Samples PR 1187 | NVIDIA - ToT VVL -  **slow**

so basically this PR is not fix/breaking things from my end... I agree this seems like something on the VVL side (we reopened the issue and I will look more into this weekend) but everything in this PR seems sane ... it should not take "hacks" to get DebugPrintf to work, if it does, we failed in VVL badly

... from a quick glance this seems to be a sync issue within how GPU-AV is working (we had to start using timeline semaphores under the hood to get things to be faster in GPU-AV, before we did a big vkQueueWaitIdle, ... but we share this code with Sync Validation, which has also been active and I feel we regressed here)

SRSaunders:
@spencer-lunarg I have raised issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8760 as requested.

> so basically this PR is not fix/breaking things from my end... I agree this seems like something on the VVL side (we reopened the issue and I will look more into this weekend) but everything in this PR seems sane

Thanks for reviewing and good to know this PR appears correct from your side.  I look forward to any discoveries you make regarding slowdowns we are seeing with nVidia GPUs.

SRSaunders:
Following interaction with the VVL team, I think this is now ready for review.  A couple of learnings:

1. The `VK_EXT_layer_settings` extension is trickier than I first realized.  On Windows and Linux it is primarily a _layer_ instance extension and not a _driver_ extension.  On macOS things are a bit different where `VK_EXT_layer_settings` is made visible by the MoltenVK _driver_ as well.  However, to detect its availability in general you need to query the _layer_ and not the _driver_ during instance enumeration.  Thanks to @spencer-lunarg + team and armed with this new (to me) information, I have modified the sample to now do proper detection.  This results in all platforms using the layer settings path independent of whether **vkconfig** is running or not.  The legacy `VkValidationFeatureEnableEXT` path is still present in the code, but will be used only with older SDKs that don't have `VK_EXT_layer_settings` within the VVL layer.
2. The above findings also made me realize that the general framework for adding a layer setting (in `[HPP]Instance()`) was not using the correct criteria to chain in the `VkLayerSettingsCreateInfoEXT` struct during instance creation.  Testing for the presence of `VK_EXT_layer_settings` in the _driver_ will not give the right answer.  And looking for which layer to check would only add a bunch of unnecessary complexity to the code.  So I simplified things and now leave it up to the sample to determine if layer settings is supported, and if so, to push a layer setting using `add_layer_setting()`.  In `[HPP]Instance()` I now only check for the presence of layer setting entries in the `required_layer_settings` vector.  This puts the onus on the sample to make sure layer settings entries are supported.  However, this is not very risky since `vkCreateInstance()` will throw away any layer settings that don't match and should not complain.
3. Lastly, the observed performance slow-downs on various platforms and SDK versions has these mitigations:
a) The slow-downs that @spencer-lunarg observed with older SDKs in https://github.com/KhronosGroup/Vulkan-Samples/pull/1187#issuecomment-2439205768 has been solved.  While I was picking the correct API version 1.2 for older SDKs, I had added code that explicitly requested support for _timeline semaphores_.  It turns out that while the VVL debugPrintfEXT feature requires this and implicitly enables it under the covers, explicitly requesting support in the sample breaks performance.  I am not sure why this is the case, but I will leave it to @spencer-lunarg to decide if this is an issue or not.  Nonetheless, I have removed this from the sample and older SDKs now work without performance degradation.
b) The slow-down that was visible for nVidia GPUs running SDK 1.3.296 is apparently solved by https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8766, which will be available with the next SDK.  I have not been able to test this yet, but will report back once I can.  **UPDATED: fix verified as working on my nVidia GPU machine**. However, the only workaround for now is to use SDK 1.3.290 for nVidia GPU users until the next SDK is released.
c) By fixing the logic to test for the VVL version (and not the instance version) when selecting API 1.1 vs. 1.2, this PR does solve performance issues observed with SDK 1.3.296 for Windows AMD, Linux AMD, and macOS.  I'm not sure about Linux nVidia as I cannot test that combination.

spencer-lunarg:
>  It turns out that while the VVL debugPrintfEXT feature requires this and implicitly enables it under the covers, explicitly requesting support in the sample breaks performance.

Is this still true? I tried to add it explicitly in this PR and didn't see it slow down (the issue was the old pre-1.3.290 SDK and should be patched now)

SRSaunders:
> > It turns out that while the VVL debugPrintfEXT feature requires this and implicitly enables it under the covers, explicitly requesting support in the sample breaks performance.
> 
> Is this still true? I tried to add it explicitly in this PR and didn't see it slow down (the issue was the old pre-1.3.290 SDK and should be patched now)

I observed this performance impact when using SDK 1.3.290 with API 1.2 and timeline semaphores explicitly enabled.  With SDK 1.3.296 using API 1.1 with timeline semaphores enabled, performance was fine.  So I removed the explicit enablement of timeline semaphores and now I get consistent performance using: a) API 1.2 with SDKs <= 1.3.290, and b) API 1.1 with SDKs >= 1.3.296 (aside from the nVidia issue mentioned above).

SRSaunders:
Given the long discussion and in case it wasn't clear, this PR is now ready to go.

Note for nVidia users (Windows) using SDK 1.3.296: A fix is also required from the VVL which will ship in the next SDK.  In the interim, nVidia users should run using this PR combined with SDK 1.3.290 for the _shader_debugprintf_ sample.

Note: This also removes the framework check in **[hpp_]instance.cpp** for `VK_EXT_layer_settings` when using `add_layer_setting()`.  This makes using the layer settings feature possible when the extension is not present in the driver, but only present in the layer.  It's now up to the sample itself to check for support in the layer if required.

SaschaWillems:
Quick update: We discussed this in a recent call and will probably wait with the merge until the next SDK is available to make sure we can simply point people to update their SDK in case of any problems with this sample. Hope that's okay with you.

SRSaunders:
@SaschaWillems that’s fine re the debugprintf sample.  My only concern is the fix for the `add_layer_setting()` framework capability.  Not sure if you want that separated out and brought forward into a new PR for merging now.  Let me know.

SaschaWillems:
Is that fix important (sorry, I got kinda lost with all the discussion in this issue)? If so, splitting it into a separate PR is fine, if not it's fine if we merge all of this once the new SDK is out.

SRSaunders:
I guess we can wait givin the `add_layer_setting()` feature is not used extensively at this point.  I will leave it up to @JoseEmilio-ARM to comment if an earlier fix would be preferable.

SRSaunders:
@SaschaWillems - a new Vulkan SDK 1.4.304.0 has been released and this PR can be retested against Nvidia GPUs.  I only have access to an AMD card at this point and haved retested this PR on Windows+AMD and macOS+AMD using the new SDK.  I will test against Linux+AMD as soon as the new SDK is available as an updated package for my distro (Manjaro Linux).

I would appreciate someone testing this PR with Windows+Nvidia against the new SDK to prove out the solution.  If this checks out perhaps this long-delayed PR could be merged.

(As an aside, the VVL team may have broken things again with the new SDK but for macOS only this time.  It seems they have added a requirement for `VkPhysicalDeviceVulkanMemoryModelFeatures::vulkanMemoryModel` and `VkPhysicalDeviceVulkanMemoryModelFeatures::vulkanMemoryModelDeviceScope` when the Validation Layer's debug printf feature is enabled.  This was not the case for previous SDKs, and unfortunately these features are not available on MoltenVK.  I plan to raise an issue with the VVL project asking why this change was made.  However, this does not affect the current PR and _shader_debugprintf_ sample which should be correct for all platforms.  The sample + this PR continue to work fine for other platforms as well as previous SDKs on macOS.  See issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9386)

SaschaWillems (dismissed):
Thank you very much. Can confirm that this no finally runs at proper frame rates on nvidia + windows + latest SDK.

SRSaunders:
> Can confirm that this no finally runs at proper frame rates on nvidia + windows + latest SDK.

Thanks very much @SaschaWillems for testing and approving.  Perhaps @JoseEmilio-ARM could review as well using the new SDK 1.4.304.0.

asuessenbach:
Unfortunately, it doesn't work as expected on my end, yet.
This sample runs fast now, no matter if the VulkanConfigurator is running or not.
But I don't see any debug output!

When I add a check against "VVL-DEBUG-PRINTF" to the `debug_utils_message_callback`, I get debug output when The VulkanConfigurator is _not_ running, but _no_ out put when the VulkanConfigurator is running. At least, it's fast in both cases.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -84,6 +71,11 @@ ShaderDebugPrintf::~ShaderDebugPrintf()
 
 void ShaderDebugPrintf::request_gpu_features(vkb::PhysicalDevice &gpu)
 {
+	// debugPrintfEXT requires fragmentStoresAndAtomics, vertexPipelineStoresAndAtomics, and shaderInt64
+	gpu.get_mutable_requested_features().fragmentStoresAndAtomics       = VK_TRUE;
+	gpu.get_mutable_requested_features().vertexPipelineStoresAndAtomics = VK_TRUE;
+	gpu.get_mutable_requested_features().shaderInt64                    = VK_TRUE;
```
asuessenbach: Shouldn't we check for support of those features before we request them? And bail out if they're not supported?
Like
```
auto const & supportedFeatures = gpu.get_features();
if (supportedFeatures.fragmentStoresAndAtomics && supportedFeatures.vertexPipelineStoresAndAtomics && supportedFeatures.sharedInt64)
{
        auto & requestedFeatures = gpu.get_mutable_requested_features();
        requestedFeatures.fragmentStoresAndAtomics       = VK_TRUE;
        requestedFeatures.vertexPipelineStoresAndAtomics = VK_TRUE;
        requestedFeatures.shaderInt64                    = VK_TRUE;
}
else
{
// some bail-out message here
}
```

asuessenbach (commented):
Everything runs fine now!
Just one little issue.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SaschaWillems: > Shouldn't we check for support of those features before we request them? And bail out if they're not supported?
Like

Isn't our framework supposed to catch that and show a message "Unsupported feature"?

asuessenbach: I think, we had this discussion a couple of times now. And we have different opinions on it.

You're right, you would get some `VK_ERROR_FEATURE_NOT_PRESENT` on device creation. But you don't get information about what feature is not present, in case you're interested. And you get the error somewhat later, while other stuff might have been done and you potentially have to clean up all that stuff.

In this specific case, there's no fallback on a missing feature. But if there was, you would check it here anyways.

To sum up: I would consider it a clean programming style to first check for support before requesting it.

SRSaunders: > ... you would get some VK_ERROR_FEATURE_NOT_PRESENT on device creation. But you don't get information about what feature is not present ...

This is correct, and here is the error message:

```
[error] Error Message: Cannot create device : ERROR_FEATURE_NOT_PRESENT
[error] Failed when running application Shader debugPrintf
Program ended with exit code: 0
```

I have pushed a new commit as suggested by @asuessenbach.  Here is the new message:

```
[error] Error Message: Selected GPU does not support features fragmentStoresAndAtomics, vertexPipelineStoresAndAtomics, and/or shaderInt64 : ERROR_FEATURE_NOT_PRESENT
[error] Failed when running application Shader debugPrintf
Program ended with exit code: 0
```

If @SaschaWillems, @asuessenbach, and @JoseEmilio-ARM could re-review that would be great.  Thx.

Sorry for all the force-pushes for the Copyright date issues - while I iniitally missed some files, I'm not sure I completely understand the algorithm used to check status when there are multiple copyright lines and date ranges.  However, should finally be good now.

SaschaWillems (dismissed):

asuessenbach (dismissed):

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -84,10 +71,25 @@ ShaderDebugPrintf::~ShaderDebugPrintf()
...
 {
+	auto const &supportedFeatures = gpu.get_features();
+	auto       &requestedFeatures = gpu.get_mutable_requested_features();
+
+	// debugPrintfEXT requires fragmentStoresAndAtomics, vertexPipelineStoresAndAtomics, and shaderInt64
+	if (supportedFeatures.fragmentStoresAndAtomics && supportedFeatures.vertexPipelineStoresAndAtomics && supportedFeatures.shaderInt64)
+	{
+		requestedFeatures.fragmentStoresAndAtomics       = VK_TRUE;
+		requestedFeatures.vertexPipelineStoresAndAtomics = VK_TRUE;
+		requestedFeatures.shaderInt64                    = VK_TRUE;
```
gary-sweet: Does this sample really require shaderInt64? My hardware doesn't support it, but if I remove this line it appears to run ok anyway.

gary-sweet (commented):
I'm also seeing a couple of validation errors:

```
[warning] 507847663 - DEBUG-PRINTF-FORMATTING: Validation Warning: [ DEBUG-PRINTF-FORMATTING ] | MessageID = 0x1e4523ef | OpString "Position = %v4f" contains a 4-wide vector modifier "%v4f", but the argument (SPIR-V Id 157) is a 3-wide vector (values might be truncated or padded)
[warning] 507847663 - DEBUG-PRINTF-FORMATTING: Validation Warning: [ DEBUG-PRINTF-FORMATTING ] | MessageID = 0x1e4523ef | OpString "Position = %v4f" contains a 4-wide vector modifier "%v4f", but the argument (SPIR-V Id 157) is a 3-wide vector (values might be truncated or padded)
```

These are easily fixed by changing the last line in scene.vert to say `debugPrintfEXT("Position = %v3f", outPos)`

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
SRSaunders: @gary-sweet thanks for your comments.  I will fix the debugPrintfEXT format in the shader.  Regarding the requested features, I could simply remove them all and have the validation layer implicitly request them as needed.  The reason I added these was to reduce the log clutter as I thought these were the mandatory ones.  However, with recent GPU-AV changes in the VVL, there may actually be even more features implicitly requested by the VVL.  To help with this decision, can you tell me what platform and SDK version are you using?  And would you mind turning off all requested features and providing a log of the VVL output showing the implicitly requested features for your machine?

SRSaunders:
Unfortunately the recent framework merge of da8ee8036573b7de8f96d20448360be46f960d93 broke this PR, so that's why it's sitting in conflict resolution mode now.  That's the danger of having a PR sit for 4+ months awaiting merge - something's going to break in between!

@asuessenbach before I fast-forward this branch, can you tell me why you added:
	`add_instance_layer("VK_LAYER_KHRONOS_shader_object");`

Shouldn't it be this instead?
	`add_instance_layer("VK_LAYER_KHRONOS_validation");`

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
spencer-lunarg: so we don't require `shaderInt64`, we will try to turn it on and if we can't we disable things like Buffer Device Address checks

We also will need it if you are trying to print out a 64-bit value, so we will warn you there

but to summarize, no you shouldn't need this feature

SRSaunders: Thanks @spencer-lunarg.  Can you tell me which features are now mandatory?

spencer-lunarg: 1. `fragmentStoresAndAtomics`/`vertexPipelineStoresAndAtomics`
2. `bufferDeviceAddress`
3. `timelineSemaphore`

SRSaunders: Thanks again.  To keep things simple, I will remove `requestedFeatures.shaderInt64` from this sample and let the VVL implicitly request `bufferDeviceAddress` and `timelineSemaphore`.

asuessenbach:
@SRSaunders You're right, it should be
`add_instance_layer("VK_LAYER_KHRONOS_validation");`

Don't know how that could have been scrambled.

SRSaunders:
@asuessenbach thanks for your quick answer.  I will update that when I fast forward the branch.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
gary-sweet: >  To help with this decision, can you tell me what platform and SDK version are you using? 

That won't help you very much as it's our own Broadcom hardware I'm afraid. SDK looks like 1.4.305

> And would you mind turning off all requested features and providing a log of the VVL output showing the implicitly requested features for your machine?

Do you still need this given Spencer's reply?

SRSaunders: @gary-sweet I should be good to go as is.  Thanks for the feedback.  Hopefully removing `shaderInt64` will help.

gary-sweet (dismissed):
Works fine for me now, thanks.

SaschaWillems (dismissed):

SaschaWillems:
This might need a rebase, as android build fails. Most probably due to the recent update of HCWPipe?

SRSaunders:
@SaschaWillems yes you are right.  I will fix in a few hours - have to step away for a bit.  Thanks for your help.

asuessenbach:
Runs fine on Win10 with NVIDIA GPU. I just get a couple of VVL warnings (not errors!):
```
[warning] 2132353751 - VALIDATION-SETTINGS: Validation Warning: [ VALIDATION-SETTINGS ] | MessageID = 0x7f1922d7 | DebugPrintf logs to the Information message severity, enabling Information level logging otherwise the message will not be seen.
[warning] 2132353751 - VALIDATION-SETTINGS: Validation Warning: [ VALIDATION-SETTINGS ] | MessageID = 0x7f1922d7 | DebugPrintf logs can possibly print many times, but duplicate_message_limit is set to 0, setting enable_message_limit to false so all logs are printed.
[warning] 1985515673 - WARNING-DEBUG-PRINTF: Validation Warning: [ WARNING-DEBUG-PRINTF ] | MessageID = 0x76589099 | vkCreateDevice(): Internal Warning: Forcing shaderInt64 to VK_TRUE
[warning] 1985515673 - WARNING-DEBUG-PRINTF: Validation Warning: [ WARNING-DEBUG-PRINTF ] | MessageID = 0x76589099 | vkCreateDevice(): Internal Warning: Adding a VkPhysicalDeviceTimelineSemaphoreFeatures to pNext with timelineSemaphore set to VK_TRUE
[warning] 1985515673 - WARNING-DEBUG-PRINTF: Validation Warning: [ WARNING-DEBUG-PRINTF ] | MessageID = 0x76589099 | vkCreateDevice(): Internal Warning: Adding a VkPhysicalDeviceVulkanMemoryModelFeatures to pNext with vulkanMemoryModel and vulkanMemoryModelDeviceScope set to VK_TRUE
[warning] 1985515673 - WARNING-DEBUG-PRINTF: Validation Warning: [ WARNING-DEBUG-PRINTF ] | MessageID = 0x76589099 | vkCreateDevice(): Internal Warning: Adding a VkPhysicalDeviceBufferDeviceAddressFeatures to pNext with bufferDeviceAddress set to VK_TRUE
```
Don't know if we want them to be resolved.

spencer-lunarg:
> Don't know if we want them to be resolved.

No, these are not errors, this is just Validation Layers being a good layer and letting you know it is adjusting settings and turning on/off features for you, so you don't get confused what is happening underneath

SRSaunders:
@marty-johnson59 I would like to ask for you to merge this ahead of #1281.  Otherwise we will have to go through another round of fixups and approvals.  There has been enough framework churn underneath this PR and it's time to close it off.

I have kindly asked @asuessenbach if he could please adjust #1281 once this is merged.  Thanks.

Approvers are likely: @SaschaWillems, @asuessenbach, and @gary-sweet.

marty-johnson59:
OK, I'll go ahead and merge this now - LMK if it causes problems and we can revert

SRSaunders:
Thanks @marty-johnson59.  Much appreciated.

PR #1189, opened 2024-10-12
Update profiles sample to use recent profiles library and roadmap 2022 profile

SaschaWillems:
## Description

This PR updates the profiles sample. It updates the profiles library to a recent version and also replaces the no longer available LunarG profile with the official Khronos roadmap 2022 milestone profile. Documentation is adjusted accordingly.

The PR also updates the Vulkan headers to a recent version, as the updated profiles library requires newer Vulkan features.

**Note**: The roadmap 2022 profile requires Vulkan 1.3, so this PR increases the requirement for devices to run this sample.

## General Checklist:


samples/tooling/profiles/profiles.cpp
```diff
@@ -98,9 +96,9 @@ std::unique_ptr<vkb::Device> Profiles::create_device(vkb::PhysicalDevice &gpu)
```
asuessenbach: Maybe move the profile support check to the beginning of this function. None of the stuff above this check is needed for it.

asuessenbach (commented):
Looks good, just one minor issue.

samples/tooling/profiles/profiles.cpp
```diff
@@ -30,14 +30,9 @@
 // The Vulkan Profiles library is part of the SDK and has been copied to the sample's folder for convenience
 #include "vulkan_profiles.hpp"
```
asuessenbach, line 31: Why is this header copied from the SDK? Wouldn't there be any version issues when the SDK is updated?

SaschaWillems: No, it's the other way round. The one from the SDK might (and has) introduce breaking changes, so if we were to rely on that one, this sample might not work for everyone. Packaging it will avoid such breakages. Also our sample are suppose to work without any SDK installed.

asuessenbach: > Also our sample are suppose to work without any SDK installed.

I see... we're using what we have in third_party/vulkan/include. And there's nothing like a vulkan_profiles.hpp.
How do we get the vulkan stuff here, any why is that file missing?

SaschaWillems: vulkan_profiles.hpp is only packaged with the SDK, it's not part of the Vulkan headers.

asuessenbach: Ah, ok. Thanks for that clarifications.

samples/tooling/profiles/profiles.cpp
SaschaWillems: That's a goo point. Adjusted the order.

asuessenbach (approved):

gary-sweet (approved):

Issue #1168, opened 2024-09-14
Clean up (and maybe extend) Hello Triangle sample


SaschaWillems:
I'm currently working on a hell triangle sample for Vulkan 1.3 and noticed that our existing hello triangle sample is in dire need of a code cleanup. While it works fine, there are several parts in the code that make it hard to follow. Other parts are unnecessary, variables are named badly and sometimes not even used at all. Also validation layers are not always properly enabled as only parts of the sample also use the debug define to enable them, others don't. Also some of the error messages are plainly wrong and may mislead readers.

SaschaWillems:
The sample is also partially outdated and uses e.g. `VK_EXT_debug_report` which has long been deprecated in favor of `VK_EXT_debug_utils `.

SaschaWillems:
I also feel the sample is lacking some crucial functionality to make it a good entry into Vulkan:

- Vertices are hard-coded in the shader, the sample does not show how to actually pass vertices to Vulkan
- No uniforms, the sample does not show how to use descriptors and how to pass data to shaders
- Compiles shaders at runtime instead of loading from SPIR-V (which is our preferred way) and as such no HLSL support

PR #1199, opened 2024-10-19
Hello triangle sample: Replace deprecated debug reports extension, fix validation and clean up code

SaschaWillems:
## Description

The hello triangle sample has been using the debug report extension for validation, which has been deprecated years ago. This PR replaces that with the debug utils extension. It also fixes an issue where validation layers weren't enabled in debug builds unless explicitly requested via the VKB_VALIDATION_LAYERS cmake option. It also fixes some minor things like wrong/misleading error messges and does some general code cleanup like removing unnecessary variables and arguments.

This fixes one of the issues of that sample documented in #1168 

## General Checklist:

### DIFF
--- a/samples/api/hello_triangle/hello_triangle.cpp
+++ b/samples/api/hello_triangle/hello_triangle.cpp
@@ -158,7 +158,7 @@ VkShaderStageFlagBits HelloTriangle::find_shader_stage(const std::string &ext)
  * @param required_instance_extensions The required Vulkan instance extensions.
  * @param required_validation_layers The required Vulkan validation layers
  */
-void HelloTriangle::init_instance(Context &context)
+void HelloTriangle::init_instance()
 {
 	LOGI("Initializing vulkan instance.");
 
@@ -308,7 +308,7 @@ void HelloTriangle::init_instance(Context &context)
  *
  * @param context A Vulkan context with an instance already set up.
  */
-void HelloTriangle::init_device(Context &context)
+void HelloTriangle::init_device()
 {
 	LOGI("Initializing vulkan device.");
 
@@ -394,7 +394,7 @@ void HelloTriangle::init_device(Context &context)
  * @param context A newly created Vulkan context.
  * @param per_frame The data of a frame.
  */
-void HelloTriangle::init_per_frame(Context &context, PerFrame &per_frame)
+void HelloTriangle::init_per_frame(PerFrame &per_frame)
 {
 	VkFenceCreateInfo info{VK_STRUCTURE_TYPE_FENCE_CREATE_INFO};
 	info.flags = VK_FENCE_CREATE_SIGNALED_BIT;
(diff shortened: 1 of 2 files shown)


samples/api/hello_triangle/hello_triangle.cpp
```diff
@@ -173,59 +169,59 @@ void HelloTriangle::init_instance(Context                         &context,
...
 	VK_CHECK(vkEnumerateInstanceExtensionProperties(nullptr, &instance_extension_count, available_instance_extensions.data()));
 
-	std::vector<const char *> active_instance_extensions(required_instance_extensions);
+	std::vector<const char *> required_instance_extensions{VK_KHR_SURFACE_EXTENSION_NAME};
 
 #if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
-	bool has_debug_report = false;
+	// Validation layers help finding wrong api usage, we enable them when explicitly requested or in debug builds
+	// For this we use the debug utils extension if it is supported
+	bool has_debug_utils = false;
```
asuessenbach, line 177: Maybe instead:
```
	bool has_debug_utils = std::any_of(available_instance_extensions.begin(),
	                                   [](auto const &ext) { return strcmp(ext.extensionName, VK_EXT_DEBUG_UTILS_EXTENSION_NAME) == 0; });
	if (has_debug_utils)
	{
		required_instance_extensions.push_back(VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
	}
	else
	{
		LOGW("{} not supported or available", VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
		LOGW("Make sure to compile the sample in debug mode and/or enable the validation layers");
	}
```

asuessenbach (commented):
Substantially improves this sample.
Just one suggestion:

samples/api/hello_triangle/hello_triangle.cpp
SaschaWillems: Tbh. I'm not a fine of such single line loops replacement. I find them hard to read. Probably a personal preference, but since this is not a Vulkan.hpp based sample, I'd prefer the loop as it currently is implemented.

asuessenbach: Of course, your code is equivalent to the STL algorithm, so it's fine to just keep it.
But if we want to use "Modern Vulkan", we might also want to use "Modern C++". And I think, using an STL algorithm not only is potentially more efficient and less error prone, it also tells the intent of the code sequence. You would not have to parse a loop and check if anything unexpected is happening there.
Note, with C++20, we could even use `std::ranges::any_of`, making it even more concise.
Note as well, this has nothing to do with Vulkan-Hpp.

SaschaWillems: I totally agree with that, but that's the scope of another sample to come. I'd rather not just change one place in this sample to use a different way for the loops than the others. If that's okay with you I'd keep it the way it is.

asuessenbach: Well, I don't quite understand your reasoning here. You change this part of the code here anyway.

asuessenbach (approved):

samples/api/hello_triangle/hello_triangle.cpp
SaschaWillems: What I meant with my comment: I did not rewrite any of the loops and simply renamed a variable. Other places in the sample also use the "old style" for loops, that's why I didn't want to use a different approach in here. I'm a fan of keeping the scope of a PR as close to the problem solution as possible ;)

gpx1000 (approved):

gary-sweet (approved):

marty-johnson59:
3 approvals - merging

PR #1200, opened 2024-10-20
Remove unavailable debug information for high level framework based samples

### LINKED ISSUE #1176: Failed assert and crash when clicking two mouse button same time when using free camera
To produce: open a sample that use free camera, click two mouse button at the same time.
Looks like there is two type of camera base class in repo, I'm talking about that used in performance samples.

Issue was getcomponent returning an empty cameras vector but this happens only when clicking two buttons at the same time for some reason.

https://github.com/KhronosGroup/Vulkan-Samples/blob/5daac4d8b0988b7a18ec8a486d7b973e3dc72d47/framework/scene_graph/hpp_scene.h#L37-L54

SaschaWillems:
## Description

The high level framework based samples have a kinda hidden debug overlay activated with right click (on desktop). It shows some general debug information. This was broken (not sure when that did happen) and would crash every time you tried to toggle that debug overlay. This was caused by the scene graph not finding some components and also not checking if those are present. While fixing this, I noticed that those components weren't available in any of those samples so I simply removed the block. Also removed a wrong assert in the has_component check.

We may want to start thinking on how to simplify the high level framework. Debugging and fixing anything in that framework is tedious.

Fixes #1176

## General Checklist:

### DIFF
--- a/framework/scene_graph/hpp_scene.h
+++ b/framework/scene_graph/hpp_scene.h
@@ -62,7 +62,6 @@ class HPPScene : private vkb::sg::Scene
 		}
 		else
 		{
-			assert(false);        // path never passed -> Please add a type-check here!
 			return false;
 		}
 	}
--- a/framework/vulkan_sample.h
+++ b/framework/vulkan_sample.h
@@ -1337,21 +1337,6 @@ inline void VulkanSample<bindingType>::update_debug_window()
 	                                                             to_string(render_context->get_swapchain().get_format()) + " (" +
 	                                                                 to_string(vkb::common::get_bits_per_pixel(render_context->get_swapchain().get_format())) +
 	                                                                 "bpp)");
-
-	if (scene != nullptr)
-	{
-		get_debug_info().template insert<field::Static, uint32_t>("mesh_count", to_u32(scene->get_components<sg::SubMesh>().size()));
-		get_debug_info().template insert<field::Static, uint32_t>("texture_count", to_u32(scene->get_components<sg::Texture>().size()));
-
-		if (auto camera = scene->get_components<vkb::sg::Camera>()[0])
-		{
-			if (auto camera_node = camera->get_node())
-			{
-				const glm::vec3 &pos = camera_node->get_transform().get_translation();
-				get_debug_info().template insert<field::Vector, float>("camera_pos", pos.x, pos.y, pos.z);
-			}
-		}
-	}
 }
 
 template <vkb::BindingType bindingType>


framework/scene_graph/hpp_scene.h
```diff
@@ -62,7 +62,6 @@ class HPPScene : private vkb::sg::Scene
 		}
 		else
 		{
-			assert(false);        // path never passed -> Please add a type-check here!
```
asuessenbach, line 65: What's wrong with this `assert`? Did you encounter it?

If at all, you should replace the complete function body with just
`		return vkb::sg::Scene::has_component(typeid(T));
`

gary-sweet (approved):

framework/vulkan_sample.h
```diff
@@ -1337,21 +1337,6 @@ inline void VulkanSample<bindingType>::update_debug_window()
...
-
-		if (auto camera = scene->get_components<vkb::sg::Camera>()[0])
-		{
-			if (auto camera_node = camera->get_node())
-			{
-				const glm::vec3 &pos = camera_node->get_transform().get_translation();
-				get_debug_info().template insert<field::Vector, float>("camera_pos", pos.x, pos.y, pos.z);
-			}
-		}
-	}
```
asuessenbach, line 1354: What's wrong with this code sequence?
Under what circumstances do you get a crash here?

Instead of just removing this block, it might be worth to improve checking here.

SaschaWillems: See my description of the issue as to why I removed it. tl;dr: that information was never working, in none of the samples. Adding checks  before the code would simply not run the code at all, because none of the requested components could be found.

framework/scene_graph/hpp_scene.h
SaschaWillems: The assert doesn't make sense. The function is used to check if a component is present, which should return true or false and not assert.

And yes, I did encounter it ;)

asuessenbach: > And yes, I did encounter it ;)

Well, in that case, some code should be added to handle those cases.
Maybe I didn't phrase that detailed enough in the comment. The assertion is just there to not miss yet unhandled cases.

framework/vulkan_sample.h
asuessenbach: > because none of the requested components could be found.

And you don't expect that to ever happen?
Instead, I would prefer to have the checks extended.

SaschaWillems: Yes, I have tested with all samples and not a single one had those working. Aside, the debug information removed with this PR is pretty useless anyway.

framework/scene_graph/hpp_scene.h
SaschaWillems: But it is a function that checks if the component is there. IMO it should never assert, or I might just misunderstand the statement above.

SaschaWillems:
@asuessenbach: As discussed during the call, this only happens with sample using the high level framework (based on the scenegraph). Those are most of the samples in the performance folder, eg. afbc, command_buffer_usage, msaa, subpasses. A simply right click in those causes an exception.

SaschaWillems:
No longer necessary with #1203

PR #1209, opened 2024-10-26
Fix pipeline stats in HPPTerrainTessellation

### LINKED ISSUE #674: HPP terrain tessellation sample reports wrong statistics
The terrain tessellations sample displays statistics using a pipeline statistics query:

image

The HPP version of that sample does the same, but displays wrong values:

image

When debugging, I can see the correct values in the HPP ones after getting the query pool results, but they seem to get lost between this and the display. I wasn't able to find out why.

Maybe you can take a look at this @asuessenbach?

yknishidate:
## Description

Fixed pipeline stats in the hpp_terrain_tessellation sample.

The bug was caused by copying `result.value` when the query result was NotReady. The original C sample (`vkGetQueryPoolResults`) also generated NotReady, but since the result value was not written to the output pointer, the same problem as the HPP version did not occur.

The bug can also be fixed by setting the `VK_QUERY_RESULT_WAIT_BIT` flag, but to match the behavior of the original C sample, I modified it to only read the value if the result was Success.

- This PR has only been tested on Windows.
- Fixes #674 

## General Checklist:

### DIFF
--- a/samples/api/hpp_terrain_tessellation/hpp_terrain_tessellation.cpp
+++ b/samples/api/hpp_terrain_tessellation/hpp_terrain_tessellation.cpp
@@ -327,11 +327,13 @@ void HPPTerrainTessellation::draw()
 	if (statistics.query_supported)
 	{
 		// Read query results for displaying in next frame
-		statistics.results =
-		    get_device()
-		        .get_handle()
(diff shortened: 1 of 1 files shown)


asuessenbach (approved):
Great catch!

Thanks for this fix.

SaschaWillems (approved):
Thank you very much. Explains (and fixes) the issue I had with the HPP variant of this sample :)

gpx1000 (approved):

PR #1211, opened 2024-10-29
Ensure to initialize image sharing mode when implicit

### LINKED ISSUE #1207: Invalid VkImageCreateInfo->sharingMode validation layer error in Release builds of msaa/16bit_arithmetic samples
I see the following validation layer error when running Release builds of the msaa/16bit_arithmetic samples when the samples are built using `-DCMAKE_BUILD_TYPE=RelWithDebInfo -DVKB_VALIDATION_LAYERS=ON`.

`
[error] 81114962 - VUID-VkImageCreateInfo-sharingMode-parameter: Validation Error: [ VUID-VkImageCreateInfo-sharingMode-parameter ] | MessageID = 0x4d5b752 | vkCreateImage(): pCreateInfo->sharingMode (23330) does not fall within the begin..end range of the VkSharingMode enumeration tokens and is not an extension added token.
The Vulkan spec states: sharingMode must be a valid VkSharingMode value (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageCreateInfo-sharingMode-parameter)
`

This can be fixed by conditionally setting sharingMode in https://github.com/KhronosGroup/Vulkan-Samples/blob/f7e97a19378255e01b51acc211bf6b31c849a34c/framework/core/image.h#L120 i.e. `create_info.sharingMode        = create_info.queueFamilyIndexCount != 0 ? VK_SHARING_MODE_CONCURRENT : VK_SHARING_MODE_EXCLUSIVE;`. I didn't raise a PR as I'm not particularly familiar with this codebase and I wasn't sure whether there was a reason this wasn't done originally.

Vulkan-Samples built with `gcc (Ubuntu 13.2.0-23ubuntu4) 13.2.0` on `Ubuntu 24.04`.

Reproduced on:
`
	driverID           = DRIVER_ID_MESA_LLVMPIPE
	driverName         = llvmpipe
	driverInfo         = Mesa 24.3~git2410240600.c0bcea~oibaf~n (git-c0bceaf 2024-10-24 noble-oibaf-ppa) (LLVM 18.1.8)
`

Tested on commit f7e97a19378255e01b51acc211bf6b31c849a34c.

zzyiwei:
## Description

Fixed a VVL violation of uninitialized image sharing mode.

Related to #1207

### DIFF
--- a/framework/core/image.h
+++ b/framework/core/image.h
@@ -115,10 +115,7 @@ struct ImageBuilder : public vkb::BuilderBaseC<ImageBuilder, VkImageCreateInfo>
 	ImageBuilder &with_implicit_sharing_mode()
 	{
 		VkImageCreateInfo &create_info = get_create_info();
-		if (create_info.queueFamilyIndexCount != 0)
-		{
-			create_info.sharingMode = VK_SHARING_MODE_CONCURRENT;
-		}
+		create_info.sharingMode        = create_info.queueFamilyIndexCount != 0 ? VK_SHARING_MODE_CONCURRENT : VK_SHARING_MODE_EXCLUSIVE;
 		return *this;
 	}
 


asuessenbach:
Following the discussion in #1207, it's not yet clear if this would not only resolve this issue but also hide some other potentially more fundamental one.
That is, we should wait 'til the discussion in #1207 is done.

zzyiwei:
> Following the discussion in #1207, it's not yet clear if this would not only resolve this issue but also hide some other potentially more fundamental one. That is, we should wait 'til the discussion in #1207 is done.

I can drop the `Fixes` tag to keep the discussion there alive. Meanwhile, it's generally not a good practice to skip assigning enums that happen to be zero. I'd prefer we take this pr still.

asuessenbach:
> Meanwhile, it's generally not a good practice to skip assigning enums that happen to be zero.

This enum is initialized with a value that happens to be zero. Perfectly valid.
Your PR would, at best, be superfluous. And at worst (as we encountered here), it would hide the actual issue.

The right move would be to prevent some compiler optimizations that are identified to be dangerous in this case. That is, using `-fno-strict-warning` for all the samples would be better.

zzyiwei:
> This enum is initialized with a value that happens to be zero. Perfectly valid.

I'm not convinced. `VK_SHARING_MODE_EXCLUSIVE` just happened to be `0` when we originally introduced it, but whatever : /

> ...And at worst (as we encountered here), it would hide the actual issue.

The actual issue being exposed by this uninitialized field is just a coincidence. There're tons of other struct members could potentially leading to unexpected render without being properly initialized. It's a big question mark to lean on this to gate proper zero-init.

The right move would be to add a unit test for the builder to guarantee zero-init under the default compiler optimization used in the project.

/cc @SaschaWillems

zhangyiwei:
> ...And at worst (as we encountered here), it would hide the actual issue.

The actual issue being exposed by this uninitialized field is just a co-incident. There're tons of other struct members could potentially leading to unexpected render if not properly initialized. It's a big question mark to lean on this to gate proper zero-init.

SaschaWillems:
Gotta admit that my C++ isn't good enough to fully understand some of the framework parts. I took a look at the builder but wasn't able to find out why or if at all it does initialize things.

So why don't we just explicitly zero initialize all the structs? Would also make it easier to understand for people trying to learn from our samples.

zzyiwei:
Me either unfamiliar with the right expectations of the builder class. I’d vote for either explicitly initialize all fields after getting the alloc or ensure zero init within the builder class ; )

PR #1217, opened 2024-11-05
ci: Reduce CI to unblock Github Action minutes

spencer-lunarg:
Recently we have found productivity slow down in other KhronosGroups repos because it can take hours for CI to run due to running out of Github Actions minutes. We have found that MacOS/iOS are charged as 10x the minute cost which really starts to add up. The goal is to not have those expensive CI actions run unless we know other things work first.

Also there are many variations of MacOS being built, I would like to know if anyone has actually seen where **only** the Debug MacOS build fails, but everything else passes. I realize it seems great to test every combination, but unfortunately we have finite CI resources currently

(There is an internal issue on this, but for short term, I have just been apply these types of YAML chanegs aroudn the repos https://gitlab.khronos.org/khronos-general/khronos-issues/-/issues/127)


.github/workflows/build.yml
```diff
@@ -108,6 +164,8 @@ jobs:
         run: NDK_CCACHE=1 NDK_CCACHE_BIN=ccache ./gradlew app:assemble${{ matrix.build_type }} --info
 
   build_ios:
+    # iOS is 10x expensive to run on GitHub machines, so only run if we know something else fast/simple passed as well
+    needs: build
```
spencer-lunarg, line 169: how this looks in practice 

image

.github/workflows/build.yml
```diff
@@ -108,11 +164,13 @@ jobs:
...
     name: "Build ios in ${{ matrix.build_type }}"
     runs-on: macos-latest
     strategy:
       matrix:
-        build_type: [Debug]
+        build_type: [Release]
```
spencer-lunarg, line 174: Release builds are about 3 times faster, this build is currently around 20+ minutes and should reduce it greatly

.github/workflows/v2.yml
```diff
@@ -1,44 +0,0 @@
...
-      - if: ${{ matrix.platform == 'ubuntu' }}
-        name: Install RandR headers
-        run: |
-          sudo apt-get update
-          sudo apt install xorg-dev libglu1-mesa-dev
-
-      - name: Configure
-        run: cmake -B"build/${{ matrix.platform }}" -DVKB_BUILD_TESTS=ON
-
-      - name: "Build Components ${{ matrix.platform }} in ${{ matrix.build_type }}"
```
tomadamatkinson, line 37: Can we move these three steps to the existing build pipeline please. That would preserve running unit tests. This should be a fast operation.

.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     name: "Build ${{ matrix.platform }} in ${{ matrix.build_type }}"
     strategy:
       matrix:
-        platform: [windows, ubuntu, macos]
+        platform: [windows, ubuntu]
```
tomadamatkinson, line 18: Any chance we can keep MacOS once merged?

Keep it in the matrix but add an if statement to the job to only run macos on main

Not sure what this would be in practice but something like

`if !macos or ( macos and branch == main )`

tomadamatkinson: We could follow with a PR to add this if it becomes an issue

.github/workflows/build.yml
tomadamatkinson: We could potentially move IOS to only run on main as well (see below for macos comment)

.github/workflows/build.yml
spencer-lunarg: > Any chance we can keep MacOS once merged?

I was thinking that, but wasn't sure what people would think, @tomadamatkinson could you make a seperate PR with how you would do it, I think you probably know what you want better, this PR was more of a "here is a way to improve things"

.github/workflows/build.yml
SaschaWillems: > We could potentially move IOS to only run on main as well (see below for macos comment)

Wouldn't that cause problems for PRs? How would people know if their changes would break iOS/Mac?

spencer-lunarg: I think changing from Debug to Release will help a lot of build times. Also just making sure we are not running it unless others are passing is huge... so ya, maybe worth just keeping it for now

tomadamatkinson:
Hey @spencer-lunarg can you rebase this. I think that will fix the antora check and we can get this merged. I will then follow with a PR that re enables macos in the same style as your IOS change

Currently we are still burning minutes

spencer-lunarg:
> Hey @spencer-lunarg can you rebase this

done

gpx1000 (approved):

SaschaWillems (approved):

tomadamatkinson (approved):

marty-johnson59:
Merging - 3 approvals

SaschaWillems:
Not sure if this did improve anything. Before this PR, builds took ~3h 30m, while the last build (from today) took ~3h 47m (actually slower).

spencer-lunarg:
> Not sure if this did improve anything

The idea was not to speed up build time, the goal was to not waste minutes on things that are going to fail (bad PRs)

so if someone pushes up a bad PR, it will fail the cheap actions first and not waste the minutes on the expensive runs (aka the Mac runs)

spencer-lunarg:
an example of it working

https://github.com/KhronosGroup/Vulkan-Samples/actions/runs/12327204395

PR #1230, opened 2024-11-18
Vulkan 1 3 hello triangle sample

alihuawei:
This Pull Request introduces a new sample, "Hello Triangle", leveraging Vulkan 1.3 features and modern C++20 language enhancements. The sample aims to provide developers with an updated and clean starting point for learning and experimenting with the latest Vulkan API functionalities.

1: Sample is using vertex buffer to load vertices compared to hardcoded in the vertex shader in hello_triangle 1.0.
2: Sample is using dynamic rendering so frame buffers and Render passes removed.
3: Used Synchronization2 in image layout transition.
3: Updated the code to use VK_DEBUG_UTILS (new) instead VK_DEBUG_REPORT(old).
4: Update the code to use Dynamic pipeline status.
5: Removed unused parameters from functions.
6: Used C++20 initializers.

The sample is built and tested on **Windows** and **Linux** OS.

## General Checklist:

- [x] This PR describes the scope and expected impact of the changes I am making:
- [x] My changes build on Windows and Linux.
## Sample Checklist

### DIFF
--- a/samples/CMakeLists.txt
+++ b/samples/CMakeLists.txt
@@ -50,7 +50,7 @@ endforeach ()
 set(ORDER_LIST
     #API Samples
     "hello_triangle"
-	"hello_triangle_1_3"
+    "hello_triangle_1_3"
     "dynamic_uniform_buffers"
     "texture_loading"
     "hdr"
--- a/samples/CMakeLists.txt
+++ b/samples/CMakeLists.txt
@@ -50,6 +50,7 @@ endforeach ()
 set(ORDER_LIST
     #API Samples
     "hello_triangle"
+	"hello_triangle_1_3"
     "dynamic_uniform_buffers"
     "texture_loading"
     "hdr"
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -815,12 +815,12 @@ void HelloTriangleV13::init_pipeline()
 	std::array<VkPipelineShaderStageCreateInfo, 2> shader_stages = {{
 	    {.sType  = VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO,
 	     .stage  = VK_SHADER_STAGE_VERTEX_BIT,
-	     .module = load_shader_module("triangle.vert", VK_SHADER_STAGE_VERTEX_BIT),
+	     .module = load_shader_module("hello_triangle_1_3/triangle.vert", VK_SHADER_STAGE_VERTEX_BIT),
 	     .pName  = "main"},        // Vertex shader stage
 	    {
 	        .sType  = VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO,
 	        .stage  = VK_SHADER_STAGE_FRAGMENT_BIT,
-	        .module = load_shader_module("triangle.frag", VK_SHADER_STAGE_FRAGMENT_BIT),
+	        .module = load_shader_module("hello_triangle_1_3/triangle.frag", VK_SHADER_STAGE_FRAGMENT_BIT),
 	        .pName  = "main"}        // Fragment shader stage
 	}};
 
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -341,7 +341,6 @@ void HelloTriangleV13::init_device()
 	vkGetPhysicalDeviceFeatures2(context.gpu, &query_device_features2);
 
 	// Check if Physical device supports Vulkan 1.3 features
-	// Check if Physical device supports Vulkan 1.3 features
 	if (!query_vulkan13_features.dynamicRendering)
 	{
 		throw std::runtime_error("Dynamic Rendering feature is missing");
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -1126,10 +1126,7 @@ void HelloTriangleV13::transition_image_layout(
 	vkCmdPipelineBarrier2(cmd, &dependency_info);
 }
 
-/**
- * @brief Tears down the Vulkan context.
- */
-void HelloTriangleV13::teardown()
+HelloTriangleV13::~HelloTriangleV13()
 {
 	// Don't release anything until the GPU is completely idle.
 	vkDeviceWaitIdle(context.device);
@@ -1198,11 +1195,6 @@ void HelloTriangleV13::teardown()
 	}
 
 	vk_instance.reset();
-}
-
-HelloTriangleV13::~HelloTriangleV13()
-{
-	teardown();
 }
 
 bool HelloTriangleV13::prepare(const vkb::ApplicationOptions &options)
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -649,12 +649,12 @@ void HelloTriangleV13::init_swapchain()
 	{
 		// Create an image view which we can render into.
 		VkImageViewCreateInfo view_info{
-		    .sType      = VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO,
-		    .pNext      = nullptr,
-		    .flags      = 0,
-		    .image      = swapchain_images[i],
-		    .viewType   = VK_IMAGE_VIEW_TYPE_2D,
-		    .format     = context.swapchain_dimensions.format,
+		    .sType            = VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO,
+		    .pNext            = nullptr,
+		    .flags            = 0,
+		    .image            = swapchain_images[i],
+		    .viewType         = VK_IMAGE_VIEW_TYPE_2D,
+		    .format           = context.swapchain_dimensions.format,
 		    .subresourceRange = {.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT, .baseMipLevel = 0, .levelCount = 1, .baseArrayLayer = 0, .layerCount = 1}};
 
 		VkImageView image_view;
@@ -1245,7 +1245,6 @@ void HelloTriangleV13::update(float delta_time)
 		{
 			LOGI("Resize failed");
 		}
-		
 	}
 	else if (res != VK_SUCCESS)
 	{
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -196,7 +196,7 @@ void HelloTriangleV13::init_instance()
 
 	std::vector<const char *> requested_validation_layers{};
 
-#ifdef defined(VKB_DEBUG) || VKB_VALIDATION_LAYERS
+#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
 	// Determine the optimal validation layers to enable that are necessary for useful debugging
 	std::vector<const char *> optimal_validation_layers = vkb::get_optimal_validation_layers(supported_validation_layers);
 	requested_validation_layers.insert(requested_validation_layers.end(), optimal_validation_layers.begin(), optimal_validation_layers.end());
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -1041,6 +1041,10 @@ void HelloTriangleV13::render_triangle(uint32_t swapchain_index)
 		                           &context.per_frame[swapchain_index].swapchain_release_semaphore));
 	}
 
+	// Using TOP_OF_PIPE here to ensure that the command buffer does not begin executing any pipeline stages
+	// (including the layout transition) until the swapchain image is actually acquired (signaled by the semaphore).
+	// This prevents the GPU from starting operations too early and guarantees that the image is ready
+	// before any rendering commands run.
 	VkPipelineStageFlags wait_stage{VK_PIPELINE_STAGE_2_TOP_OF_PIPE_BIT};
 
 	VkSubmitInfo info{
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -674,6 +674,7 @@ void HelloTriangleV13::init_swapchain()
 /**
  * @brief Helper function to load a shader module.
  * @param path The path for the shader (relative to the assets directory).
+ * @param shader_stage The shader stage flag specifying the type of shader (e.g., VK_SHADER_STAGE_VERTEX_BIT).
  * @returns A VkShaderModule handle. Aborts execution if shader creation fails.
  */
 VkShaderModule HelloTriangleV13::load_shader_module(const char *path, VkShaderStageFlagBits shader_stage)
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -615,12 +615,9 @@ void HelloTriangleV13::init_swapchain()
 			vkDestroyImageView(context.device, image_view, nullptr);
 		}
 
-		uint32_t image_count;
-		VK_CHECK(vkGetSwapchainImagesKHR(context.device, old_swapchain, &image_count, nullptr));
-
-		for (size_t i = 0; i < image_count; i++)
+		for (auto &per_frame : context.per_frame)
 		{
-			teardown_per_frame(context.per_frame[i]);
+			teardown_per_frame(per_frame);
 		}
 
 		context.swapchain_image_views.clear();
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -655,11 +655,6 @@ void HelloTriangleV13::init_swapchain()
 		    .image      = swapchain_images[i],
 		    .viewType   = VK_IMAGE_VIEW_TYPE_2D,
 		    .format     = context.swapchain_dimensions.format,
-		    .components = {
-		        .r = VK_COMPONENT_SWIZZLE_R,
-		        .g = VK_COMPONENT_SWIZZLE_G,
-		        .b = VK_COMPONENT_SWIZZLE_B,
-		        .a = VK_COMPONENT_SWIZZLE_A},
 		    .subresourceRange = {.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT, .baseMipLevel = 0, .levelCount = 1, .baseArrayLayer = 0, .layerCount = 1}};
 
 		VkImageView image_view;
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -605,7 +605,7 @@ void HelloTriangleV13::init_swapchain()
 
 	VkSwapchainKHR old_swapchain = context.swapchain;
 
-	//one bitmask needs to be set according to the priority of presentation engine
+	// one bitmask needs to be set according to the priority of presentation engine
 	VkCompositeAlphaFlagBitsKHR composite = VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR;
 	if (surface_properties.supportedCompositeAlpha & VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR)
 	{
@@ -626,7 +626,7 @@ void HelloTriangleV13::init_swapchain()
 
 	VkSwapchainCreateInfoKHR info{
 	    .sType            = VK_STRUCTURE_TYPE_SWAPCHAIN_CREATE_INFO_KHR,
-	    .surface          = context.surface,							// The surface onto which images will be presented
+	    .surface          = context.surface,                            // The surface onto which images will be presented
 	    .minImageCount    = desired_swapchain_images,                   // Minimum number of images in the swapchain (number of buffers)
 	    .imageFormat      = format.format,                              // Format of the swapchain images (e.g., VK_FORMAT_B8G8R8A8_SRGB)
 	    .imageColorSpace  = format.colorSpace,                          // Color space of the images (e.g., VK_COLOR_SPACE_SRGB_NONLINEAR_KHR)
@@ -638,7 +638,7 @@ void HelloTriangleV13::init_swapchain()
 	    .compositeAlpha   = composite,                                  // Alpha blending to apply (e.g., opaque, pre-multiplied)
 	    .presentMode      = swapchain_present_mode,                     // Presentation mode (e.g., vsync settings)
 	    .clipped          = true,                                       // Whether to clip obscured pixels (improves performance)
-	    .oldSwapchain     = old_swapchain};                             // Handle to the old swapchain, if replacing an existing one
+	    .oldSwapchain     = old_swapchain};                                 // Handle to the old swapchain, if replacing an existing one
 
 	VK_CHECK(vkCreateSwapchainKHR(context.device, &info, nullptr, &context.swapchain));
 
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -278,10 +278,7 @@ void HelloTriangleV13::init_device()
 		VkPhysicalDeviceProperties device_properties;
 		vkGetPhysicalDeviceProperties(physical_device, &device_properties);
 
-		uint32_t major_version = VK_API_VERSION_MAJOR(device_properties.apiVersion);
-		uint32_t minor_version = VK_API_VERSION_MINOR(device_properties.apiVersion);
-
-		if ((major_version < 1) || (major_version == 1 && minor_version < 3))
+		if (device_properties.apiVersion < VK_API_VERSION_1_3)
 		{
 			LOGW("Physical device '{}' does not support Vulkan 1.3, skipping.", device_properties.deviceName);
 			continue;
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -277,13 +277,6 @@ void HelloTriangleV13::init_instance()
 	}
 #endif
 
-#if (defined(VKB_ENABLE_PORTABILITY))
-	if (portability_enumeration_available)
-	{
-		instance_info.flags |= VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
-	}
-#endif
-
 	// Create the Vulkan instance
 	VK_CHECK(vkCreateInstance(&instance_info, nullptr, &context.instance));
 
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -232,7 +232,7 @@ void HelloTriangleV13::init_instance()
 
 	std::vector<const char *> requested_validation_layers{};
 
-#ifdef VKB_VALIDATION_LAYERS
+#ifdef defined(VKB_DEBUG) || VKB_VALIDATION_LAYERS
 	// Determine the optimal validation layers to enable that are necessary for useful debugging
 	std::vector<const char *> optimal_validation_layers = vkb::get_optimal_validation_layers(supported_validation_layers);
 	requested_validation_layers.insert(requested_validation_layers.end(), optimal_validation_layers.begin(), optimal_validation_layers.end());
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -1070,7 +1070,7 @@ void HelloTriangleV13::render_triangle(uint32_t swapchain_index)
 		                           &context.per_frame[swapchain_index].swapchain_release_semaphore));
 	}
 
-	VkPipelineStageFlags wait_stage{VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT};
+	VkPipelineStageFlags wait_stage{VK_PIPELINE_STAGE_2_TOP_OF_PIPE_BIT};
 
 	VkSubmitInfo info{
 	    .sType                = VK_STRUCTURE_TYPE_SUBMIT_INFO,
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -786,6 +786,7 @@ void HelloTriangleV13::init_pipeline()
 	// Specify we will use triangle lists to draw geometry.
 	VkPipelineInputAssemblyStateCreateInfo input_assembly{
 	    .sType                  = VK_STRUCTURE_TYPE_PIPELINE_INPUT_ASSEMBLY_STATE_CREATE_INFO,
+	    .topology               = VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST,
 	    .primitiveRestartEnable = VK_FALSE};
 
 	// Specify rasterization state.
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -726,17 +726,10 @@ void HelloTriangleV13::init_pipeline()
 	    .inputRate = VK_VERTEX_INPUT_RATE_VERTEX};
 
 	// Define the vertex input attribute descriptions
-	std::array<VkVertexInputAttributeDescription, 2> attribute_descriptions{};
-
-	attribute_descriptions[0].binding  = 0;
-	attribute_descriptions[0].location = 0;
-	attribute_descriptions[0].format   = VK_FORMAT_R32G32_SFLOAT;        // vec2 for position
-	attribute_descriptions[0].offset   = offsetof(Vertex, position);
-
-	attribute_descriptions[1].binding  = 0;
-	attribute_descriptions[1].location = 1;
-	attribute_descriptions[1].format   = VK_FORMAT_R32G32B32_SFLOAT;        // vec3 for color
-	attribute_descriptions[1].offset   = offsetof(Vertex, color);
+	std::array<VkVertexInputAttributeDescription, 2> attribute_descriptions = {{
+	    {.location = 0, .binding = 0, .format = VK_FORMAT_R32G32_SFLOAT, .offset = offsetof(Vertex, position)},        // position
+	    {.location = 1, .binding = 0, .format = VK_FORMAT_R32G32B32_SFLOAT, .offset = offsetof(Vertex, color)}         // color
+	}};
 
 	// Create the vertex input state
 	VkPipelineVertexInputStateCreateInfo vertex_input{.sType                           = VK_STRUCTURE_TYPE_PIPELINE_VERTEX_INPUT_STATE_CREATE_INFO,
@@ -799,19 +792,17 @@ void HelloTriangleV13::init_pipeline()
 	    .pDynamicStates    = dynamic_states.data()};
 
 	// Load our SPIR-V shaders.
-	std::array<VkPipelineShaderStageCreateInfo, 2> shader_stages{};
-
-	// Vertex stage of the pipeline
-	shader_stages[0].sType  = VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO;
-	shader_stages[0].stage  = VK_SHADER_STAGE_VERTEX_BIT;
-	shader_stages[0].module = load_shader_module("triangle.vert", VK_SHADER_STAGE_VERTEX_BIT);
-	shader_stages[0].pName  = "main";
-
-	// Fragment stage of the pipeline
-	shader_stages[1].sType  = VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO;
-	shader_stages[1].stage  = VK_SHADER_STAGE_FRAGMENT_BIT;
-	shader_stages[1].module = load_shader_module("triangle.frag", VK_SHADER_STAGE_FRAGMENT_BIT);
-	shader_stages[1].pName  = "main";
+	std::array<VkPipelineShaderStageCreateInfo, 2> shader_stages = {{
+	    {.sType  = VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO,
+	     .stage  = VK_SHADER_STAGE_VERTEX_BIT,
+	     .module = load_shader_module("triangle.vert", VK_SHADER_STAGE_VERTEX_BIT),
+	     .pName  = "main"},        // Vertex shader stage
+	    {
+	        .sType  = VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO,
+	        .stage  = VK_SHADER_STAGE_FRAGMENT_BIT,
+	        .module = load_shader_module("triangle.frag", VK_SHADER_STAGE_FRAGMENT_BIT),
+	        .pName  = "main"}        // Fragment shader stage
+	}};
 
 	// Pipeline rendering info (for dynamic rendering).
 	VkPipelineRenderingCreateInfo pipeline_rendering_info{
@@ -933,8 +924,8 @@ void HelloTriangleV13::render_triangle(uint32_t swapchain_index)
 	    VK_PIPELINE_STAGE_2_COLOR_ATTACHMENT_OUTPUT_BIT        // dstStage
 	);
 	// Set clear color values.
-	VkClearValue clear_value;
-	clear_value.color = {{0.01f, 0.01f, 0.033f, 1.0f}};
+	VkClearValue clear_value{
+	    .color = {{0.01f, 0.01f, 0.033f, 1.0f}}};
 
 	// Set up the rendering attachment info
 	VkRenderingAttachmentInfo color_attachment{
@@ -974,9 +965,11 @@ void HelloTriangleV13::render_triangle(uint32_t swapchain_index)
 	vkCmdSetViewport(cmd, 0, 1, &vp);
 
 	// Set scissor dynamically
-	VkRect2D scissor{};
-	scissor.extent.width  = context.swapchain_dimensions.width;
-	scissor.extent.height = context.swapchain_dimensions.height;
+	VkRect2D scissor{
+	    .extent = {
+	        .width  = context.swapchain_dimensions.width,
+	        .height = context.swapchain_dimensions.height}};
+
 	vkCmdSetScissor(cmd, 0, 1, &scissor);
 
 	// Since we declared VK_DYNAMIC_STATE_CULL_MODE as dynamic in the pipeline,
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -421,8 +421,9 @@ void HelloTriangleV13::init_vertex_buffer()
 	// Create the vertex buffer
 	VkBufferCreateInfo vertext_buffer_info{
 	    .sType       = VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO,
-	    .flags       = VK_BUFFER_USAGE_VERTEX_BUFFER_BIT,
+	    .flags       = 0,
 	    .size        = buffer_size,
+	    .usage       = VK_BUFFER_USAGE_VERTEX_BUFFER_BIT,
 	    .sharingMode = VK_SHARING_MODE_EXCLUSIVE};
 
 	VK_CHECK(vkCreateBuffer(context.device, &vertext_buffer_info, nullptr, &context.vertex_buffer));
--- a/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
+++ b/samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
@@ -126,42 +126,6 @@ bool HelloTriangleV13::validate_layers(const std::vector<const char *>      &req
 }
 
 /**
- * @brief Find the vulkan shader stage for a given a string.
- *
- * @param ext A string containing the shader stage name.
- * @return VkShaderStageFlagBits The shader stage mapping from the given string, VK_SHADER_STAGE_VERTEX_BIT otherwise.
- */
-VkShaderStageFlagBits HelloTriangleV13::find_shader_stage(const std::string &ext)
-{
-	if (ext == "vert")
-	{
-		return VK_SHADER_STAGE_VERTEX_BIT;
-	}
-	else if (ext == "frag")
-	{
-		return VK_SHADER_STAGE_FRAGMENT_BIT;
-	}
-	else if (ext == "comp")
-	{
-		return VK_SHADER_STAGE_COMPUTE_BIT;
-	}
-	else if (ext == "geom")
-	{
-		return VK_SHADER_STAGE_GEOMETRY_BIT;
-	}
-	else if (ext == "tesc")
-	{
-		return VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT;
-	}
-	else if (ext == "tese")
-	{
-		return VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT;
-	}
-
-	throw std::runtime_error("No Vulkan shader stage found for the file extension name.");
-};
-
-/**
  * @brief Initializes the Vulkan instance.
  */
 void HelloTriangleV13::init_instance()
@@ -715,7 +679,7 @@ void HelloTriangleV13::init_swapchain()
  * @param path The path for the shader (relative to the assets directory).
  * @returns A VkShaderModule handle. Aborts execution if shader creation fails.
  */
-VkShaderModule HelloTriangleV13::load_shader_module(const char *path)
+VkShaderModule HelloTriangleV13::load_shader_module(const char *path, VkShaderStageFlagBits shader_stage)
 {
 	vkb::GLSLCompiler glsl_compiler;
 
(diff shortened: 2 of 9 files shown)


samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1340 @@
...
+
+	// Before starting rendering, transition the swapchain image to COLOR_ATTACHMENT_OPTIMAL
+	transition_image_layout(
+	    cmd,
+	    context.swapchain_images[swapchain_index],
+	    VK_IMAGE_LAYOUT_UNDEFINED,
+	    VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL,
+	    0,                                                     // srcAccessMask (no need to wait for previous operations)
+	    VK_ACCESS_2_COLOR_ATTACHMENT_WRITE_BIT,                // dstAccessMask
+	    VK_PIPELINE_STAGE_2_TOP_OF_PIPE_BIT,                   // srcStage
```
cforfang, line 942: This transition for the swap-chain image has src = TOP, however the vkQueueSubmit waits on the acquire semaphore in pWaitDstStageMask = COLOR. These two need to align for correct sync I believe; as-is the transition can start before the acquire-semaphore has signalled (since pWaitDstStageMask = COLOR means only the COLOR stage is held back so long as the acquire semaphore isn't signalled, and TOP happens prior to COLOR).

gary-sweet (changes requested):
This seems to be missing a change to samples/CMakeLists.txt that actually exposes the sample.

I know very little about the background to this change, but it does seem a little odd that using the newer version of Vulkan results in 200 more lines of code for hello_triangle than we used to have.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
alihuawei: Yes, you're correct to point out this. I will update the code for wait_stage in vkQueueSubmit call to VK_PIPELINE_STAGE_2_TOP_OF_PIPE_BIT

SaschaWillems:
> This seems to be missing a change to samples/CMakeLists.txt that actually exposes the sample.
> 
> I know very little about the background to this change, but it does seem a little odd that using the newer version of Vulkan results in 200 more lines of code for hello_triangle than we used to have.

I think that's at least partially due to the sample doing things that the original hello triangle sample is missing. E.g. creating a proper vertex buffer instead of hardcoding it in the shaders like the non 1.3 sample does. But I agree that this sample could be an opportunity to reduce line count. It does a several not so optimal things 1:1 like the hello triangle sample. E.g. having a function to deduce the shader stage from a file extension while it only uses vertex and fragment shader stages.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1340 @@
...
+
+	uint32_t instance_layer_count;
+	VK_CHECK(vkEnumerateInstanceLayerProperties(&instance_layer_count, nullptr));
+
+	std::vector<VkLayerProperties> supported_validation_layers(instance_layer_count);
+	VK_CHECK(vkEnumerateInstanceLayerProperties(&instance_layer_count, supported_validation_layers.data()));
+
+	std::vector<const char *> requested_validation_layers{};
+
+#ifdef VKB_VALIDATION_LAYERS
```
SaschaWillems: This is missing the VKB_DEBUG define, see https://github.com/KhronosGroup/Vulkan-Samples/blob/main/samples/api/hello_triangle/hello_triangle.cpp#L237

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1340 @@
...
+	return all_found;
+}
+
+/**
+ * @brief Find the vulkan shader stage for a given a string.
+ *
+ * @param ext A string containing the shader stage name.
+ * @return VkShaderStageFlagBits The shader stage mapping from the given string, VK_SHADER_STAGE_VERTEX_BIT otherwise.
+ */
+VkShaderStageFlagBits HelloTriangleV13::find_shader_stage(const std::string &ext)
```
SaschaWillems: This could be simplified or removed all along and making the sample pass the stage to the shader loading function. The sample only uses vert and frag anyway.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1340 @@
...
+	VK_CHECK(vkCreateSwapchainKHR(context.device, &info, nullptr, &context.swapchain));
+
+	if (old_swapchain != VK_NULL_HANDLE)
+	{
+		for (VkImageView image_view : context.swapchain_image_views)
+		{
+			vkDestroyImageView(context.device, image_view, nullptr);
+		}
+
+		uint32_t image_count;
```
SaschaWillems: This is not spec compliant, and needs to be changed, see #1227

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1340 @@
...
+	};
+
+	// Record the pipeline barrier into the command buffer
+	vkCmdPipelineBarrier2(cmd, &dependency_info);
+}
+
+/**
+ * @brief Tears down the Vulkan context.
+ */
+void HelloTriangleV13::teardown()
```
SaschaWillems: This function is only called once from the destructor below. It's code could be moved to the d'tor.

SaschaWillems (changes requested):
I did add a few comments, mostly things also not that great with the VK 1.0 hello triangle sample.

But I also noticed lots of validation errors when running this sample that need to be fixed:

```
[error] -1733013297 Validation Layer: Error: VUID-VkBufferCreateInfo-flags-parameter: Validation Error: [ VUID-VkBufferCreateInfo-flags-parameter ] | MessageID = 0x98b450cf | vkCreateBuffer(): pCreateInfo->flags contains flag bits (0x80) which are not recognized members of VkBufferCreateFlagBits.
The Vulkan spec states: flags must be a valid combination of VkBufferCreateFlagBits values (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-VkBufferCreateInfo-flags-parameter)
[error] -153856805 Validation Layer: Error: VUID-VkBufferCreateInfo-None-09500: Validation Error: [ VUID-VkBufferCreateInfo-None-09500 ] | MessageID = 0xf6d454db | vkCreateBuffer(): pCreateInfo->usage is zero.
The Vulkan spec states: If the pNext chain does not include a VkBufferUsageFlags2CreateInfoKHR structure, usage must not be 0 (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-VkBufferCreateInfo-None-09500)
[error] 1895430949 Validation Layer: Error: VUID-VkGraphicsPipelineCreateInfo-topology-08773: Validation Error: [ VUID-VkGraphicsPipelineCreateInfo-topology-08773 ] Object 0: handle = 0xd175b40000000013, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0x70f9fb25 | vkCreateGraphicsPipelines(): pCreateInfos[0].pStages[0] SPIR-V (Vertex) PointSize is not written to, but Pipeline topology is set to VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the pipeline is being created with a Vertex Execution Model and no TessellationEvaluation or Geometry Execution Model, and the topology member of pInputAssembly is VK_PRIMITIVE_TOPOLOGY_POINT_LIST, and either VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state is not enabled or dynamicPrimitiveTopologyUnrestricted is VK_FALSE, a PointSize decorated variable must be written to if maintenance5 is not enabled (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-topology-08773)
[error] -1696391559 Validation Layer: Error: VUID-vkCmdBindVertexBuffers-pBuffers-00627: Validation Error: [ VUID-vkCmdBindVertexBuffers-pBuffers-00627 ] Object 0: handle = 0x1fb5f6efca0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x9ae31e79 | vkCmdBindVertexBuffers(): pBuffers[0] (VkBuffer 0xfa21a40000000003[]) was created with VkBufferUsageFlags2KHR(0) but requires VK_BUFFER_USAGE_VERTEX_BUFFER_BIT.
The Vulkan spec states: All elements of pBuffers must have been created with the VK_BUFFER_USAGE_VERTEX_BUFFER_BIT flag (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers-pBuffers-00627)
[error] -901174136 Validation Layer: Error: VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500: Validation Error: [ VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x1fb5f6efca0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xca492c88 | vkCmdDraw():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500)
[error] -1696391559 Validation Layer: Error: VUID-vkCmdBindVertexBuffers-pBuffers-00627: Validation Error: [ VUID-vkCmdBindVertexBuffers-pBuffers-00627 ] Object 0: handle = 0x1fb5f6f3830, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x9ae31e79 | vkCmdBindVertexBuffers(): pBuffers[0] (VkBuffer 0xfa21a40000000003[]) was created with VkBufferUsageFlags2KHR(0) but requires VK_BUFFER_USAGE_VERTEX_BUFFER_BIT.
The Vulkan spec states: All elements of pBuffers must have been created with the VK_BUFFER_USAGE_VERTEX_BUFFER_BIT flag (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers-pBuffers-00627)
[error] -901174136 Validation Layer: Error: VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500: Validation Error: [ VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x1fb5f6f3830, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xca492c88 | vkCmdDraw():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500)
[error] -1696391559 Validation Layer: Error: VUID-vkCmdBindVertexBuffers-pBuffers-00627: Validation Error: [ VUID-vkCmdBindVertexBuffers-pBuffers-00627 ] Object 0: handle = 0x1fb5f6efca0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x9ae31e79 | vkCmdBindVertexBuffers(): pBuffers[0] (VkBuffer 0xfa21a40000000003[]) was created with VkBufferUsageFlags2KHR(0) but requires VK_BUFFER_USAGE_VERTEX_BUFFER_BIT.
The Vulkan spec states: All elements of pBuffers must have been created with the VK_BUFFER_USAGE_VERTEX_BUFFER_BIT flag (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers-pBuffers-00627)
[error] -901174136 Validation Layer: Error: VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500: Validation Error: [ VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x1fb5f6efca0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xca492c88 | vkCmdDraw():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500)
[error] -1696391559 Validation Layer: Error: VUID-vkCmdBindVertexBuffers-pBuffers-00627: Validation Error: [ VUID-vkCmdBindVertexBuffers-pBuffers-00627 ] Object 0: handle = 0x1fb5f6efca0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x9ae31e79 | vkCmdBindVertexBuffers(): pBuffers[0] (VkBuffer 0xfa21a40000000003[]) was created with VkBufferUsageFlags2KHR(0) but requires VK_BUFFER_USAGE_VERTEX_BUFFER_BIT.
The Vulkan spec states: All elements of pBuffers must have been created with the VK_BUFFER_USAGE_VERTEX_BUFFER_BIT flag (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers-pBuffers-00627)
[error] -901174136 Validation Layer: Error: VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500: Validation Error: [ VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x1fb5f6efca0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xca492c88 | vkCmdDraw():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500)
[error] -1696391559 Validation Layer: Error: VUID-vkCmdBindVertexBuffers-pBuffers-00627: Validation Error: [ VUID-vkCmdBindVertexBuffers-pBuffers-00627 ] Object 0: handle = 0x1fb5f6f3830, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x9ae31e79 | vkCmdBindVertexBuffers(): pBuffers[0] (VkBuffer 0xfa21a40000000003[]) was created with VkBufferUsageFlags2KHR(0) but requires VK_BUFFER_USAGE_VERTEX_BUFFER_BIT.
The Vulkan spec states: All elements of pBuffers must have been created with the VK_BUFFER_USAGE_VERTEX_BUFFER_BIT flag (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers-pBuffers-00627)
[error] -901174136 Validation Layer: Error: VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500: Validation Error: [ VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x1fb5f6f3830, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xca492c88 | vkCmdDraw():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500)
[error] -1696391559 Validation Layer: Error: VUID-vkCmdBindVertexBuffers-pBuffers-00627: Validation Error: [ VUID-vkCmdBindVertexBuffers-pBuffers-00627 ] Object 0: handle = 0x1fb5f6f93d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x9ae31e79 | vkCmdBindVertexBuffers(): pBuffers[0] (VkBuffer 0xfa21a40000000003[]) was created with VkBufferUsageFlags2KHR(0) but requires VK_BUFFER_USAGE_VERTEX_BUFFER_BIT.
The Vulkan spec states: All elements of pBuffers must have been created with the VK_BUFFER_USAGE_VERTEX_BUFFER_BIT flag (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers-pBuffers-00627)
[error] -901174136 Validation Layer: Error: VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500: Validation Error: [ VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x1fb5f6f93d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xca492c88 | vkCmdDraw():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500)
[error] -1696391559 Validation Layer: Error: VUID-vkCmdBindVertexBuffers-pBuffers-00627: Validation Error: [ VUID-vkCmdBindVertexBuffers-pBuffers-00627 ] Object 0: handle = 0x1fb5f6efca0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x9ae31e79 | vkCmdBindVertexBuffers(): pBuffers[0] (VkBuffer 0xfa21a40000000003[]) was created with VkBufferUsageFlags2KHR(0) but requires VK_BUFFER_USAGE_VERTEX_BUFFER_BIT.
The Vulkan spec states: All elements of pBuffers must have been created with the VK_BUFFER_USAGE_VERTEX_BUFFER_BIT flag (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers-pBuffers-00627)
[error] -901174136 Validation Layer: Error: VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500: Validation Error: [ VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x1fb5f6efca0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xca492c88 | vkCmdDraw():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500)
[error] -1696391559 Validation Layer: Error: VUID-vkCmdBindVertexBuffers-pBuffers-00627: Validation Error: [ VUID-vkCmdBindVertexBuffers-pBuffers-00627 ] Object 0: handle = 0x1fb5f6f3830, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x9ae31e79 | vkCmdBindVertexBuffers(): pBuffers[0] (VkBuffer 0xfa21a40000000003[]) was created with VkBufferUsageFlags2KHR(0) but requires VK_BUFFER_USAGE_VERTEX_BUFFER_BIT.
The Vulkan spec states: All elements of pBuffers must have been created with the VK_BUFFER_USAGE_VERTEX_BUFFER_BIT flag (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers-pBuffers-00627)
[error] -901174136 Validation Layer: Error: VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500: Validation Error: [ VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x1fb5f6f3830, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xca492c88 | vkCmdDraw():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500)
[error] -1696391559 Validation Layer: Error: VUID-vkCmdBindVertexBuffers-pBuffers-00627: Validation Error: [ VUID-vkCmdBindVertexBuffers-pBuffers-00627 ] Object 0: handle = 0x1fb5f6f93d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x9ae31e79 | vkCmdBindVertexBuffers(): pBuffers[0] (VkBuffer 0xfa21a40000000003[]) was created with VkBufferUsageFlags2KHR(0) but requires VK_BUFFER_USAGE_VERTEX_BUFFER_BIT.
The Vulkan spec states: All elements of pBuffers must have been created with the VK_BUFFER_USAGE_VERTEX_BUFFER_BIT flag (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers-pBuffers-00627)
[error] -901174136 Validation Layer: Error: VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500: Validation Error: [ VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x1fb5f6f93d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xca492c88 | vkCmdDraw():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500)
[error] -1696391559 Validation Layer: Error: VUID-vkCmdBindVertexBuffers-pBuffers-00627: Validation Error: [ VUID-vkCmdBindVertexBuffers-pBuffers-00627 ] Object 0: handle = 0x1fb5f6f3830, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x9ae31e79 | vkCmdBindVertexBuffers(): pBuffers[0] (VkBuffer 0xfa21a40000000003[]) was created with VkBufferUsageFlags2KHR(0) but requires VK_BUFFER_USAGE_VERTEX_BUFFER_BIT.
The Vulkan spec states: All elements of pBuffers must have been created with the VK_BUFFER_USAGE_VERTEX_BUFFER_BIT flag (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers-pBuffers-00627)
[error] -901174136 Validation Layer: Error: VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500: Validation Error: [ VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x1fb5f6f3830, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xca492c88 | vkCmdDraw():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_POINT_LIST.
The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500)
```

samples/CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ endforeach ()
 set(ORDER_LIST
     #API Samples
     "hello_triangle"
+	"hello_triangle_1_3"
```
asuessenbach: Please replace tab with spaces.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1290 @@
...
+
+	uint32_t instance_layer_count;
+	VK_CHECK(vkEnumerateInstanceLayerProperties(&instance_layer_count, nullptr));
+
+	std::vector<VkLayerProperties> supported_validation_layers(instance_layer_count);
+	VK_CHECK(vkEnumerateInstanceLayerProperties(&instance_layer_count, supported_validation_layers.data()));
+
+	std::vector<const char *> requested_validation_layers{};
+
+#ifdef defined(VKB_DEBUG) || VKB_VALIDATION_LAYERS
```
asuessenbach: Should probably be
`#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
`

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1290 @@
...
+	VkDebugUtilsMessengerCreateInfoEXT debug_messenger_create_info = {VK_STRUCTURE_TYPE_DEBUG_UTILS_MESSENGER_CREATE_INFO_EXT};
+	if (has_debug_utils)
+	{
+		debug_messenger_create_info.messageSeverity = VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT | VK_DEBUG_UTILS_MESSAGE_SEVERITY_WARNING_BIT_EXT;
+		debug_messenger_create_info.messageType     = VK_DEBUG_UTILS_MESSAGE_TYPE_GENERAL_BIT_EXT | VK_DEBUG_UTILS_MESSAGE_TYPE_VALIDATION_BIT_EXT;
+		debug_messenger_create_info.pfnUserCallback = debug_callback;
+
+		instance_info.pNext = &debug_messenger_create_info;
+	}
+#endif
```
asuessenbach, line 242: Why have you skipped the portability handling here:
```
#if (defined(VKB_ENABLE_PORTABILITY))
	if (portability_enumeration_available)
	{
		instance_info.flags |= VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
	}
#endif

```

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1290 @@
...
+	VkPhysicalDeviceExtendedDynamicStateFeaturesEXT extended_dynamic_state_features{VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTENDED_DYNAMIC_STATE_FEATURES_EXT};
+	device_features2.pNext  = &vulkan13_features;
+	vulkan13_features.pNext = &extended_dynamic_state_features;
+
+	vkGetPhysicalDeviceFeatures2(context.gpu, &device_features2);
+
+	// Enable required Vulkan 1.3 features
+	vulkan13_features.dynamicRendering                   = VK_TRUE;
+	vulkan13_features.synchronization2                   = VK_TRUE;
+	extended_dynamic_state_features.extendedDynamicState = VK_TRUE;
```
asuessenbach: This is not the correct way to handle feature requests.

With vkGetPhysicalDeviceFeatures2, you get what features would be supported. You could check them here to decide if your code would be able to run with or without them.

Unconditionally setting them to VK_TRUE is wrong. If they were VK_FALSE before, device creation would just fail.

Instead, you should use feature structs, where just those features are set to VK_TRUE you really want to ask for, and use those to create the device.

asuessenbach (commented):
LGTM, just a few issues.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
alihuawei: Hmm, my thought process was i am going to check for Vulkan 1.3 which is not supported on IOS anyway but you are correct i should do portability handling so when V1.3 is available on iOS it should work.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
SaschaWillems: Not sure if that's what we want. It's a 1.3 sample and those features are a requirement. So imo it's fine to set them that way, as device creation will then fail with a proper error message.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1290 @@
...
+	{
+		resize(context.swapchain_dimensions.width, context.swapchain_dimensions.height);
+	}
+	else if (res != VK_SUCCESS)
+	{
+		LOGE("Failed to present swapchain image.");
+	}
+}
+
+bool HelloTriangleV13::resize(const uint32_t, const uint32_t)
```
SaschaWillems, line 1286: Probably already the case for the original hello triangle sample: This function returns a bool but there is no code checking the function's return code. Should be either void or the result should be checked.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1290 @@
...
+	    .imageLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL,
+	    .loadOp      = VK_ATTACHMENT_LOAD_OP_CLEAR,
+	    .storeOp     = VK_ATTACHMENT_STORE_OP_STORE,
+	    .clearValue  = clear_value};
+
+	// Begin rendering
+	VkRenderingInfo rendering_info{
+	    .sType                = VK_STRUCTURE_TYPE_RENDERING_INFO_KHR,
+	    .renderArea           = {                         // Initialize the nested `VkRect2D` structure
+	                             .offset = {0, 0},        // Initialize the `VkOffset2D` inside `renderArea`
```
SaschaWillems, line 962: Indentation here looks to be off.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1290 @@
...
+	VkViewport vp{
+	    .width    = static_cast<float>(context.swapchain_dimensions.width),
+	    .height   = static_cast<float>(context.swapchain_dimensions.height),
+	    .minDepth = 0.0f,
+	    .maxDepth = 1.0f};
+
+	vkCmdSetViewport(cmd, 0, 1, &vp);
+
+	// Set scissor dynamically
+	VkRect2D scissor{};
```
SaschaWillems: Should be changed to use designated initialization.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1290 @@
...
+	    context.swapchain_images[swapchain_index],
+	    VK_IMAGE_LAYOUT_UNDEFINED,
+	    VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL,
+	    0,                                                     // srcAccessMask (no need to wait for previous operations)
+	    VK_ACCESS_2_COLOR_ATTACHMENT_WRITE_BIT,                // dstAccessMask
+	    VK_PIPELINE_STAGE_2_TOP_OF_PIPE_BIT,                   // srcStage
+	    VK_PIPELINE_STAGE_2_COLOR_ATTACHMENT_OUTPUT_BIT        // dstStage
+	);
+	// Set clear color values.
+	VkClearValue clear_value;
```
SaschaWillems: Should be changed to use designated initialization.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1290 @@
...
+	for (const auto &physical_device : gpus)
+	{
+		// Check if the device supports Vulkan 1.3
+		VkPhysicalDeviceProperties device_properties;
+		vkGetPhysicalDeviceProperties(physical_device, &device_properties);
+
+		uint32_t major_version = VK_API_VERSION_MAJOR(device_properties.apiVersion);
+		uint32_t minor_version = VK_API_VERSION_MINOR(device_properties.apiVersion);
+
+		if ((major_version < 1) || (major_version == 1 && minor_version < 3))
```
SaschaWillems: The first expression (major_version < 1) seems to be superfluous? But I guess this could probably be replaced by something as simple as:

```if (device_properties.apiVersion < VK_API_VK_API_VERSION_1_3```

File: samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp [hunk 15]
```diff
@@ -0,0 +1,1290 @@
...
+		// Create an image view which we can render into.
+		VkImageViewCreateInfo view_info{
+		    .sType      = VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO,
+		    .pNext      = nullptr,
+		    .flags      = 0,
+		    .image      = swapchain_images[i],
+		    .viewType   = VK_IMAGE_VIEW_TYPE_2D,
+		    .format     = context.swapchain_dimensions.format,
+		    .components = {
+		        .r = VK_COMPONENT_SWIZZLE_R,
```
SaschaWillems: This whole black can be removed. Designated initialization does zero initialization, which means all components are set to `VK_COMPONENT_SWIZZLE_IDENTITY`.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1290 @@
...
+	// Define the vertex input binding description
+	VkVertexInputBindingDescription binding_description{
+	    .binding   = 0,
+	    .stride    = sizeof(Vertex),
+	    .inputRate = VK_VERTEX_INPUT_RATE_VERTEX};
+
+	// Define the vertex input attribute descriptions
+	std::array<VkVertexInputAttributeDescription, 2> attribute_descriptions{};
+
+	attribute_descriptions[0].binding  = 0;
```
SaschaWillems: Could be changed to designated initializers.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1290 @@
...
+	VkPipelineDynamicStateCreateInfo dynamic_state_info{
+	    .sType             = VK_STRUCTURE_TYPE_PIPELINE_DYNAMIC_STATE_CREATE_INFO,
+	    .dynamicStateCount = static_cast<uint32_t>(dynamic_states.size()),
+	    .pDynamicStates    = dynamic_states.data()};
+
+	// Load our SPIR-V shaders.
+	std::array<VkPipelineShaderStageCreateInfo, 2> shader_stages{};
+
+	// Vertex stage of the pipeline
+	shader_stages[0].sType  = VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO;
```
SaschaWillems: Could be changed to designated initializers.

SaschaWillems (changes requested):
No more validation errors :)

Added a few comments, but mostly (if not all) are cosmetic.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
alihuawei: resize is getting inherited from vkb::Application so changing it to void is going to break other samples.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
alihuawei: yeah tried to make it look structured but clang_format script brings this off looking indentation back

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
alihuawei: Since this sample is specifically targeting Vulkan 1.3-compatible hardware, and the code already checks for Vulkan 1.3 support on the selected device, there's no need to enable the portability extension (VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR). The portability subset is mainly used to ensure Vulkan works on devices with partial or non-standard Vulkan implementations, such as older or embedded systems.

However, in this case, I am focusing on platforms where Vulkan 1.3 is fully supported (modern desktops, mobile devices, etc.). Therefore, enabling the portability flag is unnecessary.

@SaschaWillems  what is your opinion about this ?

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
SaschaWillems: Sorry, didn't realize that. The simply ignore this :)

SaschaWillems (changes requested):
The sample itself looks good to me now :)

One thing that should be expanded though is the readme. Newer samples should come with  at least a basic tutorial or overview of what the sample does / how it works.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
asuessenbach: I'm sure, that's what we want.
I think, the samples here should show how to do it right.
You're right, the spec says that you **can** check for extension presence and you **can** check for feature support. You don't have to, as `vkCreateDevice` would fail with some appropriate error code if you ask for something not supported.
But here, you're asking for each and every feature (listed by your structure chain) that's supported by that device (plus some that might actually not be supported). You should not do that, as it might result in sub-optimal performance or 
unnecessary memory consumption.
Instead, you should just ask for the extensions you need and the features you need.
And I think, it would be a good habit to check for their support before just trying to create the device.

alihuawei: I agree that i should only enable the features that i am using. I will update the code

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
SaschaWillems: Agree 100%. MoltenVK can't do 1.3 and it's probably going to take a while before it gets 1.3 support. And I don't think we should delay this sample because of that.

So why not remove the Mac related stuff for now and add support in a different PR once MoltenVK get's support for 1.3.

asuessenbach: Ok, agreed.

alihuawei:
I accidentally force-pushed earlier without rebasing my branch on the latest main. I have now rebased my branch onto the latest main, resolved all feedback, and force-pushed the corrected changes. The branch is now clean and ready for review. Apologies for the confusion caused by the rewritten history.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+	VkPhysicalDeviceFeatures2                       query_device_features2{VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2};
+	VkPhysicalDeviceVulkan13Features                query_vulkan13_features{VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_3_FEATURES};
+	VkPhysicalDeviceExtendedDynamicStateFeaturesEXT query_extended_dynamic_state_features{VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTENDED_DYNAMIC_STATE_FEATURES_EXT};
+	query_device_features2.pNext  = &query_vulkan13_features;
+	query_vulkan13_features.pNext = &query_extended_dynamic_state_features;
+
+	vkGetPhysicalDeviceFeatures2(context.gpu, &query_device_features2);
+
+	// Check if Physical device supports Vulkan 1.3 features
+	// Check if Physical device supports Vulkan 1.3 features
```
asuessenbach: Duplicated comment?

SaschaWillems (dismissed):
LGTM. The readme is great and a nice list of things that make VK 1.3 so interesting 👍🏻

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+
+		if (!found)
+		{
+			// Output an error message for the missing extension
+			LOGE("Error: Required extension not found: {}", extension_name);
+			all_found = false;
+		}
+	}
+
+	return all_found;
```
asuessenbach, line 90: Maybe somewhat more "modern C++":
```
	return std::all_of(required.begin(),
	                   required.end(),
	                   [&available](const char *extension_name) {
		                   bool found = std::any_of(
		                       available.begin(), available.end(), [extension_name](auto const &ep) { return strcmp(ep.extensionName, extension_name) == 0; });
		                   if (!found)
		                   {
			                   // Output an error message for the missing extension
			                   LOGE("Error: Required extension not found: {}", extension_name);
		                   }
		                   return found;
	                   });
```
But it's up to you, as it's your sample.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+		}
+
+		if (!found)
+		{
+			LOGE("Error: Required layer not found: {}", layer_name);
+			all_found = false;
+		}
+	}
+
+	return all_found;
```
asuessenbach, line 125: Similar to the above:
```
	return std::all_of(required.begin(),
	                   required.end(),
	                   [&available](const char *layer_name) {
		                   bool found = std::any_of(
		                       available.begin(), available.end(), [layer_name](auto const &lp) { return strcmp(lp.layerName, layer_name) == 0; });
		                   if (!found)
		                   {
			                   // Output an error message for the missing extension
			                   LOGE("Error: Required layer not found: {}", layer_name);
		                   }
		                   return found;
	                   });
```

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
+	bool has_debug_utils = false;
+	for (const auto &ext : available_instance_extensions)
+	{
+		if (strncmp(ext.extensionName, VK_EXT_DEBUG_UTILS_EXTENSION_NAME, strlen(VK_EXT_DEBUG_UTILS_EXTENSION_NAME)) == 0)
+		{
+			has_debug_utils = true;
+			break;
+		}
+	}
```
asuessenbach, line 157: Similar to the above:
```
	bool has_debug_utils = (std::any_of(
	    available_instance_extensions.begin(),
	    [](auto const &ext) { return strncmp(ext.extensionName, VK_EXT_DEBUG_UTILS_EXTENSION_NAME, strlen(VK_EXT_DEBUG_UTILS_EXTENSION_NAME)) == 0; }));
```

asuessenbach (commented):
Thanks for correctly checking for the features. Looks good now.

Just one duplicate comment remaining and a couple of suggestions.

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
alihuawei: @asuessenbach I appreciate your suggestion, but I’d like to avoid lambda functions in this code. I’m aiming to keep the logic more explicit for readability and maintenance, so I’m going to write a named predicate function instead.

SaschaWillems (approved):

samples/api/hello_triangle_1_3/hello_triangle_1_3.cpp
SaschaWillems: Same here. While that may be modern C++, it's imo harder to read.

asuessenbach: Well, I think an explicit loop is harder to read than an algorithm.

asuessenbach (approved):

gary-sweet (approved):

marty-johnson59:
3 approvals - merging.  Thanks!

PR #1284, opened 2025-02-26
Add missing VK_KHR_portability_subset extension to [hpp_]hello_triangle[*] samples

### LINKED ISSUE #1283: VK_KHR_portability_subset device extension missing in [hpp_]hello_triangle[*] samples
I'm not sure how I missed this before, but the `VK_KHR_portability_subset` device extension is missing for the  `hello_triangle`, `hpp_hello_triangle`, and `hello_triangle_1_3` samples.  It \*is\* present in the general framework for other samples, but somehow I forgot to add it to these standalone samples.

The spec says that `VK_KHR_portability_subset` must be enabled if available in the implementation.  Leaving it out leads to validation errors on macOS/iOS when the validation layer is enabled:

```
[error] 976972960 Validation Layer: Error: VUID-VkDeviceCreateInfo-pProperties-04451: Validation Error: [ VUID-VkDeviceCreateInfo-pProperties-04451 ] Object 0: handle = 0x600000270200, type = VK_OBJECT_TYPE_PHYSICAL_DEVICE; | MessageID = 0x3a3b6ca0 | vkCreateDevice(): VK_KHR_portability_subset must be enabled because physical device VkPhysicalDevice 0x600000270200[] supports it.
The Vulkan spec states: If the VK_KHR_portability_subset extension is included in pProperties of vkEnumerateDeviceExtensionProperties, ppEnabledExtensionNames must include "VK_KHR_portability_subset" (https://vulkan.lunarg.com/doc/view/1.4.304.0/mac/1.4-extensions/vkspec.html#VUID-VkDeviceCreateInfo-pProperties-04451)
```
I will submit a PR to fix.

SRSaunders:
## Description

If portability is enabled (i.e. `VKB_ENABLE_PORTABILITY`) and the device supports the extension at runtime, this PR adds the `VK_KHR_portability_subset` device extension in the standalone samples: `hello_triangle`, `hpp_hello_triangle`, and `hello_triangle_1_3`.  This is required by the spec and prevents validation errors when the validation layer is enabled.

There is no change to the general framework where `VK_KHR_portability_subset` is already enabled as required for other non-standalone samples.

Fixes #1283

Tested on macOS Ventura and iOS (iPhone 15 physical device and simulator).

## General Checklist:

## Sample Checklist

### DIFF
--- a/samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
+++ b/samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
@@ -307,10 +307,22 @@ vk::Device HPPHelloTriangle::create_device(const std::vector<const char *> &requ
 		throw std::runtime_error("Required device extensions are missing, will try without.");
 	}
 
+	std::vector<const char *> active_device_extensions(required_device_extensions);
+
+#if (defined(VKB_ENABLE_PORTABILITY))
+	// VK_KHR_portability_subset must be enabled if present in the implementation (e.g on macOS/iOS with beta extensions enabled)
+	if (std::any_of(device_extensions.begin(),
+	                device_extensions.end(),
+	                [](vk::ExtensionProperties const &extension) { return strcmp(extension.extensionName, VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME) == 0; }))
+	{
+		active_device_extensions.push_back(VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME);
+	}
+#endif
+
 	// Create a device with one queue
 	float                     queue_priority = 1.0f;
 	vk::DeviceQueueCreateInfo queue_info({}, graphics_queue_index, 1, &queue_priority);
-	vk::DeviceCreateInfo      device_info({}, queue_info, {}, required_device_extensions);
+	vk::DeviceCreateInfo      device_info({}, queue_info, {}, active_device_extensions);
 	vk::Device                device = gpu.createDevice(device_info);
 
 	// initialize function pointers for device
--- a/samples/api/hello_triangle/hello_triangle.cpp
+++ b/samples/api/hello_triangle/hello_triangle.cpp
@@ -327,6 +327,16 @@ void HelloTriangle::init_device()
 		throw std::runtime_error("Required device extensions are missing.");
 	}
 
+#if (defined(VKB_ENABLE_PORTABILITY))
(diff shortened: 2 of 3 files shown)


samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
```diff
@@ -307,10 +307,22 @@ vk::Device HPPHelloTriangle::create_device(const std::vector<const char *> &requ
...
+
+#if (defined(VKB_ENABLE_PORTABILITY))
+	// VK_KHR_portability_subset must be enabled if present in the implementation (e.g on macOS/iOS with beta extensions enabled)
+	if (std::any_of(device_extensions.begin(),
+	                device_extensions.end(),
+	                [](vk::ExtensionProperties const &extension) { return strcmp(extension.extensionName, VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME) == 0; }))
+	{
+		active_device_extensions.push_back(VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME);
+	}
+#endif
```
asuessenbach, line 320: Instead of copying over the complete vector `required_device_extensions` (ok, it holds just one element...) I would prefer to have that extension be added to the `required_device_extension` where `create_device` is called (line 160).

SRSaunders: Thanks @asuessenbach.  I actually prototyped that option first since `required_device_extensions` is a `const`.  However I came back to creating and using `active_device_extensions` for a few reasons:

1. The spec requires you to check if `VK_KHR_portability_subset` is available before enabling, and thus needs `device_extensions` to be available, which it is not in the calling function.
2. `create_instance()` uses the same code pattern by copying `required_instance_extensions` into `active_instance_extensions` before adding additional extensions to the vector.
3. I thought it was better to encapsulate this logic in `create_device()` vs "pollute" the calling function.

asuessenbach: > create_instance() uses the same code pattern by copying required_instance_extensions into active_instance_extensions before adding additional extensions to the vector.

I see... it's been a while since I had a look at those samples.
That is: agreed, doing it this way is consistent with the rest of the code here.

Thanks for the careful editing.

asuessenbach (approved):

gary-sweet (approved):

SaschaWillems (approved):

PR #1295, opened 2025-03-11
Update astc to 4.8.0, enable SSE41 decode, and fix build issues for physical iOS targets

### LINKED ISSUE #1294: New astc library causes build problems for physical iOS targets
The new astc library from commit https://github.com/KhronosGroup/Vulkan-Samples/pull/1292/commits/7017dc07869aa0fd0eba1fd440010f10d56407b2 does not generate a static (.a) library when cross-compiling from x86_64 hosts to arm64 targets.  This causes the build to fail.  Note that native compilation from x86_64 hosts to x86_64 targets, and from arm64 hosts to arm64 targets works properly, and this is likely why the CI is still working for macOS and iOS builds.

The issue is with the CMakeLists in the third_party directory - it forces "native" on for all Apple builds.  The solution is to make sure "neon" is set for all iOS physical (not simulator) targets, and "native" otherwise.

I will push a PR to solve this.

SRSaunders:
## Description

The new astc library from commit https://github.com/KhronosGroup/Vulkan-Samples/commit/7017dc07869aa0fd0eba1fd440010f10d56407b2 does not generate a static (.a) library when cross-compiling from x86_64 hosts to arm64 targets. This PR fixes this build problem by making sure "neon" is set for all iOS physical (not simulator) targets, and "native" otherwise.

This PR also removes a previous workaround for macOS/iOS arm64 targets since the original defect is now fixed.  See fixed issue https://github.com/ARM-software/astc-encoder/issues/458.

Fixes #1294.

Tested on macOS x86_64 (Intel iMac) and arm64 (M1 MacBook Air) hosts building for native macOS execution.  Also tested on macOS x86_64 hosts (Intel iMac) building for iOS physical (arm64 iPhone 15) and simulator (x86_64 host) targets.

## General Checklist:

## Sample Checklist

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -308,8 +308,13 @@ if (NOT (DEFINED ASTC_ARCH))
     # `ld: object file .../libastcdec-avx2-static.a was built for different x86_64 sub-type (8) than link command line (3)`
     # so we fallback to native. If we're on an ARM based mac, there's no need for any SIMD extension, because they support
     # ASTC_LDR natively, so this library won't even be used.
+    # Always build for ARM when compiling or cross-compiling to iOS physical devices, otherwise fall back to native as above
     if (APPLE)
-        set(ASTC_ARCH NATIVE)
+		if(IOS AND ${CMAKE_OSX_SYSROOT} STREQUAL "iphoneos")
+			set(ASTC_ARCH NEON)
+		else()
+			set(ASTC_ARCH NATIVE)
+		endif()
     else()
         include( FindAVX2 )
         include( FindSSE2 )
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -304,23 +304,33 @@ target_include_directories(stb SYSTEM INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/stb
 # Allow the user to specify the ASTC architecture explicitly if building for a platform
 # other than the host architecture (i.e. x86 on an x64 CPU)
 if (NOT (DEFINED ASTC_ARCH))
-    # On an x86_64 macbook, AVX2 is selected, but linking fails with this message,
-    # `ld: object file .../libastcdec-avx2-static.a was built for different x86_64 sub-type (8) than link command line (3)`
-    # so we fallback to native. If we're on an ARM based mac, there's no need for any SIMD extension, because they support
-    # ASTC_LDR natively, so this library won't even be used.
-    # Always build for ARM when compiling or cross-compiling to iOS physical devices, otherwise fall back to native as above
     if (APPLE)
-		if(IOS AND ${CMAKE_OSX_SYSROOT} STREQUAL "iphoneos")
+		# Always use NEON when host is arm64 or cross-compiling to iOS physical devices,
+		# otherwise look for supported vector instructions based on x86_64 host processor.
+		# Don't use AVX2 since it builds for x86_64h which does not match x86_64 project,
+		# and lastly fall back to NATIVE if no matches found.
+		if(${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "arm64" OR ${CMAKE_OSX_SYSROOT} STREQUAL "iphoneos")
 			set(ASTC_ARCH NEON)
 		else()
-			set(ASTC_ARCH NATIVE)
+			include( FindSSE41 )
+			include( FindSSE2 )
+			if (${SSE41_FOUND})
+				set(ASTC_ARCH SSE41)
+			elseif(${SSE2_FOUND})
+				set(ASTC_ARCH SSE2)
+			else()
+				set(ASTC_ARCH NATIVE)
+			endif()
 		endif()
     else()
         include( FindAVX2 )
+		include( FindSSE41 )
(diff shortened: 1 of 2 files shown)


third_party/CMakeLists.txt
```diff
@@ -308,8 +308,13 @@ if (NOT (DEFINED ASTC_ARCH))
     # `ld: object file .../libastcdec-avx2-static.a was built for different x86_64 sub-type (8) than link command line (3)`
     # so we fallback to native. If we're on an ARM based mac, there's no need for any SIMD extension, because they support
     # ASTC_LDR natively, so this library won't even be used.
+    # Always build for ARM when compiling or cross-compiling to iOS physical devices, otherwise fall back to native as above
     if (APPLE)
-        set(ASTC_ARCH NATIVE)
+		if(IOS AND ${CMAKE_OSX_SYSROOT} STREQUAL "iphoneos")
```
gpx1000: Does this work for iPad?

Great catch by the way!

SRSaunders: @gpx1000 yes it should work for iPad as is (which is an iOS target from a developer perspective).  As far as I know, the valid Apple target platforms (cmake SysRoots) are:

Screenshot 2025-03-11 at 5 46 38 PM

This project supports three of them: _macosx_, _iphoneos_, and _iphonesimulator_.  While I have not tested on a physical iPad (don't have one), I have tested using a virtual iPad using the _iphonesimulator_ target, and that works fine.  If you have a physical iPad, perhaps you could test and let me know if there are any issues.  Thx.

Note that _iphoneos_ is already used in **.../Vulkan-Samples/app/CMakeLists.txt**, and _iphonesimulator_ is used in **.../Vulkan-Samples/bldsys/cmake/global_options.cmake** to test for iOS physical devices vs. sim  devices.

gpx1000: Yep, that looks right, the CMake docs say that it could be appleTV, iOS, MacOS, WatchOS, or visionOS etc.  I'm okay with the changes here thanks!

gpx1000 (dismissed):

third_party/CMakeLists.txt
SRSaunders: Thanks for reviewing!

SRSaunders:
As per discussion with @gpx1000 and @gary-sweet I have updated the astc library to version 4.8.0 to pick up macOS fix https://github.com/ARM-software/astc-encoder/commit/71a55e7697a72df9e004a8212288f1ef71572f9e.

In addition, I have added an enhancement to enable SSE41 decoding for all x86 platforms (Win, Linux, macOS) as a fallback option if AVX2 is not available.  This was likely not enabled previously due to a mismatch in naming convention for the target library which should be **libastcdec-sse4.1-static.a** and not **libastcdec-sse41-static.a**.  Subtle difference with "." in name.

SRSaunders:
@SaschaWillems, if you review this PR please note that I could possibly collapse the APPLE vs. Other logic for **astc** vector instruction selection into a single cmake code block.  However, I was reluctant to do this without more info since I don't really understand how Android is built.  I thought it would be a normal C++ cross-compilation target from x86 to arm64, but looking at the CI build log for Android it seems more complicated than that.

In fact I see that many libraries are not built during CI, with the message: `not building target library <...> because static libraries are not build by default`

Also the sample targets are not built and show the message: `not building target <...> because no targets are specified and library build output file is null`

Is this expected behaviour for the Android CI build?

In any case, getting back to **astc**, if Android is a normal cross-compilation target, I could simply force NEON on for Android builds just like iOS physical target builds, and then collapse the APPLE and non-APPLE x86 cmake code blocks into one.  However, if Android uses a self-hosted or virtual self-hosted dev environment then the answer may be different. Would you please let me know how Android is built: cross compilation or virtual self-hosted?  Thanks.

Note this question is only about possible code minimization, and this PR will work as-is without modification.

SaschaWillems:
I know close to noting about our build setup on Android, so sadly can't answer your question.

All I can say is that it builds for me locally and that samples that use ASTC do work as expected. So from that pov, this PR is fine.

Sadly samples on Android are so broken, that it's nigh impossible to test them right, see #1299, though that's not an issue fo this PR.

We plan on removing ASTC at some point though, as there are better options that won't take ages to transcode textures to native formats on desktop.

I guess this makes #1300 obsolete then?

SaschaWillems (dismissed):

SRSaunders:
Thanks @SaschaWillems for your answer.  I will leave things alone since it is working for you on Android, etc.

> I guess this makes https://github.com/KhronosGroup/Vulkan-Samples/pull/1300 obsolete then?

Yes, I suppose it does.  However, I would like @gpx1000 and @gary-sweet to review this PR before making that decision.  They indicated they may want to return to a stable base (astc 4.7) before any further changes.  However, if they agree with taking a single step to astc 4.8, then I would be fine with that too.

SRSaunders:
Had to rebase following merge of #1300 due to astc downgrade merge conflict.

gary-sweet (approved):
Doesn't seem to break anything for me, though I can't comment on the Mac stuff.

gpx1000 (approved):

marty-johnson59:
Merging with Steve's approval (per discussion on 24 March Samples call)

PR #1308, opened 2025-03-19
temp fix for  android issue 1299

### LINKED ISSUE #1299: Samples on Android totally unstable and prone to crashing
Not sure when this started to happen, but with current main, running a sample on Android twice from a launcher always makes the whole app crash. No matter what sample. I also noticed general instability, with the samples crashing left and right at seemingly random points. E.g. starting an extension sample that the device does not support, you sometimes get an error message or sometimes the whole app just crashes.

To reproduce this:

- Start the app on Android
- Run the "Hello Triangle" sample from the list
- Return to the sample list
- Run the "Hell Triangle" sample again

The sample now crashes in this assert:

```
	void set_platform(Platform *platform)
	{
		assert(!this->platform && platform);
		this->platform = platform;
	}
```

Tested on a Pixel 9 with Android 15.

gpx1000:
Android is a special case where the memory for the plugin doesn't go away; thus this assert will fail on any relaunch of any app from the menu.

Someone smarter than me should think about the implications of not enforcing this assert and lemme know if there's a better fix for us to use.  My tests indicate that this does solve the problem everywhere I care to test it.

## Description

Please read the [contribution guidelines](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc)

Fixes #1299 

## General Checklist:

- [ ] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Code-Style)
- [ ] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Copyright-Notice-and-License-Template)

## Sample Checklist

### DIFF
--- a/framework/platform/android/android_platform.cpp
+++ b/framework/platform/android/android_platform.cpp
@@ -344,6 +344,9 @@ AndroidPlatform::AndroidPlatform(const PlatformContext &context) :
 
 ExitCode AndroidPlatform::initialize(const std::vector<Plugin *> &plugins)
 {
+    for(auto plugin : plugins) {
+        plugin->clear_platform();
+    }
 	android_app_set_key_event_filter(app, key_event_filter);
 	android_app_set_motion_event_filter(app, motion_event_filter);
 
@@ -461,7 +464,8 @@ void AndroidPlatform::terminate(ExitCode code)
 		// Process events until app->destroyRequested is set
 	}
 
-	Platform::terminate(code);
+    plugins.clear();
+    Platform::terminate(code);
 }
 
 void AndroidPlatform::send_notification(const std::string &message)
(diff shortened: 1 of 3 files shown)


asuessenbach:
I think, fixing the issue this way is wrong.
The assert tells us, that this function is supposed to be used exactly once. Some kind of extended constructor function.
Where is it called a second time? And what part of the assertion is failing?

gpx1000:
Because we use the launcher menu in Android, the app is still alive, so when you hit the back button to exit out of a sample and get back to the menu, we've gone through the destructor.  However, there currently isn't a path which clears the pointer in this case.  This means that when you select another sample from the menu, this assert will fail.
All other platforms this assert is correct, but Android's use of a menu launcher is what trips us up here.

gpx1000:
This version is slightly better.  Now the assert is back and the bug is fixed for Android by clearing out the plugin variable on init.

gary-sweet (approved):
Can't test on Android, but works ok for me

SaschaWillems (approved):
Thanks 👍🏻 

This does properly fix the issue and switching between samples now works as expected.

PR #1322, opened 2025-03-28
fix wait fence indefinitely failed bug

JingyanChen:
## Description

In the Vulkan API, the timeout parameter of vkWaitForFences is of type uint64_t. If the API is expected to wait indefinitely (until the GPU signals the fence), the correct value to pass is UINT64_MAX (the maximum 64-bit unsigned integer), not the maximum 32-bit value (UINT32_MAX).

see vulkan spec

> // Provided by VK_VERSION_1_0
> VkResult vkWaitForFences(
>   VkDevice device,
>   uint32_t fenceCount,
>   const VkFence* pFences,
>   VkBool32 waitAll,
>   uint64_t timeout);

### DIFF
--- a/framework/fence_pool.cpp
+++ b/framework/fence_pool.cpp
@@ -66,7 +66,7 @@ VkFence FencePool::request_fence()
 	return fences.back();
 }
 
-VkResult FencePool::wait(uint32_t timeout) const
+VkResult FencePool::wait(uint64_t timeout) const
 {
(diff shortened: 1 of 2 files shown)


asuessenbach (approved):

gary-sweet:
You will, at the very least, need to update the Copyright header dates in order for this to progress I'm afraid.

PR #1337, opened 2025-04-21
fix: fix the surface resize logic when use wayland wsi

Kkaiwz:
According to the Vulkan Spec:https://registry.khronos.org/vulkan/ specs/latest/man/html/VkSurfaceCapabilitiesKHR.html

VkSurfaceCapabilitiesKHR->currentExtent can be set to 0xFFFFFFFF which indicating that surface size will be determined by the extent of swapchain targeting the surface rather than driver.

In mali umd, currentExtent are always set to 0xFFFFFFFF which is comply with spec. Since the surface extent size of wayland depends on the dynamic negotiation between the client and compositor, the Vulkan driver cannot determine the valid value of current Extent before vkQueuePresentKHR submission, so the driver alwayas return 0xFFFFFFFF when use wayland wsi.

Affected cases:
    oit_depth_peeling
    oit_linked_lists
    hpp_oit_depth_peeling
    hpp_oit_linked_lists

Signed-off-by: Ryan Zhang ryan.zhang@nxp.com

## Description

Please read the [contribution guidelines](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc)

Fixes #<issue number>

## General Checklist:


Kkaiwz:
I have signed the CLA already but the status is still pending.

SaschaWillems:
Not 100% sure, but it looks like the PR has commits from two maintainers (see https://github.com/KhronosGroup/Vulkan-Samples/pull/1337/commits/8b1db82d5a39a8f1af0b8e01757e948de3a18eae). Afaik both must ign the CLA.

Kkaiwz:
Thanks Sascha, I have both agree CLA in two account. Both of these accounts are mine. It's just that they were confused during git push, which lead to problems with CLA.

asuessenbach (commented):
Due to not having such a device, I can't verify that resizing is handled correctly with that change.
But in case it is correct, I would prefer to have one if-clause instead of two nested ones, like (for `hpp_api_vulkan_sample.cpp`):
```
	if ((surface_properties.currentExtent != vk::Extent2D{0xFFFFFFFF, 0xFFFFFFFF}) &&
	    (surface_properties.currentExtent != get_render_context().get_surface_extent()))
```

SaschaWillems:
Looks like something bad happened here. The PR now has > 1,000 changed files?

Kkaiwz:
> Looks like something bad happened here. The PR now has > 1,000 changed files?

I just modify 2 files. I don't know why result in this

SaschaWillems:
Well, you just force-pushed, which might be the reason for this mess-up.

So prob. best to do a new clean PR and close this one.

Kkaiwz:
This PR is broken. Close it and create new one.

PR #1339, opened 2025-04-23
Implement simple cache for decoded ASTC textures

### LINKED ISSUE #1202: (Meta issue) Improve sample load times
Doing a full batch run in debug mode can take up to 20 minutes. This is especially cumbersome for maintainers, as doing a batch run is now a requirement when doing changes to the framework. One culprit are assets that store textures in ASTC format. This is only supported on mobile and on desktop (where samples are actually developed) those samples need to transcode those textures. This can take a long time, and for a batch run this is done for each and every sample that uses those assets.

We should improve load times of the samples, esp. the ones that use complex scenes.

Related:
- #860

JoseEmilio-ARM:
## Description

Introduce a simple caching mechanism for decoded ASTC textures on platforms that do not support the format. Before decoding, the system will check if there are existing binaries saved locally, and otherwise create them for future runs.

Fixes #1202

## General Checklist:

### DIFF
--- a/framework/scene_graph/components/image.cpp
+++ b/framework/scene_graph/components/image.cpp
@@ -17,6 +17,7 @@
 
 #include "image.h"
 
+#include <cstddef>
 #include <mutex>
 
 #include "common/error.h"
@@ -174,7 +175,6 @@ void Image::clear_data()
 {
 	data.clear();
 	data.shrink_to_fit();
-	// preserve last known hash
 }
 
 VkFormat Image::get_format() const
@@ -321,10 +321,26 @@ std::vector<uint8_t> &Image::get_mut_data()
 
 void Image::update_hash()
 {
-	data_hash = 0;
-	for (size_t i = 0; i < data.size(); ++i)
+	static_assert(sizeof(data[0]) == 1);
+	constexpr size_t chunk_size = sizeof(size_t) / sizeof(data[0]);
+	data_hash                   = 0;
+	size_t i                    = 0;
+	if (data.size() >= chunk_size)
 	{
-		glm::detail::hash_combine(data_hash, data[i]);
+		for (; i <= (data.size() - chunk_size); i += chunk_size)
+		{
+			size_t it = 0;
+			std::memcpy(&it, &data[i], chunk_size);
+			glm::detail::hash_combine(data_hash, it);
+		}
+
+		// Handle elements smaller than sizeof(size_t) with zero padding
+		if (i < data.size())
+		{
+			size_t it = 0;
+			std::memcpy(&it, &data[i], data.size() - i);
+			glm::detail::hash_combine(data_hash, it);
+		}
 	}
 }
 
--- a/framework/scene_graph/components/image.cpp
+++ b/framework/scene_graph/components/image.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018-2024, Arm Limited and Contributors
+/* Copyright (c) 2018-2025, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
--- a/framework/scene_graph/components/image.cpp
+++ b/framework/scene_graph/components/image.cpp
@@ -325,19 +325,17 @@ void Image::update_hash()
 	constexpr size_t chunk_size = sizeof(size_t) / sizeof(data[0]);
 	data_hash                   = 0;
 	size_t offset               = 0;
-	if (data.size() >= chunk_size)
-	{
-		for (; offset + chunk_size < data.size(); offset += chunk_size)
-		{
-			glm::detail::hash_combine(data_hash, *reinterpret_cast<size_t const *>(&data[offset]));
-		}
 
-		if (offset < data.size())
-		{
-			size_t it = 0;
-			std::memcpy(&it, &data[offset], data.size() - offset);
-			glm::detail::hash_combine(data_hash, it);
-		}
+	for (; offset + chunk_size < data.size(); offset += chunk_size)
+	{
+		glm::detail::hash_combine(data_hash, *reinterpret_cast<size_t const *>(&data[offset]));
+	}
+
+	if (offset < data.size())
+	{
+		size_t it = 0;
+		std::memcpy(&it, &data[offset], data.size() - offset);
+		glm::detail::hash_combine(data_hash, it);
 	}
 }
 
--- a/framework/scene_graph/components/image.cpp
+++ b/framework/scene_graph/components/image.cpp
@@ -324,21 +324,18 @@ void Image::update_hash()
 	static_assert(sizeof(data[0]) == 1);
 	constexpr size_t chunk_size = sizeof(size_t) / sizeof(data[0]);
 	data_hash                   = 0;
-	size_t i                    = 0;
+	size_t offset               = 0;
 	if (data.size() >= chunk_size)
 	{
-		for (; i <= (data.size() - chunk_size); i += chunk_size)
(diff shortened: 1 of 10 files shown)


framework/scene_graph/components/image.cpp
```diff
@@ -164,10 +165,16 @@ const std::vector<uint8_t> &Image::get_data() const
...
+size_t Image::get_data_hash() const
+{
+	return data_hash;
+}
+
 void Image::clear_data()
 {
 	data.clear();
 	data.shrink_to_fit();
+	// preserve last known hash
```
asuessenbach: Why should the last known hash be preserved?

framework/scene_graph/components/image.cpp
```diff
@@ -312,10 +319,25 @@ std::vector<uint8_t> &Image::get_mut_data()
 	return data;
 }
 
+void Image::update_hash()
+{
+	data_hash = 0;
+	for (size_t i = 0; i < data.size(); ++i)
+	{
+		glm::detail::hash_combine(data_hash, data[i]);
```
asuessenbach: `glm::detail::hash_combine` works on hashes of type `size_t`. Maybe it's worth to loop by 4 bytes, instead of 1?

asuessenbach (commented):
First of all, to make your PR work with `vkb::scene_graph::components::HPPImage`, you would need to add the `data_hash` to that class as well.
Besides that, as the caching is an ASTC-specific thing, I would have expected that the hash lives in `vkb::sg::ASTC`.

With this PR, you calculate the hash for each and every image, and you calculate it based on the image data itself. Wouldn't it be more efficient, to just hash the URI of that file? That would probably require to just modify the `vkb::GLTFLoader`.

JoseEmilio-ARM:
> First of all, to make your PR work with `vkb::scene_graph::components::HPPImage`, you would need to add the `data_hash` to that class as well. Besides that, as the caching is an ASTC-specific thing, I would have expected that the hash lives in `vkb::sg::ASTC`.
> 
> With this PR, you calculate the hash for each and every image, and you calculate it based on the image data itself. Wouldn't it be more efficient, to just hash the URI of that file? That would probably require to just modify the `vkb::GLTFLoader`.

Thanks! Prefer to leave it as is, to support cases where the image itself changes, but not its URI.

framework/scene_graph/components/image.cpp
JoseEmilio-ARM: In case we load the data back from GPU, not a common use case though, removed the comment to avoid confusion

framework/scene_graph/components/image.cpp
JoseEmilio-ARM: Done

framework/scene_graph/components/image.cpp
```diff
@@ -312,10 +319,41 @@ std::vector<uint8_t> &Image::get_mut_data()
...
+		}
+
+		// Handle elements smaller than sizeof(size_t) with zero padding
+		if (i < data.size())
+		{
+			size_t it = 0;
+			std::memcpy(&it, &data[i], data.size() - i);
+			glm::detail::hash_combine(data_hash, it);
+		}
+	}
```
asuessenbach: There's no need to `memcpy` all the data into a `size_t` before hashing. And you're missing the unlikely case of `data.size() < chunk_size`.
Maybe something like that:
```
	size_t offset = 0;
	for (; offset + chunk_size < data.size(); offset += chunk_size)
	{
		glm::detail::hash_combine(data_hash, *reinterpret_cast<size_t const *>(&data[offset]));
	}
	if (offset < data.size())
	{
		size_t it = 0;
		std::memcpy(&it, &data[offset], data.size() - offset);
		glm::detail::hash_combine(data_hash, it);
	}
```

JoseEmilio-ARM: Perfect, updated, thank you!

SaschaWillems:
This PR breaks batch mode for me. When reacing the hlsl_shaders sample (which worked fine last time I did batch with master) it now asserts in create_vk_image. 

I'm also getting a lot of (silent?) exceptions when running samples that decode astc now:

image

Not sure if any of these is related, but can't approve this without being able to do a batch run.

JoseEmilio-ARM:
> This PR breaks batch mode for me. When reacing the hlsl_shaders sample (which worked fine last time I did batch with master) it now asserts in create_vk_image.

I can reproduce the issue, and found a workaround in my latest commit, but cannot really explain what is wrong, any ideas @asuessenbach ?
 
> I'm also getting a lot of (silent?) exceptions when running samples that decode astc now:

These are probably the ones I introduced when a problem occurs trying to load an image from the cache. The first time the cache does not exist, so they will be thrown, but in later runs they should not be there anymore

framework/scene_graph/components/image.h
```diff
@@ -146,6 +144,8 @@ class Image : public Component
 	std::unique_ptr<core::Image> vk_image;
 
 	std::unique_ptr<core::ImageView> vk_image_view;
+
+	size_t data_hash{0};
```
asuessenbach: Sorry, but this doesn't fix the hpp vs. non-hpp issue.
The data layout of those two classes need to be identical. That is, you need to add the data_hash at the same position to both, the Image and the HPPImage class.

JoseEmilio-ARM: Yes that was it, fixed, thank you!

asuessenbach:
Regarding all those exceptions: I don't think, it's the right approach to use exceptions for non-exceptional cases.
You could instead just use `std::filesystem::exists()` to check if the cached file exists.

And... having that `data_hash` in `HPPImage` as well, at the right position, made the hpp-based samples run correctly. But only because there's no hpp-based sample using an ASTC texture. If we get something like that, we'd need all those hash-related functions from `Image` there as well. So, would be great if you could add them.

SaschaWillems:
Any chance of speeding this up? I'm currently doing large changes to the framework, and testing is extremely tedious as all those high-level framework samples use ASTC and every time it needs to decode them. That makes testing very slow.

SaschaWillems (dismissed):
LGTM 👍🏻 

A minor suggestion: Add the astc cache folder to the .gitignore

framework/scene_graph/components/image/astc.cpp
```diff
@@ -164,18 +213,131 @@ Astc::Astc(const Image &image) :
...
+		{
+			return false;
+		}
+		try
+		{
+			if (!fs->exists(path))
+			{
+				LOGW("ASTC cache file {} does not exist. ASTC image will be decoded", path.string())
+				return false;
+			}
```
asuessenbach, line 239: Maybe add some
`LOGI("Loading ASTC cache data from file: {}", path.string());`

asuessenbach (commented):
Maybe you could delete the warning in `gltf_loader.cpp`, line 1489.

Besides that (and the comment below), it looks great now.

gary-sweet (dismissed):

JoseEmilio-ARM:
> LGTM 👍🏻
> 
> A minor suggestion: Add the astc cache folder to the .gitignore

Done, thanks!

framework/scene_graph/components/image/astc.cpp
JoseEmilio-ARM: Done, thanks!

asuessenbach (approved):

SaschaWillems (approved):

gary-sweet (approved):

marty-johnson59:
3 approvals, merging

PR #1354, opened 2025-05-23
Increase base version to Vulkan 1.1

### LINKED ISSUE #1320: Vulkan 1.0 samples fail on Android 15 when using RenderDoc
Before this is immediately closed because it only happens with RenderDoc's capture layer running - hear me out.

For Android 15 the Android Vulkan loader [was changed](https://cs.android.com/android/_/android/platform/frameworks/native/+/9d84eae592f9760b1985e628a59dc40a424717b4:vulkan/libvulkan/api_gen.cpp;dlc=6557491ca6b2f80cd3d8e0b820232b6067b6917d) so that `vkGetDeviceGroupPresentCapabilitiesKHR`, `vkGetDeviceGroupSurfacePresentModesKHR`, and `vkAcquireNextImage2KHR` become hard requirements (as in their addresses need to available).  This isn't a problem because from Android 10 it is hard requirement for Vulkan 1.1 support, this is important because those functions are a bit weird in that prior to 1.1 they were only available when _two_ extensions were requested: `VK_KHR_swapchain` or `VK_KHR_surface`, and `VK_KHR_device_group`.  After 1.1 only `VK_KHR_swapchain` was needed.

However, RenderDoc is a stickler for the rules, and the Vulkan Samples 'Hello Triangle' (and no doubt many others) set the Vulkan API version [to 1.0](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/samples/api/hello_triangle/hello_triangle.cpp#L222).  RenderDoc sees this and so does not make those functions' addresses available without `VK_KHR_device_group` being explicitly requested.  Of course `VK_KHR_device_group` is not requested by the sample because it doesn't need it, it's Android that needs it.

1. The sample calls `vkCreateDevice`
2. The Android Vulkan loader propagates the call to RD
3. RD creates the `VkDevice` successfully and then initialises the dispatch table for the device-level functions.  Explicitly excluding those functions Android needs because the sample didn't ask nicely enough
4. On return the loader then attempts to [initialise its own device-level dispatch table](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/native/vulkan/libvulkan/api.cpp;l=889;drc=3785475f912c360b93faffd111e89eed47715b80?q=vulkan::api::Creat), this fails because RD intercepts the `vkGetDeviceProcAddr` function and duly returns `NULL`
5. The loader freaks out and returns `VK_ERROR_INITIALIZATION_FAILED` to the sample, which triggers an exception

This isn't a problem when not ran under RD because the loader can get the func address straight from the driver, which at least on Mali, isn't checking which extensions have been enabled.

As from Android 10, [Vulkan 1.1 is an official requirement](https://developer.android.com/games/develop/vulkan/native-engine-support#recommendations), I'd argue that it is Vulkan Samples who are in the wrong here.  Ideally the Android SDK of the target device should be queried in the Java and then passed to native code, whereby the samples will set `VkApplicationInfo::apiVersion` to whatever the min Android version is required.  If that's a too onerous, you could probably get away with just using a compile-time define configured from Gradle's `targetSdk`, so anyone wanted to target older devices could still set it (albeit manually).

### LINKED ISSUE #1340: Update Vulkan version min. requirement to 1.1
Slang shaders (might) require a SPIR-V 1.4 enviroment. That requires at least Vulkan 1.1 + the SPIR-V (and adjacent) extension. So before starting to add slang shaders to the samples. we need to make sure we adhere to this requirement by e.g. increasing the base api version to 1.1 when slang is selected as a shading language.

SaschaWillems:
## Description

As discussed on our samples calls, this PR will increase the base version for all samples to 1.1.

Fixes #1340
Fixes #1320

## General Checklist:

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -52,6 +52,10 @@ endif::[]
 xref:samples/performance/README.adoc[Performance samples] show the recommended best practice together with real-time profiling information.
 They are more advanced but also contain a detailed tutorial with more in-detail explanations.
 
+=== Requirements
+
+The samples are written in C{pp} and require a compiler that supports at least C{pp}20. To run the samples, a device that supports at least Vulkan 1.1 or newer is required. Some samples might require a higher Vulkan version and/or support for certain extensions.
+
 === Goals
 
 * Create a collection of resources that demonstrate best-practice recommendations in Vulkan
--- a/framework/README.adoc
+++ b/framework/README.adoc
@@ -1,5 +1,5 @@
 ////
-- Copyright (c) 2023, Sascha Willems
+- Copyright (c) 2023-2025, Sascha Willems
 -
 - SPDX-License-Identifier: Apache-2.0
 -
@@ -75,7 +75,7 @@ There is no need to manually get extension pointer functions.
 
 === Changing the Vulkan api version
 
-By default all samples create a Vulkan 1.0 instance.
+By default all samples create a Vulkan 1.1 instance.
 Higher versions can be requested in the constructor of a sample:
 
 [,cpp]
--- a/framework/core/hpp_instance.h
+++ b/framework/core/hpp_instance.h
@@ -56,7 +56,7 @@ class HPPInstance
 	            const std::unordered_map<const char *, bool> &requested_extensions    = {},
 	            const std::unordered_map<const char *, bool> &requested_layers        = {},
 	            const std::vector<vk::LayerSettingEXT>       &required_layer_settings = {},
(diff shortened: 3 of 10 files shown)


gary-sweet (approved):

gpx1000 (approved):

asuessenbach (approved):

marty-johnson59:
3 approvals - merging

Issue #1201, opened 2024-10-21
(Meta issue) Simplify the framework


SaschaWillems:
A common criticism with our samples is the complicated framework. Esp. the samples that use the high level framework have a lot of abstraction, making it hard to debug, follow or learn from the code. While there is no direct action to take here we should look at ways to simplify the framework like removing multiple levels of abstraction and agreeing on ways to write code. Using modern C++ features might look nice, but if they result in code that's hard to follow we should not use in the samples.

SaschaWillems:
Looking through the framework makes me wonder why we did certain things the way we do. We have different design patterns, heavily templated code, lots of abstractions which in the end do nothing more than to fill out a Vk*CreateInfo structure. I wonder how we can simplify that in such a way that esp. the higher level framework becomes easy to follow  and debug again? I don't consider myself a C++ power user, but imo the code should be readable by someone with average C++ skills (like me). Yet we have whole header/implementation files that I don't understand at all (and sadly those files have little or often no code comments at all).

SRSaunders:
A side effect of the current project structure is a relatively long compile time.  A clean build on macOS takes about 5 minutes on my 8700K machine (6x2 cores using parallel compilation). This also results in longer CI times which has been a topic here before.  Can things be restructured to use precompiled headers or other componentization methods to reduce both clean and incremental build times?  Even if the frameworks continue to take time, speeding up sample compilation could really help.  In the current setup using my machine, each sample takes about 16-18 sec which quickly adds up even with parallel compilation enabled (12 threads). Single-threaded serialized compilation illustrates the issue more starkly, taking about 30 minutes on my machine with 9-10 seconds per sample.

SaschaWillems:
Indeed, it's something we need to address. We did some improvements, but still compilation times are too long compared to other similar projects. We'll probably move to C++20 (for other reasons) and then start using modules which should reduce build times, esp. for anything using Vulkan.hpp.

Windows is a bit better, but building this repo still takes almost 5 minutes in my setup, compared to building my own samples, which only takes ~40s.

SaschaWillems:
We also do a lot of abstractions that are IMO totally unnecessary. Even for simple, non-Vulkan related things. This is causing us issues that are night impossible to debug.

marty-johnson59:
Closing per discussion on 24 August meeting..

PR #1359, opened 2025-06-05
Replace runtime shader compilation with offline compiled SPIR-V shaders

### LINKED ISSUE #1107: [Shading language support] Remove shader variants for high-level framework samples
The high-level framework currently uses a few shader variants and along with that reflection to map shader inputs to the C++ side of things. That is contrary to the api framework samples, where every sample uses it's own set of shaders and no runtime reflection-

This prevents us from adding support for offline compiled shaders and HLSL shader support to the high-level framework samples-

We should completely remove shader variants and replace them with explicit shaders per sample instead.

Examples are most of the shaders in the shader root directory, e.g.:

* base.frag
* base.vert
* lighting.h
* pbr.frag
* pbr.vert

These should be moved to explicit per-sample folders and adjusted per sample so that we no longer need variants and runtime reflection.

### LINKED ISSUE #1108: [Shading language support] Add support for loading SPIR-V shaders for high-level framework samples
With #990 we added support for offline compilation of HLSL shaders via CMake. So any HLSL shader added to a sample's CMakeLists.txt will be compiled to SPIR-V when running CMake.

With #961 we added support for selecting between GLSL (compiled at runtime) and HLSL (compiled offline, loading SPIR-V) to the api based samples. A new command line argument was introduced to select what shader type to load. Default is GLSL, but with `--shading-language HLSL`  one can load HLSL shaders instead if they're present for a sample.

For this to work in a portable way that requires little modification to a sample both the `ApiVulkanSample` as well as the `HPPApiVulkanSample` got a new overload for loading shaders. Instead of providing a direct shader file path, path and shader name are separate arguments, and the function then loads either the glsl file (and compiles it at runtime) or loads the offline compiled SPIR-V file from an HLSL shader.

Before:

```cpp
shader_stages[0] = load_shader("dynamic_uniform_buffers\glsl\base.vert", VK_SHADER_STAGE_VERTEX_BIT);
```

After:

```cpp
shader_stages[0] = load_shader("dynamic_uniform_buffers", "base.vert", VK_SHADER_STAGE_VERTEX_BIT);
```

The latter will select the folder to load the shader from based on the currently selected shading language (glsl or HLSL). If HLSL is selected, shaders are loaded from offline compiled SPIR-V.

The ARM framework samples should be adjusted accordingly and should check for the value of the `--shading-language` argument and depending on that should select either GLSL (and compile at runtime) or HLSL (and load SPIR-V instead). Hopefully code can be leveraged from the api sample functions, see [ApiVulkanSample::load_shader](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/framework/api_vulkan_sample.cpp#L440).

This probably requires a change in the `ShaderSource` class.

### LINKED ISSUE #1110: [Shading language support] Replace runtime GLSL shader compilation with offline compilation
Ultimately after implementing #1107, #1108 and #1109 runtime GLSL shader compilations should be replaced with offline compilation in the same way we do with HLSL.

So instead of compiling shaders at runtime, GLSL shaders are compiled with CMake. We should also add the compilid SPIR-V files to the project (like we do with HLSL), so people wanting to try our samples can do so without the need of having a glsl compiler installed.

That way we can simplify some code and switching between GLSL and HLSL (or other shading languages to be added) would be no more than simply changing folders.

SaschaWillems:
## Description

This PR will replace runtime GLSL compilation with loading offline compiled SPIR-V shaders instead. That's a requirement for getting full support for Slang and HLSL into our samples.

**This is a pretty big PR including lots of cleanup, fixes and simplification of the framework.**
**I would appreciate if we could merge this timely, or at least don't merge any other PR that affects the framework.**

What this PR does (among other minor things):

* Removed remaining shader variants
  * All related framework functionality has been removed
* Completely removed runtime compilation
  * This also allows us to remove glsl as a (kinda heavy) dependency
* Removed real-time shader selection plugin (wasn't used anywhere)
* Removed the HLSL shading standalone example
  * This is no longer useful with HLSL support being part of the framework 
* Samples exclusively load shaders from offline compiled SPIR-V 
* GLSL shader compilation via CMake
  * Including ways of adding additional glsl compiler options (just like with DXC), e.g. required for samples that use higher SPIR-V target versions (ray tracing, mesh shading)
* The offline shader compilation projects (GLSL, HLSL and Slang) have been restructured to things less convoluted. It's one project per sample, instead of one project per shader file (which caused issues on Windows due to path size limiations)
  * image
  * Fixes a lot of the samples missing used shaders in the CMake-File (which wouldn't be compiled)
* Introduces a function to load binary shaders as a vector of uint32_t to replace the old uint8_t variant that requires unncessary conversion

As for testing, I did multiple batch runs on Windows 11(nvidia RTX 4070) and also did test several samples on Android 15 (Google Pixel 9).

Fixes #1107
Fixes #1108 
Fixes #1110
Ref #1201

## General Checklist:


SaschaWillems:
Working on that part of the framework is kinda pretty rough. Lots of nasty dependencies, long build and loading times (astc conversion) are making changes hard and tedious 😡

SaschaWillems:
Another odd thing I noticed with the framework: SPIR-V uses 32 bit unsigned integers, but loading SPIR-V files in the frameworks uses 8 bit unsigned integers. That then always requires some odd in-memory conversions.

Not sure why this is done in the framework, but I'll add a more native version with this PR to make things easier to understand.

SaschaWillems:
The shaders in the high-level framework could use some cleanup. There are things in those shaders that don't make sense at all, but it's kinda hard to judge if it's okay to just remove. So for now I'll got with "it compiles, it's fine".

SaschaWillems:
Another annoyance is how much duplicated code we have. We often have a C and a C++ (Vulkan.hpp) based version of a framework file and sample. That means having to do the same thing twice.

Wish we could decide on how to unifiy all of that and then actually implement that. Would make the samples much easier to mantain and also easier to understand for people wanting to learn from them.

SaschaWillems:
The "Constant data" sample is a very odd one. While removing variants from it I noticed that all paths, even those not using push contants, rely on the push constant size limit. That feels wrong, but was already the case before this PR. Might be good to rework that sample in a separate PR.

components/filesystem/src/legacy.cpp
```diff
@@ -109,14 +109,16 @@ std::vector<uint8_t> read_asset(const std::string &filename)
...
+std::string read_text_file(const std::string &filename)
 {
 	return vkb::filesystem::get()->read_file_string(path::get(path::Type::Shaders) + filename);
 }
 
-std::vector<uint8_t> read_shader_binary(const std::string &filename)
+std::vector<uint32_t> read_shader_binary_u32(const std::string &filename)
 {
-	return vkb::filesystem::get()->read_file_binary(path::get(path::Type::Shaders) + filename);
+	auto buffer = vkb::filesystem::get()->read_file_binary(path::get(path::Type::Shaders) + filename);
```
asuessenbach, line 119: Maybe add an assertion that the file actually exists before reading it.

components/filesystem/src/legacy.cpp
```diff
@@ -109,14 +109,16 @@ std::vector<uint8_t> read_asset(const std::string &filename)
...
+	auto spirv  = std::vector<uint32_t>(reinterpret_cast<uint32_t *>(buffer.data()), reinterpret_cast<uint32_t *>(buffer.data()) + buffer.size() / sizeof(uint32_t));
```
asuessenbach: Maybe add an assertion that `buffer.size()` is a multiple of `sizeof(uint32_t)`.

framework/platform/application.h
```diff
@@ -124,14 +113,12 @@ class Application
 	 */
 	static vkb::ShadingLanguage get_shading_language();
 
-  protected:
 	/**
-	 * @brief Stores a list of shaders for the active sample, used by plugins to dynamically change the shader
-	 * @param shader_language The shader language for which the shader list will be provided
-	 * @param list_of_shaders The shader list, where paths and shader types are provided
+	 * @brief Returns the defaultfolder for the currentl set shading language
```
asuessenbach: Typo: currentl -> currently

samples/CMakeLists.txt
```diff
@@ -124,7 +123,6 @@ set(ORDER_LIST
 	"hpp_dynamic_uniform_buffers"
```
asuessenbach: Maybe, while you're touching this file, unify the indenting?

samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp
```diff
@@ -543,39 +542,16 @@ vk::RenderPass HPPHelloTriangle::create_render_pass()
  * @param path The path for the shader (relative to the assets directory).
  * @returns A vk::ShaderModule handle. Aborts execution if shader creation fails.
  */
-vk::ShaderModule HPPHelloTriangle::create_shader_module(const char *path)
+vk::ShaderModule HPPHelloTriangle::create_shader_module(const char *path, vk::ShaderStageFlagBits stage)
```
asuessenbach: The added argument `stage` isn't used in this function.

samples/extensions/debug_utils/debug_utils.cpp
```diff
@@ -192,31 +192,15 @@ void DebugUtils::set_object_name(VkObjectType object_type, uint64_t object_handl
...
-	}
-
-	std::string shader_file_name = "debug_utils/" + shader_folder + "/" + file;
+	std::string shader_file_name = "debug_utils/" + get_shader_folder() + "/" + file;
 
 	VkPipelineShaderStageCreateInfo shader_stage = {};
 	shader_stage.sType                           = VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_CREATE_INFO;
 	shader_stage.stage                           = stage;
-	shader_stage.module                          = vkb::load_shader((shader_file_name + shader_extension).c_str(), get_device().get_handle(), stage, src_language);
+	shader_stage.module                          = vkb::load_shader((shader_file_name).c_str(), get_device().get_handle(), stage);
```
asuessenbach: `vkb::load_shader` takes a `std::string const &` as its first argument. So no need to get the `c_str()` here.

samples/extensions/ray_tracing_reflection/CMakeLists.txt
```diff
@@ -33,5 +30,8 @@ add_sample_with_tags(
         "ray_tracing_reflection/raygen.rgen"
         "ray_tracing_reflection/miss.rmiss"
         "ray_tracing_reflection/missShadow.rmiss"
-        "ray_tracing_reflection/closesthit.rchit")
+        "ray_tracing_reflection/closesthit.rchit"
+    GLSLC_ADDITIONAL_ARGUMENTS
+        "--target-spv=spv1.4"
```
asuessenbach: Doesn't that need to be `spv1.5` here?

shaders/async_compute/lighting.h
asuessenbach, line 1: This file (`shaders/async_compute/lighting.h`) and `shader/16bit_storage_input_output/lighting.h` are identical. Maybe use just one of them, at some common place?

shaders/constant_data/lighting.h
asuessenbach, line 1: That's the third identical `lighting.h`. A unique version of it at a common position would be better.

shaders/shader_object/material_gen_normals.geom
```diff
@@ -1,6 +1,6 @@
 #version 460
 /*
- * Copyright 2023 Nintendo
+ * Copyright 2023-2025 Nintendo
```
asuessenbach, line 3: Just the copyright data changed?

shaders/shader_object/material_pass_sin_offset.geom
```diff
@@ -1,6 +1,6 @@
 #version 460
 /*
- * Copyright 2023 Nintendo
+ * Copyright 2023-2025 Nintendo
```
asuessenbach, line 3: Again, just the copyright date changed?

shaders/shader_object/material_pass_through.geom
```diff
@@ -1,6 +1,6 @@
 #version 460
 /*
- * Copyright 2023 Nintendo
+ * Copyright 2023-2025 Nintendo
```
asuessenbach, line 3: Copyright date again?

samples/api/hpp_compute_nbody/hpp_compute_nbody.cpp
asuessenbach, line 1: Line 448: Missing ".spv"

samples/api/hpp_oit_depth_peeling/hpp_oit_depth_peeling.cpp
asuessenbach, line 1: Line 417: missing ".spv"

asuessenbach (commented):
What a tiny little change!
And just a couple of mostly minor issues.

Besides that, it would be great, if the shader files would be listed in the VS solution. Currently, I see the `Shaders` folder, containing `glsl` and `hlsl` subfolders, but they are empty:
image

SaschaWillems:
Thank you very much for the fast review👍🏻 Glad it's mostly minor things, gotta admit working on that part of the framework wasn't easy 😉

SaschaWillems:
> What a tiny little change! And just a couple of mostly minor issues.
> 
> Besides that, it would be great, if the shader files would be listed in the VS solution. Currently, I see the `Shaders` folder, containing `glsl` and `hlsl` subfolders, but they are empty: image

Actually they should be, but somehow sometimes VS refuses to load them for some of the projects. My guess is that this happens if a shader is used in multiple projects and then it only shows up for one of them, Haven't found a solution for that yet (if there is one after all).

Afaik this has always been a problem and is not related to this PR.

samples/CMakeLists.txt
SaschaWillems: Where exactly? Indenting looks fine for me in that file. At least in Visual Studio.

samples/extensions/ray_tracing_reflection/CMakeLists.txt
SaschaWillems: Good catch. That sample actual does not run, but it looks like the batch run still continues for errors that are gracefully handled.

shaders/async_compute/lighting.h
SaschaWillems: That's a good point, but maybe something we could/should do at a later point. Actually I think we should completely get rid of that include.

But for this PR I went with the route that Tom started when doing the initial PRs for offline compilation.

components/filesystem/src/legacy.cpp
SaschaWillems: Tbh. I'm not sure how to do that in a way that works accross all supported platforms. Working on these kind of framework units is always tricky and extremly time consuming. Might take a look, but can't promise anything.

samples/CMakeLists.txt
asuessenbach: Ah, I see...
Most of the entries in this file are indented by spaces, and some by tab. When your tab size is different from 4 you see the difference.

components/filesystem/src/legacy.cpp
asuessenbach: I think, `std::filesystem::exists` should work on every supported platform?

shaders/async_compute/lighting.h
SaschaWillems: Looks like I found a solution that actually isn't too complicated. Will remove the duplicated files and replace with a single include :)

samples/CMakeLists.txt
SaschaWillems: Done. The file now uses 4 spaces in all places. I think that's what we generally use.

components/filesystem/src/legacy.cpp
SaschaWillems: That's a good idea. Added.

components/filesystem/src/legacy.cpp
asuessenbach: Have you tried this?
```
	auto path = path::get(path::Type::Shaders) + filename;
	assert(std::filesystem::exists(path) && "Shader file does not exist");
	auto buffer = vkb::filesystem::get()->read_file_binary(path);
```
Or this?
```
	assert(is_file(path) && "Shader file does not exist");
```

SaschaWillems: github seemed to have issues, so my comment was lost :/

I think it's not required in here. The framework has it's own filesystem unit used to load all files, and that one already throws an error if a file is not present.

shaders/README.adoc
```diff
@@ -22,6 +22,10 @@
 
 This folder contains the textual shaders for the samples. All samples come with GLSL shaders and some optionally with HLSL shaders. For samples that support both shader language this is a good way to compare GLSL to HLSL syntax when targeting SPIR-V for Vulkan.
 
+== Compling shaders
```
gary-sweet: typo: Compiling

gary-sweet (commented):
Runs fine for me, just spotted one small typo

shaders/README.adoc
SaschaWillems: Well, that's an easy fix :)

Thank you very much for your review 👍🏻

asuessenbach (dismissed):

gpx1000 (dismissed):
Thanks for the amazing effort on this PR.  Tested in Linux, Windows, Android, and iOS.  Works here.

gary-sweet (dismissed):

asuessenbach (approved):

gary-sweet (approved):

PR #1366, opened 2025-06-25
Improve and fix async compute sample

### LINKED ISSUE #863: Add sample utilizing VK_EXT_layer_settings
As an example it is now possible to set layer settings via a programmatic interface.

See [Configuring the layer settings using VK_EXT_layer_settings](https://github.com/KhronosGroup/Vulkan-Utility-Libraries/blob/main/docs/layer_configuration.md#configuring-the-layer-settings-using-vk_ext_layer_settings) for documentation on how to use the new official extension.

I think a good sample would be showing how to enable vendor specific best practices via VK_EXT_layer_settings.

### LINKED ISSUE #1331: Why isn't queue family ownership transferred for blur_chain[1] vkimage in the async compute sample?
In my opinion, in that sample, the only resources that require queue family ownership transfer are color_targets[1 and 2] and blur_chain[1]. The color_targets[] are properly handled with queue family ownership transfer logic in place, so there are no issues. However, for blur_chain[1], there is no ownership transfer explicitly defined anywhere in the code. Yet, the program seems to run without any warnings or errors.

What could be the reason for this?

To clarify, blur_chain[1] is a VkImage created with VK_SHARING_MODE_EXCLUSIVE, and it is written to by a dispatch command recorded in the compute queue, then read by a draw command recorded in the graphics queue. Proper semaphore synchronization is in place between these two commands.

iagoCL:
## Description

This PR adds missing barriers to the async_compute sample as highlighted by #1331. We have also added comments to the code to better explain how the sample works. A few validation synchronization errors have also been fixed.

When looking at this sample I found some problems in the framework that I also fix in this PR:

- Barriers were actively ignoring `old_queue_family ` and `new_queue_family`. The async compute sample is the only place using them so I re-enable them and I stop ignoring them. I also renamed them to `src_queue_family` and `dst_queue_family` to be closer to the vulkan spec.
- `VKB_VALIDATION_LAYERS_GPU_ASSISTED`, `VKB_VALIDATION_LAYERS_BEST_PRACTICES `and `VKB_VALIDATION_LAYERS_SYNCHRONIZATION `were being ignored. I use this chance to update how we control the validation layers, replacing `VK_EXT_validation_features `with `VK_EXT_layer_settings`. This should close #863
- I noticed that some gltfs have images without a name, this makes more difficult to debug the samples using render doc, so images without a name will use the URI instead.

Tested on Android using a Mali G715 and on Windows using a Nvidia RTX 4070 Super

Fixes #1331 #863 

## General Checklist:

## Sample Checklist

- [NA] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements)
- [NA] Any dependent assets have been merged and published in downstream modules
- [NA] For new samples, I have added a paragraph with a summary to the appropriate chapter in the readme of the folder that the sample belongs to [e.g. api samples readme](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/samples/api/README.adoc)
- [NA] For new samples, I have added a tutorial README.md file to guide users through what they need to know to implement code using this feature. For example, see [conditional_rendering](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/extensions/conditional_rendering)
- [NA] For new samples, I have added a link to the [Antora navigation](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/antora/modules/ROOT/nav.adoc) so that the sample will be listed at the Vulkan documentation site

### DIFF
--- a/framework/core/hpp_instance.cpp
+++ b/framework/core/hpp_instance.cpp
@@ -139,6 +139,36 @@ bool enable_extension(const char                                 *requested_exte
 	return is_available;
 }
 
+bool enable_layer_setting(const vk::LayerSettingEXT        &requested_layer_setting,
+                          const std::vector<const char *>  &enabled_layers,
+                          std::vector<vk::LayerSettingEXT> &enabled_layer_settings)
+{
+	// We are checking if the layer is available.
+	// Vulkan does not provide a reflection API for layer settings. Layer settings are described in each layer JSON manifest.
+	bool is_available =
+	    std::ranges::any_of(enabled_layers,
+	                        [&requested_layer_setting](auto const &available_layer) { return strcmp(available_layer, requested_layer_setting.pLayerName) == 0; });
+	if (!is_available)
+	{
+		LOGW("Layer: {} not found. Disabling layer setting: {}", requested_layer_setting.pLayerName, requested_layer_setting.pSettingName);
+		return false;
+	}
+
+	bool is_already_enabled =
+	    std::ranges::any_of(enabled_layer_settings,
+	                        [&requested_layer_setting](VkLayerSettingEXT const &enabled_layer_setting) { return (strcmp(requested_layer_setting.pLayerName, enabled_layer_setting.pLayerName) == 0) && (strcmp(requested_layer_setting.pSettingName, enabled_layer_setting.pSettingName) == 0); });
+
+	if (is_already_enabled)
+	{
+		LOGW("Ignoring duplicated layer setting {} in layer {}.", requested_layer_setting.pSettingName, requested_layer_setting.pLayerName);
+		return false;
+	}
+
+	LOGI("Enabling layer setting {} in layer {}.", requested_layer_setting.pSettingName, requested_layer_setting.pLayerName);
+	enabled_layer_settings.push_back(requested_layer_setting);
+	return true;
+}
+
 bool enable_layer(const char                             *requested_layer,
                   const std::vector<vk::LayerProperties> &available_layers,
                   std::vector<const char *>              &enabled_layers)
@@ -169,7 +199,7 @@ bool enable_layer(const char                             *requested_layer,
 HPPInstance::HPPInstance(const std::string                            &application_name,
                          const std::unordered_map<const char *, bool> &requested_extensions,
                          const std::unordered_map<const char *, bool> &requested_layers,
-                         const std::vector<vk::LayerSettingEXT>       &required_layer_settings,
+                         const std::vector<vk::LayerSettingEXT>       &requested_layer_settings,
                          uint32_t                                      api_version)
 {
 	std::vector<vk::ExtensionProperties> available_instance_extensions = vk::enumerateInstanceExtensionProperties();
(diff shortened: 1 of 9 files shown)


framework/core/hpp_instance.cpp
```diff
@@ -194,12 +194,17 @@ HPPInstance::HPPInstance(const std::string                            &applicati
 	bool portability_enumeration_available = enable_extension(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME, available_instance_extensions, enabled_extensions);
 #endif
 
+#ifdef USE_VALIDATION_LAYERS
+	const char *validation_layer_name = "VK_LAYER_KHRONOS_validation";
+#endif
```
asuessenbach: Maybe we should expand the `#ifdef USE_VALIDATION_LAYERS`-section around the following `#ifdef USE_VALIDATION_LAYER_FEATURES`-section.

framework/core/hpp_instance.cpp
```diff
@@ -194,12 +194,17 @@ HPPInstance::HPPInstance(const std::string                            &applicati
...
 #ifdef USE_VALIDATION_LAYER_FEATURES
 	bool validation_features = false;
 	{
-		std::vector<vk::ExtensionProperties> available_layer_instance_extensions = vk::enumerateInstanceExtensionProperties(std::string("VK_LAYER_KHRONOS_validation"));
-
-		enable_extension(VK_EXT_VALIDATION_FEATURES_EXTENSION_NAME, available_layer_instance_extensions, enabled_extensions);
+		std::vector<vk::ExtensionProperties> available_layer_instance_extensions = vk::enumerateInstanceExtensionProperties(std::string(validation_layer_name));
+		if (!available_layer_instance_extensions.empty())
```
asuessenbach: You don't need to check for emptiness here.
If there are no instance extensions available, `enable_extension` will return `false`, which is the correct behaviour.

framework/core/hpp_instance.cpp
```diff
@@ -268,6 +273,23 @@ HPPInstance::HPPInstance(const std::string                            &applicati
...
+		bool is_available =
+		    std::ranges::any_of(enabled_layers,
+		                        [&layer_setting](auto const &available_layer) { return strcmp(available_layer, layer_setting.pLayerName) == 0; });
+		if (is_available)
+		{
+			enabled_layer_settings.push_back(layer_setting);
+		}
+		else
+		{
+			LOGE("Ignoring LayerSetting {}: Layer: {} not found.", layer_setting.pSettingName, layer_setting.pLayerName);
```
asuessenbach: I don't get the layer settings handling here.
They are either required or (implicitly optionally) requested.
If they are required, you don't need to copy the available ones into a new vector. They either are all available or not. The error message should then be something like "Required LayerSetting {} not available, cannot run", or so.
If they are just requested, we should rename the argument accordingly, and change the `LOGE` here into a `LOGW`.

framework/core/instance.cpp
asuessenbach, line 1: The same comments as on hpp_instance.cpp hold here, of course.

asuessenbach (commented):
Great improvement, and very helpful comments in `async_compute.cpp`!

Just a few comments in the framework part.

framework/core/hpp_instance.cpp
iagoCL: You are right, changed it to make them optional.

gary-sweet (dismissed):
I am seeing some warning with this sample:
```
[warning] Render Pass creation: size of attachment list and load/store info list does not match: 1 vs 2
[warning] Render Pass creation: size of attachment list and load/store info list does not match: 1 vs 2
```
It looks like those existed before your change though.

iagoCL:
> I am seeing some warning with this sample:
> 
> ```
> [warning] Render Pass creation: size of attachment list and load/store info list does not match: 1 vs 2
> [warning] Render Pass creation: size of attachment list and load/store info list does not match: 1 vs 2
> ```
> 
> It looks like those existed before your change though.

The warning was present before my changes, but I’ve pushed a fix for it.
We could have ignored it, as the default load_store_op behavior was correct. The warning was triggered because the framework default configuration expects two attachments. In this case, the sample only used one attachment. The unused load_store_op was ignored.

asuessenbach (approved):
LGTM.

gary-sweet (approved):
Thanks

SaschaWillems (approved):

marty-johnson59:
3 approvals - merging

PR #1396, opened 2025-08-15
Don't reopen redirected standard handles on Windows

AnyOldName3:
## Description

If the calling process has redirected a standard handle (e.g. `vulkan_samples.exe sample msaa > out.log`), then reopening it will detach it from the file/pipe and connect it to the console instead. We don't want to do that.

Instead, the type of the file should be tested before attaching the parent console. If it's already a disk file or a pipe, then we want to leave it alone.

Note that for PowerShell, this doesn't make `> file` work, but `| Out-File file` does, and `>` works in other shells like Command Prompt and MSYS2 Bash. IIRC from past investigation when I've made similar changes to other projects, it's because PowerShell closes the file handle for `/SUBSYSTEM:WINDOWS` applications immediately.

## General Checklist:

- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Copyright-Notice-and-License-Template) *Note: the script wanted to extend the year range of the first line, which I'm not sure is right, as the person it specifies didn't make any changes to this file in 2025.*
- [ ] I have tested every sample to ensure everything runs correctly *I didn't bother with this as it should have the same effect on all samples and doesn't affect Vulkan itself*

 - [ ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements) *It builds on Windows and is a Windows-specific file, so wouldn't be expected to build elsewhere*

## Sample Checklist

N/A

### DIFF
--- a/components/windows/src/context.cpp
+++ b/components/windows/src/context.cpp
@@ -1,4 +1,5 @@
-/* Copyright (c) 2023, Thomas Atkinson
+/* Copyright (c) 2023-2025, Thomas Atkinson
+ * Copyright (c) 2025, Chris Djali
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -87,13 +88,12 @@ WindowsPlatformContext::WindowsPlatformContext(HINSTANCE hInstance, HINSTANCE hP
 	_temp_directory             = get_temp_path_from_environment();
 	_arguments                  = get_args();
 
-	auto isRedirected = [](DWORD stdHandle)
-	{
+	auto isRedirected = [](DWORD stdHandle) {
 		DWORD fileType = GetFileType(GetStdHandle(stdHandle));
 		return fileType == FILE_TYPE_DISK || fileType == FILE_TYPE_PIPE;
 	};
 
-	bool inRedirected = isRedirected(STD_INPUT_HANDLE);
+	bool inRedirected  = isRedirected(STD_INPUT_HANDLE);
 	bool outRedirected = isRedirected(STD_OUTPUT_HANDLE);
 	bool errRedirected = isRedirected(STD_ERROR_HANDLE);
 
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Did try this on windows, but neither Powershell/Terminal nor CMD work as expected. On CMD it breaks things as it seems to be stuck in output, even when returning to the CMD prompt. Powershell with Out-File is kinda works, but leaves you with an empty prompt where you have to press a key for it to exit and actually write to the file.

IMO we need to completely rework this part of our framework. It never worked well, even when not outputting to a file.

AnyOldName3:
Because it's a `/SUBSYSTEM:WINDOWS` application, shells emit the prompt for the next command immediately, before the application's had a chance to print anything. You should still be able to type the next command and press enter to run it while the Vulkan sample is still running or once it's finished. It's not great, but that's just how Windows works unless you pass the `/SUBSYSTEM:CONSOLE` flag to the linker, in which case it opens a console when the application starts up, with no way to opt out.

If I had to guess, maybe printing an end-of-file character when the application exits would resolve the problem with PowerShell, although I don't remember that being something that happened in my testing, so it could be a difference between Windows PowerShell and PowerShell Core.

marty-johnson59:
Hi @AnyOldName3, the team decided not to take this change at this time.  We may rework this part of the framework in the future to clean this up.  LMK if you strongly disagree.  Thanks

AnyOldName3:
If the bug isn't being fixed in the immediate future (and this, something very similar, or switching the samples to `/SUBSYSTEM:CONSOLE`, are the only ways to fix it), it at least should be added to the issue tracker as a bug - I didn't open a report as it was easier to fix than explain.

SaschaWillems:
See #1459

PR #1426, opened 2025-10-27
Set default queue priorities to 0.5f

SaschaWillems:
## Description

As discussed on the last sample call, always setting queue priorities to 1.0f (max.) is not a good practice. This PR adjusts all places where queue priorities are set to use 0.5f instead. Won't make a difference anyway, as we rarely (if at all) use multiple queues from the same family.

Refs https://github.com/KhronosGroup/Vulkan-Tutorial/pull/166

## General Checklist:

### DIFF
--- a/framework/core/device.h
+++ b/framework/core/device.h
@@ -634,7 +634,7 @@ inline void Device<bindingType>::init(std::unordered_map<const char *, bool> con
 		if (gpu.has_high_priority_graphics_queue() &&
 		    (vkb::common::get_queue_family_index(queue_family_properties, vk::QueueFlagBits::eGraphics) == queue_family_index))
 		{
-			queue_priorities.back()[0] = 0.0f;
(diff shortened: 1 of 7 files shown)


asuessenbach:
I've probably missed something, but why is a queuePriority of 1.0f not a good practice?
As we never have more than one queue from the same family, the actual value doesn't make any difference. Why not keeping it at 1.0f?

SaschaWillems:
We discussed this on one of our recent calls and agreed to change/make it consistent across all samples (and the tutorial). This is similar to other computing areas where telling people to set the highest priority for a resource is usually considered bad practice. People will copy that and at some point might wonder why things don't work out.

gary-sweet (dismissed):
I might have gone with 0.5, but 0.0 shouldn't cause any problems.

gary-sweet (approved):

PR #1435, opened 2025-11-01
Apple compatibility fixes for driver optionality (MoltenVK & KosmicKrisp)

### LINKED ISSUE #1434: Expected Incompatibilities with KosmicKrisp driver for macOS
The Vulkan-Samples project currently assumes that MoltenVK will be the Vulkan driver for macOS and iOS.   Going forward there will be two options available in the Vulkan SDK: MoltenVK (portability driver on macOS arm64/x86_64 and iOS arm64) and KosmicKrisp (conformant macOS driver on arm64 and modern OS/Metal versions only).

In anticipation of a coming SDK release containing both of these drivers, it is prudent to take a look at what needs to be updated in the Vulkan-Samples project to ensure compatibility with both.

I plan to propose a WIP PR containing anticipated changes and will seek feedback from interested parties (e.g. @aitor-lunarg and others).  Once we have validation of the necessary changes we could move the PR from WIP to ready for review.

Fortuntately the framework of the current project was already setup to provide this flexibility.  I do not anticipate major challenges with this, but mainly assumption changes and adjustments using currently available settings.

SRSaunders:
## Description

This WIP PR removes the Apple = MoltenVK driver assumption and provides compatibility for the coming KosmicKrisp driver on supported platforms (macOS, arm64). In addition, it provides a simple mechanism to allow the user to build for either situation. The already-existant CMake variable `VKB_ENABLE_PORTABILITY` is now externalized as a CMake command line option ~~for users to select building for MoltenVK vs. KosmicKrisp~~.  The option is set in this PR to default = ON for backwards compatibility.  To build for KosmicKrisp the user will likely have to execute a specific **setup-env.sh** from the SDK (or set the `VK_DRIVER_FILES` environment variable) and then run CMake using the macOS instructions ~~adding the option `-DVKB_ENABLE_PORTABILITY=OFF`~~.

Specifics are:

1. Change `VKB_ENABLE_PORTABILITY` from being an internally set CMake variable to a cmdline option (default ON).
2. Guard a small number of Apple-specific layer settings-related code blocks used for configuring MoltenVK.
3. Remove unneeded Apple-specific #ifdef legacy code blocks from the _descriptor_indexing_ sample.
4. Update the _profiles_ sample to run with VKB_ENABLE_PORTABILITY set to ON or OFF on Apple platforms.  The only issue here is that I had to remove a hard-coded `#ifdef __APPLE__` code block from **vulkan_profiles.hpp** to provide this flexibility.  I realize this is a generated file, but the .hpp file makes a bad assumption that Apple = PORTABILITY and turns on the portability extensions by itself.  This is not correct going forward and the sample itself should control that, not the hpp file.  This really should be changed in the generator script, but I do not have access to that within this project.
5. Add a few missing null object checks in the _[hpp\_]hello_triangle_ samples so they do not crash on exit if no valid driver is found.  I discovered this during regression testing.
6. [_new_] Update comments and messages for iOS config and build framework.  These changes retain the assumption that MoltenVK is used for iOS, but the framework now exists for adding alternative iOS drivers.  Comments welcome.
7. [_new_] Update _profiles_ sample to include previously missing items: a) enable `VK_KHR_portability_subset` device extension when `VKB_ENABLE_PORTABILITY=ON`, b) allocate descriptors from update-after-bind pool for descriptor indexing scalability, and c) destroy images and sampler on exit - all to eliminate validation errors.

Fixes #1434

Regression tested for MoltenVK on macOS Ventura on x86_64 + AMD 6600 and on macOS Sequoia on M1 Air using SDK 1.4.328.1 as well as the latest development build of MoltenVK.

@aitor-lunarg if you could review and/or test this functionality with KosmicKrisp that would be great.  See macOS [build instructions](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/docs/build.adoc#macos). ~~Simply add `-DVKB_ENABLE_PORTABILITY=OFF` to the command line.~~  There are some unknowns regarding project dependencies on components like the Vulkan **DynamicLoader** and **volk**, but I am hoping those work as-is for KosmicKrisp.

Once this PR has been verified as working by others, I will add an additional commit for macOS build instructions.

## General Checklist:

## Sample Checklist

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2024, Arm Limited and Contributors
+# Copyright (c) 2019-2025, Arm Limited and Contributors
 #
 # SPDX-License-Identifier: Apache-2.0
 #
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -135,7 +135,7 @@ if(IOS)
                 XCODE_EMBED_FRAMEWORKS_CODE_SIGN_ON_COPY		"YES"
         )
     endif ()
-    # No need to search for Vulkan package or MoltenVK library, Vulkan cache variables already defined on Apple platforms by global_options.cmake
+    # Vulkan cache variables already defined by main project CMakeLists and updated on Apple platforms by global_options.cmake
     if(Vulkan_LIBRARY AND ${Vulkan_VERSION} VERSION_GREATER_EQUAL 1.3.278)
         target_sources(${PROJECT_NAME} PRIVATE
                 ${Vulkan_Target_SDK}/iOS/share/vulkan
@@ -163,17 +163,24 @@ if(IOS)
     set(FRAMEWORKS_TO_EMBED)
 	if(Vulkan_MoltenVK_LIBRARY)
 		list(APPEND FRAMEWORKS_TO_EMBED "${Vulkan_MoltenVK_LIBRARY};")
+		message(STATUS "Embedding Vulkan driver: ${Vulkan_MoltenVK_LIBRARY}")
+	# add support for potentially other iOS Vulkan drivers here...
+	#elseif(OTHER_IOS_VULKAN_DRIVER)
+		#list(APPEND FRAMEWORKS_TO_EMBED "${OTHER_IOS_VULKAN_DRIVER};")
+		#message(STATUS "Embedding Vulkan driver: ${OTHER_IOS_VULKAN_DRIVER}")
 	else()
-		message(FATAL_ERROR "Can't find MoltenVK library. Please install the Vulkan SDK or MoltenVK project and set VULKAN_SDK.")
+		message(FATAL_ERROR "Can't find Vulkan driver for iOS. Please install the Vulkan SDK and run: 'source <MY_SDK_PATH>/iOS/setup-env.sh'")
 	endif()
 	if(Vulkan_LIBRARY)
 		list(APPEND FRAMEWORKS_TO_EMBED "${Vulkan_LIBRARY};")
+		message(STATUS "Embedding Vulkan loader: ${Vulkan_LIBRARY}")
 	endif()
 	if(Vulkan_Layer_VALIDATION)
 		# trouble is can't turn this on/off if XCode decides to build debug and we're configured for release.  Need to revist
 		# note the Vulkan validation layer must be present and enabled even in release mode for the shader_debugprintf sample
 		#if(("${VKB_DEBUG}" STREQUAL "ON") OR ("${VKB_VALIDATION_LAYERS}" STREQUAL "ON"))
 			list(APPEND FRAMEWORKS_TO_EMBED "${Vulkan_Layer_VALIDATION}")
+			message(STATUS "Embedding Vulkan Validation Layer: ${Vulkan_Layer_VALIDATION}")
 		#endif()
     endif()
     set_target_properties(${PROJECT_NAME} PROPERTIES
--- a/bldsys/cmake/global_options.cmake
+++ b/bldsys/cmake/global_options.cmake
@@ -105,6 +105,8 @@ if(APPLE)
 			set(CMAKE_SUPPRESS_REGENERATION ON)
 		endif()
 	endif()
+else()
+    option(VKB_ENABLE_PORTABILITY "Enable portability enumeration and subset features in the framework.  This is default off on non-Apple platforms." OFF)
 endif()
 
 set(VKB_WARNINGS_AS_ERRORS ON CACHE BOOL "Enable Warnings as Errors")
--- a/bldsys/cmake/global_options.cmake
+++ b/bldsys/cmake/global_options.cmake
@@ -38,7 +38,7 @@ if(APPLE)
 	if(VKB_ENABLE_PORTABILITY)
 		find_package(Vulkan QUIET OPTIONAL_COMPONENTS MoltenVK)
 		if(USE_MoltenVK OR (IOS AND (NOT Vulkan_MoltenVK_FOUND OR ${CMAKE_OSX_SYSROOT} STREQUAL "iphonesimulator")))
-			# if using MoltenVK, or MoltenVK for iOS was not found, or using iOS Simulator, look for MoltenVK in the Vulkan SDK and MoltenVK project locations
+			# if using MoltenVK standalone, or MoltenVK for iOS not found or using iOS Simulator, look for MoltenVK in Vulkan SDK and MoltenVK project paths
 			if(NOT Vulkan_MoltenVK_LIBRARY)
 				# since both are available in the Vulkan SDK and MoltenVK github project, make sure we look for MoltenVK framework on iOS and dylib on macOS
 				set(_saved_cmake_find_framework ${CMAKE_FIND_FRAMEWORK})
@@ -72,13 +72,13 @@ if(APPLE)
 				else()
 					message(FATAL_ERROR "Vulkan library found in MoltenVK search path. Please set VULKAN_SDK to the MoltenVK project install location.")
 				endif()
-				message(STATUS "Using MoltenVK: ${Vulkan_MoltenVK_LIBRARY}")
+				message(STATUS "Using MoltenVK standalone: ${Vulkan_MoltenVK_LIBRARY}")
 			else()
 				message(FATAL_ERROR "Can't find MoltenVK library. Please install the Vulkan SDK or MoltenVK project and set VULKAN_SDK.")
 			endif()
-		elseif(IOS)
-			# if not using MoltenVK standalone on iOS, set up global Vulkan Library define for iOS Vulkan loader
-			add_compile_definitions(_HPP_VULKAN_LIBRARY="vulkan.framework/vulkan")
+		#else()
+			# if not using MoltenVK standalone, retain find_package() results for MoltenVK, Vulkan loader, and Validation Layer library defines
+			# no need to override with _HPP_VULKAN_LIBRARY in this case since Vulkan DynamicLoader will find/load Vulkan library on macOS & iOS
 		endif()
 	endif()
 
--- a/bldsys/cmake/global_options.cmake
+++ b/bldsys/cmake/global_options.cmake
@@ -33,53 +33,57 @@ endif()
 
 if(APPLE)
     cmake_minimum_required(VERSION 3.24)
-    option(VKB_ENABLE_PORTABILITY "Enable portability enumeration and subset features in the framework.  This is required for MoltenVK on Apple platforms." ON)
+    option(VKB_ENABLE_PORTABILITY "Enable portability enumeration and subset features in the framework. This is default ON for Apple platforms." ON)
 
-	if(VKB_ENABLE_PORTABILITY)
-		find_package(Vulkan QUIET OPTIONAL_COMPONENTS MoltenVK)
-		if(USE_MoltenVK OR (IOS AND (NOT Vulkan_MoltenVK_FOUND OR ${CMAKE_OSX_SYSROOT} STREQUAL "iphonesimulator")))
-			# if using MoltenVK standalone, or MoltenVK for iOS not found or using iOS Simulator, look for MoltenVK in Vulkan SDK and MoltenVK project paths
-			if(NOT Vulkan_MoltenVK_LIBRARY)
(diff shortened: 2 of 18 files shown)


SaschaWillems:
We just discussed this, also in light of what you did for my repo. If I did understand you correct, if MoltenVK could be adjusted, we could get rid of having to handle it different than KosmicKrisp? Is that correct? If so we could forward that to the people still working on MoltenVK.

SRSaunders:
> ...if MoltenVK could be adjusted, we could get rid of having to handle it different than KosmicKrisp? Is that correct?

Not really.  The main end-user difference between MoltenVK and KosmicKrisp is recognition of MoltenVK being a Vulkan Portability driver with the need for enabling portability extensions (`VK_KHR_portability_enumeration` at the instance level and `VK_KHR_portability_subset` at the device level).  Fortunately this is encapsulated in the Vulkan-Samples project with the `VKB_ENABLE_PORTABILITY` CMake variable and identically-named compiler define.  KosmicKrisp will not need or support these extensions as it is being built as a conformant (i.e. non-portability) driver from the outset.

What I have tried to do in this PR is:
1. Leverage the `VKB_ENABLE_PORTABILITY` CMake variable by changing it from hard-coded ON on Apple into an cmake option switch when configuring the project.  `VKB_ENABLE_PORTABILITY` is set to default ON to support MoltenVK, but will need to be specified OFF on the cmake command line for KosmicKrisp.  The default could change later, but this should be a good start.  Once this PR is tested by the KosmicKrisp developers, I will add a doc commit here to help explain this.
2. Remove as many MoltenVK-specifics from the samples as possible and guard any remaining necessary differences with `VKB_ENABLE_PORTABILITY` to ensure cross-driver compatibility.
3. A special note re the _profiles_ sample: The **vulkan_profiles.hpp** generated file contains a bad assumption that all Apple platforms require the portability extensions be enabled.  Going forward this will be incorrect.  I have taken a quick and dirty approach in this PR and removed that bad assumption from the generated file, leaving enablement of those extensions in the sample itself.  However, this should really be fixed in the generator script (different project) for the **vulkan_profiles.hpp** file.

Regarding your comment about asking the MoltenVK folks to handle things differently (i.e. portability vs conformant), I think that may be unreasonable or perhaps impossible at this point.  Portability was a major design assumption of MoltenVK and has allowed it to advance its Vulkan version support by leveraging the subsetting nature of the Vulkan spec.  I think MoltenVK is fine as is, supporting macOS, iOS and arm64/x86_64 across a wide range of OS versions.  My goal here was not to force change back down the pipe, but to accommodate both approaches in the Vulkan-Samples project.  Note that KosmicKrisp will start with a narrower support matrix of newer macOS versions on arm64 only, and I don't believe iOS support is planned at least for now.

SRSaunders:
Not sure why Antora doc build is failing.  I didn't touch anything related with this PR.  Strange...

SaschaWillems:
The Doc build failure is our fault, see https://github.com/KhronosGroup/antora-ui-khronos/pull/30

gpx1000:
Fixed it.

gpx1000:
Also, I agree with @SRSaunders, there's nothing for MoltenVK to fix.  That's the way MoltenVK is designed to work.  During the meeting today, we were under the impression that @SRSaunders found some bug in MoltenVK having to do with non-standard's compliant layer handling in Sascha's samples.  I think that's what we were wondering if we needed to escalate knowledge about to get fixed in the MoltenVK team.

aitor-lunarg:
Just tested and it works with no issues, thank you!

I do have one suggestion, but feel free to ignore it. Changing drivers is as easy as setting a new value for `VK_DRIVER_FILES`. Would it be possible not to have something like `VKB_ENABLE_PORTABILITY` and just decide at run time if portability is needed or not based on available drivers? Mainly with the idea of letting users easily switch between MoltenVK, KosmicKrisp and Lavapipe as their Vulkan backend.

SRSaunders:
> Just tested and it works with no issues, thank you!

@aitor-lunarg thanks for testing quickly and delivering excellent news!  I was hoping external dependencies like the Vulkan `DynamicLoader` (vulkan.hpp) and `volk` would work as-is, and I am glad to hear they do.

> Would it be possible not to have something like VKB_ENABLE_PORTABILITY and just decide at run time if portability is needed or not based on available drivers? 

This is an interesting idea, but would require changing the code from being a build-time decision to a runtime one.  This may indeed be possible, but I would have to investigate if there are any limitations or downsides here.  On the positive side, part of this concept is already in place - the portability extensions are always checked for availability at runtime prior to enabling them.  However, we also need to understand the implications of always including **vulkan_beta.h** and things like `VkPhysicalDevicePortabilitySubsetFeaturesKHR` on non-portability drivers.  Please let me have a look at this and do some prototyping before I answer.

> Changing drivers is as easy as setting a new value for VK_DRIVER_FILES

This is also good to hear.  Can you tell me how a future SDK will handle setting up this env variable for KosmicKrisp vs. MoltenVK?  Currently there is a single **setup-env.sh** file at the SDK root level (and a different one in the iOS subfolder) that sets up important env variables including `VULKAN_SDK`, `VK_DRIVER_FILES`, etc.  Going forward will there be multiple **setup-env.sh** files?  Or perhaps an updated script with some optionality?  In other words have you figured out how SDK co-residency and env variable setup will work for selection of either KosmicKrisp or MoltenVK?

Also will CMake's `find_package(Vulkan OPTIONAL_COMPONENTS ...)` be able to handle the differences?  I presume it will key off of `VULKAN_SDK`, but will there be new optional components for KosmicKrisp with new variables similar to `Vulkan_MoltenVK_FOUND` and `Vulkan_MoltenVK_LIBRARY`?

SRSaunders:
> > Would it be possible not to have something like VKB_ENABLE_PORTABILITY and just decide at run time if portability is needed or not based on available drivers?
> 
> Please let me have a look at this and do some prototyping before I answer.

@aitor-lunarg I have updated this PR to ensure there are runtime checks for the portability extensions and features.  However, I have left `VKB_ENABLE_PORTABILITY` in place as a developer build option, and have it turned ON by default for Apple and OFF by default for other platforms.  There should be no need to expose this to end users or in the docs.  The runtime checks should ensure KosmicKrisp can run without special configuration (i.e. can follow current instructions) other than setting a new value for `VK_DRIVER_FILES`.

However, you may now see runtime warning messages that `VK_KHR_portability_subset` or certain portability features are not available.  This is a by-product of leaving `VKB_ENABLE_PORTABILITY=ON`.  These messages would not appear with `VKB_ENABLE_PORTABILITY=OFF`, but I think the benefit of simplicity may be worth the trade-off. The messages are informational only and will not affect performance in any way.

Note that I am still interested in how env variables will be configured for MoltenVK vs. KosmicKrisp in a co-resident SDK via **setup-env.sh**.  If you have information about this I could update the docs to provide instructions to Vulkan-Samples users when running on macOS.  I presume iOS will not change for now.

If you want to test this updated PR with KosmicKrisp that would be great. If you come back with positive results, I could move this PR out of WIP status into ready for review. Thanks.

gpx1000 (approved):
This works well for me on KosmicKrisp and MoltenVK.  Also tested in Linux to ensure everything still works everywhere.

aitor-lunarg:
> Note that I am still interested in how env variables will be configured for MoltenVK vs. KosmicKrisp in a co-resident SDK via setup-env.sh. If you have information about this I could update the docs to provide instructions to Vulkan-Samples users when running on macOS. I presume iOS will not change for now.

My understanding is that it will just be added to the path used by the loader to search for drivers. Right now, setup-env.sh will export `VK_DRIVER_FILES="$VULKAN_SDK/share/vulkan/icd.d/MoltenVK_icd.json"`, so that will eventually also contain KK. Then it's up to the developer to choose which driver they want to use as you would in any other environment with multiple Vulkan drivers.

richard-lunarg:
The plan for the first Alpha release of KosmicKrisp is to not include it in the system global folders the way MoltenVK is. The rationale is it may create havok if suddenly there are two ICD's to choose from now for developers using that option. An icd json file is included and people who want to experiment can embed it in their application bundles, or copy it to the system folder where MoltenVK is themselves. KosmicKrisp also does not support direct linking the way MoltenVK does. You have to use the Vulkan Loader. The ICD can then be placde in the system folders or (preferably) in the application bundle. I hope no one is actually shipping applications or games and assuming MoltenVK (or KosmicKrisp in the future) is installed in a system folder.

SRSaunders:
Thanks @richard-lunarg for the explanation on initial KosmicKrisp installation decisions for the SDK.  For the Vulkan-Samples project at least, the end user is asked to run setup-env.sh to define the path to the driver icd file (as well as to other things like the VVL).  Up to this point that has been for MoltenVK only.  Similar to @aitor-lunarg's comment above, will there be a second setup-env.sh available somewhere in an upcoming SDK which specifies VK_DRIVER_FILES and defines the path to KosmicKrisp_icd.json within the SDK (e.g. in a separate KosmicKrisp folder similar to what is done for iOS)?  Or will that be entirely manual and up to the developer for the Alpha release?

SRSaunders:
Thanks @gpx1000 for testing this PR and approving.  I was waiting for someone to provide feedback re the added runtime checks for portability extensions and features, which provides compatibility for MoltenVK and KosmicKrisp without prior configuration.  Given your positive results, I will now remove the WIP label on this PR.

gpx1000:
Thanks for the hard work on it!  Your help on Apple has been really appreciated!

richard-lunarg:
>Or will that be entirely manual and up to the developer for the Alpha release?

Manual for now, but all they have to do is set the one environment variable. I'll put something in the release notes about it. After the alpha and some other feedback, we'll probably finesse this a bit more for the next SDK release. The response to KosmicKrisp has been really strong, we are all very gratified by how much enthusiasm the community is showing for this project.

SRSaunders:
> Manual for now, but all they have to do is set the one environment variable. I'll put something in the release notes about it.

Thanks @richard-lunarg, sounds good.  I won't add any additional docs on the Vulkan-Samples side for now.

And @gpx1000 I'm happy to help out.  Thanks for your ongoing feedback, testing, and support re Apple.

richard-lunarg:
Heads up, we did have a change in plan late in testing for the next SDK. If KosmicKrisp is selected, it will be placed in the system folders along side MoltenVK. The loader sees MoltenVK first, so it should not break anything, but all you will need to do is run:

export VK_DRIVER_FILES=$VULKAN_SDK/share/vulkan/icd.d/libkosmickrisp_icd.json

To switch to KosmicKrisp from MoltenVK from the command line. There are multiple ways to use KosmicKrisp, and see the 'Getting Started Guide" for macOS in the SDK for details next week.

SRSaunders:
> If KosmicKrisp is selected, it will be placed in the system folders along side MoltenVK.

Will this be on arm64 hosts only?  I presume this will not be the case for x86_64 hosts.  Alternatively, will the SDK's KosmicKrisp installation option be available only on arm64?

> export VK_DRIVER_FILES=$VULKAN_SDK/share/vulkan/icd.d/libkosmickrisp_icd.json

This should work well for command line invocation of Vulkan-Samples.  However, for Xcode-based builds and execution, the user will have to update the Vulkan-Samples Xcode schema (under Environment Variables) to modify the definition of `VK_DRIVER_FILES`.  That's because shell definitions are not available when running under Xcode.  Alternatively, if the user executes the above line _after_ running setup-env.sh and _before_ cmake configuration for the project, the new definition will automatically be populated in the project's Xcode schema.

richard-lunarg:
>Will this be on arm64 hosts only? I presume this will not be the case for x86_64 hosts. 
>Alternatively, will the SDK's KosmicKrisp installation option be available only on arm64?

Ah, good point/question. The installer does not distinguish between Apple Silicon and x86_64, but KosmicKrisp is only for Apple Silicon. I've added a note at least to indicate this on the installer and in the release notes.

richard-lunarg:
>for Xcode-based builds and execution

They can also include KosmicKrisp in the application bundle, which is the recommended method for distributable applications as well.

SRSaunders:
Thanks @richard-lunarg for your answers.

gpx1000:
We're going to meet on Monday, to review, I wanted to check if you guys have anything pending you'd like for the PR prior to accepting it?  Best I can tell this PR is ready, but I don't want to cut anyone off if there's pending changes we need.  Maybe can do another PR for those if they're needed?  I'm hoping to get KosmicKrisp support or Samples out there.

PR #1443, opened 2025-11-11
Resolve VVL warning on incompatible OpTypeImage vs. image view format

asuessenbach:
## Description

One possible way to resolve the VVL warnings on incompatible `OpTypeImage`, without the need of additional extensions.
Just needs to switch the generated image from BGRA8 to RGBA8, and overwrites the function `create_render_context` to enforce selection of an RGBA surface format.

Build tested on Win10 with VS2022. Run tested on Win10 with NVidia GPU.

Fixes #1423

## General Checklist:

Please ensure the following points are checked:

- [x] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Code-Style)
- [x] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Copyright-Notice-and-License-Template)
- [x] I have commented any added functions (in line with Doxygen)
- [x] I have commented any code that could be hard to understand
- [x] My changes do not add any new compiler warnings
- [x] My changes do not add any new validation layer errors or warnings
- [x] I have used existing framework/helper functions where possible
- [x] My changes do not add any regressions
- [x] I have tested every sample to ensure everything runs correctly
- [x] This PR describes the scope and expected impact of the changes I am making

 Note: The Samples CI runs a number of checks including:
 - [x] I have updated the header Copyright to reflect the current year (CI build will fail if Copyright is out of date)
 - [ ] My changes build on Windows, Linux, macOS and Android. Otherwise I have [documented any exceptions](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements)

 If this PR contains framework changes:
 - [ ] I did a full batch run using the `batch` command line argument to make sure all samples still work properly

## Sample Checklist

If your PR contains a new or modified sample, these further checks must be carried out *in addition* to the General Checklist:
- [x] I have tested the sample on at least one compliant Vulkan implementation
- [ ] If the sample is vendor-specific, I have [tagged it appropriately](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#General-Requirements)
- [x] I have stated on what implementation the sample has been tested so that others can test on different implementations and platforms
- [ ] Any dependent assets have been merged and published in downstream modules
[...]

PR #1452, opened 2025-12-07
Remove need to specify image formats for ray tracing storage  images

### LINKED ISSUE #1423: Validation layer warnings on ray tracing samples
With the ray tracing samples `ray_tracing_basic`, `ray_tracing_extended`, `ray_tracing_position_fetch`, and `ray_tracing_reflection` I get this validation layer warning:
```
vkCmdTraceRaysKHR(): the storage image descriptor [VkDescriptorSet 0x6c000000006c, Set 0, Binding 1, Index 0, variable "image"] is accessed by a OpTypeImage that has a Format operand Rgba8 (equivalent to VK_FORMAT_R8G8B8A8_UNORM) which doesn't match the VkImageView 0x500000000050 format (VK_FORMAT_B8G8R8A8_UNORM). Any loads or stores with the variable will produce undefined values to the whole image (not just the texel being accessed). While the formats are compatible, Storage Images must exactly match. Few ways to resolve this are
1. Set your ImageView to VK_FORMAT_R8G8B8A8_UNORM and swizzle the values in the shader to match the desired results.
2. Use the Unknown format in your shader (will need the widely supported shaderStorageImageWriteWithoutFormat feature)
Spec information at https://docs.vulkan.org/spec/latest/chapters/textures.html#textures-format-validation
```
Simply changing the `VkImage`'s and `VkImageView`'s format to `VK_FORMAT_R8G8B8A8_UNORM` would resolve the warning, but of course changes the visual output of those samples.

SaschaWillems:
## Description

This is an alternative fox for #1443. Instead of adjusting image formats in the C++ side of the code, this uses widely supported features to remove the need to specify format qualifiers at all.

Works with glsl, slang and hlsl.

Fixes #1423

## General Checklist:


asuessenbach (approved):
I would prefer the approach from #1443.

Besides that: Approved.

SaschaWillems:
Indeed. It feels odd that the correct way is more verbose. Something that might be worth fixing with future Vulkan versions. I've raised in internal issue.

gary-sweet (approved):
shaderStorageImageWriteWithoutFormat is only supported by more recent hardware for us, but since we don't support ray tracing, it doesn't actually matter.

gary-sweet (commented):
Sorry, I shouldn't actually approve this since I can't test it

gpx1000 (approved):

PR #1473, opened 2026-01-22
New GLTF API Sample Base Class & GLTF Sample

### LINKED ISSUE #848: glTF loader for api samples is severely limited
While working on a new sample requiring me to render a complete glTF scene I noticed that the glTF loader for the api samples is severely limited. It can only load (and display) a single mesh with a single primitive from a glTF file. That's a real hindrance and something we should address near-term. It also does not apply any matrices, so scenes aren't scaled properly.

Sadly the glTF loader initially taken from my code has been partially wired to the framework, making it non-trivial to fix this without needing to change a lot of things.

HaoranXu-ARM:
## Description

Add a new GLTF API Sample base class to the framework, as well as a new GLTF sample instance to demonstrate its usage.

### Note
We made the following changes to the framework:

- Added new sample base class containing helper functions for loading and **rendering** a glTF scene. 

Fixes #848

The sample builds and runs successfully on Windows (NVIDIA) and Android (Mali).

## General Checklist:

## Sample Checklist


samples/general/gltf/gltf.cpp
```diff
@@ -0,0 +1,639 @@
...
+ */
+
+#include "gltf.h"
+
+GLTF::GLTF()
+{
+	title = "GLTF";
+	add_instance_extension(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+	add_device_extension(VK_KHR_CREATE_RENDERPASS_2_EXTENSION_NAME);
+	add_device_extension(VK_KHR_MAINTENANCE2_EXTENSION_NAME);
```
asuessenbach: No need to add the same extensions that are already added in `GLTFApiVulkanSample`.

samples/general/gltf/gltf.cpp
```diff
@@ -0,0 +1,639 @@
...
+}
+
+GLTF::~GLTF()
+{
+}
+
+bool GLTF::prepare(const vkb::ApplicationOptions &options)
+{
+	if (!ApiVulkanSample::prepare(options))
```
asuessenbach: You should call the base class' prepare here, that is `GLTFApiVulkanSample::prepare`

framework/gltf_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,580 @@
...
+	vkDestroyPipelineLayout(device_handle, pipeline_data.pipeline_layout, nullptr);
+	vkDestroyDescriptorSetLayout(device_handle, pipeline_data.set_layout, nullptr);
+	pipeline_data.pipeline        = VK_NULL_HANDLE;
+	pipeline_data.pipeline_layout = VK_NULL_HANDLE;
+	pipeline_data.set_layout      = VK_NULL_HANDLE;
+}
+
+GLTFApiVulkanSample::GLTFApiVulkanSample()
+{
+	title = "GLTFApiVulkanSample";
```
asuessenbach: As this class is designed as a base class for sample, you don't need to set a `title` here. That should be done in the actual sample.

framework/gltf_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,580 @@
...
+
+		vkDestroySampler(device_handle, samplers.nearest, nullptr);
+
+		vkDestroyFramebuffer(device_handle, main_pass.framebuffer, nullptr);
+	}
+}
+
+bool GLTFApiVulkanSample::prepare(const vkb::ApplicationOptions &options)
+{
+	return false;
```
asuessenbach: You don't need to override functions from a base class that doesn't do anything.

Besides that, if at all, should probably return `true` here.

framework/gltf_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,580 @@
...
+	    .unnormalizedCoordinates = VK_FALSE,
+	};
+
+	VK_CHECK(vkCreateSampler(get_device().get_handle(), &sampler_create_info, nullptr, &samplers.nearest));
+}
+
+void GLTFApiVulkanSample::prepare_pipelines()
+{}
+
+void GLTFApiVulkanSample::build_command_buffers()
```
asuessenbach: As with `prepare`: no need to override a function that actually does nothing.

framework/gltf_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,580 @@
...
+	view_changed();
+
+	prepared = true;
+	return true;
+}
+
+void GLTFApiVulkanSample::reset_gpu_data()
+{}
+
+void GLTFApiVulkanSample::render(float delta_time)
```
asuessenbach: As above.

framework/gltf_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,580 @@
...
+{}
+
+void GLTFApiVulkanSample::load_assets()
```
asuessenbach: Maybe make this function a bit more versatile by having the scene file to load as an argument?

framework/gltf_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,580 @@
...
+	vkb::GLTFLoader loader{get_device()};
+	sg_scene = loader.read_scene_from_file(scene_file);
+	assert(sg_scene);
+
+	camera.type              = vkb::CameraType::FirstPerson;
+	const float aspect_ratio = 1.0f;        // dummy value. Reset by update_extents.
+	camera.set_perspective(50.0f, aspect_ratio, 4000.0f, 1.0f);
+	camera.set_rotation(glm::vec3(230.0f, 101.0f, -5.0f));
+	camera.set_translation(glm::vec3(115.0f, -390.0f, 18.0f));
+	camera.translation_speed = 100.0f;
```
asuessenbach, line 169: Maybe try to extract a camera from the scene file?

framework/gltf_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,580 @@
...
+	    .maxLod                  = 0.0f,
+	    .borderColor             = VK_BORDER_COLOR_FLOAT_TRANSPARENT_BLACK,
+	    .unnormalizedCoordinates = VK_FALSE,
+	};
+
+	VK_CHECK(vkCreateSampler(get_device().get_handle(), &sampler_create_info, nullptr, &samplers.nearest));
+}
+
+void GLTFApiVulkanSample::prepare_pipelines()
+{}
```
asuessenbach: Why do you introduce this as a virtual function, but don't call it in this class?

framework/gltf_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,580 @@
...
+	VkDescriptorPoolCreateInfo descriptor_pool_create_info =
+	    vkb::initializers::descriptor_pool_create_info(
+	        vkb::to_u32(pool_sizes.size()),
+	        pool_sizes.data(), max_sets);
+
+	VK_CHECK(vkCreateDescriptorPool(get_device().get_handle(), &descriptor_pool_create_info, nullptr, &main_pass.descriptor_pool));
+	debug_utils.set_debug_name(get_device().get_handle(), VK_OBJECT_TYPE_DESCRIPTOR_POOL, get_object_handle(main_pass.descriptor_pool), "Main pass descriptor pool");
+}
+
+void GLTFApiVulkanSample::setup_additional_descriptor_pool()
```
asuessenbach: As with `prepare_pipelines`: a virtual function that doesn't get called here?

asuessenbach (commented):
I think, having some GLTFApiVulkanSample base class would be a good extension to the class hierarchy.
But the responsibilities between `GLTFApiVulkanSample` and `GLTF` seems to be a bit fuzzy:
- empty virtual functions introduced with `GLTFApiVulkanSample`, but only called in `GLTF`?
- elements (like `present.pipeline.pipeline`) that are members in `GLTFApiVulkanSample`, but created and used in `GLTF`; and destroyed in `GLTFApiVulkanSample`.
Maybe more, stopped reviewing at that point.

If we want to have something like `GLTFApiVulkanSample`, the tasks and responsibilities of that class should be a bit cleaner.
Or maybe we could introduce some GLTF class that provides all those capabilities and could be used by a sample, without introducing another layer above ApiVulkanSample? Would probably make such a class more widely usable.

samples/general/gltf/gltf.cpp
HaoranXu-ARM: Done

samples/general/gltf/gltf.cpp
HaoranXu-ARM: Done

framework/gltf_api_vulkan_sample.cpp
HaoranXu-ARM: Done

framework/gltf_api_vulkan_sample.cpp
HaoranXu-ARM: Makes sense, removed.

framework/gltf_api_vulkan_sample.cpp
HaoranXu-ARM: Done

framework/gltf_api_vulkan_sample.cpp
HaoranXu-ARM: Sure, I added scene_file as an optional parameter of this function.

framework/gltf_api_vulkan_sample.cpp
HaoranXu-ARM: I tried to directly extract a camera and parse it into a Vulkan Samples camera, but the transformations and axes conventions are very confusing. Since the primary purpose of this PR is to refactor the previous FDM sample, I prefer we leave this as the default implementation and save the camera extraction for a future PR :)

framework/gltf_api_vulkan_sample.cpp
HaoranXu-ARM: Changed it to be called by reset_gpu_data().

framework/gltf_api_vulkan_sample.cpp
HaoranXu-ARM: Changed it to be called by reset_gpu_data().

framework/gltf_api_vulkan_sample.cpp
HaoranXu-ARM: Done

HaoranXu-ARM:
> * elements (like `present.pipeline.pipeline`) that are members in `GLTFApiVulkanSample`, but created and used in `GLTF`; and destroyed in `GLTFApiVulkanSample`.
>   Maybe more, stopped reviewing at that point.

Fixed the mismatch in site of creation and destroy.

> If we want to have something like `GLTFApiVulkanSample`, the tasks and responsibilities of that class should be a bit cleaner. Or maybe we could introduce some GLTF class that provides all those capabilities and could be used by a sample, without introducing another layer above ApiVulkanSample? Would probably make such a class more widely usable.

I think the goal of `GLTFApiVulkanSample` is to provide a basic framework for rendering a glTF scene, but leave the application-specific methods such as pipeline creation to the individual samples. `GLTF` is directly created from the sample template that comes with this PR, which builds upon it by filling the gaps with a very simple application set-up. 

The expected use case will be that developers use the `generate.py` script to generate a working sample that loads & renders a glTF scene straight out-of-the-box, and they can start from there by overriding the functionalities. So I think the additional layer above ApiVulkanSample is still useful in its own right.

framework/gltf_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,581 @@
...
+	add_device_extension(VK_KHR_CREATE_RENDERPASS_2_EXTENSION_NAME);
+	add_device_extension(VK_KHR_MAINTENANCE2_EXTENSION_NAME);
+}
+
+GLTFApiVulkanSample::~GLTFApiVulkanSample()
+{
+	if (has_device())
+	{
+		VkDevice device_handle = get_device().get_handle();
+		vkDestroyDescriptorPool(device_handle, main_pass.descriptor_pool, nullptr);
```
asuessenbach, line 57: You explicitly set all destroyed resources but `main_pass.descriptor_pool` to `VK_NULL_HANDLE`. For a reason?

samples/general/gltf/gltf.cpp
```diff
@@ -0,0 +1,650 @@
...
+	if (has_device())
+	{
+		VkDevice device_handle = get_device().get_handle();
+
+		vkDestroyPipeline(device_handle, present.pipeline.pipeline, nullptr);
+		present.pipeline.pipeline = VK_NULL_HANDLE;
+		vkDestroyPipeline(device_handle, main_pass.sky_pipeline.pipeline, nullptr);
+		present.pipeline.pipeline = VK_NULL_HANDLE;
+		vkDestroyPipeline(device_handle, main_pass.meshes.pipeline.pipeline, nullptr);
+		present.pipeline.pipeline = VK_NULL_HANDLE;
```
asuessenbach: `present.pipeline.pipeline` is set to `VK_NULL_HANDLE` three times?

framework/gltf_api_vulkan_sample.cpp
```diff
@@ -0,0 +1,581 @@
...
+	get_device().wait_idle();
+
+	// Notify derived class.
+	view_changed();
+
+	prepared = true;
+	return true;
+}
+
+void GLTFApiVulkanSample::reset_gpu_data()
```
asuessenbach: `reset_gpu_data`? What does that mean? Is here anything not related to the GPU?

Besides that: this function is overridden by GLTF, which does _not_ call into its parent implementation. That is, whatever you have in this function is more or less meaningless?

framework/gltf_api_vulkan_sample.cpp
HaoranXu-ARM: Fixed now

samples/general/gltf/gltf.cpp
HaoranXu-ARM: Fixed now

framework/gltf_api_vulkan_sample.cpp
HaoranXu-ARM: Renamed to `recreate_swapchain_resources` and removed the default implementation.

SaschaWillems:
Not sure about this. What we're looking for is an improved glTF loader class that adds stuff like animations. Adding another sample base class (we already have two) might not be the right direction.

Issue #1462, opened 2026-01-04
Layout discrepancy in dynamic_rendering_local_read


mbty-hi:
There is a discrepancy between the layout `swapchain_buffers[i].image` gets [transitioned to](https://github.com/KhronosGroup/Vulkan-Samples/blob/97fcdeecf2db26a78b432b285af3869a65bb00bd/samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp#L846) before the render/gets [transitioned from](https://github.com/KhronosGroup/Vulkan-Samples/blob/97fcdeecf2db26a78b432b285af3869a65bb00bd/samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp#L952) after it, and what [is declared](https://github.com/KhronosGroup/Vulkan-Samples/blob/97fcdeecf2db26a78b432b285af3869a65bb00bd/samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp#L849C1-L860C66) as part of the related `VkRenderingAttachmentInfoKHR` in the [dynamic_rendering_local_read](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/extensions/dynamic_rendering_local_read) sample (`VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL` vs. `VK_IMAGE_LAYOUT_RENDERING_LOCAL_READ_KHR`). Despite that, the sample runs without warnings on machine, though it must be said that it still runs fine if I replace the `VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL` with about any other layout, or even if I wholly remove the two explicit transitions for this resource. Similarly, I can replace `VK_IMAGE_LAYOUT_RENDERING_LOCAL_READ_KHR` with other layouts without Vulkan complaining. However, switching the nearby `VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_OPTIMAL` for a `VK_LAYOUT_UNDEFINED` results in a black screen (understandably). I wonder if this could be a matter of layouts actually being the same behind the scenes (resulting in ignored transitions/accidental compatibility) + incomplete layers, though I may well be misunderstanding something more fundamental.

SaschaWillems:
Thanks for bringing this up. We'll take a look and fix if necessary.

[...]

PR #1483, opened 2026-02-11
Adding GUI Support for Dynamic Rendering

ellioman:
## Description

Adds framework support for rendering the GUI overlay when samples use dynamic rendering instead of render passes, and introduces a cleaner `get_gui_subpass()` pattern for samples that render the GUI in a non-zero subpass.

### Changes
**Gui class (gui.h):**
- Added `prepare()` overload that accepts color/depth formats instead of a render pass, creating the GUI pipeline with `VkPipelineRenderingCreateInfoKHR`
- Added `draw()` overload that manages its own rendering pass internally for dynamic rendering
- Added `subpass` parameter to the render pass `prepare()` overload (default 0), replacing the old `set_subpass()` method
- Removed `set_subpass()` method and `subpass` member variable — subpass is now passed directly at prepare time

**ApiVulkanSample / HppApiVulkanSample (api_vulkan_sample.h/cpp, hpp_api_vulkan_sample.h):**
- Added `uses_dynamic_rendering()` helper that checks if `render_pass == VK_NULL_HANDLE`
- Updated `prepare_gui()` to automatically detect dynamic rendering and call the appropriate `Gui::prepare()` overload
- `prepare_gui()` now passes `get_gui_subpass()` to `Gui::prepare()` for the render pass path
- Added `virtual uint32_t get_gui_subpass() const` (returns 0) — samples override this to render the GUI in a later subpass
- Added `draw_ui(cmd, swapchain_index)` overload for dynamic rendering
- Updated destructor to use the `uses_dynamic_rendering()` check

**dynamic_rendering sample:**
- Removed `prepare_gui()` override — the base class now handles both dynamic rendering and render pass GUI paths
- Added `setup_render_pass()` and `setup_framebuffer()` overrides that skip creation when using dynamic rendering, ensuring `uses_dynamic_rendering()` returns the correct value
- Added `draw_ui(cmd)` call in the render pass command buffer path (was previously missing)
- Fixed `clearValueCount` to use `clear_values.size()` instead of a hardcoded 3

**dynamic_rendering_local_read sample:**
- Replaced `prepare_gui()` override with `get_gui_subpass()` override returning 2
- Uses the new `draw_ui(cmd, swapchain_index)` overload for the dynamic rendering path

**color_write_enable sample:**
- Replaced `prepare_gui()` override with `get_gui_subpass()` override returning 1

### Usage
For samples using dynamic rendering, call `draw_ui()` after ending your main rendering pass:

```cpp
vkCmdEndRenderingKHR(cmd);
draw_ui(cmd, swapchain_image_index);
```

Unlike render passes which support subpasses with different attachment configurations, dynamic rendering has a fixed attachment configuration per vkCmdBeginRenderingKHR block. Since the GUI renders to a single color attachment but samples may use multiple attachments (e.g., G-buffer), the GUI must render in its own separate rendering pass - requiring the main pass to be ended first. Otherwise you'll get validation errors, like in the dynamic_rendering_local_read sample.

For samples using render passes, the existing `draw_ui(cmd)` call remains unchanged. If the GUI needs to render in a non-zero subpass, override `get_gui_subpass()` instead of `prepare_gui()`.

### Screenshots
Below is a screenshot from the dynamic_rendering sample.
dynamic_rendering

Below are two screenshots, from the dynamic_rendering_local_read sample, first one using renderpasses and the other dynamic rendering.

image

image

## General Checklist:

## Sample Checklist

## Jira Task
https://jira.arm.com/browse/STEGFX-369

### DIFF
--- a/framework/api_vulkan_sample.cpp
+++ b/framework/api_vulkan_sample.cpp
@@ -1,5 +1,5 @@
-/* Copyright (c) 2019-2025, Sascha Willems
- * Copyright (c) 2024-2025, Arm Limited and Contributors
+/* Copyright (c) 2019-2026, Sascha Willems
+ * Copyright (c) 2024-2026, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -81,9 +81,21 @@ bool ApiVulkanSample::prepare(const vkb::ApplicationOptions &options)
 void ApiVulkanSample::prepare_gui()
 {
 	create_gui(*window, nullptr, 15.0f, true);
-	get_gui().prepare(pipeline_cache, render_pass,
-	                  {load_shader("uioverlay/uioverlay.vert.spv", VK_SHADER_STAGE_VERTEX_BIT),
-	                   load_shader("uioverlay/uioverlay.frag.spv", VK_SHADER_STAGE_FRAGMENT_BIT)});
+
+	std::vector<VkPipelineShaderStageCreateInfo> shader_stages = {
+	    load_shader("uioverlay/uioverlay.vert.spv", VK_SHADER_STAGE_VERTEX_BIT),
+	    load_shader("uioverlay/uioverlay.frag.spv", VK_SHADER_STAGE_FRAGMENT_BIT)};
+
+	if (uses_dynamic_rendering())
+	{
+		VkFormat color_format = get_render_context().get_swapchain().get_format();
+		VkFormat depth_fmt    = depth_format;
+		get_gui().prepare(pipeline_cache, color_format, depth_fmt, shader_stages);
+	}
+	else
+	{
+		get_gui().prepare(pipeline_cache, render_pass, shader_stages, get_gui_subpass());
+	}
 }
 
 void ApiVulkanSample::update(float delta_time)
@@ -496,6 +508,19 @@ void ApiVulkanSample::draw_ui(const VkCommandBuffer command_buffer)
 	}
 }
 
+void ApiVulkanSample::draw_ui(const VkCommandBuffer command_buffer, uint32_t swapchain_image_index)
+{
+	if (has_gui())
+	{
+		const VkViewport viewport = vkb::initializers::viewport(static_cast<float>(width), static_cast<float>(height), 0.0f, 1.0f);
+		const VkRect2D   scissor  = vkb::initializers::rect2D(width, height, 0, 0);
+		vkCmdSetViewport(command_buffer, 0, 1, &viewport);
+		vkCmdSetScissor(command_buffer, 0, 1, &scissor);
+
+		get_gui().draw(command_buffer, swapchain_buffers[swapchain_image_index].view, width, height);
+	}
+}
+
 void ApiVulkanSample::prepare_frame()
 {
 	if (get_render_context().has_swapchain())
@@ -582,7 +607,7 @@ ApiVulkanSample::~ApiVulkanSample()
 			vkDestroyDescriptorPool(get_device().get_handle(), descriptor_pool, nullptr);
 		}
 		destroy_command_buffers();
-		if (render_pass != VK_NULL_HANDLE)
+		if (!uses_dynamic_rendering())
 		{
 			vkDestroyRenderPass(get_device().get_handle(), render_pass, nullptr);
 		}
--- a/framework/api_vulkan_sample.h
+++ b/framework/api_vulkan_sample.h
@@ -1,5 +1,5 @@
-/* Copyright (c) 2019-2025, Sascha Willems
- * Copyright (c) 2025, Arm Limited and Contributors
+/* Copyright (c) 2019-2026, Sascha Willems
+ * Copyright (c) 2025-2026, Arm Limited and Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -149,6 +149,12 @@ class ApiVulkanSample : public vkb::VulkanSampleC
 
 	// Global render pass for frame buffer writes
 	VkRenderPass render_pass = VK_NULL_HANDLE;
+
+	// Returns true if using dynamic rendering
+	bool uses_dynamic_rendering() const
+	{
+		return render_pass == VK_NULL_HANDLE;
+	}
 
 	// List of available frame buffers (same as number of swap chain images)
 	std::vector<VkFramebuffer> framebuffers;
@@ -372,6 +378,16 @@ class ApiVulkanSample : public vkb::VulkanSampleC
 	void draw_ui(const VkCommandBuffer command_buffer);
 
 	/**
+	 * @brief Draw UI for dynamic rendering mode
+	 * @param command_buffer A valid command buffer (after ending your main rendering pass)
+	 * @param swapchain_image_index The swapchain image index to render onto
+	 *
+	 * This method starts its own rendering pass, draws the UI, and ends the pass.
+	 * Call this after ending your main rendering pass (vkCmdEndRenderingKHR).
+	 */
+	void draw_ui(const VkCommandBuffer command_buffer, uint32_t swapchain_image_index);
+
+	/**
 	 * @brief Prepare the frame for workload submission, acquires the next image from the swap chain and
 	 *        sets the default wait and signal semaphores
 	 */
@@ -392,6 +408,15 @@ class ApiVulkanSample : public vkb::VulkanSampleC
 	 * @brief Initializes the UI. Can be overridden to customize the way it is displayed.
 	 */
 	virtual void prepare_gui();
+
+	/**
+	 * @brief Returns the subpass index used for GUI rendering (render pass path only).
+	 *        Override in samples that use a multi-subpass render pass with the GUI in a later subpass.
+	 */
+	virtual uint32_t get_gui_subpass() const
+	{
+		return 0;
+	}
 
   private:
 	/** brief Indicates that the view (position, rotation) has changed and buffers containing camera matrices need to be updated */
--- a/framework/gui.h
+++ b/framework/gui.h
@@ -1,6 +1,6 @@
-/* Copyright (c) 2018-2025, Arm Limited and Contributors
- * Copyright (c) 2019-2025, Sascha Willems
- * Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+/* Copyright (c) 2018-2026, Arm Limited and Contributors
+ * Copyright (c) 2019-2026, Sascha Willems
+ * Copyright (c) 2025-2026, NVIDIA CORPORATION. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -99,6 +99,8 @@ class Gui
 
 	using CommandBufferType                 = typename std::conditional<bindingType == BindingType::Cpp, vk::CommandBuffer, VkCommandBuffer>::type;
 	using DescriptorSetType                 = typename std::conditional<bindingType == BindingType::Cpp, vk::DescriptorSet, VkDescriptorSet>::type;
+	using FormatType                        = typename std::conditional<bindingType == BindingType::Cpp, vk::Format, VkFormat>::type;
+	using ImageViewType                     = typename std::conditional<bindingType == BindingType::Cpp, vk::ImageView, VkImageView>::type;
 	using PipelineType                      = typename std::conditional<bindingType == BindingType::Cpp, vk::Pipeline, VkPipeline>::type;
 	using PipelineCacheType                 = typename std::conditional<bindingType == BindingType::Cpp, vk::PipelineCache, VkPipelineCache>::type;
 	using PipelineLayoutType                = typename std::conditional<bindingType == BindingType::Cpp, vk::PipelineLayout, VkPipelineLayout>::type;
@@ -194,6 +196,18 @@ class Gui
 	 */
 	void draw(vkb::core::CommandBuffer<bindingType> &command_buffer);
 
+	/**
+	 * @brief Draws the Gui for dynamic rendering mode
+	 * @param command_buffer Command buffer to register draw-commands
+	 * @param swapchain_view The swapchain image view to render UI onto
+	 * @param width Render area width
+	 * @param height Render area height
+	 *
+	 * This method starts its own rendering pass, draws the UI, and ends the pass.
+	 * Call this after ending your main rendering pass (vkCmdEndRenderingKHR).
+	 */
+	void draw(CommandBufferType command_buffer, ImageViewType swapchain_view, uint32_t width, uint32_t height);
+
 	Drawer &get_drawer();
 
 	vk::ImageView get_font_image_view() const;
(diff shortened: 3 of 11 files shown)


framework/gui.h
```diff
@@ -1044,6 +1122,106 @@ inline void Gui<bindingType>::prepare_impl(vk::PipelineCache
...
+	                                                       .pInputAssemblyState = &input_assembly_state,
+	                                                       .pViewportState      = &viewport_state,
+	                                                       .pRasterizationState = &rasterization_state,
+	                                                       .pMultisampleState   = &multisample_state,
+	                                                       .pDepthStencilState  = &depth_stencil_state,
+	                                                       .pColorBlendState    = &color_blend_state,
+	                                                       .pDynamicState       = &dynamic_state,
+	                                                       .layout              = pipeline_layout->get_handle(),
+	                                                       .renderPass          = VK_NULL_HANDLE,
+	                                                       .basePipelineIndex   = -1};
```
ellioman: Note that this entire function is almost exactly the same as the one for renderpasses except

* creating the `pipeline_rendering_info`
* setting it as `.pNext`
* setting `.renderPass` to `VK_NULL_HANDLE`.

I wasn't sure whether we'd like to have the two almost identical functions, one for renderpasses and one for dynamic rendering, or whether we would like to create a helper for the common things. I decided this was an acceptable duplication as future changes might not need to be applied for both.

I can easily change this if requested.

asuessenbach:
With this change, I still get the same error as mentioned in #1462.

ellioman:
> With this change, I still get the same error as mentioned in #1462.

Good to know. That issue though seems to be unrelated to the work done in this PR. 

I'm working on a new sample, using dynamic rendering, and I need the GUI to work with that so I created this PR to add that support. I used the dynamic_rendering_local_read sample to test with, on Android + Windows, and didn't see any errors so I enabled it.

SaschaWillems:
Thank you very much for this!

LTGM, but can you also add the UI to the dynamic_rendering sample? That way both dynamic rendering samples do display a user interface.

framework/api_vulkan_sample.cpp
```diff
@@ -81,9 +81,21 @@ bool ApiVulkanSample::prepare(const vkb::ApplicationOptions &options)
...
-	                   load_shader("uioverlay/uioverlay.frag.spv", VK_SHADER_STAGE_FRAGMENT_BIT)});
+
+	std::vector<VkPipelineShaderStageCreateInfo> shader_stages = {
+	    load_shader("uioverlay/uioverlay.vert.spv", VK_SHADER_STAGE_VERTEX_BIT),
+	    load_shader("uioverlay/uioverlay.frag.spv", VK_SHADER_STAGE_FRAGMENT_BIT)};
+
+	if (uses_dynamic_rendering())
+	{
+		vk::Format color_format = static_cast<vk::Format>(get_render_context().get_swapchain().get_format());
+		vk::Format depth_fmt    = static_cast<vk::Format>(depth_format);
```
asuessenbach: ApiVulkanSample derives from VulkanSampleC. That is, you should use the C-binding types here, instead of the C++-binding types:
```
		VkFormat color_format = get_render_context().get_swapchain().get_format();
		VkFormat depth_fmt    = depth_format;
```

framework/api_vulkan_sample.h
```diff
@@ -150,6 +150,12 @@ class ApiVulkanSample : public vkb::VulkanSampleC
 	// Global render pass for frame buffer writes
 	VkRenderPass render_pass = VK_NULL_HANDLE;
 
+	// Returns true if using dynamic rendering
+	bool uses_dynamic_rendering() const
+	{
+		return render_pass == VK_NULL_HANDLE;
```
asuessenbach, line 156: It's not obvious to me why this condition means, dynamic rendering is used.

framework/gui.h
```diff
@@ -194,6 +194,18 @@ class Gui
...
+	 * @brief Draws the Gui for dynamic rendering mode
+	 * @param command_buffer Command buffer to register draw-commands
+	 * @param swapchain_view The swapchain image view to render UI onto
+	 * @param width Render area width
+	 * @param height Render area height
+	 *
+	 * This method starts its own rendering pass, draws the UI, and ends the pass.
+	 * Call this after ending your main rendering pass (vkCmdEndRenderingKHR).
+	 */
+	void draw(CommandBufferType command_buffer, vk::ImageView swapchain_view, uint32_t width, uint32_t height);
```
asuessenbach: gui is a class templated on the binding type.
That is, you should neither use C-binding types nor C++-binding types in its interface, but instead introduce some compile time conditional type, like
`	using ImageViewType = typename std::conditional<bindingType == BindingType::Cpp, vk::ImageView, VkImageView>::type;
`
and use that here:
`	void draw(CommandBufferType command_buffer, ImageViewType swapchain_view, uint32_t width, uint32_t height);
`

framework/gui.h
```diff
@@ -215,8 +227,12 @@ class Gui
 	 */
 	void new_frame();
 
+	// Prepare with a render pass
 	void prepare(PipelineCacheType pipeline_cache, RenderPassType render_pass, std::vector<PipelineShaderStageCreateInfoType> const &shader_stages);
 
+	// Prepare for dynamic rendering
+	void prepare(PipelineCacheType pipeline_cache, vk::Format color_format, vk::Format depth_format, std::vector<PipelineShaderStageCreateInfoType> const &shader_stages);
```
asuessenbach: Similar to the `draw` function above, you should use some compile time conditional type here
`	using FormatType = typename std::conditional<bindingType == BindingType::Cpp, vk::Format, VkFormat>::type;
`
giving
`	void prepare(PipelineCacheType pipeline_cache, FormatType color_format, FormatType depth_format, std::vector<PipelineShaderStageCreateInfoType> const &shader_stages);
`

framework/gui.h
```diff
@@ -514,6 +531,48 @@ inline void Gui<bindingType>::draw(CommandBufferType command_buffer)
 	draw(command_buffer, pipeline, pipeline_layout->get_handle(), descriptor_set);
 }
 
+template <vkb::BindingType bindingType>
+inline void Gui<bindingType>::draw(CommandBufferType command_buffer, vk::ImageView swapchain_view, uint32_t width, uint32_t height)
```
asuessenbach: `swapchain_view` should be a `ImageViewType` (introduced above), which should be handled similar to the `command_buffer`, being a `CommandBufferType`.
Or maybe introduce some function `draw_impl`, working on C++-binding types.

framework/gui.h
```diff
@@ -952,6 +1011,25 @@ inline void Gui<bindingType>::prepare(PipelineCacheType
...
+		prepare_impl(pipeline_cache, color_format, depth_format, shader_stages);
+	}
+	else
+	{
+		prepare_impl(static_cast<vk::PipelineCache>(pipeline_cache),
+		             color_format,
+		             depth_format,
+		             reinterpret_cast<std::vector<vk::PipelineShaderStageCreateInfo> const &>(shader_stages));
+	}
+}
```
asuessenbach, line 1041: As this function should take some `FormatType`, instead of `vk::Format`, it should probably look like
```
template <vkb::BindingType bindingType>
inline void Gui<bindingType>::prepare(PipelineCacheType                                     pipeline_cache,
                                      FormatType                                            color_format,
                                      FormatType                                            depth_format,
                                      std::vector<PipelineShaderStageCreateInfoType> const &shader_stages)
{
	if constexpr (bindingType == BindingType::Cpp)
	{
		prepare_impl(pipeline_cache, color_format, depth_format, shader_stages);
	}
	else
	{
		prepare_impl(static_cast<vk::PipelineCache>(pipeline_cache),
		             static_cast<vk::Format>(color_format),
		             static_cast<vk::Format>(depth_format),
		             reinterpret_cast<std::vector<vk::PipelineShaderStageCreateInfo> const &>(shader_stages));
	}
}
```

framework/gui.h
asuessenbach: I would vote for some helper(s) to do the common work.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
```diff
@@ -320,12 +321,19 @@ void DynamicRenderingLocalRead::setup_render_pass()
...
+	vk::Format color_format = static_cast<vk::Format>(get_render_context().get_swapchain().get_format());
+	vk::Format depth_fmt    = static_cast<vk::Format>(depth_format);
+	get_gui().prepare(pipeline_cache, color_format, depth_fmt, shader_stages);
+#else
 	get_gui().set_subpass(2);
-	get_gui().prepare(pipeline_cache, render_pass,
-	                  {load_shader("uioverlay/uioverlay.vert.spv", VK_SHADER_STAGE_VERTEX_BIT),
-	                   load_shader("uioverlay/uioverlay.frag.spv", VK_SHADER_STAGE_FRAGMENT_BIT)});
+	get_gui().prepare(pipeline_cache, render_pass, shader_stages);
 #endif
```
asuessenbach: This function is almost identical to `ApiVulkanSample::prepare_gui`, you've introduced with this PR as well.
Is that intended?

asuessenbach (commented):
There are a couple of mix-ups with C- and C++-binding types. I tried to clarify them in the comments below.
Besides that, could you please adjust the same stuff you're changing in ApiVulkanSample in HPPApiVulkanSample?

framework/gui.h
ellioman: I've split the common work into two helpers: `prepare_descriptors()` for all the descriptor setup, and `create_gui_pipeline()` for the pipeline state and creation. The two `prepare_impl` functions are now just a few lines each, they call the helpers and only differ in how they configure the render pass vs dynamic rendering fields.

ellioman:
@SaschaWillems 
> LTGM, but can you also add the UI to the dynamic_rendering sample? That way both dynamic rendering samples do display a user interface.

Done, I've added the UI to that sample and I also added a screenshot to the description for this PR.

framework/gui.h
ellioman: You're right, I missed that. 

Added `using ImageViewType = ...` and updated the `draw` signature accordingly.

framework/api_vulkan_sample.cpp
ellioman: Done, same approach as with ImageViewType. 
Added `using FormatType = ...` and updated the calls.

framework/gui.h
ellioman: Fixed.

framework/gui.h
ellioman: Makes sense, I've moved the rendering pass logic into a new `draw_impl` overload so the public `draw` just converts types, same as the other overloads do.

framework/gui.h
ellioman: Fixed.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
ellioman: Wasn't intentional, no. The only thing this sample needs differently is `set_subpass(2)` for the render pass path, so I've trimmed it down to just that and a call to `ApiVulkanSample::prepare_gui()`.

framework/api_vulkan_sample.h
ellioman: The reason I went with this is that the `dynamic_rendering_local_read` sample already explicitly sets `render_pass = VK_NULL_HANDLE` in [its constructor, with the comment](https://github.com/ellioman/Vulkan-Samples/blob/be52510029d2fa81aa43ef3b93cdc989573f57dc/samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp#L46-L47)
```
// Dynamic rendering doesn't use render passes
// To make sure that framework related classes like the user interface are aware of this, we explicitly st the base class' renderpass to a null handle
render_pass = VK_NULL_HANDLE;
```


So samples that use dynamic rendering skip render pass creation entirely and the handle stays null. 
`uses_dynamic_rendering()` just gives that existing convention a readable name instead of having raw null checks in multiple places. 

Happy to change the approach though if you have something better in mind, or I can remove it if you'd prefer.

ellioman:
@asuessenbach 
> There are a couple of mix-ups with C- and C++-binding types. I tried to clarify them in the comments below. Besides that, could you please adjust the same stuff you're changing in ApiVulkanSample in HPPApiVulkanSample?

Thank you for the review. 
First time making a PR for these samples and clearly many things that I wasn't aware of.

I've hopefully addressed the issues you found, including adding the changes to `HPPApiVulkanSample`, so when you have a moment, please take another look.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
asuessenbach: Unfortunately, this does not work when `USE_DYNAMIC_RENDERING` is not defined: `get_gui()` gives you a nullptr.
And calling `ApiVulkanSample::prepare_gui()` before `get_gui().set_subpass(2)` works, for some unknown reason, but setting the subpass is essentially a nop: it sets the subpass, but the gui pipeline has already been created with subpass == 0.

That is, it seems, you actually need the complete prepare_gui here, just as you had it before. Maybe with some additional comment why it's done the way it's done...

Or, maybe, you could extend the framework a little, by introducing some `virtual uint32_t ApiVulkanSample::get_gui_subpass() const`, or so, with a default implementation returning 0, and override in this sample, returning 2.
You could then completely skip the `prepare_gui` here, and should remove `Gui<bindingType>::set_subpass` and the the `subpass` member in `Gui`.

gpx1000 (dismissed):
This works in Lnux.  Looks good to me.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
ellioman: Understood. Decided to do the extension idea.

Added a virtual uint32_t get_gui_subpass() const (defaulting to 0) in ApiVulkanSample and HPPApiVulkanSample, and threaded the subpass value as a parameter through Gui::prepare(), so I dropped set_subpass() and the subpass member. DynamicRenderingLocalRead now just overrides the getter to return 2 (and ColorWriteEnable to return 1), so neither sample needs its own prepare_gui() anymore.

Did some local testing:
- DynamicRendering (Dynamic & Render Pass) on Windows and Android
- DynamicRenderingLocalRead (Dynamic & Render Pass) on Windows and Android
- ColorWriteEnable on Windows

No errors and GUI rendered correctly.

asuessenbach: Looks great!

asuessenbach (approved):

ellioman:
@gpx1000 Thanks for reviewing the PR.
I had to do some minor changes as part of other PR review comments. 
When you have a moment, please take another look :)

SaschaWillems (approved):
Works great. Thank you very much for this 👍🏻

gpx1000 (approved):

marty-johnson59:
3 approvals - merging

PR #1491, opened 2026-02-25
Limit number of columns in framework to 120

asuessenbach:
## Description

Every now and then, I'm surprised about the length of some code lines. I would suggest to limit that to, say, 120.
The only manual change in this PR is in `.clang-format`, setting `ColumnLimit : 120`. Then all *.cpp and *.h files are formatted accordingly.

In addition to the column limit I changed the following option:
`BinPackArguments: false`
`BinPackParameters: false`

Restricted to files in framework/common, for now.

Build tested on Win11 with VS2022. Run tested on Win11 with NVidia GPU.

## General Checklist:

- [ ] My code follows the [coding style](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Code-Style)
- [ ] I have reviewed file [licenses](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/CONTRIBUTING.adoc#Copyright-Notice-and-License-Template)

## Sample Checklist


gpx1000:
It looks like we're gonna need to fix all the instances where the columns are long or limit it from CI being the cause of failing.  But I'm all for the idea!

asuessenbach:
... and it seems, I'm using a different clang-format version (21.1.0) than the CI. Is version to use specified, somewhere?

SaschaWillems:
clang-format version is listed [here](https://github.com/KhronosGroup/Vulkan-Samples/blob/19a5668f115ee5decae73e7784ecbf63f8b2d033/CONTRIBUTING.adoc#code-style). Not sure if that's up-to-date though.

SaschaWillems:
Looking at the numerous changes: Increasing the column limit might make sense, but some of the changes make code harder to read. Not sure if this is something we want to apply across the board to all samples.

asuessenbach:
> Increasing the column limit might make sense

You mean, even more than 160 columns?

> some of the changes make code harder to read.

Could you point me at some of those places? Maybe with some additional clang-format settings, it looks better, again.

> Not sure if this is something we want to apply across the board to all samples.

I think, it should either be done for the complete repository, or not at all.

asuessenbach:
> clang-format version is listed [here](https://github.com/KhronosGroup/Vulkan-Samples/blob/19a5668f115ee5decae73e7784ecbf63f8b2d033/CONTRIBUTING.adoc#code-style). Not sure if that's up-to-date though.clang-format version is listed [here](https://github.com/KhronosGroup/Vulkan-Samples/blob/19a5668f115ee5decae73e7784ecbf63f8b2d033/CONTRIBUTING.adoc#code-style). Not sure if that's up-to-date though.

It's up-to-date. In check.yml, you'll find
`    container: ghcr.io/khronosgroupactions/clang-tools:15.0.0
`

asuessenbach:
Well, in fact it's clang-format 15.0.6, that is used. And for some unknown reasons, it required a few manual adjustments.
But at least, the CI accepts the formatting now.

But I definitely won't crank up the copyright of all those formatted files.

app/plugins/data_path/data_path.cpp
```diff
@@ -23,13 +23,9 @@
 namespace plugins
 {
 DataPath::DataPath() :
-    DataPathTags("Data Path Override",
-                 "Specify the folder containing the sample data folders.",
-                 {vkb::Hook::OnAppStart},
-                 {},
+    DataPathTags("Data Path Override", "Specify the folder containing the sample data folders.", {vkb::Hook::OnAppStart}, {},
```
SaschaWillems: I'm not really sure if this improves readability. I actually prefer the way it was written before this change.

SaschaWillems: Any way to fix this? There are several spots that have had their wrapping changed in a way that at least IMO hurts code readabilit.

app/plugins/batch_mode/batch_mode.cpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright (c) 2020-2025, Arm Limited and Contributors
+/* Copyright (c) 2020-2026, Arm Limited and Contributors
  * Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
```
asuessenbach, line 2: Apparently, neither the copyright.py fixes such Copyright strings, nor the copyright check of the CI catches them?

app/plugins/data_path/data_path.cpp
asuessenbach: With `BinPackParameters: false`, that looks better again.

asuessenbach:
Well, no matter what clang-format version (15.0.x, 0 <= x <= 7 ), win64, I tried, there are four files that needs some manual adjustments to keep the format check happy.
Could someone extent the `check.yml` in a way that the actually used clang-format version is displayed? I don't even know where the clang-format.exe lives in that context.

app/plugins/data_path/data_path.cpp
tomek-brcm: I think the original was easier to read too. When we enter the realm of automatic line wrapping we've gone wrong somewhere. I'd be tempted to be much more conservative in terms of line width as seeing 2 versions side-by-side is often handy.

asuessenbach: > I think the original was easier to read too.

Do you refer to the original of this specific code part? I think, they're pretty close, now.
Or do you have other part is mind? If so, could you hint me to some of them? Maybe I can improve the readability there as well.

> When we enter the realm of automatic line wrapping we've gone wrong somewhere.

I don't understand what you mean with that.

> I'd be tempted to be much more conservative in terms of line width as seeing 2 versions side-by-side is often handy.

Do I read that like "It would be better to reduce the column limit to, say, 120"?

tomek-brcm: I was referring to this specific bit.

In general when line ends are added by hand or by a code formatting tool, the code is easier to read, compared to auto-wrapping done by the viewer or horizontal scrolling to see the rest of it.

I think 120 is about right. Less than this will likely result in a massive change to the existing codebase, more will be hard to work with on a laptop screen.

asuessenbach: > I was referring to this specific bit.

ok... and have you seen it with the latest adjustment from last week? Do you think, it's ok that way?

> I think 120 is about right.

If there's more than just your request, I can easily cut column length down to 120.

framework/core/pipeline.cpp
```diff
@@ -251,24 +239,18 @@ GraphicsPipeline::GraphicsPipeline(vkb::core::DeviceC &device,
...
-	    VK_DYNAMIC_STATE_SCISSOR,
-	    VK_DYNAMIC_STATE_LINE_WIDTH,
-	    VK_DYNAMIC_STATE_DEPTH_BIAS,
-	    VK_DYNAMIC_STATE_BLEND_CONSTANTS,
-	    VK_DYNAMIC_STATE_DEPTH_BOUNDS,
-	    VK_DYNAMIC_STATE_STENCIL_COMPARE_MASK,
-	    VK_DYNAMIC_STATE_STENCIL_WRITE_MASK,
+	    VK_DYNAMIC_STATE_VIEWPORT,          VK_DYNAMIC_STATE_SCISSOR,      VK_DYNAMIC_STATE_LINE_WIDTH,           VK_DYNAMIC_STATE_DEPTH_BIAS,
+	    VK_DYNAMIC_STATE_BLEND_CONSTANTS,   VK_DYNAMIC_STATE_DEPTH_BOUNDS, VK_DYNAMIC_STATE_STENCIL_COMPARE_MASK, VK_DYNAMIC_STATE_STENCIL_WRITE_MASK,
 	    VK_DYNAMIC_STATE_STENCIL_REFERENCE,
```
tomek-brcm, line 272: nit: this was a bit more future-proof, it's easier to see additions and deletions with one state per line.

tomek-brcm (commented):
The updated pull request reads easier than the old one. I think 180 columns is too much, especially for a side-by-side diff I'm looking at right now. I don't want to impose my personal preferences on anybody here, but I'd give 120 columns a try

framework/core/pipeline.cpp
asuessenbach: Added `BinPackArguments: false` to resolve that.

asuessenbach:
I can't get what difference there is between the clang-format I'm using and the clang-format of the CI.
And I can't get what the CI is trying to tell me about the differences.
If anybody can actually decipher the CI messages on the failing formatting step, I'd be happy to manually adjust those (hopefully just a couple of) files.

Otherwise, I'm essentially out on this PR and would have to just close it.

gpx1000:
https://github.com/KhronosGroup/Vulkan-Samples/actions/runs/25072289035/artifacts/6692580894  <-- I setup CI to upload a diff of what it found.

Also, see if the script helps: scripts/clang-format.py <branch>

I'll take a real look when I get a chance and try to help.

asuessenbach:
> Also, see if the script helps: scripts/clang-format.py

Well, that python script changes a lot of files. But the CI is still failing.

> I setup CI to upload a diff of what it found.

Sure. But how would I get what has to be changed out of that blob of characters?

SaschaWillems:
>  I think 180 columns is too much

I second this. What I do very often is having the .cpp and .h side-by-side with the solution explorer visible. With my text zoom (~120%) on WQHD that leaves every source pane with ~100 columns. Increasing the current limit to e.g. 160 (or even more) would make things worse for me.

Personally I'm fine with things are right now, at least for those samples that don't make excessive use of C++ and are closer to C. They're fine to read. The actual issue is when modern C++ features are used, which make code very hard to read in general. Maybe we should fix that separate?

asuessenbach:
> I think 180 columns is too much

I never asked for 180 columns, I tried with 160 here. But it's just a number, we can easily cut it down to 120, or 100, or even 80.

asuessenbach:
Rebased, still failing the format check.

asuessenbach:
Reduced to framework, still failing the format check.

gpx1000:
[quality_fixes.patch](https://github.com/user-attachments/files/29508988/quality_fixes.patch)
Try this patch out.

asuessenbach:
Unfortunately, that patch doesn't fit. On what state is it based on?
All I get when I try is
```
git.exe am --3way --ignore-space-change --keep-cr -- "C:/Users/asuessenbach/Downloads/quality_fixes.patch"
Patch format detection failed.

Fail
```

gpx1000:
Oh, sorry I did that yesterday.  Lemme generate a new one.

gpx1000:
[quality_fixes.patch](https://github.com/user-attachments/files/29513886/quality_fixes.patch)
Try this one

asuessenbach:
Unfortunateyl, got the same `Patch format detection failed`.

As most (?) of the problems seem to be with comments, I tried with `ReflowComments : false` now.

gpx1000:
[quality_fixes.patch](https://github.com/user-attachments/files/29518476/quality_fixes.patch)
This patch version should be in git am format.  Sorry, I was using a different format.

asuessenbach:
@gpx1000 I could apply your patch, but the clang-format check is still failing.

SaschaWillems:
Looking at the CI, that patch oddly made it worse. I also don't think that the error is in comments, but rather how function result types are align.

This seems to be the culprit in device.h (commit before the patch was applied):

```cpp
	vkb::core::HPPQueue const &get_queue_by_flags_impl(vk::QueueFlags queue_flags, uint32_t queue_index) const;
	void init(std::unordered_map<const char *, bool> const &requested_extensions,
	          std::function<void(vkb::core::PhysicalDevice<bindingType> &)> request_gpu_features);
```

Manually running the clang_format.py script, this is adjusted as follows for me:

```cpp
vkb::core::HPPQueue const &get_queue_by_flags_impl(vk::QueueFlags queue_flags, uint32_t queue_index) const;
void                       init(std::unordered_map<const char *, bool> const                 &requested_extensions,
	                                std::function<void(vkb::core::PhysicalDevice<bindingType> &)> request_gpu_features);
```

Which seems proper .

Sadly can't test CI easily, so this is more of a guess.

So could you revert the last patch, and run the clang_format.p script in the repo like this:

```
python .\scripts\clang_format.py main
```

Hope that'll help fix the issue.

asuessenbach:
Ok, reformatted again, without the patch from @gpx1000, and run the clang_format.py script. Still failing.

The first point I could identify to be an issue is in framework/api_vulkan_sample.cpp, in the function `ApiVulkanSample::setup_depth_stencil`.
Before my change, it looks like
```
	image_create_info.sType       = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO;
	image_create_info.imageType   = VK_IMAGE_TYPE_2D;
	image_create_info.format      = depth_format;
	image_create_info.extent      = {get_render_context().get_surface_extent().width, get_render_context().get_surface_extent().height, 1};
	image_create_info.mipLevels   = 1;
	image_create_info.arrayLayers = 1;
```
With my change:
```
	image_create_info.sType     = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO;
	image_create_info.imageType = VK_IMAGE_TYPE_2D;
	image_create_info.format    = depth_format;
	image_create_info.extent    = {
	    get_render_context().get_surface_extent().width, get_render_context().get_surface_extent().height, 1};
	image_create_info.mipLevels   = 1;
	image_create_info.arrayLayers = 1;
```
But I can't tell from the error message how it's supposed to look like.

asuessenbach:
By the way, it seems we have a lot of lines with more than 120 columns but less than 140.
In other words, limiting the number of columns to 140 would have a much smaller impact.

SaschaWillems:
This all looks like we're now trying to enfore clan-format to source files that were added before we enabled clang-format in the repo. Makes me wonder if it's actually possible to fix these issues and if so, how we can avoid contributors running into those issues?

What good is using clang-format, if the formatting produced by the local script that's running as e.g. a pre-commit hook doesn't match what CI/CD expects? 

Shouldn't we start by trying to get those two inline frist?

gpx1000:
Clang format is a really old version in CI which doesn't match the one on the local system which explains why the script doesn't fix it.  I'll try a pr that updates clang format

asuessenbach:
For easier human digesting, I reduced the scope of this PR even more to just `framework/common`.

SaschaWillems (dismissed):

gpx1000 (dismissed):

SaschaWillems (approved):

tomek-brcm (approved):

marty-johnson59:
Merging per agreement on Monday's Samples call

PR #1493, opened 2026-02-27
Add new sample for VK_EXT_descriptor_heap

### LINKED ISSUE #1486: Add new sample for descriptor heap
Just to make sure people are aware: I'm working on a new sample for the recent descriptor heap extension.

SaschaWillems:
## Description

This PR adds a new sample for the recently released VK_EXT_descriptor_heap extension. It shows off different heap types (resources, samplers). It only does the descriptor set and binding mapping way of using heaps. Untyped pointers might follow at a later point of certain things have been resolved (see our internal discussions).

Also comes with documentation/tutorial material.

Tested on Win11 with an Nvidia RTX 4070.

Fixes #1486 

## General Checklist:

## Sample Checklist


SaschaWillems:
Please read:

* https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_descriptor_heap.html
* https://docs.vulkan.org/guide/latest/descriptor_heap.html
* https://www.khronos.org/blog/vulkan-introduces-roadmap-2026-and-new-descriptor-heap-extension

There is nothing misleading about this example.

This PR is also a **DRAFT**.

SaschaWillems:
This PR is now ready to review. I also added documentation/tutorial.

A variation using untyped pointers will follow independently at a later pointer.

samples/extensions/descriptor_heap/descriptor_heap.cpp
```diff
@@ -0,0 +1,575 @@
...
+		uniform_data.model_matrix[i] = cubeMat;
+	}
+
+	uniform_buffers[current_buffer]->convert_and_update(uniform_data);
+}
+
+void DescriptorHeap::create_descriptor_heaps()
+{
+	// Descriptor heaps have varying offset, size and alignment requirements, so we store it's properties for later user
+	descriptor_heap_properties.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DESCRIPTOR_HEAP_PROPERTIES_EXT;
```
asuessenbach, line 160: Maybe initialize `descriptor_heap_properties.sType` in the header, like 	`VkPhysicalDeviceDescriptorHeapPropertiesEXT      descriptor_heap_properties{.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DESCRIPTOR_HEAP_PROPERTIES_EXT};`?

samples/extensions/descriptor_heap/descriptor_heap.cpp
```diff
@@ -0,0 +1,575 @@
...
+	// There are two descriptor heap types: One that can store resources (buffers, images) and one that can store samplers
+	// We create heaps with a fixed size that's guaranteed to fit in the few descriptors we use
+	const VkDeviceSize heap_buffer_size = aligned_size(2048 + descriptor_heap_properties.minResourceHeapReservedRange, descriptor_heap_properties.resourceHeapAlignment);
+
+	descriptor_heap_resources = std::make_unique<vkb::core::BufferC>(get_device(),
+	                                                                 heap_buffer_size,
+	                                                                 VK_BUFFER_USAGE_DESCRIPTOR_HEAP_BIT_EXT | VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT,
+	                                                                 VMA_MEMORY_USAGE_CPU_TO_GPU);
+
+	const VkDeviceSize heap_sampler_size = aligned_size(2048 + descriptor_heap_properties.minSamplerHeapReservedRange, descriptor_heap_properties.samplerHeapAlignment);
```
asuessenbach, line 175: Magic `2048`?

Maybe, with some code reordering, use `sampler_create_infos.size() * sampler_descriptor_size` instead?

samples/extensions/descriptor_heap/descriptor_heap.cpp
```diff
@@ -0,0 +1,575 @@
...
+	// Descriptor heaps have varying offset, size and alignment requirements, so we store it's properties for later user
+	descriptor_heap_properties.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DESCRIPTOR_HEAP_PROPERTIES_EXT;
+	VkPhysicalDeviceProperties2 device_props_2{
+	    .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2,
+	    .pNext = &descriptor_heap_properties};
+	vkGetPhysicalDeviceProperties2(get_device().get_gpu().get_handle(), &device_props_2);
+
+	// There are two descriptor heap types: One that can store resources (buffers, images) and one that can store samplers
+	// We create heaps with a fixed size that's guaranteed to fit in the few descriptors we use
+	const VkDeviceSize heap_buffer_size = aligned_size(2048 + descriptor_heap_properties.minResourceHeapReservedRange, descriptor_heap_properties.resourceHeapAlignment);
```
asuessenbach, line 168: Magic `2048`?

Maybe, with some code reordering, use `image_heap_offset + cubes.size() * image_descriptor_size` instead?

samples/extensions/descriptor_heap/descriptor_heap.cpp
```diff
@@ -0,0 +1,575 @@
...
+	        .addressModeV  = VK_SAMPLER_ADDRESS_MODE_REPEAT,
+	        .addressModeW  = VK_SAMPLER_ADDRESS_MODE_REPEAT,
+	        .maxAnisotropy = 16.0f,
+	        .maxLod        = (float) cubes[1].texture.image.get()->get_mipmaps().size(),
+	    }};
+
+	for (auto i = 0; i < static_cast<uint32_t>(sampler_create_infos.size()); i++)
+	{
+		host_address_ranges_samplers[i] = {
+		    .address = (uint8_t *) (descriptor_heap_samplers->get_data()) + sampler_heap_offset + sampler_descriptor_size * i,
```
asuessenbach, line 215: We should not use a C-cast to cast the const away!

Either use `const_cast` or introduce a new function to `vkb::allocated::Allocated`, that gets you a non-const pointer.
I would prefer the latter.

samples/extensions/descriptor_heap/descriptor_heap.cpp
```diff
@@ -0,0 +1,575 @@
...
+	for (auto i = 0; i < uniform_buffers.size(); i++)
+	{
+		device_address_ranges_uniform_buffer[i]        = {.address = uniform_buffers[i]->get_device_address(), .size = uniform_buffers[i]->get_size()};
+		resource_descriptor_infos[resource_heap_index] = {
+		    .sType = VK_STRUCTURE_TYPE_RESOURCE_DESCRIPTOR_INFO_EXT,
+		    .type  = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER,
+		    .data  = {
+		         .pAddressRange = &device_address_ranges_uniform_buffer[i]}};
+		host_address_ranges_resources[resource_heap_index] = {
+		    .address = (uint8_t *) (descriptor_heap_resources->get_data()) + buffer_descriptor_size * i,
```
asuessenbach, line 244: As above.

samples/extensions/descriptor_heap/descriptor_heap.cpp
```diff
@@ -0,0 +1,575 @@
...
+		    .size    = buffer_descriptor_size};
+
+		resource_heap_index++;
+	}
+
+	// Images
+	std::array<VkImageViewCreateInfo, 2>    image_view_create_infos{};
+	std::array<VkImageDescriptorInfoEXT, 2> image_descriptor_infos{};
```
asuessenbach, line 252: There are supposed to be as many `VkImageViewCreateInfo`s and `VkImageDescriptorInfoEXT`s as there are `cubes`?
Maybe introduce some `static const uint32_t CUBES_COUNT = 2`, or so, and use that for all three `std::array`s.
Or use `std::tuple_size<decltype(cubes)>::value` as the size for the two arrays here.
Or just change them to be `std::vector`s, sized by `cubes.size()`.

samples/extensions/descriptor_heap/descriptor_heap.cpp
```diff
@@ -0,0 +1,575 @@
...
+	camera.type = vkb::CameraType::LookAt;
+	camera.set_position({0.f, 0.f, -4.f});
+	camera.set_rotation({0.f, 180.f, 0.f});
+	camera.set_perspective(60.f, static_cast<float>(width) / static_cast<float>(height), 256.f, 0.1f);
+
+	load_assets();
+	prepare_uniform_buffers();
+	create_descriptor_heaps();
+	create_pipeline();
+	build_command_buffers();
```
asuessenbach, line 63: No need to call `build_command_buffers` when you don't prebuild them at all.

samples/extensions/descriptor_heap/descriptor_heap.cpp
```diff
@@ -0,0 +1,575 @@
...
+	VK_CHECK(vkQueueSubmit(queue, 1, &submit_info, VK_NULL_HANDLE));
+	ApiVulkanSample::submit_frame();
+}
+
+void DescriptorHeap::build_command_buffers()
+{
+	// This sample doesn't use prebuilt command buffers
+}
+
+void DescriptorHeap::build_command_buffer()
```
asuessenbach, line 422: `build_command_buffers` and `build_command_buffer` are visually pretty close.
Maybe name it `build_command_buffer_dynamic`, or so?

asuessenbach (commented):
Looks great!

Just a couple of mostly minor issues:

samples/extensions/descriptor_heap/descriptor_heap.cpp
SaschaWillems: I'd like to keep that name in preparation for #1525

asuessenbach: > I'd like to keep that name in preparation for https://github.com/KhronosGroup/Vulkan-Samples/issues/1525

I don't see that connection. Could you elaborate a bit?

asuessenbach (commented):
I think, we should not use `const_cast` in our samples. It's dangerous and not a good coding style.

But if nobody else complains, I'm ok with it.

SaschaWillems:
> I think, we should not use `const_cast` in our samples. It's dangerous and not a good coding style.

As discussed, I added a non-const function in the allocated class and removed all const casts from the sample.

framework/core/allocated.h
```diff
@@ -176,6 +176,14 @@ class Allocated : public vkb::core::VulkanResource<bindingType, HandleType>
 	 * @note This performs no checking that the memory is actually mapped, so it's possible to get a nullptr
 	 */
 	const uint8_t *get_data() const;
+
+	/**
+	 * @brief Retrieves a pointer to the host visible memory as an unsigned byte array.
+	 * @return The pointer to the host visible memory.
+	 * @note Non-const. This performs no checking that the memory is actually mapped, so it's possible to get a nullptr
+	 */
+	uint8_t *get_mapped_data() const;
```
asuessenbach, line 185: A function returning a non-const pointer should not be `const`.

Moreover, it's hard to see the difference to `get_data()`, just by looking at the name.
Maybe you could change `get_data()` to `uint8_t const * get_const_data() const;` and change the new function to `uint8_t * get_data();`?

SaschaWillems: If okay, I'd do that after this PR has been merged. Don't want to defer this any further.

samples/extensions/descriptor_heap/README.adoc
```diff
@@ -0,0 +1,303 @@
...
+vkCmdPushDataEXT(draw_cmd_buffer, &push_data_info);
+----
+
+== Accessing heaps
+
+The extension lets us access descriptors in shaders in two different ways:
+
+=== Descriptor set and binding mappings
+
+link:https://docs.vulkan.org/refpages/latest/refpages/source/VkDescriptorSetAndBindingMappingEXT.html[VkDescriptorSetAndBindingMappingEXT] can be used to simplify porting existing codebases to descriptor heaps. By using this, existing shaders can be used without having to modify them. The syntax for binding descriptors, regardless of shading language, stays the same.
```
iagoCL, line 206: I would add a  small note:

We are only showing `VK_DESCRIPTOR_MAPPING_SOURCE_HEAP_WITH_CONSTANT_OFFSET_EXT` but there are more binding-mapping options. Developers should check https://docs.vulkan.org/guide/latest/descriptor_heap.html#_mapping_the_heap_to_exisiting_shaders and consider the best strategy for their use case.

I do not want to delay this so I am fine with merging it as it is.

iagoCL (dismissed):
I think the PR is good.

I could only test this on Nvidia hardware, but the code looks good and is a nice overview of the basic parts of how to use descriptor heap. 

Happy for this to be merged.

shaders/descriptor_heap/glsl/cube.frag
```diff
@@ -0,0 +1,35 @@
...
+layout(push_constant) uniform PushConsts {
+	int samplerIndex;
+	int frameIndex;
+} pushConsts;
+
+layout (location = 0) out vec4 outFragColor;
+
+void main() 
+{
+	outFragColor = texture(sampler2D(textureImage[inInstanceIndex], textureSampler[pushConsts.samplerIndex]), inUV);
```
Guang-035: Hello @SaschaWillems , I think `inInstanceIndex` should be marked as non-uniform here.

This shader uses regular set/binding declarations with the mapping API, and does not declare `GL_EXT_descriptor_heap` / `DescriptorHeapEXT`, so I think the usual descriptor indexing rules still apply (see [Issues 4803#note_606359](https://gitlab.khronos.org/vulkan/vulkan/-/issues/4803#note_606359)). Since `inInstanceIndex` can vary between instances in the same draw, `textureImage[inInstanceIndex]` should use `nonuniformEXT`.

Please correct me if I am missing anything.

SaschaWillems: Fixed (GLSL, HLSL and Slang). Thx for noticing!

samples/extensions/descriptor_heap/README.adoc
SaschaWillems: That's a good point. Had to touch this branch anyway (to fix CI), so quickly added a note similar to what you proposed.

samples/extensions/descriptor_heap/descriptor_heap.cpp
```diff
@@ -0,0 +1,576 @@
...
+}
+
+void DescriptorHeap::create_command_pool()
+{
+	VkCommandPoolCreateInfo command_pool_info{
+	    .sType            = VK_STRUCTURE_TYPE_COMMAND_POOL_CREATE_INFO,
+	    .flags            = VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT,
+	    .queueFamilyIndex = get_device().get_queue_by_flags(VK_QUEUE_GRAPHICS_BIT | VK_QUEUE_COMPUTE_BIT, 0).get_family_index()};
+	VK_CHECK(vkCreateCommandPool(get_device().get_handle(), &command_pool_info, nullptr, &cmd_pool));
+}
```
gpx1000, line 571: This looks the same as the overridden function.  Do we still need to override it?

SaschaWillems: The one in the [base class](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/framework/api_vulkan_sample.cpp#L668) does not set the VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT flag. That's why I had to do this override.

This sample, unlike older ones, doesn't use pre-built command buffers but rather recreates them each frame, hence why I require this flag.

gpx1000: Ahh, okay I missed that.  LGTM.

gpx1000 (approved):

PR #1522, opened 2026-04-19
Set VK_LAYER_KHRONOS_validation to Optional in case layer not available

### LINKED ISSUE #1523: Some samples crash on cleanup if VVL not available in Debug mode
The samples _texture_loading, texture_mipmap_generation, hpp_pipeline_cache, pipeline_cache_  crash during cleanup if the VVL is not available when running in Debug mode.  This can happen in a few cases: a) on Windows with no SDK installed, and b) on the iOS Simulator where the VVL is not built for that target.

The root cause of the crash is that the VVL is marked as Required for Debug mode vs. Optional, and because of this the Vulkan device is never defined.  The above samples should check if the Vulkan device is defined before running their cleanup code.

I will submit a PR that addresses both issues:  a) VVL marked as Optional, and b) fixes for the cleanup code.

SRSaunders:
## Description

This PR fixes a few small issues found when testing on the iOS Simulator, and on Windows 11 with no SDK installed:

1. Changed `VK_LAYER_KHRONOS_validation` to be Optional for debug builds vs. Required. This applies to all platforms but allows debug builds to run even when the validation layer is not available. This is important for several use cases: a) Windows users who do not install the SDK, and b) on the iOS Simulator where the VVL library is not built or available for that target. I debated this change before making it, but I believe it is correct given that Windows build docs do not mention or require installation of the SDK, and the default build config for Visual Studio is debug. I am interested in feedback from maintainers on this regarding project intentions: i.e. Required forces a runtime error that prevents samples from running in debug mode for the above scenarios, or Optional which allows execution with a runtime warning message instead. I believe the latter is more appropriate.
2. Updated some samples (_texture_loading_, _texture_mipmap_generation_, _hpp_pipeline_cache_, _pipeline_cache_) to guard against an undefined device on exit. These samples assume the device is defined and attempt cleanup without checking first, which can result in a crash on exit when the sample does not run successfully. I found this in testing item 1 above, i.e. when running in debug mode with `VK_LAYER_KHRONOS_validation` set to the original Required.
3. Added MoltenVK configuration code to allow _dynamic_rendering_local_read_ to render properly on the iOS Simulator.  This is identical code to a few other samples that already use this iOS-specific #ifdef config block for correct rendering on the iOS Simulator: _layout_transitions_, _pipeline_barriers_, and _subpasses_.
4. Updated the gui pipeline creation code to use `VK_FORMAT_UNDEFINED` for the depth format when using dynamic rendering.  While Vulkan technically allows the pipeline to be created with a defined depth format, even while rendering uses a null depth attachment, this is not necessarily the best practice.  Looking at docs and recommendations for dynamic rendering, the both the pipeline and rendering steps _should_ indidate that no depth attachment will be used in cases like gui overlay rendering as implemented in this project.  Note the gui pipeline already specifies `depthTestEnable = false` and `depthWriteEnable = false` so I felt it was appropriate to also set `depthAttachmentFormat  = VK_FORMAT_UNDEFINED`.  This has the added benefit of allowed the VVL to identify a mismatch case if the gui ever erroneously specifies a non-null depth attachment during dynamic rendering for the gui.  Note I found this issue not from the VVL, but because Metal validation picks it up.

Fixes #1523.

Tested on Windows 11, macOS Sequoia, iOS 26, and iOS Simulator running iOS 26.

## General Checklist:

## Sample Checklist


samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
```diff
@@ -772,6 +772,24 @@ void DynamicRenderingLocalRead::prepare_pipelines()
...
 
+#if defined(PLATFORM__MACOS) && TARGET_OS_IOS && TARGET_OS_SIMULATOR
+void DynamicRenderingLocalRead::request_instance_extensions(std::unordered_map<std::string, vkb::RequestMode> &requested_extensions) const
+{
+	// On iOS Simulator use layer setting to disable MoltenVK's Metal argument buffers - otherwise incorrect rendering for this sample
+	vkb::VulkanSampleC::request_instance_extensions(requested_extensions);
+	requested_extensions[VK_EXT_LAYER_SETTINGS_EXTENSION_NAME] = vkb::RequestMode::Optional;
+}
+
+void DynamicRenderingLocalRead::request_layer_settings(std::vector<VkLayerSettingEXT> &requested_layer_settings) const
```
asuessenbach, line 783: With #1507, `request_layer_settings` has changed.
Needs to be adjusted accordingly.

framework/vulkan_sample.h
```diff
@@ -1442,7 +1442,8 @@ template <vkb::BindingType bindingType>
 inline void VulkanSample<bindingType>::request_layers(std::unordered_map<std::string, vkb::RequestMode> &requested_layers) const
 {
 #if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
-	requested_layers["VK_LAYER_KHRONOS_validation"] = vkb::RequestMode::Required;
+	// VK_LAYER_KHRONOS_validation must be optional in case layer not available for debug builds (e.g. running with no SDK installed or on iOS Simulator)
+	requested_layers["VK_LAYER_KHRONOS_validation"] = vkb::RequestMode::Optional;
```
tomek-brcm, line 1477: This is misleading to me. I'd expect validation layers to be present when VKB_VALIDATION_LAYERS is on. A build error when validation layers can't be found seems the right answer. The VKB_DEBUG can have optional validation layers. I'd prefer required for VKB_VALIDATION_LAYERS and optional for VKB_DEBUG.

samples/extensions/dynamic_rendering_local_read/dynamic_rendering_local_read.cpp
SRSaunders: Thanks @asuessenbach for the heads up on this.  I also fixed a small typo in your change (vkb vs vbk) that would break iOS Simulator builds.  I appreciate you updating the other samples with this #ifdef code.

framework/vulkan_sample.h
SRSaunders: @tomek-brcm you have a good point here.  I have made a change to accommodate your feedback.  Note none of these settings can generate a build error.  However, there will now be a runtime error if `VKB_VALIDATION_LAYERS` is enabled with no validation layer available.

asuessenbach (approved):

tomek-brcm (approved):

gpx1000 (approved):

PR #1526, opened 2026-04-23
Fix usage of vkCreateDebugUtilsMessengerEXT/vk::Instance::createDebugUtilsMessengerEXT.

asuessenbach:
## Description

Using `VkDebugUtilsMessengerCreateInfoEXT` needs special consideration, depending on where it is used.
- `VkDebugUtilsMessengerCreateInfoEXT` is specified to extend `VkInstanceCreateInfo`, and multiple instances of it in that extension chain are allowed. That is, with multiple `VkDebugUtilsMessengerCreateInfoEXT`s in that chain, multiple callbacks are created on instance creation, but just for a very short period of time.
- `VkDebugUtilsMessengerCreateInfoEXT` is not specified to extend itself! That is, with multiple instances of it in the pNext-chain for vkCreateDebugUtilsMessengerEXT, just the very first `VkDebugUtilsMessengerCreateInfoEXT` in that chain is used to create a callback. Actually, I would have expected that the validation layers would fire a warning with such a usage.

The sample `shader_debugprintf` used an additional `VkDebugUtilsMessengerCreateInfoEXT` for both, Instance and DebugUtilsMessenger creation. Now, a separate local DebugUtilsMessenger is created to have both callbacks, the one from the sample and the one from the framework, alive.

Build tested on Win11 with VS2022. Run tested on Win11 with NVidia GPU.

## General Checklist:

## Sample Checklist

### DIFF
--- a/framework/vulkan_sample.h
+++ b/framework/vulkan_sample.h
@@ -198,8 +198,6 @@ class VulkanSample : public vkb::Application
 	virtual void draw_renderpass(vkb::core::CommandBuffer<bindingType> &command_buffer, vkb::rendering::RenderTarget<bindingType> &render_target);
 
 	virtual void                    extend_instance_create_info(vkb::StructureChainBuilder<bindingType, InstanceCreateInfoType> &scb) const;
-	virtual void                    extend_debug_report_callback_create_info(vkb::StructureChainBuilder<bindingType, DebugReportCallbackCreateInfoType> &scb) const;
-	virtual void                    extend_debug_utils_messenger_create_info(vkb::StructureChainBuilder<bindingType, DebugUtilsMessengerCreateInfoType> &scb) const;
 	virtual uint32_t                get_api_version() const;
 	virtual InstanceCreateFlagsType get_instance_create_flags(std::vector<std::string> const &enabled_extensions) const;
 
@@ -341,8 +339,8 @@ class VulkanSample : public vkb::Application
 	static void set_viewport_and_scissor_impl(vkb::core::CommandBufferCpp const &command_buffer, vk::Extent2D const &extent);
 
 #if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
-	vk::DebugReportCallbackCreateInfoEXT get_debug_report_callback_create_info() const;
-	vk::DebugUtilsMessengerCreateInfoEXT get_debug_utils_messenger_create_info() const;
+	vk::DebugReportCallbackCreateInfoEXT const &get_debug_report_callback_create_info() const;
+	vk::DebugUtilsMessengerCreateInfoEXT const &get_debug_utils_messenger_create_info() const;
 #endif
 
 	/**
@@ -836,14 +834,6 @@ inline void VulkanSample<bindingType>::extend_instance_create_info_impl(vkb::Str
 }
 
 template <vkb::BindingType bindingType>
-inline void VulkanSample<bindingType>::extend_debug_report_callback_create_info(vkb::StructureChainBuilder<bindingType, DebugReportCallbackCreateInfoType> &scb) const
-{}
-
-template <vkb::BindingType bindingType>
-inline void VulkanSample<bindingType>::extend_debug_utils_messenger_create_info(vkb::StructureChainBuilder<bindingType, DebugUtilsMessengerCreateInfoType> &scb) const
-{}
-
-template <vkb::BindingType bindingType>
 inline uint32_t VulkanSample<bindingType>::get_api_version() const
 {
 	return VK_API_VERSION_1_1;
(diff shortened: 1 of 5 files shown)


samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -458,27 +458,38 @@ bool ShaderDebugPrintf::prepare(const vkb::ApplicationOptions &options)
...
+	ApiVulkanSample::extend_debug_utils_messenger_create_info(scb);
+
 	// Register a sample specific debug utils callback in addition to the one registered by the base class
-	static VkDebugUtilsMessengerCreateInfoEXT local_debug_utils_messenger_create_info{.sType           = VK_STRUCTURE_TYPE_DEBUG_UTILS_MESSENGER_CREATE_INFO_EXT,
-	                                                                                  .pNext           = ApiVulkanSample::get_debug_utils_messenger_create_info(),
-	                                                                                  .messageSeverity = VK_DEBUG_UTILS_MESSAGE_SEVERITY_INFO_BIT_EXT,
-	                                                                                  .messageType     = VK_DEBUG_UTILS_MESSAGE_TYPE_VALIDATION_BIT_EXT,
-	                                                                                  .pfnUserCallback = debug_utils_message_callback};
-	return &local_debug_utils_messenger_create_info;
+#if 0
```
SaschaWillems: Why is this define required?

framework/vulkan_sample.h
```diff
@@ -346,6 +346,11 @@ class VulkanSample : public vkb::Application
 	void        request_layer_settings_impl(std::vector<vk::LayerSettingEXT> &requested_layer_settings, vkb::StructureChainBuilderCpp<vk::InstanceCreateInfo> &scb) const;
 	static void set_viewport_and_scissor_impl(vkb::core::CommandBufferCpp const &command_buffer, vk::Extent2D const &extent);
 
+#if defined(VKB_DEBUG) || defined(VKB_VALIDATION_LAYERS)
```
SaschaWillems, line 341: Not sure if it's a good idea to hide these behind those defines. What about samples like debugprintf that are supposed to work without validation layers (to be used with e.g. RenderDoc)?

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
```diff
@@ -458,27 +458,38 @@ bool ShaderDebugPrintf::prepare(const vkb::ApplicationOptions &options)
...
 {
 	ApiVulkanSample::extend_instance_create_info(scb);
 
 	// Register a sample specific debug utils callback in addition to the one registered by the base class
-	VkDebugUtilsMessengerCreateInfoEXT debug_utils_messenger_create_info{.sType           = VK_STRUCTURE_TYPE_DEBUG_UTILS_MESSENGER_CREATE_INFO_EXT,
-	                                                                     .messageSeverity = VK_DEBUG_UTILS_MESSAGE_SEVERITY_INFO_BIT_EXT,
-	                                                                     .messageType     = VK_DEBUG_UTILS_MESSAGE_TYPE_VALIDATION_BIT_EXT,
-	                                                                     .pfnUserCallback = debug_utils_message_callback};
-	scb.add_struct(debug_utils_messenger_create_info);
+	scb.add_struct(get_debug_utils_messenger_create_info());
```
SaschaWillems, line 486: Not a fan of this. Samples like this are supposed to explicitly show how to set up things. Moving them to the framework and making things implicit makes it harder to follow and/or adopt such samples.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
asuessenbach: It is not. I'll remove it.

framework/vulkan_sample.h
asuessenbach: Those two functions hidden by those defines are private helper functions of the `VulkanSample` class. They are not virtual, and nobody out of this class actually need to know about them.

In contrast to that, `extend_instance_create_info` and `extend_debug_utils_messenger_create_info` are protected virtual functions in `VulkanSample`, not hidden by any defines. The `shader_debugprintf` sample, for example, overrides them and calls it own local helper function `get_debug_utils_messenger_create_info`, which is not hidden behind any defines.

samples/extensions/shader_debugprintf/shader_debugprintf.cpp
asuessenbach: > Not a fan of this.

What exactly don't you like here?
The usage of the StructureChainBuilder to ease structure chaining? Or the use of a local helper function to get the very same `VkDebugUtilsMessengerCreateInfoEXT` in two separate functions?

SaschaWillems: Kinda hard to describe, but the way that the structures are now set up feels odd and I can imagine that this will be hard to follow for people wanting to learn how to use those.

This esp.:

```cpp
VkDebugUtilsMessengerCreateInfoEXT tmp = *scb.get_struct<VkDebugUtilsMessengerCreateInfoEXT>();
scb.set_anchor_struct(get_debug_utils_messenger_create_info());
scb.add_struct(tmp);
```

Requires to reader to understand what the structure builder (behind scb) is and how it works. You can also no longer simply copy over code.

When I write Vulkan samples (this one was done by me), I try to keep abstractions to a minimum to make it easy to follow for people to follow as possible. Even if that means not using all the (smart) features that our framework has to offer or having to duplicate code.

tl;dr: I think this hurts readability of the sample and I'd prefer to keep it like it was before.

asuessenbach: You're right, that code was hard get understand. And also wrong.

I finally figured out, how `VkDebugUtilsMessengerCreateInfoEXT` should be used in that case. See the adjusted Description above.

asuessenbach:
Note: I still had to manually undo some formatting so that the “clang-format” CI step could be completed successfully.

SaschaWillems (dismissed):

tomek-brcm (dismissed):

tomek-brcm (dismissed):

tomek-brcm (dismissed):

PR #1538, opened 2026-05-02
Clean up project solution structure

SaschaWillems:
## Description

With some recent additions, the Visual Studio solution generated via CMake became cluttered, confusing people as to what's the right project to actually launch:

image

This PR cleans up that project structure by moving the recently added vgf components to the ThirdParty folder and framework related projects to the framework folder. That only leaves the samples project at the top level, making it clear that this is the project people are supposed to start from the IDE:

image

Note that this is a pure visual change and doesn't change anything about the source code structure itself and does not affect the build in any way. It simple tells CMake that, if the IDE supports it, those projects should be displayed in their respective sub folders.

## General Checklist:


asuessenbach:
Looks much better!

But I would not expect the `framework` in a `components` folder.

And while you're at it: maybe you could move all those `Shaders-<Language>` folders into some `Shaders` folder?
Like
```
Shaders
|--- GLSL
|--- HLSL
|--- SLANG
|--- SPVASM
```

SaschaWillems:
I have updated this and move a new third party library added recently to said folder. Framework is now back at top-level.

There is a CTest project that seems to come from the astc third party dependency which can't be moved, so it's still there on top-level.

Steve mentioned that IDEs other than to Visual Studio don't support multi-level structures, so moving shaders into a shaders folder wouldn't work everywhere.

From my POV this is now ready for review.

asuessenbach:
> Steve mentioned that IDEs other than to Visual Studio don't support multi-level structures, so moving shaders into a shaders folder wouldn't work everywhere.

That's surprising. How are the samples listed on such an IDE? With VS, we have 5 folders ("Api", "Extensions",...) beneath "Samples"

asuessenbach:
... and, maybe totally unrelated: why do we have all those shader projects at all? Why don't we compile the shaders on-the-fly?

SaschaWillems:
> ... and, maybe totally unrelated: why do we have all those shader projects at all? Why don't we compile the shaders on-the-fly?

That's what we did in the early days. But that would mean we'd have to ship shader compiler libraries for GLSL, Slang and DXC. Esp. the later have limited OS support and they're not available on all our target platforms. That's why we went for the compile shaders at project compilation time and provide pre-compiled SPIR-V.

SaschaWillems:
> That's surprising. How are the samples listed on such an IDE? With VS, we have 5 folders ("Api", "Extensions",...) beneath "Samples"

No idea. I think this is related to how the folders are structured, but I don't have access to other IDEs outside of VS and VSCode (where this works fine).

So maybe @gpx1000 can chime in. If this does work in other IDEs just fine, I can add it back.

tomek-brcm (approved):
Nothing is broken for us with this change.

PR #1543, opened 2026-05-15
Fix image usage for screenshot

yknishidate:
## Description

I fixed an issue where taking screenshots caused a validation error because the swapchain image usage did not include `TRANSFER_SRC`.

```
[error] -8823274 - VUID-VkImageMemoryBarrier-oldLayout-01212: vkCmdPipelineBarrier(): pImageMemoryBarriers[0].newLayout (VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL) is not compatible with VkImage 0x380000000038 usage flags VK_IMAGE_USAGE_TRANSFER_DST_BIT|VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT.
```

## General Checklist:

## Sample Checklist

### DIFF
--- a/samples/extensions/rasterization_order_attachment_access/rasterization_order_attachment_access.cpp
+++ b/samples/extensions/rasterization_order_attachment_access/rasterization_order_attachment_access.cpp
@@ -60,7 +60,7 @@ bool RasterizationOrderAttachmentAccess::prepare(const vkb::ApplicationOptions &
 	}
 
 	// Add INPUT_ATTACHMENT_BIT so fragment shaders can read the color attachment with subpassLoad()
-	update_swapchain_image_usage_flags({VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT});
+	update_swapchain_image_usage_flags({VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT, VK_IMAGE_USAGE_TRANSFER_SRC_BIT});
 
 	camera.type = vkb::CameraType::LookAt;
 	camera.set_position({0.0f, 0.0f, -4.0f});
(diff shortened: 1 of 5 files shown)


SaschaWillems:
Afaik blit source is not mandatory for (some of) the formats we use for the swapchain. So this should only add said flag if that format flag is supported.

yknishidate:
Thanks for the comment!

I agree that `TRANSFER_SRC` should not be used if it is not supported by the surface.

However, from what I understand, swapchain images include `TRANSFER_SRC` by default in the common swapchain setup:

https://github.com/KhronosGroup/Vulkan-Samples/blob/51c094308493da5908dddb0f4a2eed999c4a2b45/framework/core/swapchain.h#L97

The validation error is caused by `update_swapchain_image_usage_flags()` dropping `TRANSFER_SRC` from the default usage. My intention was to preserve the flag required by the screenshot, not to blindly add an unsupported usage flag.

Also, the support check is already performed when recreating the swapchain, so I think it should be reasonable for the sample code to request `TRANSFER_SRC` without performing its own capability check, similar to the setup process.

NOTE: As a possible follow-up improvement, it might be worth considering the requested image usage flags during swapchain format selection, instead of selecting the format first and validating the usage afterward.

asuessenbach:
Your PR nicely fixes the error desribed above.

But then I get this error (with the `shader_object` sample), initiated by `Screenshot::on_post_draw` of the screenshot plugin:
```
vkQueueSubmit(): pSubmits[0] performs a layout transition on presentable VkImage 0xaf00000000af, but the image has not been acquired from VkSwapchainKHR 0xae00000000ae (either never or since the last present operation).
The Vulkan spec states: Use of a presentable image must occur only after the image is returned by vkAcquireNextImageKHR, and before it is released by vkQueuePresentKHR. This includes transitioning the image layout and rendering commands (https://docs.vulkan.org/refpages/latest/refpages/source/VkSwapchainKHR.html#_description)
```

tomek-brcm (approved):
The shader_object is not supported for Broadcom so it works for us,

marty-johnson59:
Hi @yknishidate, can you reply to @asuessenbach' comment above.  Would like to move this forward if possible.  Thanks!

PR #1557, opened 2026-07-29
Fix GUI buffer update with deferred allocation

### LINKED ISSUE #1508: Bug: framework/gui.h/update_buffers() crashes when "explicit_update == false" because vertex_buffer/index_buffer are null
When explicit_update is false , the update_buffers() function (line 1412) can still be called. vertex_buffer and index_buffer are initially null because persistent buffers are not created. However, the current code directly accesses vertex_buffer->get_handle() and vertex_buffer->get_size() without checking if the pointers are valid, causing a null pointer dereference.
Proposed Fix
Add null pointer checks before accessing vertex_buffer and index_buffer:

``` cpp
if (!vertex_buffer || !vertex_buffer->get_handle() || (vertex_buffer_size != vertex_buffer->get_size()))
{
    // Create buffer if needed
}
```

swayam8624:
## Summary

Fixes #1508.

`Gui::update_buffers()` can be called while `explicit_update` is false. In that mode the persistent vertex and index buffers are intentionally null until the first update, but the method dereferenced both pointers before allocating them.

This change short-circuits the two conditions when the corresponding buffer is null, allowing the existing allocation and upload path to initialize both buffers.

## Reproduction

On macOS Apple Silicon (M2 Pro, macOS 26.6, Apple Clang 21, Vulkan SDK 1.4.341), an AddressSanitizer harness using the real `GuiCpp::update_buffers()` implementation and non-empty ImGui draw data reproduced a deterministic null dereference in the parent implementation:

```text
SEGV on unknown address 0x000000000028
vk::Buffer::operator!()
vkb::Gui<(vkb::BindingType)1>::update_buffers()
```

## Validation

- `cmake --build build/repro --config Debug --target framework vulkan_samples -j4` — passed
- `ctest --test-dir build/repro --output-on-failure` — no tests found in this repository configuration
- `python3 scripts/copyright.py main --fix` — passed
- `git diff --check` — passed
- `./scripts/clang_format.py main` — could not run because `clang-format` is not installed

The diff is limited to `framework/gui.h`; no public API or dependency changes are included.

### DIFF
--- a/framework/gui.h
+++ b/framework/gui.h
@@ -1406,7 +1406,7 @@ inline bool Gui<bindingType>::update_buffers()
 	}
 
 	bool updated = false;
-	if (!vertex_buffer->get_handle() || (vertex_buffer_size != vertex_buffer->get_size()))
+	if (!vertex_buffer || !vertex_buffer->get_handle() || (vertex_buffer_size != vertex_buffer->get_size()))
 	{
 		vertex_buffer.reset();
 		vertex_buffer = std::make_unique<vkb::core::BufferCpp>(render_context.get_device(), vertex_buffer_size,
@@ -1416,7 +1416,7 @@ inline bool Gui<bindingType>::update_buffers()
 		updated = true;
 	}
 
-	if (!index_buffer->get_handle() || (index_buffer_size != index_buffer->get_size()))
+	if (!index_buffer || !index_buffer->get_handle() || (index_buffer_size != index_buffer->get_size()))
 	{
 		index_buffer.reset();
 		index_buffer = std::make_unique<vkb::core::BufferCpp>(render_context.get_device(), index_buffer_size,


asuessenbach:
Your fix might be correct. But it might as well just hide the actual error.

That is: does anybody know what that `explicit_update` flag is supposed to control?
The comment says "If true, update buffers every frame".  But none of the two `update_buffers` functions in `Gui` seem to care about that flag.
In the `Gui`-constructor, it controls the initialization of the `vertex_buffer` and the `index_buffer`. But both are used unconditionally at several places.

swayam8624:
Thanks for checking the lifecycle semantics. The `explicit_update` flag selects the persistent-buffer path: the constructor creates placeholder buffers, `update_buffers()` replaces/resizes and uploads them, and `draw_impl()` binds them. When the flag is false, `draw_impl()` instead allocates per-frame buffers through `update_buffers(command_buffer)`. The two overloads therefore do not need to inspect the flag themselves; the caller/path selection does.\n\nThe null guards in this PR address the reported deferred-allocation case where a zero-sized ImGui frame leaves the persistent buffers null before a later draw. I agree the flag comment could be clearer, but I would prefer to keep that separate from this narrowly scoped fix unless you recommend combining them.

asuessenbach (approved):

SaschaWillems:
Also for future reference: Can you use the PR template we provide? It has a checklist of all the things that need testing, which is missing here.

tomek-brcm (approved):

swayam8624:
Thanks for the reviews and approvals. All current CI checks are green. I’ll make sure to use the repository PR template/checklist for future contributions. Please let me know if anything else is needed from my side before merge.

SaschaWillems (approved):
Successfully tested on Android.

gpx1000 (approved):

marty-johnson59:
Merging - 4 approvals

PR #1577, opened 2026-09-11
Remove C++20 requirement

### LINKED ISSUE #1576: A single line of code introduces C++20 dependency
The samples code already contains string formatting utility similar to std::format, use that instead.

tomek-brcm:
A trivial fix that removes C++20 requirement.

Fixes #1576


SaschaWillems:
We explicitly bumped the min. C++ requirement to 20, see https://github.com/KhronosGroup/Vulkan-Samples#requirements

asuessenbach:
We're on C++20, right. But I vaguely remember that there was at least one platform that supports C++20, but not `std::format`. That's why we're still using `fmt::format`, instead of `std::format`.

asuessenbach (approved):

PR #1580, opened 2026-09-16
Remove unused PostProcessingComputePass

### LINKED ISSUE #841: framework/rendering/postprocessing_computepass is never used - should it be removed?
As the Title says.
Removing this class does not lead to any compile-time errors.

This class was introduced by #178, but was not used at that time. And it seems, it has never been used since.

### LINKED ISSUE #1570: PostProcessingComputePass: dangling iterators from get_resources() temporaries (found via clang 23 -Wdangling-gsl)
Building with clang 23 and warnings-as-errors fails in framework/rendering/postprocessing_computepass.cpp:

```
framework/rendering/postprocessing_computepass.cpp:154:33: error:
object backing the pointer will be destroyed at the end of the
full-expression [-Werror,-Wdangling-gsl]
```

Looking at the code, I think the warning is a true positive rather than noise. `PipelineLayout::get_resources()` returns
`const std::vector<ShaderResource>` by value (a fresh copy each call), and the transition code calls it three times:

```cpp
auto resource = std::find_if(pipeline_layout.get_resources().begin(),
                             pipeline_layout.get_resources().end(),
                             [&storage](const auto &res) { ... });
if (resource == pipeline_layout.get_resources().end())
    ...
const bool readable = !(resource->qualifiers & ...);
```

So `begin()` and `end()` come from two different temporary vectors, both are destroyed at the end of the statement, the later `end()`
comparison uses a third temporary, and `resource->qualifiers` dereferences an iterator into a destroyed vector.

It probably never crashes in practice because, as #841 noted, this class appears to be unused - but it is still compiled everywhere, so
clang 23 makes it a build failure for -Werror configurations. 

Two possible ways out:

1. Call get_resources() once and keep it in a local:

```cpp
const auto resources = pipeline_layout.get_resources();
auto resource = std::find_if(resources.begin(), resources.end(),
                             [&storage](const auto &res) { ... });
if (resource == resources.end())
```

or

2. remove the class, as #841 already suggested.

Happy to send a PR for option 1 if that is the preferred direction.

asuessenbach:
## Description

Build tested on Win11 with VS2026. Run tested on Win11 with NVidia GPU.

Fixes #1570
Fixes #841

## General Checklist:

## Sample Checklist


asuessenbach:
Clang Tidy Check is failing here as well. No idea why.

tomek-brcm (approved):

